Issue #1917, opened 2014-12-02
Update dashboard links


david-yz-liu:
Right now, dashboard links for assignments and spreadsheets go to their settings pages; they should instead point to the submissions table/grades table.

paymog:
To see if I understand what I should do correctly: dashboard links for assignments should point to the submissions table and dashboard links for spreadsheets should point to the grades table?

PR #1969, opened 2015-01-09
Fixed dashboard links for assignments and spreadsheets

paymog:
issue-1917 indicated that dashboard links for assignments should be point to thesubmission table for that particular assignment. Similarly, dashboard links for spreadsheets should point to the grades table. This has been implmeneted.

This is the first time I've committed to this repo and worked with rails. I wouldn't be surprised if I missed something (rails related, git related, workflow related, etc...) but I don't know what that might be. Would definitely apprciate a more-than-typical thorough review of what I've done.

Link to issue: https://github.com/MarkUsProject/Markus/issues/1917

### DIFF
--- a/app/views/assignments/_assignment_summary.html.erb
+++ b/app/views/assignments/_assignment_summary.html.erb
@@ -4,7 +4,7 @@
       <%= link_to I18n.t(:assignment_link,
                          short_id: assignment.short_identifier,
                          description: assignment.description),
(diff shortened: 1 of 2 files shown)


david-yz-liu:
@paymog Everything looks great. One GitHub tip for the future: https://help.github.com/articles/closing-issues-via-commit-messages/

Issue #1741, opened 2014-09-03
Checkboxes not reset after adding students to a group


david-yz-liu:
1. Go to the Groups view for any assignment.
2. Add a student to a group (checking the relevant checkboxes).
3. After the page reloads, the checkboxes are still checked. 

We should change this so the checkboxes are reset after the page loads.

PR #1972, opened 2015-01-13
Issue 1741 reset checkboxes on groups

Eitredes:
The groups page wouldn't clear checkboxes when you added users.
Fix:
Added a way for checkboxes in the Table class to be cleared, included syncing up the headers and footers.
Also, made a successful request clear the manager's list of selected students and groups.
Issue #1741


david-yz-liu:
Nice!

PR #1982, opened 2015-01-15
Issue 1980 - Changed 'Assignment Type' to 'Group Properties'

SoftwareDev:


SoftwareDev:
@david-yz-liu Looks like another race condition issue with the failing test. [Markus/test/unit/submission_collector_test.rb:219]

david-yz-liu:
I think we should remove "assignment.type" from the locale files too, now.

Issue #1919, opened 2014-12-02
Submission can be marked as complete even if not all rubric criteria are marked


david-yz-liu:
Haven't checked this for flexible criteria, though.

peterzhao333:
@david-yz-liu I took a look at the grader's interface and just wanted to clarify some things about this issue. From what I understand it is referring to the Grader's interface for marking submissions, where the issue lies in that the grader can mark the submission as completed even if he has not assigned a grade to all criteria laid out in the marking rubric?

If that is the case, what should be the desired behaviour? An prompt (overridable?) notifying the user that not all criteria have been met? Or should submissions not be allowed to be marked as 'complete' unless all rubric criteria have been assigned a mark?

david-yz-liu:
@peterzhao333 The change shouldn't be allowed at all. I'm fine with an alert appearing.

PR #1984, opened 2015-01-15
Fixed issue #1919: User cannot set marking state to 'complete' unless all rubric criteria are marked

peterzhao333:
Issue #1919 : Marking state of submissions can be set as 'complete' even when not all rubric criteria have been marked.

Fix :
If user attempts to set the submission's marking state to 'complete' when not all rubric criteria have been, the change is not completed and the user is prompted with the following error message:

"Unable to change marking status to complete, marks missing. Check the marking rubric and make sure that each criterion has been given a mark."

### DIFF
--- a/app/models/result.rb
+++ b/app/models/result.rb
@@ -118,8 +118,8 @@ class Result < ActiveRecord::Base
     if marks.where(mark: nil).first &&
           marking_state == Result::MARKING_STATES[:complete]
       errors.add(:base, I18n.t('common.criterion_incomplete_error'))
-      false
+      return false
     end
-    true
+    return true
   end
 end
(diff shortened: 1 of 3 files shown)


app/models/result.rb
```diff
@@ -118,8 +118,8 @@ def check_for_nil_marks
     if marks.where(mark: nil).first &&
           marking_state == Result::MARKING_STATES[:complete]
       errors.add(:base, I18n.t('common.criterion_incomplete_error'))
-      false
+      return false
     end
-    true
+    return true
```
david-yz-liu: @peterzhao333 So Hound is our Ruby style-checker. It's saying that this change was not necessary, and doesn't conform with the Ruby style guidelines.

Issue #1965, opened 2015-01-08
Adding new users has the wrong I18n text


david-yz-liu:
All three of Students, Graders, and Admins say "Add New File".

ishanthukral:
Working on this

ishanthukral:
@david-yz-liu, should it say "Add New Student" for Students, or "Ass New Grader" for Graders, or just "Add New" like in the French version? Thanks.

PR #1991, opened 2015-01-16
fixes issue #1965

ishanthukral:


david-yz-liu:
This changes the button text in the student submission view (where they go to submit new files). You should make separate text tags for each of the other user adding.

ishanthukral:
@david-yz-liu Hi David, thanks for pointing that out. I thought I had found all uses for it. Should it say "Add New Student" or just "Add New".

david-yz-liu:
I think it should be "Add New Student", "Add New TA", and "Add New Admin". You might want to do a quick search in en.yml; the tags might already exist.

ishanthukral:
@david-yz-liu, I added the new strings for "Add new Admin", "Add new Grader" and "Add new Student". There were existing strings but the words "Grader" and "Admin" were pluralized. I felt that since these links create one new student/admin/grader, they wouldn't be the best suited. Please let me know if you'd like me to user the existing strings instead. Thanks!

david-yz-liu:
Looks great.

PR #1998, opened 2015-01-17
File size restriction on the file submission

MatakuT:
File size restriction on the file submission


MatakuT:
I had to edit "file_manager.html.erb" to fix one issue:
"added negation to the previous .nil fix. Without this negation the submission controller size limiting portion was breaking"
I was able to successfully commit, but I cannot see this fix here. Is it because the branch is closed? After creating the new branch an running git diff on that one file I got empty result. I assume the change is here. It's very important to push it through, so I want to know for sure.

david-yz-liu:
I already made this change, don't worry.

MatakuT:
Didn't see this comment. Thank you. I'm deleting maryna_branch

PR #2001, opened 2015-01-17
Fix duplicate criteria name error message

### LINKED ISSUE #1795: Duplicate criterion error message is awful
Try creating two criteria with the same name for either Rubric or Flexible criteria modes. We don't seem to have a standard error message for this.

It really should be the same error message for both cases.

paymog:
Fixes #1795. The error message presented to the user upon attempting
to create a new criteria (for either flexible or rubric) was awful. This
has been fixed, internationalized and tested.


david-yz-liu:
@paymog Looks like there are some Hound comments. Also, this change caused some old tests to fail. (See Travis build.)

Rather than simply fix them, could you please spend some extra time converting these tests to RSpec? This will be greatly beneficial in the long run.

paymog:
All of the failures are in the same file (flexible_criteria_controller_test.rb). Do you envision just converting the failing tests or the whole test file?

david-yz-liu:
Whew, looks good! I know this ended up being more work than intended, but I really appreciate it.

Issue #1621, opened 2014-07-25
Assignment Submission Rule fields require validation


david-yz-liu:
Right now, if you try to add a grace period that's empty, MarkUs will crash. Same for Penalty Decay Formula and Penalty Formula.

@arkon?

jiangha4:
I'll take a look at this.

PR #2007, opened 2015-01-18
Fix for #1621: Submission Rule fields are now required fields

peterzhao333:
Submission rule fields are now required to be filled in the Assignment edit/create form if a submission rule is chosen.

Fix for #1621


peterzhao333:
The issue still seems to persist even when the Submission Rule fields are populated. I'll try to look into it more and find the root cause.

PR #2012, opened 2015-01-21
added negation to the previous .nil fix. 

MatakuT:
added negation to the previous .nil fix. 
Without this negation the submission controller size limiting portion was breaking


david-yz-liu:
@MatakuT This was already fixed, you just need to update your branch.

Issue #76, opened 2010-09-29
Disallow non-AssignmentFiles option in Assignment Properties


benjaminvialle:
From Byron:

"Should have an option when setting up the assignment whether files not listed in the assignment spec will be accepted."

This goes against our "accept all" policy. Might have to ask Karen about this.

MatakuT:
I'm working on it :)

MatakuT:
Almost done - the skeleton is working, just have to work on the checkbox message. Got stuck with making an option for disallowing non-assignment files to show in the right place under the right action. My goal is to show the checkmark after clicking on "Add a Required File". I made a flag (false) in the controller that is passed to the assignment form, then to the helper method responsible for processing the mini-form triggered by clicking on "Add a Required File". Then in the helper method the flag is set to true and passed to the partial rendered by the original form. If the value is true then the checkbox is shown, otherwise it's hidden. The problem is that this helper method is called independent from whether "Add a Required File" is clicked, so the flag is set to true every time the assignment form is rendered. It has to only be set to true if the partial form is rendered - I still have to find the rule/way for checking that.

MatakuT:
done and pushed

PR #2013, opened 2015-01-21
Iss76

MatakuT:
Disallow non-AssignmentFiles option in Assignment Properties #76

### DIFF
--- a/app/assets/javascripts/create_assignment.js
+++ b/app/assets/javascripts/create_assignment.js
@@ -68,6 +68,10 @@ jQuery(document).ready(function() {
       document.getElementById('assignment_group_min').value = this.value;
     }
   });
+  
+  //hide the checkbox for "Required Files" by default
+  jQuery('#assignment_is_required').hide();
+  jQuery('#assignment_is_required_label').hide();
 });
 
 
--- a/app/assets/javascripts/create_assignment.js
+++ b/app/assets/javascripts/create_assignment.js
@@ -69,7 +69,7 @@ jQuery(document).ready(function() {
     }
   });
   
-  //hide the checkbox for "Required Files" by default
+  // Hide the checkbox for "Required Files" by default
   jQuery('#assignment_is_required').hide();
   jQuery('#assignment_is_required_label').hide();
 });
--- a/app/assets/javascripts/create_assignment.js
+++ b/app/assets/javascripts/create_assignment.js
@@ -72,6 +72,12 @@ jQuery(document).ready(function() {
   // Hide the checkbox for "Required Files" by default
   jQuery('#assignment_only_required_files').hide();
   jQuery('#assignment_only_required_files_label').hide();
+  
+  
+  jQuery('#assignment_only_required_files_edit').show();
+  jQuery('#assignment_only_required_files_edit_label').show();
+  
+  
 });
 
 
--- a/app/assets/javascripts/create_assignment.js
+++ b/app/assets/javascripts/create_assignment.js
@@ -70,8 +70,8 @@ jQuery(document).ready(function() {
   });
   
   // Hide the checkbox for "Required Files" by default
-  jQuery('#assignment_is_required').hide();
-  jQuery('#assignment_is_required_label').hide();
+  jQuery('#assignment_only_required_files').hide();
+  jQuery('#assignment_only_required_files_label').hide();
 });
 
 
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -197,7 +197,6 @@ class AssignmentsController < ApplicationController
   # Called on editing assignments (GET)
   def edit
     @assignment = Assignment.find_by_id(params[:id])
-
     @past_date = @assignment.section_names_past_due_date
     @assignments = Assignment.all
     @sections = Section.all
@@ -293,7 +292,6 @@ class AssignmentsController < ApplicationController
         flash[:success] = I18n.t('assignment.create_success')
       end
     end
-
     redirect_to action: 'edit', id: @assignment.id
   end
 
@@ -759,6 +757,7 @@ class AssignmentsController < ApplicationController
         :group_name_displayed,
         :invalid_override,
         :section_groups_only,
+        :is_required,
         section_due_dates_attributes: [:_destroy,
                                        :id,
                                        :section_id,
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -757,7 +757,7 @@ class AssignmentsController < ApplicationController
         :group_name_displayed,
         :invalid_override,
         :section_groups_only,
-        :is_required,
+        :only_required_files,
         section_due_dates_attributes: [:_destroy,
                                        :id,
                                        :section_id,
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -273,7 +273,7 @@ class SubmissionsController < ApplicationController
     @file_manager_errors = Hash.new
     assignment_id = params[:assignment_id]
     @assignment = Assignment.find(assignment_id)
-    required_assignments = AssignmentFile.where(
+    required_files = AssignmentFile.where(
                            assignment_id: @assignment).pluck(:filename)
     students_filename = []
     @path = params[:path] || '/'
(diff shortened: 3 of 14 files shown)


app/assets/javascripts/create_assignment.js
```diff
@@ -68,6 +68,10 @@ jQuery(document).ready(function() {
       document.getElementById('assignment_group_min').value = this.value;
     }
   });
+  
+  //hide the checkbox for "Required Files" by default
```
david-yz-liu: Add a space after the `//`

app/controllers/assignments_controller.rb
```diff
@@ -759,6 +757,7 @@ def assignment_params
         :group_name_displayed,
         :invalid_override,
         :section_groups_only,
+        :is_required,
```
david-yz-liu: This attribute name isn't very good. I would suggest "only_required_files" instead.

app/controllers/submissions_controller.rb
```diff
@@ -365,7 +369,17 @@ def update_files
                                 ' for assignment ' +
                                 "'#{@assignment.short_identifier}'.")
         end
-
+        #
+        unless students_filename.length < 1
+          if required_assignments != students_filename &&
+             @assignment.is_required == true
+            @file_manager_errors[:size_conflict] =
+            'The list of uploaded files does not match' +
```
david-yz-liu: Please make this an internationalized string (i.e., make a new value in en.yml, fr.yml, and pt.yml). You can use Google translate for the other languages.

app/controllers/submissions_controller.rb
```diff
@@ -365,7 +369,17 @@ def update_files
                                 ' for assignment ' +
                                 "'#{@assignment.short_identifier}'.")
         end
-
+        #
+        unless students_filename.length < 1
+          if required_assignments != students_filename &&
+             @assignment.is_required == true
```
david-yz-liu: There should be no need to compare this attribute to "true", since it's always either true or false.

app/controllers/submissions_controller.rb
```diff
@@ -273,6 +273,9 @@ def update_files
     @file_manager_errors = Hash.new
     assignment_id = params[:assignment_id]
     @assignment = Assignment.find(assignment_id)
+    required_assignments = AssignmentFile.where(
```
david-yz-liu: Shouldn't this variable be "required_files"?

app/views/assignments/_assignment_file.html.erb
```diff
@@ -1,8 +1,8 @@
 <div class="assignment_file">
 <%= form.fields_for :assignment_files, assignment_file do |f| %>
   <p>
-    <%= f.label :filename, t(:filename).html_safe %>
-    <%= f.text_field :filename %>
+    <%= f.label :filename, t(:filename).html_safe %>
```
david-yz-liu: Please remove the trailing whitespace from these two lines.

app/views/assignments/_form.html.erb
```diff
@@ -212,13 +212,23 @@
       <%= f.fields_for :assignment_files do |assignment_file_form| %>
         <%= render partial: 'assignment_file',
                    locals: { assignment_file: assignment_file_form.object,
-                             form: f } %>
+                             form: f } %>
```
david-yz-liu: Please remove the trailing whitespace here as well.

app/views/assignments/_form.html.erb
```diff
@@ -212,13 +212,23 @@
...
                    locals: { assignment_file: assignment_file_form.object,
-                             form: f } %>
+                             form: f } %> 
       <% end %>
     </div>
-
      <p id='assignment_file_add_link'>
         <%= add_assignment_file_link t('assignment.add_required_file'), f %>
      </p>
+     </ br>
```
david-yz-liu: This one looks a bit odd; you don't need the "`/` ", just the `br`.

app/views/assignments/_form.html.erb
```diff
@@ -212,13 +212,23 @@
...
+     <%= f.check_box :is_required,
+       checked: @assignment.is_required,
+       onclick: 'set_onbeforeunload(true);' 
+     %>
+     <%= @filename %>
+     <%= f.label :is_required,
+       "Accept no other files besides the listed required files.
```
david-yz-liu: Please make this be an internationalized string as well.

app/views/assignments/_form.html.erb
```diff
@@ -212,13 +212,23 @@
...
+                             form: f } %> 
       <% end %>
     </div>
-
      <p id='assignment_file_add_link'>
         <%= add_assignment_file_link t('assignment.add_required_file'), f %>
      </p>
+     </ br>
+     <%= f.check_box :is_required,
+       checked: @assignment.is_required,
```
david-yz-liu: The alignment in this and the f.label tags is off. The arguments should all be vertically aligned, so here `checked` should be aligned with `:is_required`, etc.

app/views/assignments/_form.html.erb
```diff
@@ -212,13 +212,23 @@
...
+       onclick: 'set_onbeforeunload(true);' 
+     %>
+     <%= @filename %>
+     <%= f.label :is_required,
+       "Accept no other files besides the listed required files. 
+       This would also restrics the file extensions.",
+       {id: "assignment_is_required_label"}
```
david-yz-liu: In addition to alignment, please put spaces after the `{` and before the `}`.

db/migrate/20150118173952_add_is_required_to_assignments.rb
```diff
@@ -0,0 +1,5 @@
+class AddIsRequiredToAssignments < ActiveRecord::Migration
+  def change
+    add_column :assignments, :is_required, :boolean, default: 'false'
```
david-yz-liu: I'll repeat my above comment; I think a better column name is something like `only_required_files`.

david-yz-liu:
Hi @MatakuT, I've made some minor style comments that I'd like you to address.

Also, the one main error in functionality is that when there are multiple required files, you're currently forcing students to submit them all at once. However, students should still be able to submit a proper subset of the required files, even when the checkbox is clicked.

david-yz-liu:
Oh, and your migration should have changed schema.rb as well; we're storing that in the repo too, so please add that to your commit.

MatakuT:
David, please test.
I want to change few things later as well. The error message:
"The list of uploaded files does not match the assignment required files" I want to append the list of required files at the end of the message
Also, I want to make file extensions more flexible. For example, if "assignment.txt" is required, I want to allow .doc and .pdf + few excel formats. Also add .jpg/jpeg, .gif and .png for image submissions.

app/controllers/assignments_controller.rb
david-yz-liu: @MatakuT You forgot to change this.

david-yz-liu:
@MatakuT Looking good.

You need to add the internationalized strings to `en.yml`, `fr.yml`, and `pt.yml`, using Google translate probably. I don't think I need to see a list of the required files in the error message right now, and I'm not interested in being flexible with file formats. It is often the case that a particular format (pdf vs. txt) is actually required.

Once you add the strings and address the comment left above, I'll merge this in!

MatakuT:
Oh, I added the strings, but forgot to add .yml files. Will push that in a bit

david-yz-liu:
There are also two failing tests to deal with, it seems.

MatakuT:
the tests should pass now.

david-yz-liu:
It still seems like you aren't checking in the controller the attribute that determines whether students can only submit required files, or whether they can also submit other files.

MatakuT:
Sorry David. Completely misunderstood your comment. The check was added. I had to also change some names that still had 'is_required' string in them.

MatakuT:
That is indeed much simpler :)

david-yz-liu:
One more issue - it looks like the checkbox doesn't appear after a new assignment has been created, even if there's a required file and it's been checked.

The checkbox should appear if and only if there is at least one required file field, at any time. Also, it would be nice if the checkbox were automatically unchecked (set to false) if there are no required files.

MatakuT:
I have noticed that as well. Right now the checkbox appears when a user clicks "Add a Required File" link. I will change a couple rules to make it appear on edit assignments without a need to click on "Add a Required File". In this case the checkbox will appear under the link.

david-yz-liu:
Okay. Not sure if it's easier to make it always appear, but just disable it when not applicable. I don't particularly care which.

PR #2015, opened 2015-01-22
1975 populate on remark

Eitredes:
Marks are now autofilled to their original values when you remark an assignment.


david-yz-liu:
Great work, @Eitredes

PR #2017, opened 2015-01-25
Issue#989 annotation

spring-ryanc:
Preview of the line annotation work. My goal is to replace the use of the highlight function and do the span splitting myself. This will keep the element structure intact. Using highlight is unpredictable and has produced several edge cases I did not expect.


david-yz-liu:
@spring-ryanc Don't modify the schema.rb file directly; instead, create a migration (see Rails guide for details).

Issue #1760, opened 2014-09-15
Student Interface: Spreadsheet marks table should be updated to React


david-yz-liu:
I suppose we're using an old template in there, that was overlooked during the style overhaul this past summer.

PR #2018, opened 2015-01-28
Fix for issue 1760: Updated Student Interface Spreadsheet marks table to React

peterzhao333:
Fix for issue #1760 : 
Updated student interface's spreadsheet marks table to React


peterzhao333:
Two of the grade_entry_forms_controller_tests seem to be failing because they are using a regular expression to search for the expected result string from the response of a student_interface GET request.

However, because we are using React tables now, I believe they are not compiling in the test and the table data is not updated, and a matching string is not found.

Any ideas what we should do with these @david-yz-liu?

app/views/grade_entry_forms/student_interface.html.erb
```diff
@@ -12,42 +12,136 @@
...
-                  <% end %>
-                </td>
-              <% end %>
-            </tr>
-          </tbody>
-       </table>
-      </div>
+      <div id='marks_table'></div>
+
+      <script type='text/jsx'>
```
arkon: Should this all be moved to a separate file...?

peterzhao333: Thanks for the suggestion @arkon , I moved it to its own separate file.

david-yz-liu:
@peterzhao333 I'm okay removing those asserts (not the full tests, just the RegEx asserts).

Everything's looking good, except the spreadsheet columns should be sortable.

david-yz-liu:
Thanks, @peterzhao333!

PR #2024, opened 2015-02-05
Partially working issue-1738

tsimica:
This code hasn't been tested with having only one person in a group but has been with the default seeds file. Also, another issue will have to be made with respect to having either a Penalty Submission Rule or a Penalty Decay Rule as the schema does not enforce the order in which penalties are enforced (and thus can't be enumerated). 

I will work on this after the Rails 4 issues are completed.


app/views/assignments/_read.html.erb
```diff
@@ -11,12 +11,47 @@
...
   <div class='sub_block'>
     <span class='prop_label'><%= t('due_date') %>:</span>
     <%= I18n.l(@due_date, format: :long_date ) %>
   </div>
+    
+  
+  <div class='sub_block'>
+    <span class='prop_label'>Lateness Penalty:</span>
+    <% if @penalty.type == "GracePeriodSubmissionRule" %>
+      Grace Period
```
david-yz-liu: These should be internationalized strings.

app/views/assignments/_read.html.erb
```diff
@@ -11,12 +11,47 @@
...
+    <% elsif @penalty.type == "PenaltyPeriodSubmissionRule" %>
+      Penalty Period 
+    <% else @penalty.type == "NoLateSubmissionRule" %>
+      Late submissions are <b>not</b> allowed
+    <% end %>
+  </div>
+
+
+  <div class='sub_block'>
+    <span class='prop_label'>Penalty:</span>
```
david-yz-liu: Would be nice if this didn't appear when no late submissions are allowed. Otherwise students might be scared by the word "penalty".

app/views/assignments/_read.html.erb
```diff
@@ -11,12 +11,47 @@
...
+    <% if @penalty.type == "GracePeriodSubmissionRule" %>
+      <% if !@grouping.nil? %>
```
david-yz-liu: Use `unless` here

app/controllers/assignments_controller.rb
```diff
@@ -77,6 +77,10 @@ def student_interface
 
     @student = current_user
     @grouping = @student.accepted_grouping_for(@assignment.id)
+    @penalty = SubmissionRule.find_by_assignment_id(@assignment.id)
+    
+    # sort if it's penaltydecay or penalty period
+    @enum_penalty = Period.find_all_by_submission_rule_id(@penalty.id).sort_by{|obj| obj.hours}
```
david-yz-liu: I think I'd prefer sorting by obj.id. It's still wrong, but less obviously so.

app/views/assignments/_read.html.erb
```diff
@@ -11,12 +11,47 @@
...
+          <%= raw(t('student.group_credits',
+                      available_grace_credits: @grouping.available_grace_credits)) %>
+      <% end %>
+    <% elsif @penalty.type == "PenaltyDecayPeriodSubmissionRule" %>
+      Penalty Decay
+    <% elsif @penalty.type == "PenaltyPeriodSubmissionRule" %>
+      <ul>
+      <% @enum_penalty.each do |p| %> 
+        <li> After <%= p.hours < 1 ? (p.hours * 60).round : p.hours.round %> <%= p.hours < 1 ? "minutes" : (p.hours == 1 ? "hour" : "hours") %>, <%= p.deduction %>% will be deducted. </li>
```
david-yz-liu: Is the rounding necessary here? I don't see why we wouldn't just stick with what the instructors put in.

Also, it would be good if the reported "after X hours" was cumulative.

app/views/assignments/_read.html.erb
```diff
@@ -11,12 +11,47 @@
...
+
+  <div class='sub_block'>
+    <span class='prop_label'>Penalty:</span>
+    <% if @penalty.type == "GracePeriodSubmissionRule" %>
+      <% if !@grouping.nil? %>
+          <%= raw(t('student.group_credits',
+                      available_grace_credits: @grouping.available_grace_credits)) %>
+      <% end %>
+    <% elsif @penalty.type == "PenaltyDecayPeriodSubmissionRule" %>
+      Penalty Decay
```
david-yz-liu: Even though it's primitive, I'm okay for now using the same presentation as for PenaltyPeriodSubmissionRule

app/views/assignments/student_interface.html.erb
```diff
@@ -255,6 +229,48 @@
           <% end %>
         <% end %>
 
+
+        <% if !@group.nil? %>
```
david-yz-liu: It seems like the indentation in the whole section is a little off.
Also, it would be nice to internationalize the strings.

config/locales/en.yml
```diff
@@ -681,6 +681,7 @@ en:
       form_group: "The groups must be from <strong>%{group_min}</strong> to <strong>%{group_max}</strong> people."
       decline_invitation2: "This will decline the invitation you have. Are you sure?"
       your_group: "Your group"
+      your_group_header: "Your Group"
```
david-yz-liu: Can you double-check where `your_group` is used? It might be worth merging these two strings.
Also, when you add the internationalized strings, please add them to the fr and pt files.

tsimica: Merged in commit. Removed your_group_header.

david-yz-liu:
Three comments:

1. The penalties in "penalty period" are being accumulated in reverse order
2. The time for the penalties should be cumulative, in addition to the actual penalties themselves. This can be signalled by adding the phrase "after XX hours *after the due date*, ..."
3. Change the "Penalty" field name to something like "Additional information". The current heading doesn't make sense when grace tokens are used.

david-yz-liu:
@tsimica Good work, but there is still this comment:

The time for the penalties should be cumulative, in addition to the actual penalties themselves. This can be signalled by adding the phrase "after XX hours after the due date, ..."

david-yz-liu:
@tsimica Just tried again, same problem.

david-yz-liu:
Thanks, @tsimica!

PR #2026, opened 2015-02-05
Iss218

MatakuT:
Write tests for calculate_collection_time and calculate_grouping_collection_time in SubmissionRule.
Written few rules, at least one more to come


david-yz-liu:
I think the tests look good, but unfortunately right now they're bundled with your changes for issue 76.

We can wait for that PR to be merged, or you can redo your changes on a fresh branch off of master. (It's also possible to use rebase, although that's a bit more complicated than simply saving a temp version of the new RSpec file.)

PR #2035, opened 2015-02-10
Fixes Non Web-Submit Group Creation

SoftwareDev:
Fixed a gitolite config option.


SoftwareDev:
Included a fix from master branch from @peterzhao333 for that pesky test

PR #2036, opened 2015-02-11
Remove members functionality now working

SoftwareDev:
Gitolite config changes.


SoftwareDev:
Build will start to pass after f0bc428 is applied in other PR

PR #2040, opened 2015-02-12
Key Pairs Page

SoftwareDev:
Template for the new "Key Pairs" page where students will manage their SSH keys for Git repo access.


david-yz-liu:
@SoftwareDev The page should appear for Admins as well.

SoftwareDev:
@david-yz-liu You're right. It'll just have a different layout where they can manage all keys and upload as well. I'll get on that.

SoftwareDev:
@david-yz-liu A lot of this was auto-generated through scaffold, will be throwing out most of it.

PR #2044, opened 2015-02-15
Configure hound for javascript and scss

david-yz-liu:
Fix @2019

@arkon I followed this document: https://houndci.com/configuration

Do you know if this was the right way to do this?


arkon:
I haven't actually tried doing it, but it seems right?

david-yz-liu:
Okay well let's give it a go.

PR #2051, opened 2015-02-20
Issue#1936 annotation

spring-ryanc:
Turns out I refactored out some support for the multi-line annotations. I wasn't able to recreate the other issue but I tested it pretty thoroughly and found some other small things that snuck by (including one that would have been existing for the previous annotation setup).

### DIFF
--- a/app/assets/javascripts/SourceCodeGlower/SourceCodeLine.js
+++ b/app/assets/javascripts/SourceCodeGlower/SourceCodeLine.js
@@ -23,35 +23,107 @@ function SourceCodeLine(line_node) {
 }
 
 // Increase a Source Code Line's glow depth
+// The base of the new glow functionality was found here:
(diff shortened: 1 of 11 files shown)


spring-ryanc:
Yikes, I think I got all the major hound fixes. I'm not sure if I should change those two camel cases issues because they are from another class and that might make hound examine it as well. A bunch of these comments were from things I didnt even change

david-yz-liu:
@spring-ryanc I'm okay not fixing those changes (Hound is a bit odd). But the tests are failing - could you look into this?

arkon:
@david-yz-liu You can create a config file based on [the defaults](http://robots.thoughtbot.com/hound-reviews-javascript-for-style-violations) to turn things like camel case variables off.

david-yz-liu:
@spring-ryanc Great job!

PR #2052, opened 2015-02-20
added several unit tests to iss218

MatakuT:


david-yz-liu:
@MatakuT Your new tests seem to be failing (see Travis build)

MatakuT:
I think I forgot to upload one important file

david-yz-liu:
@MatakuT Okay, it would be good to have tests which actually check the value of the time returned by the methods.

MatakuT:
The time returned is always different, so I cannot compare it to a fixed string. If that is what you mean by checking the value

david-yz-liu:
Even though the methods probably use the current time, which does change, it would be acceptable to check the computed time to within a certain tolerance (even in hours).

david-yz-liu:
There seems to be a failing test.

MatakuT:
That test had something to do with FactoryGirl naming conventions. They supposed to be unique for all tests. I assume that it is failing because of group names not being unique. Will see if my commit will fix that

app/models/submission_rule.rb
```diff
@@ -27,13 +27,13 @@ def get_collection_time
...
   end
-
+  
   def calculate_collection_time
     assignment.latest_due_date + hours_sum.hours
   end
 
   def calculate_grouping_collection_time(grouping)
-    if grouping.inviter.section
+    if !grouping.inviter.nil?
```
david-yz-liu: @MatakuT This change breaks existing functionality. SectionDueDate should only be queried when a grouping is part of a section.

PR #2054, opened 2015-02-21
1933 fix save button

Eitredes:
Fixes issue 1933, and also includes a chunk of refactoring and misc tweaks.


Eitredes:
Realized it still needs some work; namely, there are marker and student versions of the moved files (I had assumed just student). May revert the movement if I can't find an alternate solution.

Eitredes:
Ended up just undoing the move; I felt it was better to maintain consistency, since this pattern is fairly common, and a merge of the two files would (at best) create massive piles of if-checking based on roles in between some basic boilerplate.

PR #2055, opened 2015-02-22
routes_spec testing for issue 664

MatakuT:


david-yz-liu:
@MatakuT This is a good start, although there are some outstanding Hound comments

MatakuT:
Will fix houndci and potential Travis CI's complaints a bit later.
Will also revisit commented-out tests

spec/routing/routes_spec.rb
```diff
@@ -0,0 +1,1590 @@
...
+          action: 'global_actions',
+          assignment_id: assignment.id.to_s,
+          locale: 'en')
+      end
+    end
+  end
+  # end assignment graders route tests
+
+  context 'annotation_categories' do
+    let(:annot_path) {
```
MatakuT: I have researched and there seems to be no way to re-write 'let' with 'do end' syntax

PR #2056, opened 2015-02-23
Made the remarks able to have marks added

Eitredes:
...re two extra marks lists - the old list, and the remark. When there's a remark, the old marks list appears before the new one in the html, so the add extra mark button puts the dialogue on that html (you'll have to look at the source of the page to see it. So, I swapped their order. Also, made the old marks hidden by default, since they weren't before.

Fixes issue 1869.


david-yz-liu:
@Eitredes The issue seems to be fixed, but after the remark has been released, the old marks aren't hidden any more.

Issue #1978, opened 2015-01-14
Hide section columns from all tables where there are no sections


david-yz-liu:
We don't really use Sections at UofT, so the empty column is just taking up space.

PR #2058, opened 2015-02-26
#1978 hide sections

spring-ryanc:


spring-ryanc:
@david-yz-liu Is it possible to undo only the last commit? It introduces an issue for the marks_graders page since its not using react. If not it's something that will be fixed when I finish the other bug Im working on (1673)

PR #2060, opened 2015-02-26
We now delete remarks.

Eitredes:
Very small fix. Fixes issue 1017


david-yz-liu:
Is there also a "remark_request" we need to delete?

Eitredes:
No. The request text is a column in the submissions table, and gets set to null on deletion (this was the original behaviour, the addition just gets rid of the results row that was created for the remark).

PR #2065, opened 2015-02-28
Fix for Issue 2031: Updated Admin/Grader Spreadsheet marks table to React

david-yz-liu:
Updated the marks spreadsheet table for the Admin/Grader's view to React


peterzhao333:
@david-yz-liu Let me know what else needs to be done before this can be merged. Could you elaborate on what you meant when you said that the marks can't be entered right now?

david-yz-liu:
@peterzhao333 I think we talked about the marks not changing the React state last meeting, but even more importantly, it doesn't seem like entering marks is persisted to the database. If I enter a mark and refresh the page, the entered mark doesn't appear.

peterzhao333:
@david-yz-liu Gotcha. That's weird though, because the grade entry seems to be work working fine on my end. I'll look into this

david-yz-liu:
Browser? I'm using Chrome right now...

PR #2071, opened 2015-03-04
Key Pairs Page

SoftwareDev:
- "Add a key pair" page only allows the current user to upload their own public key.
- Created a new constant for the key-dir directory.
(Sorry about last PR)

### DIFF
--- a/app/controllers/key_pairs_controller.rb
+++ b/app/controllers/key_pairs_controller.rb
@@ -44,13 +44,13 @@ class KeyPairsController < ApplicationController
   # Creates the KEY_STORAGE directory if it does not yet exist
   def upload_key_file(file, user_name)
     Dir.mkdir(KEY_STORAGE) unless File.exists?(KEY_STORAGE)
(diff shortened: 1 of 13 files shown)


app/controllers/key_pairs_controller.rb
```diff
@@ -37,18 +37,48 @@ def edit
...
   # POST /key_pairs
   # POST /key_pairs.json
   def create
-    @key_pair = KeyPair.new(key_pair_params)
+    # Upload the file
+    upload_key_file(key_pair_params[:file], @current_user.user_name)
+
+    # Save the record
+    @key_pair = KeyPair.new(
+        key_pair_params.merge(user_name: @current_user.user_name,
```
SoftwareDev: @david-yz-liu Been trying to figure out what exactly it's complaining about here for awhile now...

david-yz-liu: It should be indented two spaces, not four

Issue #1974, opened 2015-01-13
Add "Bonus" columns to spreadsheets


david-yz-liu:
Right now every column in a spreadsheet contributes to the total mark (and thus the average calculation).

It would be nice (e.g., for bonus questions on exams) to mark a column as "bonus".

peterzhao333:
@david-yz-liu Could you elaborate on the behaviour for bonus columns when it comes to the total mark, and computing class average calculations?

Are marks in bonus columns supposed to be added to the student's total mark? (If not, I was wondering how would the student know their final mark with bonus marks included? )

Are marks in bonus columns supposed to be added to class average calculations?

david-yz-liu:
They can be added to the total mark, and hence average calculations. The main point is the point value for the column shouldn't be added to the point total for the whole spreadsheet.

peterzhao333:
When you say point value of the column, and point value for the whole spreadsheet, are you referring to the Grade/Admin view of the Editable Spreadsheet?

david-yz-liu:
Yes, but this (should) also be what's used to calculate averages, which the students see.

peterzhao333:
@david-yz-liu I updated the pull request with the change. Let me know if this was what you meant.

PR #2072, opened 2015-03-04
Fix for Issue-1974 - Added Bonus Columns

peterzhao333:
Fix for #1974 
Added Bonus Column attribute to Grade Entry Items, and a checkbox to toggle the bonus attribute.

Excluded Bonus Columns from Total Calculation


david-yz-liu:
@peterzhao333 I actually meant that the bonus mark should not be added to the total mark for the spreadsheet, not the individual user.

So if the midterm is out of 20, that doesn't include a bonus question (which could be out of 5).

Issue #2075, opened 2015-03-06
parse_csv action crashes in grade_entry_form controller


MatakuT:
upload_csv should gracefully fail on non-csv file with .csv extension, but right now it throws an exception. upload_csv only checks for correct file extension and passes the file to parse_csv which cannot candle non-csv files. 

Edit: Here is the error that I'm getting
     CSV::MalformedCSVError:
       Illegal quoting in line 2.

david-yz-liu:
Thanks, @MatakuT. @paymog this is basically the same error as before.

paymog:
Sounds good. Wanna assign it to me?

paymog:
I've fixed this problem but one of the tests is now failing. Specifically, a test in ```grade_entry_forms_controller_test.rb``` which ```should 'have invalid values in database after an upload of an ISO-8859-1 encoded file parsed as UTF-8'```

I must be misunderstanding something because it seems that the test is expecting to DB to have invalid values when the user messes up. Surely this isn't what's actually desired.

david-yz-liu:
Yes, you can remove the test.

PR #2077, opened 2015-03-08
added numerous tests for file upload/download

MatakuT:


MatakuT:
I tried to cut the line length as much as possible to make the code still readable. 
Also, there are two failing tests that will pass as soon as the tickets that I issued will be handled.
The ticket numbers: #2078 , #2075

david-yz-liu:
Cut the line lengths to 80, please.

david-yz-liu:
Comment out the failing tests and add a note to those two issues, then.

MatakuT:
I have one outstanding houndci comment for this issue. But I'm not sure how to fix it.
Avoid using {...} for multi-line blocks.
Block body expression is on the same line as the block start. For the following lines of code:
let(:csv_options) { {
      filename: "#{grade_entry_form.short_identifier}_grades_report.csv",
      disposition: 'attachment',
      type: 'application/vnd.ms-excel'
    } }

spec/controllers/grade_entry_forms_controller_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+      expect(response.status).to eq(302)
+      expect(flash[:error]).to_not be_empty
+      expect(response).to redirect_to(
+        grades_grade_entry_form_path(grade_entry_form, locale: 'en'))
+    end
+  end
+
+  context 'CSV_Downloads' do
+    let(:csv_data) { grade_entry_form.get_csv_grades_report }
+    let(:csv_options) {
```
MatakuT: I have researched and there seems to be no way to re-write 'let' with 'do end' syntax

PR #2079, opened 2015-03-08
Catch exceptions for csv uploads in entire app

### LINKED ISSUE #2075: parse_csv action crashes in grade_entry_form controller

paymog:
Now, whenever a user uploads a bad csv file (either malformed
or a binary file with csv extension) the problem will be caught
and gracefully handled by flashing an error. This error message
has been localized with Google Translate. Tests have been written
for all of the affected files. Fixes #2075.


paymog:
Should a separate issue be opened to get proper localization?

david-yz-liu:
@paymog Yay! Don't worry about localization right now.

PR #2082, opened 2015-03-12
2037 expired session modal

Eitredes:
Adds an alert when a session is 5 minutes from expiring. Fixes issue 2037.


david-yz-liu:
@Eitredes After some experimentation, I don't think you've set the path to main/check_timeout correctly. Depending on the page I'm on when I log in, it seems that the periodic request just appends main/check_timeout to that URL. You probably want to use a path helper, something like `check_timeout_main_path()`.

Please let me know if you don't understand the problem I'm referring to.

Eitredes:
Got it. Should be fixed now.

david-yz-liu:
@Eitredes I'm not sure if you made the change somewhere that I missed, but I don't see it?

Eitredes:
Oops, forgot to push it.

david-yz-liu:
Yes.

PR #2084, opened 2015-03-12
confirm box added if file size is greater than 8 mb

MatakuT:


reidka:
I think the message is fine.
It sounds from @DustinWehr that it may not address his main concern which is pdf files that the MarkUs viewer can't open.  I think we could go ahead with this fix, and continue to work on that issue.

PR #2085, opened 2015-03-13
Issue1673 Marks Graders react table

spring-ryanc:
This is an early pull request so users can benefit from the update. I still need to go back and remove the files that are no longer used. I also noticed that the controller doesn't provide the most helpful error messages so I can fix that up. And last the check all check boxes stay checked (this seems to be an issue on other tables as well)


app/views/graders/_graders_manager.js.jsx.erb
```diff
@@ -108,7 +108,7 @@
       to_render.push(<input type='checkbox'
         checked={this.state.assign_to_criteria}
         onChange={this.criteriaCheckboxClicked}><%= j raw I18n.t('graders.assign_to_criteria') %></input>);
-      to_render.push(<GroupsTableManager 
+        to_render.push(<GroupsTableManager
```
david-yz-liu: This was properly indented before.

app/views/marks_graders/_marks_grader_table.js.jsx.erb
```diff
@@ -0,0 +1,128 @@
...
+    },
+    componentWillMount: function() {
+      this.refresh();
+    },
+    // Asks for new info from backend and sets props accordingly.
+    refresh: function() {
+      document.getElementById('working').style.display = '';
+
+      jQuery.ajax({
+        url: '/en/grade_entry_forms/1/marks_graders/populate_graders_table',
```
david-yz-liu: Won't this only work for the grade_entry_form with id 1?

app/views/marks_graders/_marks_graders_student_table.js.jsx.erb
```diff
@@ -0,0 +1,170 @@
...
+    },
+    componentWillMount: function() {
+      this.refresh();
+    },
+    // Asks for new info from backend and sets props accordingly.
+    refresh: function() {
+      document.getElementById('working').style.display = '';
+
+      jQuery.ajax({
+        url: '/en/grade_entry_forms/1/marks_graders/populate_students_table',
```
david-yz-liu: Won't this only work for the grade_entry_form with id 1?

spring-ryanc: You're right, I had it hard coded for testing, I'll fix that ASAP

app/views/marks_graders/index.html.erb
```diff
@@ -34,14 +32,14 @@
   </div>
 </div>
 
-<div class='wrapper'>
-  <%= render 'shared/flash_message' %>
-</div>
+<div class="wrapLeft"></div>
```
david-yz-liu: Use single quotes

config/database.yml.mysql
```diff
@@ -1,50 +0,0 @@
-# MySQL.  Versions 4.1 and 5.0 are recommended.
```
david-yz-liu: This file... shouldn't have been removed.

david-yz-liu:
@spring-ryanc Nice work. The conflict seems to be coming from divergent versions of the schema, but that's fine - when you do a pull from master, you can run a `rake db:migrate` to generate the file.

PR #2090, opened 2015-03-19
added header for the csv

lifejuggler:
added csv headers to the download section of the assignment


david-yz-liu:
Please make the first column header "Username" rather than "Student Username". Also, if you run the tests locally, you'll see that one functional test now fails.

PR #2091, opened 2015-03-19
764 correct error code

Eitredes:
Lets us serve a redirect page with a 401 code, which then bounces you back to the login.


Eitredes:
Per discussion, we've decided to leave the behaviour as-is (a 302 gets served and auto-redirects you to the login screen).

Issue #99, opened 2010-09-29
Simple Grade Entry: Create unit tests and functional tests for CSV upload/download


benjaminvialle:
Unit tests and functional tests need to be created for uploading/downloading the grades table as a CSV file. (see review request#497)

MatakuT:
made a progress on this issue: #2077

david-yz-liu:
@MatakuT You could also write tests in which correct files are uploaded, and make sure all the marks are there.

MatakuT:
Will work on that

MatakuT:
made a progress: #2092 
There seem to be a conflicting pull request. Does it conflict with the previous fix: #2077?
This conflict does not allow Travis CI to do run the testing.

david-yz-liu:
@MatakuT You should pull from master and resolve the conflict yourself.

MatakuT:
@david-yz-liu I tried the following to find out what is conflicting (I assume it's grade_entry_form_spec.rb file)
Switched to branch 'iss99_additional_tests'
moskalem:Markus alzu$ git merge master
Already up-to-date.
moskalem:Markus alzu$ git merge origin/master
Already up-to-date.

I have never resolved a merge conflict before, so not quite sure what to do next. Found few suggestions to force merge/rebase into remote master, but I don't know git too well and too afraid to screw things up.

david-yz-liu:
You need to be pulling from upstream master, not origin/master. (Origin is you, upstream is MarkUsProject.)

After that, you can consult this: https://help.github.com/articles/resolving-a-merge-conflict-from-the-command-line/

MatakuT:
After spending 4 hours and not being able to resolve the conflict I created a new branch. This one 100% in synch with MarkUsProject master branch. #2092 has been closed and iss99_additional_tests branch was deleted. I still have a local copy just in case.
I realize that it is not a good practice, but I just got into a dead-end and decided to take an easy approach. I learned my lesson with why it was conflicting in the first place and will try to not repeat similar mistake again.
The new PR request can be found here: #2098

david-yz-liu:
@MatakuT It's not bad to have conflicts - it's pretty normal, in fact. You can keep your branch up to date by pulling from MarkUsProject/master every few days.

PR #2092, opened 2015-03-20
added test for the working csv data files

MatakuT:
Continuing working on: 
Simple Grade Entry: Create unit tests and functional tests for CSV upload/download #99


MatakuT:
The following tests are failing:
it does not accept files with wrong column name (FAILED - 1)
it does not accept files with additional columns (FAILED - 2)
it does not accept files with wrong grade total (FAILED - 3)

Will create separate tickets for columns/grade totals (maybe one ticket for all of them?) to handle those cases. Right now upload_csv accepts an improper file and replaces original Simple Grade Spreadsheet with it.

PR #2095, opened 2015-03-20
Using a path helper now.

Eitredes:
Now using a path helper.


david-yz-liu:
@Eitredes One more small thing - could you double-check to see if multiple timeout checks are spawned when you switch between pages and/or controllers? I thought I saw this happen to me, but I'm not sure and don't have time to investigate right now.

Eitredes:
You got it.

PR #2097, opened 2015-03-21
Preview new annotation with math

paymog:
When a marker creates a new annotation with math in it,
a preview of the annotation will be shown to the marker
in the create-an-annotation popup.

This preview will only be shown if there is math content. ie: if there exists something to the effect of ```$$\sum$$```. Specifically, there need to be at least two sets of ```$$```.

### DIFF
--- a/app/assets/javascripts/MathJax/mathjax_helpers.js
+++ b/app/assets/javascripts/MathJax/mathjax_helpers.js
@@ -4,6 +4,7 @@ function reloadDOM() {
     MathJax.Hub.Queue(["Typeset", MathJax.Hub]);
 }
 
+var timeout;
 function updateAnnotationPreview(){
     var updatedAnnotation = document.getElementById("new_annotation_content").value;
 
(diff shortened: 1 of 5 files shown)


david-yz-liu:
The problem with this approach is the MathJaX is re-rendered on every keyup, which doesn't look great.

I'd prefer having two tabs, like "Edit" and "Preview", where the MathJaX is only rendered when the user clicks on the Preview tab.

paymog:
Is there a reason you'd like tabs other than the current rendering issue? The reason I ask is because I first gave tabs a shot but couldn't figure out a UI and on more thought decided that the UX of tabs isn't all that great (takes the user out of the context of writing the annotation). 

I have two different ideas for fixing the rendering issue:
1) Render on a timeout. ie: render the math once the user takes a 3 second pause from typing
2) Have the user press a button to see the preview which gets hidden on typing.

What do you think?

david-yz-liu:
Couple of style things for the Javascript:

1. Please use single-quotes instead of double-quotes
2. Use `!==` instead of `!=`
3. Put spaces after `if`, before `{`, and after `}`

The annotation preview is certainly smoother now, though the fact that the text appears instantly but only renders in MathJaX every 3 seconds is still odd. I think I button would be best. Do you still have time to work on this?

paymog:
I do. I should be able to finish that up tomorrow.

david-yz-liu:
@paymog Yay! Looked great; I made a couple of small style tweaks, which you can checkout in the extra commit. Well done.

PR #2098, opened 2015-03-22
added several tests that use real csv data

MatakuT:


MatakuT:
The following tests are failing:
it does not accept files with wrong column name (FAILED - 1)
it does not accept files with additional columns (FAILED - 2)
it does not accept files with wrong grade total (FAILED - 3)

Will create separate tickets for columns/grade totals (maybe one ticket for all of them?) to handle those cases. Right now upload_csv accepts an improper file and replaces original Simple Grade Spreadsheet with it.

MatakuT:
Opened three tickets to resolve some of the issues with the current file upload:
Simple Grade Spreadsheet: uploaded file overrides column name #2101
Simple Grade Spreadsheet: uploaded file appends extra columns #2102
Simple Grade Spreadsheet: uploaded file overrides grade total #2103

david-yz-liu:
@MatakuT Please remove the three commented out tests, and replace them with tests which check for the actual desired behaviour (overriding existing data).

MatakuT:
I still think that overriding the template via file upload can be dangerous. What if TA by accident uploads wrong file which would pass username check. For example, upload lab spreadsheet into main course spreadsheet (the students taking lab will be also the students from the main course, so the username check would pass).
All the correct data will be replaced with the wrong data. Is there a backup copy of the previously uploaded csv stored on server? Or it's TA's responsibility to maintain up-to-date copies of the spreadsheet on their computers?

david-yz-liu:
Yes, it is dangerous. If you feel strongly about it, you can put in a warning message.

MatakuT:
@david-yz-liu, I can create a message saying "The uploaded file will change a column name/grade total. Do you wish to proceed?" I think it's best to extract the original column name/grade total and the one that will override it to put them in this warning message. So a user knows what triggers this warning. I will take a look at how easily this can be accomplished. If it's not very time consuming, I will proceed.

PR #2100, opened 2015-03-22
Removed unused modal files and converted coverage/criteria dialogs to erb

spring-ryanc:


spring-ryanc:
@david-yz-liu The upload/download dialogs for the graders were coded directly into the UI since they have somewhat static content (the upload shows the criteria when needed). Since they weren't being used anymore I removed the methods from the controller and the supporting files. However it now fails tests that expected the controller methods. I just want to make sure this is okay before I remove the tests. I looked pretty thoroughly for anything that might reference the controller methods and couldn't find anything. The tests appear to mainly be checking to ensure the students dont have access to the forms.

david-yz-liu:
@spring-ryanc That sounds fine.

spring-ryanc:
@david-yz-liu can you re-trigger the build for me? It seems like there was some strange connection issue during the test

PR #2115, opened 2015-03-31
added the warning message on attempted csv template overwrite

MatakuT:


app/controllers/grade_entry_forms_controller.rb
```diff
@@ -440,7 +479,29 @@ def csv_upload
...
     end
-    redirect_to action: 'grades', id: @grade_entry_form.id
+    if !matching_names && params[:id].to_s != @grade_entry_form.id.to_s
+      redirect_to csv_overwrite_grade_entry_form_path(params[:id], with: @grade_entry_form.id)
+    else
+      redirect_to action: 'grades', id: @grade_entry_form.id
+    end
+  end
+
+  def csv_overwrite
```
MatakuT: Does it complain that the function is too complex? Is there any need to simplify it/suggestions?

MatakuT:
There are failing tests that rely on previous csv_upload functionality when a newly uploaded file simply overwrites the old one. If my proposed warning message implementation will be accepted - those tests should be re-written into functional tests adapting the new behaviour. I think it would make sense to re-write them in Rspec vs modifying the existing rake tests. The old tests will be removed.

MatakuT:
@david-yz-liu , please let me know if you approve my solution, so I can finish up fixing the failing tests

app/controllers/grade_entry_forms_controller.rb
david-yz-liu: It's fine

david-yz-liu:
Okay, seems fine to me!

david-yz-liu:
I took an existing spreadsheet, downloaded it, changed one column name, reuploaded. Got the overwrite warning - good.

When I clicked "OK", the spreadsheet had the new column name, but no marks had been entered (all blank). Weird.

Then I tried to upload the same CSV, and got the overwrite warning again, even though the headers and totals seemed to be fine.

This seems like too much effort; I would go with the original approach of just displaying a warning message in the upload modal.

PR #2117, opened 2015-04-02
900 single member group display

Eitredes:
Solves 900. Removed the code to add lists of members when group names are shown in the results view and the submissions table.


david-yz-liu:
@Eitredes Thanks, I think you may have misunderstood me, but I used your first commit to great success.

PR #2120, opened 2015-04-06
Catch exceptions when populating submissions table

oneohtrix:
Before, when any submission got into an invalid state, the helper
function gathering all the table data would fail. This caused the table
to not load at all. Now the error is caught, and an error boolean is
placed in the returned object, which the front-end then checks and
displays an errored submission (a light red row) in the table.
Population of other submissions should continue unaffected.


app/helpers/submissions_helper.rb
```diff
@@ -42,12 +41,20 @@ def get_submissions_table_info(assignment, groupings)
...
+        g[:class_name] = get_any_tr_attributes(grouping)
+        g[:repository] = get_grouping_repository(assignment, grouping)
+        g[:marking_state] = get_grouping_marking_state(assignment, grouping)
+        g[:grace_credits_used] = get_grouping_grace_credits_used(grouping)
+        g[:final_grade] = get_grouping_final_grades(grouping)
+        g[:state] = get_grouping_state(grouping)
+        g[:section] = get_grouping_section(grouping)
+        g[:tags] = get_grouping_tags(grouping)
+      rescue
+        g[:error] = true
```
david-yz-liu: At the very least we should log the grouping and the error that was raised. (MarkUsLogger)

app/helpers/submissions_helper.rb
```diff
@@ -31,10 +31,9 @@ def set_release_on_results(groupings, release)
 
   def get_submissions_table_info(assignment, groupings)
     groupings.map do |grouping|
+
       g = grouping.attributes
-      g[:class_name] = get_any_tr_attributes(grouping)
       g[:group_name] = get_grouping_group_name(assignment, grouping)
-      g[:repository] = get_grouping_repository(assignment, grouping)
       begin
```
david-yz-liu: Fold these into your error handling below

oneohtrix:
@david-yz-liu Fixed

app/helpers/submissions_helper.rb
```diff
@@ -42,17 +39,30 @@ def get_submissions_table_info(assignment, groupings)
...
+        g[:group_name] = get_grouping_group_name(assignment, grouping)
+        g[:class_name] = get_any_tr_attributes(grouping)
+        g[:repository] = get_grouping_repository(assignment, grouping)
+        g[:marking_state] = get_grouping_marking_state(assignment, grouping)
+        g[:grace_credits_used] = get_grouping_grace_credits_used(grouping)
+        g[:final_grade] = get_grouping_final_grades(grouping)
+        g[:state] = get_grouping_state(grouping)
+        g[:section] = get_grouping_section(grouping)
+        g[:tags] = get_grouping_tags(grouping)
+      rescue
```
david-yz-liu: I mean that up here, we should be storing the error and logging its message (or some other info).

PR #2124, opened 2015-04-14
Added support for different js event handling

spring-ryanc:
Chrome uses the event.srcElement while firefox uses event.currentTarget to get the element that triggered the event. By checking which was present the hover over issue is handled.


spring-ryanc:
Last pull request of UCOSP and no hound or Travis messages :D

david-yz-liu:
Thanks, @spring-ryanc!

PR #2127, opened 2015-04-20
Added test for issue 1199

MatakuT:


MatakuT:
The reason why I changed the lib file is that the test was failing when the following line in memory_repository.rb (line 449):
if path.nil? || (!path.nil? && @timestamps_revisions[Marshal.dump(best_match)].revision_at_path(path))

ERROR: revision_at_path was called on nil

Marshal.dump(best_match) was producing the right result, but its format was slightly different from the format of the actual hash key. I think that's due to Marshal not completely interpreting Time class correctly. I was trying to find a way to convert Marshal.dump(best_match) to match the key, but wasn't able to find the way to do so. I ended up converting the actual key. Calling dump function on key did the trick. I'm not sure if that is going to break any other functionality - it seems to be working just fine for assignment class

david-yz-liu:
Thanks, @MatakuT. I'm really not sold on this forking anyway, so hopefully we'll be able to resolve this in a more robust way later.

PR #2132, opened 2015-04-24
Fix for issue 2104: Use match_array to make sure tests pass even if data is returned in a different order

akbiggs:


david-yz-liu:
Thanks, @akbiggs!

PR #2134, opened 2015-04-28
Added translations for Key Pairs pages

SoftwareDev:
Translations were completed using Google Translate.


SoftwareDev:
@david-yz-liu I'll deal with the failing rspec tests in another PR... They will require some reworking because gitolite is being used now.

PR #2147, opened 2015-05-08
Use Autoprefixer for implicit CSS vendor prefixes

arkon:
Instead of using custom SASS mixins to handle common vendor prefixes, just use [autoprefixer-rails](https://github.com/ai/autoprefixer-rails/) to deal with it.

Also added a config file for the SASS linting.


david-yz-liu:
Thanks, @arkon!

david-yz-liu:
@arkon Can you check out the student file manager? The buttons look quite grey, and I don't think that was the intended effect. I worry that the `linear-gradient` isn't working.

PR #2148, opened 2015-05-09
rewrote Assignment#latest_due_date

kubasub:
I rewrote `Assignment#latest_due_date` in order to:
- handle edge case upfront
- remove conditional branching
- remove "redundant `self`", as @houndci puts it

I ran the `Assignment` spec and everything there still passes.


david-yz-liu:
Thanks, @kubasub!

kubasub:
My pleasure, really!

PR #2152, opened 2015-05-11
File manager: various syntax and style fixes, and multiple file uploads

arkon:


arkon:
Things that still need to be fixed:

- Column with checkboxes should be hidden if the assignment is to be interacted with via VCS
- Filenames should link to the files like before
- Last revised datetimes should be nicely formatted

david-yz-liu:
Thanks, @arkon!

Your TODO list is on point. I'll make an issue after I merge the branch into master.

arkon:
@david-yz-liu Another issue is that after selecting a file and deleting it, the table refreshes and a different row is selected.

Issue #1834, opened 2014-10-15
Display the absolute final due date (after late periods) to students?


david-yz-liu:
Even though I think I communicated the absolute deadline quite well online and in lecture, perhaps it would be good to have it displayed in MarkUs as well so that there is no confusion.

What I mean by "absolute deadline" is the time after which no submission has a chance of being collected, regardless of the late penalties applied.

@reidka

david-yz-liu:
@tsimica This one is related to the work you've done in #2024, and would be a good next step.

PR #2153, opened 2015-05-12
Issue #1834 - Display the absolute final due date

tsimica:
#1834


tsimica:
Does "absolute due date" or "absolute deadline" make more sense?

david-yz-liu:
@tsimica can you please look into the failing test?

reidka:
I prefer "absolute deadline"

reidka:
Deadline including maximum late penalty
Deadline using all remaining grace tokens

tsimica:
I also added a check to not display "Additional Information" if the grouping was invalid because it didn't make sense to have a field with no value.

david-yz-liu:
@tsimica there should be different messages for whether a late penalty or grace periods are being used.

Issue #2155, opened 2015-05-13
Student file manager issues


david-yz-liu:
@arkon has nicely pointed out stylistic issues in the new student file manager:
- [x] Column with checkboxes should be hidden if the assignment disallows web submits
- [x] Filenames should link to the files (like the old file manager)
- [x] Last revised datetimes should be nicely formatted (this is a simple internationalization fix)
- [x] Searchbox could use some placeholder text
- [x] File table doesn't appear when web submits are not allowed.

arkon:
> Filenames should link to the files (like the old file manager)

Somebody just brought this up to me again today. Seems like it would be a nice feature to bring back.

arkon:
I think the last item ("File table doesn't appear when web submits are not allowed.") is okay now.

david-yz-liu:
Yes, thanks for the reminder.

PR #2157, opened 2015-05-13
Student file manager fixes

arkon:
Deals with 3 of the issues in https://github.com/MarkUsProject/Markus/issues/2155:

- Column with checkboxes should be hidden if the assignment disallows web submits
- Last revised datetimes should be nicely formatted (this is a simple internationalization fix)
- Searchbox could use some placeholder text


david-yz-liu:
Thanks, @arkon! Looks like there's a problem rendering the table when web submits aren't allowed. Oops. If you have time to take a look, that would be great, if not then I will in a few days.

Issue #1825, opened 2014-10-10
Students wanting to delete a valid group need to know to contact instructor


reidka:
Once students form a valid group, there is no longer a button to delete the group.  There should be a note telling them that if they want to delete the group, they need to contact the instructor to change group membership or remove the group.

david-yz-liu:
@tsimica This should be included in the group properties panel.

PR #2163, opened 2015-05-13
Changed functionality of group delete button

tsimica:
#1825


david-yz-liu:
@tsimica There shouldn't be a button at all; rather, just some text inside the group properties panel.

PR #2168, opened 2015-05-21
Issue 1766

hootiehoo:
When uploading a csv, the auth file now only gets rewritten once at the end, rather than once per group. This is done using the new method __generate_authz_file in the SubversionRepository class.

These changes speed up the process considerably. Before, uploading 600 names with allow_web_submits=true took 1:35 seconds. It now takes 35 seconds. And where uploading 600 names with allow_web_submits=false took 3 minutes, now takes 1.

I also added the names in students.csv into longstudents.csv, so students.csv is a proper subset of longstudents.csv. This should prevent future bugs when testing with longstudents.csv.


app/models/assignment.rb
```diff
@@ -313,7 +314,6 @@ def add_group(new_group_name=nil)
         group = Group.create(group_name: new_group_name)
```
david-yz-liu: Should there be a call to `set_repo_permissions` here too?

david-yz-liu:
@hootiehoo You can replace Karen's old longstudents file with yours. Also, please change the seeds file back to the original 216-student list (though you can leave the new one as a comment to enable easy toggling).

PR #2174, opened 2015-06-01
Issue 2129 and spreadsheet order

Lysette:
Spreadsheets are now ordered (in dashboard, assignment dropdown and assignment page) in order they were created in. Date is now optional for spreadsheets.


david-yz-liu:
Rather than commenting out the validation, I think we can make the following two changes:

1) In the controller, check the `date` parameter. If it's the empty string, change it to `nil`.
2) In the date validator (`lib/validations`), add an extra check to see if the value is nil, and if so, don't add any errors.

app/models/grade_entry_form.rb
```diff
@@ -17,7 +17,7 @@ class GradeEntryForm < ActiveRecord::Base
   # Call custom validator in order to validate the date attribute
   # date: true maps to DateValidator (custom_name: true maps to CustomNameValidator)
   # Look in lib/validators/* for more info
-  validates                 :date, date: true
+  # validates                 :date, date: true
```
david-yz-liu: Uncomment this!

lib/validators/date_validator.rb
```diff
@@ -2,7 +2,7 @@
 class DateValidator < ActiveModel::EachValidator
 
   def validate_each(record, attribute, value)
-    if Time.zone.parse(value.to_s).nil?
+    if Time.zone.parse(value.to_s).nil? && !value.nil?
```
david-yz-liu: Unfortunately, this doesn't pass all of the tests, as the value is *automatically* converted to `nil` when there is a date in an invalid format. This was quite surprising to me - but we learn something new every day!

Replace the second check with `!record.date_before_time_cast.blank?` to check whether the date string before trying to convert it was empty or not.

david-yz-liu:
@Lysette Interesting, looks like assignments also use the same validator, but it has a different attribute name (`due_date` rather than `date`).

To fix this, we shouldn't hard code `date` into the attribute access `date_before_type_cast`, but instead probably use what you'd looked up, `read_attribute_before_type_cast`, passing it the name of the attribute (which hopefully is what the `attribute` parameter is in the validator function).

david-yz-liu:
Thanks, @Lysette!

PR #2181, opened 2015-06-11
Download all submissions warning changed

Lysette:
"Note: Not all of the assignments have been marked. Are you sure you want to proceed?" has been changed to "Note: This will only download the submissions that have been collected. Are you sure you want to proceed?"


david-yz-liu:
Thanks, @Lysette!

PR #2186, opened 2015-06-25
changed date format and added date placeholder

Lysette:
The date format is now yyyy-mm-dd for English and yyyy-dd-mm for French and Portuguese. There is also a date placeholder so users know what format to put the date in.


david-yz-liu:
Thanks, @Lysette!

PR #2187, opened 2015-06-26
Added instruction page links for Key Pairs

SoftwareDev:
Once https://github.com/MarkUsProject/Wiki/pull/73 is merged then the links on the Key Pairs page should work.

When clicked they open a new window to not interrupt the user's MarkUs session.


david-yz-liu:
@SoftwareDev the links should point to something like `MarkUsProject/Markus/wiki/...` - the links don't exist yet, but they will once @reidka updates the wiki.

PR #2202, opened 2015-07-23
Issue 2173

Lysette:
Shows which submission is collected in repo browser.


app/views/submissions/repo_browser.html.erb
```diff
@@ -62,6 +62,9 @@
         (<span id='current_revision_timestamp_display'>
            <%= l(@revision_timestamp, format: :long_date) %>
         </span>)
+        <br>
+        <%= raw(t('browse_submissions.currently_collected')) %>
```
david-yz-liu: You shouldn't need to use `raw` here

app/views/submissions/repo_browser.html.erb
```diff
@@ -62,6 +62,9 @@
         (<span id='current_revision_timestamp_display'>
            <%= l(@revision_timestamp, format: :long_date) %>
         </span>)
+        <br>
+        <%= raw(t('browse_submissions.currently_collected')) %> 
+        <%= @grouping.is_collected ? "#" + raw(@grouping.submissions.last.revision_number) + " ( " + l((@revisions_history.select {|rev| rev[:num] == @grouping.submissions.last.revision_number }).first[:date], format: :long_date) + " )" : "none" %>
```
david-yz-liu: This line is too long and needs to be split up.

app/views/submissions/repo_browser.html.erb
```diff
@@ -62,6 +62,11 @@
         (<span id='current_revision_timestamp_display'>
            <%= l(@revision_timestamp, format: :long_date) %>
         </span>)
+        <br>
+        <%= t('browse_submissions.currently_collected') %> 
+        <%= @grouping.is_collected ? "#" + raw(@grouping.submissions.last.revision_number) +
```
david-yz-liu: Hmm okay, after looking at this more closely, I have three comments:

1. I'd rather use an `if` here than the ternary operator. 
    ```
      <% if @grouping.is_collected%>
        <% ="#" + @grouping.submissions.last.revision_nubmer %>
      ...
    ```
2. Probably don't need raw for the revision number either
3. I *think* that you can do something like `@grouping.submissions.last.revision_timestamp` to get the timestamp for the submission. I was just looking at that today.

Lysette: <code>@grouping.submissions.last.revision_timestamp</code> returns incorrect time which is why I had to use <code>@revisions_history</code>.

PR #2209, opened 2015-07-27
Updating marks through API

hootiehoo:
Allows TAs/Admins to update marks via api.
 
Valid requests are of the form:
curl -H "Authorization: MarkUsAuth auth_key" -X PUT "http://example.com/api/assignments/assignment_id/groups/group_id/update_marks?criterion1=4&criterion2=3"
where criterionX is a properly formatted title of the desired criterion. Eg if the criterion is "Aspernatur dolorem.", we would put ?Aspernatur%20dolorem.=4 .


app/controllers/api/groups_controller.rb
```diff
@@ -80,5 +80,54 @@ def include_students(fields)
       end
     end
 
+    # Update the group's marks for the given assignment.
+    def update_marks
+      assignment = Assignment.find_by_id(params[:assignment_id])
```
david-yz-liu: Just use `find`

app/controllers/api/groups_controller.rb
```diff
@@ -80,5 +80,54 @@ def include_students(fields)
...
+          'No group exists with that id' }, status: 404
+        return
+      elsif group.submissions.first.nil?
+        render 'shared/http_status', locals: { code: '404', message:
+          'No submissions exist for that group' }, status: 404
+        return
+      end
+
+      all_criteria = assignment.rubric_criteria
+      all_criteria.push(*assignment.flexible_criteria)
```
david-yz-liu: I think you want `concat`, not `push`

app/controllers/api/groups_controller.rb
```diff
@@ -80,5 +80,54 @@ def include_students(fields)
...
+          'No assignment exists with that id' }, status: 404
+        return
+      end
+
+      group = Group.find_by_id(params[:id])
+      if group.nil?
+        render 'shared/http_status', locals: { code: '404', message:
+          'No group exists with that id' }, status: 404
+        return
+      elsif group.submissions.first.nil?
```
david-yz-liu: We really want to be using the `grouping` properties `has_submission?` and `current_submission_used` here. This code may be incorrect when there are multiple submissions.

app/controllers/api/groups_controller.rb
```diff
@@ -80,5 +80,54 @@ def include_students(fields)
...
+      all_criteria = assignment.rubric_criteria
+      all_criteria.push(*assignment.flexible_criteria)
+      if all_criteria.empty?
+        render 'shared/http_status', locals: { code: '404', message:
+          'No criteria exist for that assignment' }, status: 404
+        return
+      end
+
+      result = group.submissions.first.get_latest_result
+      params.keys.each do |param_key|
```
david-yz-liu: This loop structure is unnecessary.

Set up a query like `RubricCriterion.where(assignment: assignment, rubric_criterion_name: params[:name])` to do the search. (Need another one for FlexibleCriterion too.)

app/controllers/api/groups_controller.rb
```diff
@@ -80,5 +80,62 @@ def include_students(fields)
...
+      end
+
+      group = Group.find(params[:id])
+      if group.nil?
+        render 'shared/http_status', locals: { code: '404', message:
+          'No group exists with that id' }, status: 404
+        return
+      end
+      if group.groupings.first.has_submission?
+        result = group.groupings
```
david-yz-liu: Use `group.get_grouping_for`

app/controllers/api/groups_controller.rb
```diff
@@ -80,5 +80,62 @@ def include_students(fields)
...
+        mark_to_change = result.marks
+                               .where(markable_id: crit.id)
+                               .first
+        if crit.is_a?(FlexibleCriterion)
+          mark_to_change.mark = params[crit.flexible_criterion_name].to_f
+        else
+          mark_to_change.mark = params[crit.rubric_criterion_name].to_i
+        end
+        unless mark_to_change.save
+          # Some error occurred
```
david-yz-liu: Don't put this check in.

david-yz-liu:
@hootiehoo the function is okay, but one improvement would be taking the actual mark updating part (*after* the database has been queried) and putting it into a separate function.

app/controllers/api/groups_controller.rb
```diff
@@ -80,5 +80,59 @@ def include_students(fields)
...
+                      .current_submission_used
+                      .get_latest_result
+      else
+        render 'shared/http_status', locals: { code: '404', message:
+          'No submissions exist for that group' }, status: 404
+        return
+      end
+
+      matched_criteria = RubricCriterion
+                           .where(rubric_criterion_name: params.keys)
```
david-yz-liu: I think I meant to make a comment here and then didn't. This isn't searching criteria by a particular assignment, which isn't good if two assignments have a criterion with the same name.

app/controllers/api/groups_controller.rb
```diff
@@ -73,6 +73,15 @@ def show
       end
     end
 
+
+    def get_group_by_name
+    end
+
+
```
hootiehoo: Oops, these shouldn't be here.

david-yz-liu:
Thanks, @hootiehoo!

PR #2212, opened 2015-07-29
Improved checking validity of marks submitted via API.

hootiehoo:
Added some more value checks. 
- Trying to submit any non-whole number (within the valid range) to an assignment with a rubric will now display an error message, instead of failing silently. 
- Can no longer submit marks to an submission whose marking state is complete.
- Fixed odd off-by-one error in mark callback.

Thanks

### DIFF
--- a/app/models/mark.rb
+++ b/app/models/mark.rb
@@ -22,7 +22,7 @@ class Mark < ActiveRecord::Base
   belongs_to :markable, polymorphic: true
   validates_numericality_of :markable_id,
                             only_integer: true,
-                            greater_than: 0,
+                            greater_than_or_equal_to: 0,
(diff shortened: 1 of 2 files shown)


app/models/mark.rb
```diff
@@ -31,7 +31,8 @@ class Mark < ActiveRecord::Base
   # Custom validator for checking the upper range of a mark
   def valid_mark
     unless mark.nil?
-      if markable_type == 'RubricCriterion' && mark > 4
+      if markable_type == 'RubricCriterion' &&
+        (mark > 4 || (mark % 1 != 0))
```
hootiehoo: This was the easiest way I could think to ensure the mark was a whole number.

david-yz-liu: Use `is_a?`

PR #2214, opened 2015-08-02
Implement summaries csv pat

PatGuan:
Changed the location of the download CSV report to the summaries (Danny).
Refactored the function download_csv_report. Spreadsheet now displays the total marks in the course with respect to their criterion weights.

### DIFF
--- a/app/controllers/course_summaries_controller.rb
+++ b/app/controllers/course_summaries_controller.rb
@@ -24,10 +24,12 @@ class CourseSummariesController < ApplicationController
       assignments.each do |assignment|
         header.push(assignment.short_identifier)
       end
+      header.push("Total marks")
       csv << header
       students.each do |student|
         row = []
         row.push(student.user_name)
+        total_spreadsheet_mark = 0
         assignments.each do |assignment|
           out_of = assignment.total_mark
           grouping = student.accepted_grouping_for(assignment.id)
@@ -43,10 +45,12 @@ class CourseSummariesController < ApplicationController
                 row.push('')
               else
                 row.push(total_mark_percentage)
+                total_spreadsheet_mark += total_mark_percentage
               end
             end
           end
         end
+        row.push(total_spreadsheet_mark)
         csv << row
       end
     end
--- a/app/controllers/course_summaries_controller.rb
+++ b/app/controllers/course_summaries_controller.rb
@@ -15,4 +15,44 @@ class CourseSummariesController < ApplicationController
   def get_marking_scheme_details
     redirect_to url_for(controller: 'marking_schemes', action: 'populate')
   end
+
+  def download_csv_grades_report
+    assignments = Assignment.order(:id)
+    students = Student.all
+    csv_string = CSV.generate do |csv|
+      header = ['Username']
+      assignments.each do |assignment|
+        header.push(assignment.short_identifier)
+      end
+      csv << header
+      students.each do |student|
+        row = []
+        row.push(student.user_name)
+        assignments.each do |assignment|
+          out_of = assignment.total_mark
+          grouping = student.accepted_grouping_for(assignment.id)
+          if grouping.nil?
+            row.push('')
+          else
+            submission = grouping.current_submission_used
+            if submission.nil?
+              row.push('')
+            else
+              total_mark_percentage = submission.get_latest_result.total_mark / out_of * 100
+              if total_mark_percentage.nan?
+                row.push('')
+              else
+                row.push(total_mark_percentage)
+              end
+            end
+          end
+        end
+        csv << row
+      end
+    end
+    course_name = "#{COURSE_NAME}"
+    course_name_underscore = course_name.squish.downcase.tr(" ", "_")
+    send_data csv_string, disposition: 'attachment',
+              filename: "#{course_name_underscore}_grades_report.csv"
+  end
 end
--- a/app/controllers/course_summaries_controller.rb
+++ b/app/controllers/course_summaries_controller.rb
@@ -19,19 +19,19 @@ class CourseSummariesController < ApplicationController
   def download_csv_grades_report
     assignments = Assignment.order(:id)
     csv_string = CSV.generate do |csv|
-      #Populates the entire CSV with headers, students and their marks
+      # Populates the entire CSV with headers, students and their marks
       populated_csv = populate_students_and_marks(csv)
     end
     course_name = "#{COURSE_NAME}"
     course_name_underscore = course_name.squish.downcase.tr(" ", "_")
     send_data csv_string, disposition: 'attachment',
-              filename: "#{course_name_underscore}_grades_report.csv"
+                          filename: "#{course_name_underscore}_grades_report.csv"
   end
 end
 
-#---- HELPER FUNCTIONS FOR DOWNLOAD_CSV_GRADES_REPORT BELOW-----#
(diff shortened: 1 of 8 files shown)


david-yz-liu:
One general comment is that because the routes have changed, you'll likely need to update some tests. Do a search through the `test/` and `spec/` directories to update the tests. You can use `bundle exec rake test` and `bundle exec rspec` to run the tests locally to make sure they all pass.

app/controllers/course_summaries_controller.rb
```diff
@@ -15,4 +15,116 @@ def populate
...
     redirect_to url_for(controller: 'marking_schemes', action: 'populate')
   end
+
+  def download_csv_grades_report
+    csv_string = CSV.generate do |csv|
+      # Populates the entire CSV with headers, students and their marks
+      populate_students_and_marks(csv)
+    end
+    course_name = "#{COURSE_NAME}"
+    course_name_underscore = course_name.squish.downcase.tr(' ', '_')
```
david-yz-liu: Please make this functionality (squish, downcase, tr) a helper function.

app/controllers/course_summaries_controller.rb
```diff
@@ -15,4 +15,116 @@ def populate
...
+                          filename: "#{course_name_underscore}_grades_report.csv"
+  end
+end
+
+# ---- HELPER FUNCTIONS FOR DOWNLOAD_CSV_GRADES_REPORT BELOW-----#
+
+# Populates the headers in the CSV
+def populate_header_titles(assignments)
+  header = ['Username']
+  assignments.each do |assignment|
```
david-yz-liu: Use `map` rather than `each` if you're building a list.

app/controllers/course_summaries_controller.rb
```diff
@@ -15,4 +15,116 @@ def populate
...
+    header.push(assignment.short_identifier)
+  end
+  header.push('Total marks')
+end
+
+# Checks grouping, submission and it's mark for nil.
```
david-yz-liu: Wrong "its"

app/controllers/course_summaries_controller.rb
```diff
@@ -15,4 +15,116 @@ def populate
...
+  end
+  row
+end
+
+# Loads up the CSV with headers, students and marks.
+def populate_students_and_marks(csv)
+  students = Student.all
+  # Populates the csv header titles here
+  csv << populate_header_titles(Assignment.order(:id))
+  students.each do |student|
```
david-yz-liu: Again, use `map` instead of `each`

app/controllers/course_summaries_controller.rb
```diff
@@ -15,4 +15,116 @@ def populate
...
+def populate_header_titles(assignments)
+  header = ['Username']
+  assignments.each do |assignment|
+    header.push(assignment.short_identifier)
+  end
+  header.push('Total marks')
+end
+
+# Checks grouping, submission and it's mark for nil.
+def fill_row_with_marks(grouping, row)
```
david-yz-liu: Couple of comments:

1. This should either mutate `row` or return a new row, but not both.
2. The name is misleading, given that only one mark is ever pushed.

I suggest making this helper simply return either `''` or the mark, and actually pushing the mark in line 65.

app/controllers/course_summaries_controller.rb
```diff
@@ -15,4 +15,116 @@ def populate
...
+    row.push(total_mark)
+    csv << row
+  end
+  csv
+end
+
+# Return the total marks for that student with its appropriate weights
+def get_total_marks_for_student(grouping)
+  current_total_marks = 0
+  if not check_for_nil_fields(grouping)
```
david-yz-liu: Unnecessary mutation here. Structure the code to something like

```
if check_for_nil_fields(grouping)
  0
else
  ...
end
```

app/controllers/course_summaries_controller.rb
```diff
@@ -15,4 +15,116 @@ def populate
...
+      row = fill_row_with_marks(grouping, row)
+      total_mark += get_total_marks_for_student(grouping)
+    end
+    row.push(total_mark)
+    csv << row
+  end
+  csv
+end
+
+# Return the total marks for that student with its appropriate weights
```
david-yz-liu: Again, the pluralization is confusing given that it seems only one mark is returned.

app/controllers/course_summaries_controller.rb
```diff
@@ -15,4 +15,116 @@ def populate
...
+def get_total_marks_for_student(grouping)
+  current_total_marks = 0
+  if not check_for_nil_fields(grouping)
+    # Create Hashmap with Criteria ID as keys and it's weight as values
+    criteria_id_to_weight = return_criteria_to_weight_for_grouping(grouping)
+    get_assignment_marks(grouping).each do |mark|
```
david-yz-liu: I'd suggest using the `sum` function here.

app/controllers/course_summaries_controller.rb
```diff
@@ -15,4 +15,116 @@ def populate
...
+      current_total_marks += mark.mark *
+                              criteria_id_to_weight[mark.markable_id].to_f
+    end
+  end
+  current_total_marks
+end
+
+# Function that is called to check for nill grouping, submissions and marks
+def check_for_nil_fields(grouping)
+  if grouping.nil? || grouping.current_submission_used.nil? ||
```
david-yz-liu: This is redundant; just use the condition as the return expression.

Also, the docstring has a typo.

app/controllers/course_summaries_controller.rb
```diff
@@ -15,4 +15,116 @@ def populate
...
+    (grouping.current_submission_used.get_latest_result.total_mark /
+                          grouping.assignment.total_mark * 100).nan?
+    return true
+  else
+    return false
+  end
+end
+
+# Returns the appropriate criteria for a group assignment
+def return_criteria_to_weight_for_grouping(grouping)
```
david-yz-liu: Don't use "return" in the name of a method.

app/controllers/course_summaries_controller.rb
```diff
@@ -15,4 +15,116 @@ def populate
...
+  criteria_id_to_weight = {}
```
david-yz-liu: Unnecessary variable. In the `if` block, just have the `map_criteria_id_to_weight` function calls - they'll be returned automatically because they're the last expression in the block.

app/controllers/course_summaries_controller.rb
```diff
@@ -15,4 +15,116 @@ def populate
...
+  elsif grouping.assignment.marking_scheme_type == 'flexible'
+    criteria_id_to_weight = map_criteria_id_to_weight('flexible',
+                                                      grouping.assignment.flexible_criteria)
+  end
+  criteria_id_to_weight
+end
+
+# Returns the marks that the group got on the assignment
+def get_assignment_marks(grouping)
+  grouping.current_submission_used.results.first.marks
```
david-yz-liu: Don't use `.results.first` - there should be a `get_latest_result` or something you can use.

app/controllers/course_summaries_controller.rb
```diff
@@ -15,4 +15,116 @@ def populate
...
+end
+
+# Load a hashmap with the criteria id to its weight according to its criterion
+def map_criteria_id_to_weight(type, criterias)
```
david-yz-liu: Use `criteria`, not `criterias`

PR #2218, opened 2015-08-10
API: Setting marks to nil, and group_ids_by_name function.

hootiehoo:
- When updating marks, can now set a mark to nil (unmarked). 

Example: curl --header "Authorization: MarkUsAuth YourAuthKey" -X PUT \
    --data "Lorem%20Ipsum=nil" \
    "http://example.com/api/assignments/1/groups/5/update_marks"

- Added api function 'group_ids_by_name' that returns json/xml consisting of key:value pairs 
  of group_name:group_id for all groups associated with a given assignment.

Example: curl -H "Authorization: MarkUsAuth YourAuthKey" "http://example.com/api/assignments/1/groups/group_ids_by_name"
will return
\<?xml version="1.0" encoding="UTF-8"?>
\<groups>
  \<group_name1>31\</group_name1>
  \<group_name2>32\</group_name2>
  ...
\</groups>

### DIFF
--- a/app/controllers/api/groups_controller.rb
+++ b/app/controllers/api/groups_controller.rb
@@ -166,9 +166,12 @@ module Api
                              dict
                            end
       respond_to do |format|
-        format.xml { render xml: reversed.to_xml(root: 'groups',
-                                                 skip_types: 'true') }
-        format.json { render json: reversed.to_json }
+        format.xml do
+          render xml: reversed.to_xml(root: 'groups', skip_types: 'true')
+        end
+        format.json do
+          render json: reversed.to_json
+        end
       end
     end
   end # end GroupsController
--- a/app/controllers/api/groups_controller.rb
+++ b/app/controllers/api/groups_controller.rb
@@ -142,11 +142,33 @@ module Api
 
     def set_mark_by_criteria(criteria, mark_to_change)
       if criteria.is_a?(FlexibleCriterion)
-        mark_to_change.mark = params[criteria.flexible_criterion_name].to_f
+        if params[criteria.flexible_criterion_name] == "nil"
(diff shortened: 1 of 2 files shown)


app/controllers/api/groups_controller.rb
```diff
@@ -142,11 +142,34 @@ def update_marks
 
     def set_mark_by_criteria(criteria, mark_to_change)
```
hootiehoo: I don't like my use of branching if statements here, but I can't see a simpler way.

david-yz-liu: It's okay, it's really an instance of the flexible/rubric criterion inheritance not being implemented in a nice way.

app/controllers/api/groups_controller.rb
```diff
@@ -142,11 +142,37 @@ def update_marks
...
+        else
+          mark_to_change.mark = params[criteria.rubric_criterion_name]
+        end
       end
       mark_to_change.save
     end
+
+    # Return key:value pairs of group_name:group_id
+    def group_ids_by_name
+      reversed = Assignment.find_by_id(params[:assignment_id])
```
david-yz-liu: Just use `find`

app/controllers/api/groups_controller.rb
```diff
@@ -142,11 +142,37 @@ def update_marks
...
+                           .groups
+                           .inject({}) do |dict, group|
```
david-yz-liu: This is okay (fold!) but I'd rather you use the `Hash` constructor, which takes a nested list `[ [key, val], ... ]` and returns a Hash.

david-yz-liu:
Great, thanks @hootiehoo!

PR #2219, opened 2015-08-10
Student spreadsheet date error

Lysette:
Fixed students home page error from a spreadsheet not having a date.


app/views/grade_entry_forms/_list.html.erb
```diff
@@ -52,7 +56,8 @@
                   <% else %>
                     <strong> <%= t('grade_entry_forms.students.detailed_marks_message') %> </strong>
                   <% end %>
-                <% elsif grade_entry_form.date < Time.zone.now.to_date %>
+                <% elsif grade_entry_form.date.nil? or
```
david-yz-liu: In general, use `||` instead of `or` (the precedence rules are more intuitive).

PR #2224, opened 2015-08-18
Flash message

Lysette:
Replaced uses of handle_error and generic_error with flash_message. Flash messages show up in response header.


david-yz-liu:
Thanks, @Lysette! (I fixed the Hound comments.)

Issue #2081, opened 2015-03-09
download with annotations broken


DustinWehr:
I don't see the annotations. This is for downloading a single group's submission, from the marking page.
I tried download in Firefox and Chrome. I tried viewing the downloaded PDF in Adobe Reader and OSX's Preview.

PR #2243, opened 2015-09-18
#2081 - added warning that pdfs don't include annotations

jeremygoh:
A temporary measure for #2081 

This PR informs the user that annotations for PDFs are not currently downloaded via a trigger on checkbox checked.


app/assets/javascripts/Results/file_selector.js
```diff
@@ -3,3 +3,10 @@ function with_annotations() {
   document.getElementById('download_include_annotations').value =
         document.getElementById('include_annotations').checked ? 'true' : 'false';
 }
+
+function annotations_included() {
+
+  if (document.getElementById('include_annotations').checked) {
+    alert("Warning! Annotations on PDFs are not currently downloaded.");
```
david-yz-liu: @jeremygoh This string should be localized. This means you'll want to put this function inside an ERB file like `app/views/results/marker/_boot.js.erb` and edit the `config/locales/*.yml` files

PR #2244, opened 2015-09-18
Issue 2235

### LINKED ISSUE #2235: Cap number of error messages in CSV upload
Some people have run into a "cookie overflow" issue when there's too many invalid lines in a CSV upload.

melneubert:
Closes #2235 

Also fixed error when uploading duplicate users in the csv file.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -161,7 +161,7 @@ class User < ActiveRecord::Base
           next if CSV.generate_line(row).strip.empty?
           if processed_users.include?(row[0])
             if result[:invalid_lines].count < max_invalid_lines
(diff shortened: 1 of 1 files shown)


app/models/user.rb
```diff
@@ -145,6 +145,7 @@ def self.generate_csv_list(user_list)
   end
 
   def self.upload_user_list(user_class, user_list, encoding)
+    max_invalid_lines = 10
```
WilsonChiang: Wondering if we should hard code this, or if it should be configurable somewhere? Not sure how it all works.

david-yz-liu:
Note: the same problem happens in many places, so the issue is not yet closed.

Might be good to refactor the CSV code and make it more general.

PR #2248, opened 2015-09-18
close #2141 Replace input with p.

### LINKED ISSUE #2141: Spreadsheets total column should not appear editable
Really, there shouldn't even be an `input` element in there at all.

wkwan:


app/views/grade_entry_forms/_react_grades_table.js.jsx.erb
```diff
@@ -40,9 +40,12 @@
     },
     render: function() {
       return(
-        <input id={this.props.grade_id} type="text" size="4" 
-         className="grade-input" value={this.props.value} 
-         onClick={this.handleChange} readonly/>
+        <p
```
david-yz-liu: @wkwan Can you change this to a `span`? The `p` has vertical padding that I don't like the look of.

reidka:
Can you right justify the column so the digits line up?

PR #2256, opened 2015-09-18
Fix #2245: Set the max file size of a submission to be configurable

### LINKED ISSUE #2245: Make the "max file size" that students can upload a configurable option
See `app\controllers\submissions_controller.rb`, line 311.

binuri:
Assumed 1MB = 1000000 Bytes (similar to past behaviour)


david-yz-liu:
@binuri Please address the Hound style comments

david-yz-liu:
@binuri Can you please add the new key for all three environments? Currently, it's only in development.rb.

PR #2259, opened 2015-09-19
Refactor flash message for browse submissions to use partial, closes #2258

### LINKED ISSUE #2258: Refactor flash message when browsing submissions
Refactor the code used to generate flash messages on the browse submissions page to use the general `ApplicationHelper` for setting the flash message, and the `shared/flash_message` partial for rendering it

jeremygoh:
Switch to use `ApplicationHelper` to set our flash notice message. 

Moved logic for determining the notice message into the controller, use shared partial for flash messages to render.

**Test Plan**
Manual testing. Verified notice messages appropriate to the assignment's status appeared.

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -270,9 +270,9 @@ class SubmissionsController < ApplicationController
     if @assignment.past_collection_date?
       notice_text = t('browse_submissions.grading_can_begin')
     else
+      collection_time = @assignment.submission_rule.calculate_collection_time
       notice_text = t('browse_submissions.grading_can_begin_after',
-           time: I18n.l(@assignment.submission_rule.calculate_collection_time,
-           format: :long_date))
+                      time: I18n.l(collection_time, format: :long_date))
     end
 
     flash_now(:notice, notice_text)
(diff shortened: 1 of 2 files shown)


app/controllers/submissions_controller.rb
```diff
@@ -2,6 +2,8 @@
 
 class SubmissionsController < ApplicationController
   include SubmissionsHelper
+  include PaginationHelper
```
david-yz-liu: Does this really need to be included?

jeremygoh: Thanks for the catch. I must have missed it or made a mistake during a rebase.

app/controllers/submissions_controller.rb
```diff
@@ -2,6 +2,7 @@
 
 class SubmissionsController < ApplicationController
   include SubmissionsHelper
+  include ApplicationHelper
```
david-yz-liu: @jeremygoh Actually, can you try without the `ApplicationHelper` as well? I seem to recall it's already included in ApplicationController.

jeremygoh: @david-yz-liu - thanks! Removed the duplicate

PR #2267, opened 2015-09-19
Switch to jQuery AJAX request for group invite cancellation

WilsonChiang:
- This is part of a task to remove dependencies for prototype.js
- Also removed unused mapping.js file


app/views/assignments/_member.html.erb
```diff
@@ -20,9 +20,8 @@ member.user.last_name %> <%= I18n.t("student.inviter") %></li>
    <% if grouping.is_inviter?(@current_user) %>
      <%= link_to I18n.t(:cancel), '#',
         onclick: "
-          new Ajax.Request('" +
-             disinvite_member_assignment_path(id: @assignment.id, membership: member.id) + "');",
-
+          jQuery.ajax({
+          url: '" + disinvite_member_assignment_path(id: @assignment.id, membership: member.id) +"'});",
```
david-yz-liu: This seems not to be indented properly. Also, use Ruby string interpolation.

WilsonChiang:
@david-yz-liu is it ok to remove the prototype files now or are there other dependencies? Namely, `rails_pt_adapters.js` and `prototype-snippet.js`.

david-yz-liu:
Probably still other dependencies.

WilsonChiang:
Ok. That should've been the only AJAX request from prototype then.

Issue #2268, opened 2015-09-19
Refactor layout for assignment scoped views


jeremygoh:
There are a lot of common elements in the views for `assignments/:id/...`

We should refactor and extract common elements for the assignments scoped url into their own layout
Specifically
- [ ] Settings
- [ ] Groups
- [ ] Graders
- [ ] Tags
- [x] Submissions (browse)
- [ ] Summaries

Later on, we may want to consider using this layout consistently for other more generally scoped views like `/students`

jeremygoh:
I'm currently looking at this

jeremygoh:
When logged in as a student, the file manager is rendered without a layout. I'll put up a fix for this shortly.

david-yz-liu:
@b2simms note: the comment from last year no longer applies, so you should be working just on using the new `assignment_content` layout.

mishaschwartz:
Reason for marking as invalid:

- the work was done but this issue was not closed

PR #2269, opened 2015-09-19
Create generic layout for assignment content - browsing assignments

jeremygoh:
For part of #2268 
- created new layout to be used for assignment scoped views
- applied this to `assignments/:id/browse`


jeremygoh:
@david-yz-liu  - I moved the rendering of flash message into layout as discussed.

Issue #2264, opened 2015-09-19
Create a menu layout for links at top-right of some pages.


david-yz-liu:
See 3.4.5-3.4.7 of http://guides.rubyonrails.org/layouts_and_rendering.html#using-nested-layouts for the best approach.

PR #2277, opened 2015-09-19
Render heading buttons in browse submissions using partials, for #2264

jeremygoh:
For #2264 

Continued building on top of browse submission's layout file, extracting heading buttons and rendering them using a partial and a spacer template.

@david-yz-liu  - Hound is going to complain about the line length. I'm not sure how I can make things a little neater. It seems like breaking things up would just be more verbose in this case.

If this approach looks alright, we can use the same partial to render heading buttons on other pages.

### DIFF
--- a/app/views/submissions/browse.html.erb
+++ b/app/views/submissions/browse.html.erb
@@ -22,27 +22,26 @@
   <% @heading_buttons +=
   [
     {
-         link_text: t('collect_submissions.collect_all'),
-         link_path: collect_all_submissions_assignment_submissions_path(
-                    @assignment.id),
+      link_text: t('collect_submissions.collect_all'),
+      link_path: collect_all_submissions_assignment_submissions_path(@assignment.id),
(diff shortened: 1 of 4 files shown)


app/views/submissions/browse.html.erb
```diff
@@ -14,62 +14,84 @@
...
+<% @heading_buttons = [] %>
+
+<% if @current_user.admin? %>
+  <% @heading_buttons +=
+  [
+    {
+         link_text: t('collect_submissions.collect_all'),
+         link_path: collect_all_submissions_assignment_submissions_path(
+                    @assignment.id),
+      html_options: { confirm: t('collect_submissions.results_loss_warning') }
```
david-yz-liu: Please align the keys rather than the colons (it's more consistent with other dictionaries in MarkUs).

jeremygoh: Sure thing!

Is there a style guide you recommend for future reference?

PR #2282, opened 2015-09-20
add apply on late penalty option on manual collection

wkwan:
This is related to issue 2136 as it involves the late penalty originally not being applied on manual collection, but I haven't yet reproduced the bug where the late penalty was not applied because of a remark request without manual collection.

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -179,9 +179,9 @@ class SubmissionsController < ApplicationController
   def manually_collect_and_begin_grading
     @grouping = Grouping.find(params[:id])
     @revision_number = params[:current_revision_number].to_i
-    @apply_late_penalty = params[:apply_late_penalty]
+    apply_late_penalty = params[:apply_late_penalty]
     submission = SubmissionCollector.instance.manually_collect_submission(
-      @grouping, @revision_number, @apply_late_penalty, false)
+      @grouping, @revision_number, apply_late_penalty, false)
     redirect_to edit_assignment_submission_result_path(
       assignment_id: @grouping.assignment_id,
       submission_id: submission.id,
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -179,8 +179,9 @@ class SubmissionsController < ApplicationController
   def manually_collect_and_begin_grading
     @grouping = Grouping.find(params[:id])
     @revision_number = params[:current_revision_number].to_i
+    @apply_late_penalty = params[:apply_late_penalty]
     submission = SubmissionCollector.instance.manually_collect_submission(
-      @grouping, @revision_number, false)
+      @grouping, @revision_number, @apply_late_penalty, false)
     redirect_to edit_assignment_submission_result_path(
       assignment_id: @grouping.assignment_id,
       submission_id: submission.id,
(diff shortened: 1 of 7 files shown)


app/controllers/submissions_controller.rb
```diff
@@ -179,8 +179,9 @@ def populate_file_manager_react
   def manually_collect_and_begin_grading
     @grouping = Grouping.find(params[:id])
     @revision_number = params[:current_revision_number].to_i
+    @apply_late_penalty = params[:apply_late_penalty]
```
david-yz-liu: Make this local - don't preprend the `@`

app/views/submissions/repo_browser.html.erb
```diff
@@ -97,6 +97,8 @@
                    id: @grouping.id,
                    assignment_id: @assignment.id) do %>
       <%= hidden_field_tag 'current_revision_number', @revision_number %>
+      <label><%= check_box_tag(:apply_late_penalty) %><%=t('apply_late_penalty')%></label>
```
david-yz-liu: You want to use one `check_box_tag` and then one `label_tag` for the label. See http://apidock.com/rails/v4.2.1/ActionView/Helpers/FormTagHelper/label_tag

david-yz-liu:
@wkwan Don't forget to fix the Hound comments. :)

wkwan:
I'm not sure why this new_submission variable was being assigned to in the first place but I added the late penalty in the Windows-specific part of the function, so I figured it should be here as well in case calling those functions causes state changes.

david-yz-liu:
@wkwan Okay that's fine.

reidka:
@wkwan I tested this, and it works fine with the penalty decay submission rule.  It doesn't quite work with the grace days.
Example: run collect all, where someone is using one or more grace periods. Then go to the repo view, switch to an earlier revision, and collect without checking the apply penalty box.  The grace periods penalty doesn't change.

wkwan:
@reidka By the grace periods penalty, do you mean deducting grace credits? So then the expected behaviour is that if you check "apply late penalty", it deducts grace credits when applicable, and when you don't, the grace credits remain the same?

reidka:
@wkwan We have to make sure we aren't deducting grace credits multiple times for the same assignment if it gets collected multiple times.  (I don't think we are) Conversely, if I don't apply the late penalty then any grace credits deducted from a previous collection of this assignment should be given back to the student.  I hope that makes sense.

app/models/submission_collector.rb
```diff
@@ -184,10 +184,25 @@ def collect_next_submission
...
+                                         new_submission)
+    if apply_late_penalty
+      new_submission = grouping.assignment.submission_rule
+                               .apply_submission_rule(new_submission)
+    elsif grouping.assignment.submission_rule.is_a?(GracePeriodSubmissionRule)
+      # Return any grace credits previously deducted for this grouping.
+      grouping.assignment.submission_rule
+                                       .remove_deductions(new_submission)
+    end
+    return new_submission
```
david-yz-liu: Remove the `return`.

app/models/submission_collector.rb
```diff
@@ -212,6 +228,9 @@ def manually_collect_submission(grouping, rev_num, async=true)
     grouping.save
 
     new_submission = Submission.create_by_revision_number(grouping, rev_num)
+    new_submission = apply_penalty_or_add_grace_credits(grouping,
```
david-yz-liu: Yeah, get rid of the assignment.

david-yz-liu:
Thanks, @wkwan. I made some tweaks, which you can see at 7715d10a0ec729f84e6f02fdf6d6498144f96a16

wkwan:
Great thanks!

PR #2283, opened 2015-09-20
Fixed 404 and 500 error for remark requests

melneubert:


app/models/submission.rb
```diff
@@ -197,8 +197,9 @@ def make_remark_result
     end
 
     original_result.marks.each do |mark|
-      remark_result.marks.create(mark.attributes.merge(
-                                   result: remark, created_at: Time.zone.now))
+      remark_result.marks.create(result: remark, created_at: Time.zone.now,
```
david-yz-liu: Put each key on a new line

david-yz-liu:
@melneubert I think the `extra_marks` creation should change as well, right?

PR #2284, opened 2015-09-20
#2261 Fixes the deprecation warnings from the unit and functional tests

binuri:


david-yz-liu:
@binuri Don't forget about the Hound comment. :)

Issue #2274, opened 2015-09-19
Tags list view doesn't load


WilsonChiang:
The list of tags available for an assignment does not load properly in `assignments/{pk}/tags`.

WilsonChiang:
Investigated this issue a little. Looks like it can't find the user for some reason.

```
ActiveRecord::RecordNotFound in TagsController#index

Couldn't find User with 'id'=#<Admin:0x0000000b5e42e0>
```

WilsonChiang:
Looks like tag creator is being inserted incorrectly. The foreign key is a string too, which seems odd?

```
2.1.3 :003 > Tag.all
  Tag Load (0.3ms)  SELECT "tags".* FROM "tags"
 => #<ActiveRecord::Relation [#<Tag id: 1, name: "test", description: "tag", user: "#<Admin:0x0000000b5e42e0>">, #<Tag id: 2, name: "test", description: "test", user: "#<Admin:0x0000000b5bab48>">, #<Tag id: 3, name: "test", description: "test", user: "#<Admin:0x0000000b211190>">, #<Tag id: 4, name: "test", description: "test", user: "#<Admin:0x0000000882f2f0>">, #<Tag id: 5, name: "testing", description: "testing\r\n", user: "#<Admin:0x000000080e7600>">]>
```

WilsonChiang:
Download tag list CSV still broken.

PR #2295, opened 2015-09-28
Fix tags not loading

WilsonChiang:
- We are passing a string of the memory address into user.find instead of an actual identifier

part of https://github.com/MarkUsProject/Markus/issues/2274

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -25,7 +25,7 @@ class TagsController < ApplicationController
     new_tag = Tag.new(
       name: params[:create_new][:name],
       description: params[:create_new][:description],
-      user: @current_user.id.to_s)
+      user: @current_user)
 
     if new_tag.save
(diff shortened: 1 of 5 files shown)


WilsonChiang:
So I'm a bit confused at the original intent behind the schema. User is stored as a string. Should this not be an int foreign key to user id? I haven't changed it yet because I can't quite figure out what was happening before... but this should have never worked?

david-yz-liu:
@WilsonChiang Good find. I don't believe your fix is the best possible; I wonder if the problem lies in how Rails is thinking about the `user` attribute for tags. Can you please try instead to add a line to the Tag model that says `belongs_to :user`?

WilsonChiang:
@david-yz-liu 

Seems to complain about a missing `user_id` once I do that. Perhaps I should be changing the schema so that it stores the user_id integer instead?

david-yz-liu:
@WilsonChiang Sorry I didn't reply, but you got it right!

Issue #2301, opened 2015-10-02
Student file manager view no longer links to the file


reidka:
A file in the file manager view should be link to the contents of the file.  This used to work, and still does in the repo view for the admins.  (I'm guessing this happened in the switch to react.)

arkon:
This is mentioned in #2155 too.

PR #2302, opened 2015-10-02
fixing #2301

reidka:
Uses repo browser code so that students can view the contents of files after they have submitted them in the browser file manager.

Please see comments, there is one context that I'm not sure about.

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -153,26 +153,25 @@ class SubmissionsController < ApplicationController
     @path = params[:path] || '/'
     @previous_path = File.split(@path).first
 
-    user_group.access_repo do |repo|
-      if revision_number.nil?
-        @revision = repo.get_latest_revision
-      else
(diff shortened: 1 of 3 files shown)


app/controllers/submissions_controller.rb
```diff
@@ -153,26 +153,25 @@ def populate_file_manager_react
...
+      files = @revision.files_at_path(full_path)
+      files_info = get_files_info(files, @assignment.id, revision_number, @path,
+                                  @grouping.id)
+
+      directories = @revision.directories_at_path(full_path)
+      directories_info = get_directories_info(directories, revision_number,
+                                              @path, @grouping.id)
+      render json: files_info + directories_info
+    else
+      return []
```
reidka: I'm not sure if this is a relevant condition, or what it should return in this case.

david-yz-liu: Yeah, rendering an error response is probably better, but I think we should make that another issue.

Issue #1496, opened 2014-06-02
Transition from Prototype to jQuery (TODO)


arkon:
Since the Prototype JavaScript framework is a bit antiquated and slow, we're switching to fully using jQuery (which is now the standard anyway). It should also be noted that we are also using other JS frameworks such as React (by Facebook) for tables.

Of course, vanilla JS is welcome too (which is more efficient than using jQuery for obvious reasons).

Progress/things that still need to be translated:

- [ ] Rails 3 seems to rely on some Prototype code (`prototype-snippet.js`)
- [ ] Inline Prototype code here and there in the HTML files
- [x] RJS files in `app/views/` (which use Prototype) (converted in PR #2309)
- [x] FilterTables (mostly replaced with React.js tables; see issue https://github.com/MarkUsProject/Markus/issues/1696)
- [x] In `app/assets/javascripts/`:
  - [x] `grades_table.js` (converted in PR https://github.com/MarkUsProject/Markus/pull/1639)
  - [x] `manage_graders.js`
  - [x] `manage_groups.js`
  - [x] `users_manager.js`  (deleted in switch to React.js)
- [x] https://github.com/MarkUsProject/Markus/pull/1326 (sort of replaced by React.js stuff)
- [x] https://github.com/MarkUsProject/Markus/pull/1323 (replaced by PR https://github.com/MarkUsProject/Markus/pull/1604)
- [x] https://github.com/MarkUsProject/Markus/pull/1325 (replaced by PR https://github.com/MarkUsProject/Markus/pull/1601)
- [x] Replacing Bluff for drawing graphs with [Chart.js](http://www.chartjs.org/) (done in PR https://github.com/MarkUsProject/Markus/pull/1519)
- [x] calendardateselect (replaced with jQuery UI date/time picker in https://github.com/MarkUsProject/Markus/pull/1511)
- [x] PeriodDeltaChain (uses jQuery as of PR https://github.com/MarkUsProject/Markus/pull/1511)
- [x] SourceCodeGlower, used for annotations (uses jQuery as of PR https://github.com/MarkUsProject/Markus/pull/1517)
- [x] SyntaxHighlighter adapter/mods (uses jQuery as of PR https://github.com/MarkUsProject/Markus/pull/1517)
- [x] The draggable bar in the marks/code view (replaced in PR https://github.com/MarkUsProject/Markus/pull/1529)
[...]

PR #2309, opened 2015-10-06
Convert rjs files in views to js.erb

melneubert:
Related to #1496


david-yz-liu:
@melneubert Looking at the file names, I'm not so sure the files aren't used. E.g., "decline_invitation" - there is a method in the Assignments controller with that name, which will load the rjs file you deleted.

melneubert:
I looked at that and was a little unsure. For decline_invitation, I thought that it was taken care of with the redirect_to call that is made on line 327 of the assignments controller? It seems to redirect to the same action that the rjs file did, and when I manually test a student declining a group invitation, it redirects to the student interface as expected.

If this isn't the case let me know and I can get these files back and just convert them.

david-yz-liu:
@melneubert Ah okay, you're right. I didn't look hard enough. Thanks!

PR #2310, opened 2015-10-08
Test framework update

jeremygoh:
@david-yz-liu - submitting a PR as requested.

Here's the point we've gotten to so far in terms of the merge. I had a look again at some issues what @WilsonChiang documented, also at issues around being unable to create new assignments and have fixed the bad merge which caused it.

@david-yz-liu I am sure that there are still areas of functionality which need to be looked into more thoroughly but this PR is just a placeholder so you and @reidka can get an idea of where we're at.

There are still a large number of failing RSpec tests and so the next step is to work through those.

Thanks.


app/views/results/marker/_marker_panes.html.erb
```diff
@@ -51,6 +51,7 @@
               <pre id='test_result_content' cols='80' rows='50'><%= t('marker.select_a_test') %></pre>
             </div>
           </div>
+>>>>>>> master
```
david-yz-liu: @jeremygoh  @WilsonChiang This line should be deleted

WilsonChiang:
https://docs.google.com/document/d/1IFfGQ3VIYq8p2ZSy13GJPkOAVoaAw-zRcrQOj9LyKuQ/edit#

Putting this here for now so we have a link to the document.

Issue #2306, opened 2015-10-04
Annotation Overlay For JPGs/PDFs is Slightly Off


wkwan:
The overlay area which an annotation can be made on an JPG or PDF file is a little bit off the actual boundaries of the image. The area seems to be roughly the same size as the actual image but a little bit shifted. This is more problematic when you rotate the image with an annotation. Because of the slight offset between the annotation overlay and the actual image, rotation causes the annotation to no longer be above the exact same place in the image.

Example with annotation going over the left side, then getting rotated 180 degrees.
![screen shot 2015-10-04 at 3 49 52 am](https://cloud.githubusercontent.com/assets/2907892/10266937/12f4bbd2-6a4b-11e5-8aad-a06ccb510f75.png)

![screen shot 2015-10-04 at 3 50 01 am](https://cloud.githubusercontent.com/assets/2907892/10266939/17df531e-6a4b-11e5-82a2-b05007e6642e.png)

PR #2312, opened 2015-10-11
close #1109

### LINKED ISSUE #1109: Change landscape / portrait mode on images and PDF
As title goes.

wkwan:
Due to #2306 the annotations are very slightly off when rotated. Will fix that along with the rest of the PDF issues.

I undid all the code I wrote for rotating JPGs.


david-yz-liu:
@wkwan I don't think the annotations are rotating correctly.

wkwan:
So I've been testing it again and it seems to work like I expected (besides the slight imprecision). Here are some screenshots of the same annotations in all 4 orientations, is there something wrong with this behaviour or do you see something different?

screen shot 2015-12-02 at 9 18 23 am

screen shot 2015-12-02 at 9 18 33 am

screen shot 2015-12-02 at 9 18 42 am

screen shot 2015-12-02 at 9 18 51 am


Also, I realize that it doesn't work if you rotate the PDF by pressing the 'r' key instead of the rotate button. Apparently this hotkey is enabled by the PDF library by default and extra code I wrote to rotate the PDFs only gets called if you press the button. So if this was the problem, I can fix that.

david-yz-liu:
@wkwan I didn't see the annotations moving at all - could this be a browser issue? Can you please try on both Firefox and Chrome?

wkwan:
@david-yz-liu Wow you're right. It doesn't work on Firefox.

wkwan:
Actually I'm not sure if that was it or not. I refreshed the page and it started working. And it kept working after I restarted the server and uploaded new PDFS. Will keeping looking.

wkwan:
OK I think I found the problem. When the zooming is set to automatic zoom, fit page, or full width, the rotation doesn't work. I was only testing with the scale set to specific percentages.

david-yz-liu:
@wkwan Looked good to me - again, I cleaned up a few things that you can check out in the latest commit.

wkwan:
@david-yz-liu Ah ok actually I made the same changes getting rid of the commented out code and the console.log but I just pushed it to my branch now. The only difference is that I added some comments and factored out the angle calculation (was previously done separately in PdfAnnotationManager.js and _annotations.js.erg). You want me to make another pull request?

david-yz-liu:
Sure!

PR #2314, opened 2015-10-11
close #2043

### LINKED ISSUE #2043: Long single line annotation overflows grey popup
See image below. Image shows issue on PDF, have also repro'd on text and image files.
screen shot 2015-02-14 at 1 43 54 pm

To repro:
1) Select area (or text) for annotation
2) Create a new annotation adding to a category (ie: not one time) with a lot of text on a single line without spaces. I used the following: ```sfghjhgfdsghjkjytresdcvbju76543wsdvnjki876543wsdfghu7654esdfghu```
3) Add the annotation
4) Hover over newly added annotation

wkwan:
Don't allow more than 50 consecutive letters in an annotation to avoid the text overflowing the speech bubble.

### DIFF
--- a/app/assets/stylesheets/grader.scss
+++ b/app/assets/stylesheets/grader.scss
@@ -26,6 +26,7 @@
   white-space: pre-wrap;
   width: auto;
   z-index: 100000;
+  word-wrap: break-word;
 }
 
 
--- a/app/views/results/common/_annotations.js.erb
+++ b/app/views/results/common/_annotations.js.erb
@@ -408,6 +408,8 @@
       alert("<%= I18n.t('marker.annotation.provide_text') %>");
       return false;
     }
+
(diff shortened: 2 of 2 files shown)


arkon:
Can't you just use [word-wrap](https://developer.mozilla.org/en-US/docs/Web/CSS/word-wrap)?

wkwan:
Maybe. It already word wraps though, the overflow only happens when you type continuously w/o spaces or new lines, which I don't think anyone will want to do, so I just prevented it.

-----Original Message-----

arkon:
`word-wrap: break-word;` should actually break a long, continuous stream of letters too. It might be a better option than enforcing it via JS -- especially if the bubble size changes, or the font size is different.

wkwan:
Thanks it works!

arkon:
👌

david-yz-liu:
Thanks, @wkwan and @arkon!

PR #2322, opened 2015-10-18
close #2320

### LINKED ISSUE #2320: Remark requests don't preserve newlines
When a student types in a remark request, even if they hit Enter the newlines are not preserved. This makes remarks difficult to process.

@wkwan We have had a request to patch this ASAP. Can you please look into this one?

wkwan:
This simple_format function should probably be called on some other strings as well. The problem was that when we have <%=...%> in a div, the new lines in the string just cause new lines in the generated html, which doesn't do anything. simple_format causes those new lines to turn into p tags.


david-yz-liu:
Thanks, @wkwan!

PR #2323, opened 2015-10-18
Removed old js files for graders and groups tables

melneubert:
Looked into these files.. from what I can tell they are no longer referenced anywhere in the code and have been replaced by _groups_manager.js.jsx.erb in app/views/groups and by _graders_manager.js.jsx.erb in app/views/graders. Checked the history of index.html.erb files for groups and graders views and noticed that javascript inclusion of the files I deleted disappeared when the React tables were created.


david-yz-liu:
Thanks, @melneubert!

Issue #2318, opened 2015-10-13
Cannot remove grader from course


bruceiv:
There was a testing grader erroneously left on one of our Markus courses, and there seems to be no way in the UI to delete such a grader. Could that feature be added?

david-yz-liu:
I don't foresee major obstacles with this, so we can definitely get someone working on this.

jeremygoh:
@david-yz-liu I can take a look at this.

PR #2325, opened 2015-10-19
Issue 2318 -  Functionality to delete grader

jeremygoh:
Functionality to delete a grader for #2318

### DIFF
--- a/app/controllers/tas_controller.rb
+++ b/app/controllers/tas_controller.rb
@@ -17,6 +17,17 @@ class TasController < ApplicationController
     @user = Ta.find_by_id(params[:id])
   end
 
+  def destroy
+    @user = Ta.find_by_id(params[:id])
+    if @user.destroy
+      flash[:success] = I18n.t('tas.delete.success',
+                               user_name: @user.user_name)
+      redirect_to action: :index
+    else
+      flash[:error] = I18n.t('tas.delete.error')
+    end
+  end
+
   def update
     @user = Ta.find_by_id(params[:user][:id])
     # update_attributes supplied by ActiveRecords
--- a/app/controllers/tas_controller.rb
+++ b/app/controllers/tas_controller.rb
@@ -18,8 +18,8 @@ class TasController < ApplicationController
   end
 
   def destroy
-    @user = Ta.find_by_id(params[:id])
-    if @user.destroy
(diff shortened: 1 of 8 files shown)


app/controllers/tas_controller.rb
```diff
@@ -17,6 +17,17 @@ def edit
     @user = Ta.find_by_id(params[:id])
   end
 
+  def destroy
+    @user = Ta.find_by_id(params[:id])
```
david-yz-liu: You should be able to use `find` here

app/views/tas/_tas_table.js.jsx.erb
```diff
@@ -65,6 +71,7 @@
         t['user_name'] = ta.user_name;
         t['first_name'] = ta.first_name;
         t['edit_link'] = <a href={ta.edit_link}><%= j raw I18n.t(:edit) %></a>
+        t['delete_link'] = <a href={ta.delete_link} data-method="delete" rel="nofollow" data-confirm="Are you sure?"><%= j raw I18n.t(:delete) %></a>
```
david-yz-liu: The "Are you sure" should be internationalized.

app/controllers/tas_controller.rb
```diff
@@ -17,6 +17,17 @@ def edit
...
+    if @user.destroy
```
david-yz-liu: I think this is a runtime error if the TA isn't found - you can flash an error in that case.

david-yz-liu:
Thanks, @jeremygoh. The other thing which is missing is a call to renegerate the repo permissions - if a TA is deleted, they should lose access to any repos they had been given access to previously.

jeremygoh:
Thanks for the quick review @david-yz-liu !

I've made the changes to the `destroy` method in the `tas_controller`, and also internationalized the confirmation dialog as requested (I knew I was forgetting to internationalize somewhere!).

For the regeneration of the repo permissions, is this call sufficient?
https://github.com/jeremygoh/Markus/blob/issue-2318-delete-grader/app/models/ta.rb#L10
https://github.com/jeremygoh/Markus/blob/issue-2318-delete-grader/app/models/user.rb#L299-L306

david-yz-liu:
Thanks, @jeremygoh!

PR #2326, opened 2015-10-20
Fixing several problems with svn permissions:

reidka:
- when an individual creates a group, the permissions for TAs and admins was not added
- the auth file was still being written multiple times when csv_upload was used.  This
has been fixed by moving the auth file generation out of the transaction, and by
changing how the membership was created (grouping.add_member writes to the auth file,
and probably should continue.)
- removed a couple of methods that are no longer being called
- fixed an issue with seeding the development environment where TAs and admins were
not being added to repos

### DIFF
--- a/app/models/student.rb
+++ b/app/models/student.rb
@@ -172,7 +172,7 @@ class Student < User
       # after memberships have been established.
       @grouping.update_repository_permissions
       # Add permissions for TAs and admins.  This completely rerwrites the auth file
-      # but that should be a big deal in this case.
+      # but that shouldn't be a big deal in this case.
       @group.set_repo_permissions
     end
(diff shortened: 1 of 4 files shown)


app/models/student.rb
```diff
@@ -171,6 +171,9 @@ def create_group_for_working_alone_student(aid)
       # Update repo permissions if need be. This has to happen
       # after memberships have been established.
       @grouping.update_repository_permissions
+      # Add permissions for TAs and admins.  This completely rerwrites the auth file
+      # but that should be a big deal in this case.
```
david-yz-liu: @reidka Do you mean that *shouldn't* be a big deal?

PR #2337, opened 2015-11-02
Marking State Dropdown and Release Checkbox Changed to Buttons

wkwan:
Also, merged the marking states "unmarked", "partial" to "incomplete".


app/assets/javascripts/Grader/marking.js
```diff
@@ -78,8 +78,8 @@ jQuery(document).ready(function() {
   });
 
   // Update server status
-  var state = document.getElementById('marking_state');
-  update_status(state, state.value);
+  // var state = document.getElementById('marking_state');
```
david-yz-liu: Remove these two lines.

app/controllers/results_controller.rb
```diff
@@ -403,7 +412,7 @@ def view_marks
       @result = @submission.remark_result
       # if remark result's marking state is 'unmarked' then the student has
       # saved a remark request but not submitted it yet, therefore, still editable
-      if @result.marking_state != Result::MARKING_STATES[:unmarked] && !@result.released_to_students
+      if @result.marking_state != Result::MARKING_STATES[:incomplete] && !@result.released_to_students
```
david-yz-liu: use `== Result::MARKING_STATES[:complete]`

david-yz-liu:
@wkwan It appears that the old "unmarked" state was used for remark requests which the student had created, but had not submitted. Please create a third state "unsubmitted" which should *only* be used for such remark requests.

wkwan:
Ok will do.

wkwan:
@david-yz-liu What would the toggle button say be if submission is in the unmarked state? Since the student hasn't submitted the request yet, I would expect the grader to still see it as complete, but then if they press the button to revert it to incomplete, then wouldn't that just override the unmarked state? Or do we need to show 3 states again.

wkwan:
Gah, wasn't intending to push that last commit yet.

PR #2340, opened 2015-11-02
Issue 2334

binuri:
Added an extra column to detailed csv report which shows the student mark/ total possible mark for the assignment


david-yz-liu:
@binuri I would rather see a total number of marks earned by itself, rather than both the mark and the total possible.

I.e., show "10" rather than "10/15".

david-yz-liu:
Thanks, @binuri!

PR #2343, opened 2015-11-08
Only create new groups from CSV file uploads

binuri:


app/models/assignment.rb
```diff
@@ -379,42 +379,68 @@ def clone_groupings_from(assignment_id)
...
+        membership_error = I18n.t('csv.memberships_not_unique',
+                                  { group_name: row[0],
+                             student_user_name: errors.get(:groupings).first })
+        errors.delete(:groupings)
       else
-        # Student name does not exist, use provided repo_name
-        group.repo_name = row[1]
+        # student_membership error set if a member does not exist
+        membership_error = I18n.t('csv.member_does_not_exist',
+                                  { group_name: row[0],
```
david-yz-liu: @binuri This means you should be able to remove the `{` and `}`

reidka:
@binuri You implemented exactly what I asked for.  Unfortunately, now that I'm testing it, I'm discovering subtle and not so subtle problems with the strategy.  I put a comment in the issue, but I'm going to have to investigate a bit further.

app/models/assignment.rb
```diff
@@ -9,7 +9,7 @@ class Assignment < ActiveRecord::Base
 
   has_many :rubric_criteria,
            -> { order(:position) },
-           class_name: 'RubricCriterion', 
+           class_name: 'RubricCriterion',
```
binuri: When I put double quotes, it tells me "Prefer single-quoted strings when you don't need string interpolation or special symbols.". So which quotations do i use?

app/models/assignment.rb
```diff
@@ -379,42 +379,125 @@ def clone_groupings_from(assignment_id)
...
+        if any_grouping.nil?
+          # CASE: Group exists with same repo name but has no grouping
+          #       associated with it for any assignment
+          # Use existing group and create a new grouping between the existing
+          # group and the given students and return without error
+          add_new_grouping_for_group(row, group)
+          return
+        else
+          grouping_for_current_assignment = group.grouping_for_assignment(id)
+          if grouping_for_current_assignment.nil?
```
binuri: I think the code is easier to understand with the 3 levels

app/models/assignment.rb
```diff
@@ -379,42 +379,125 @@ def clone_groupings_from(assignment_id)
...
-        group.repo_name = row[0]
+    # If any of the given members do not exist or is part of another group,
+    # an error is returned without creating a group
+    unless membership_unique?(row)
+      if !errors[:groupings].blank?
+        # groupings error set if a member is already in different group
+        membership_error = I18n.t('csv.memberships_not_unique',
+                                  group_name: row[0],
+                                  student_user_name: errors
+                                                         .get(:groupings).first)
```
binuri: Once i place the . with the method call receiver, i get the comment to place the . with the method call.

app/models/assignment.rb
```diff
@@ -379,42 +379,125 @@ def clone_groupings_from(assignment_id)
...
-    # order, so that everything works the way we want it to.
     group = Group.where(group_name: row.first).first
-    if group.nil?
-      group = Group.new
-      group.group_name = row[0]
+
+    unless group.nil?
+      if group.repo_name != row[1]
+        # CASE: Group already exits but the repo name is different
+        duplicate_group_error = I18n.t('csv.group_with_different_repo',
```
david-yz-liu: @binuri You can ignore all of these ones. Not sure why they're appearing.

app/models/assignment.rb
david-yz-liu: Try doing

```
student_user_name: errors.get(:groupings)
                         .first
```

app/models/assignment.rb
david-yz-liu: Okay, this is fine.

reidka:
I have tested this this and I think it works as expected.  I'm slightly nervous about this because of the implications.  The changes fix the concerns I had.

Issue #2335, opened 2015-10-31
Less than or equal sign does not appear in comments.


ostrokach:
Whenever I put a "<=" in a comment in student code, it does not get rendered. If I put "<=" inside a \<code\> block, nothing gets rendered on the line after "<=".

reidka:
This happens when annotation pdf files.

ostrokach:
I think I was annotating python .py files?

reidka:
Hmm.  I can't replicate it when adding annotations to code files, but I do see the problem when adding an annotation for a pdf file.

WilsonChiang:
Working on this. Also could not replicate this on .py files.

PR #2345, opened 2015-11-08
Fix issue with PDF annotations missing less than sign

WilsonChiang:
For https://github.com/MarkUsProject/Markus/issues/2335

@david-yz-liu, @reidka wanted to get your thoughts on this for a fix. Doesn't seem super elegant but seems to fix the problem. This is the first point when I traced the stacktrace in which the `<=` sign is lost. Doesn't seem like the javascript causes it to be lost at any point. I created the super method as I didn't want the user to see the `&lt;` when they go to edit the annotation later.

I thought about sanitizing the input as well, but it seems like the suggested method is to sanitize the output. It would also result in `&lt;` showing up when you go to edit an annotation.

### DIFF
--- a/app/models/annotation_text.rb
+++ b/app/models/annotation_text.rb
@@ -24,4 +24,10 @@ class AnnotationText < ActiveRecord::Base
   def html_content
     content.gsub(/\n/, '<br/>').html_safe
   end
+
+  # Fixes the bug in PDF annotation where '<=' is stripped
+  def html_content_symbols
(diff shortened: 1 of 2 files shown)


david-yz-liu:
1. I don't think this fixes the problem when the annotation is first created and rendered (doesn't go through the server at all in this case). The problem does seem to be fixed when the page refreshes, though.

2. I want to start using markdown (#2382), so I don't think it's worth looking at this issue too much right now.

PR #2355, opened 2015-11-15
Issue 2050

### LINKED ISSUE #2050: Flag empty submissions?
It would be nice to be able to flag submissions where students have not submitted any files.  Sometimes this is easy to see based on the last commit date, but other times it is a bit trickier.

melneubert:
Closes #2050 

Instructors can optionally choose to disallow adding graders to groups with no submission. If this is the case, a message will be displayed to them and all other valid grader -> group assignments will go through.

After submissions are collected, those with empty submissions will be flagged with an icon.

### DIFF
--- a/app/controllers/graders_controller.rb
+++ b/app/controllers/graders_controller.rb
@@ -191,7 +191,7 @@ class GradersController < ApplicationController
         if found_empty_submission
           assign_all_graders(filtered_grouping_ids, grader_ids)
           render text: I18n.t('assignment.group.group_submission_no_files'),
-          status: 200
+                 status: 200
         else
           assign_all_graders(grouping_ids, grader_ids)
           head :ok
--- a/app/controllers/graders_controller.rb
+++ b/app/controllers/graders_controller.rb
@@ -177,12 +177,13 @@ class GradersController < ApplicationController
       case params[:global_actions]
       when 'assign'
         if grader_ids.blank?
-          render text: I18n.t('assignment.group.select_a_grader'), status: 400
+          render text: I18n.t('assignment.group.select_a_grader'),
+                 status: 400
         end        
-        # If the instructor wants to skip empty submissions, remove those groups
-        # from the list of grouping_ids to assign graders to
         if params[:skip_empty_submissions] == 'true'
-          @found_empty_submission = false;
+          # If the instructor wants to skip empty submissions, remove those groups
+          # from the list of grouping_ids to assign graders to
+          @found_empty_submission = false
           grouping_ids = find_empty_submissions(grouping_ids)
         end
         assign_all_graders(grouping_ids, grader_ids)
@@ -191,7 +192,6 @@ class GradersController < ApplicationController
         else
           head :ok
         end
-        
       when 'unassign'
         if params[:grader_memberships].blank?
           render text: I18n.t('assignment.group.select_a_grader'), status: 400
@@ -204,12 +204,13 @@ class GradersController < ApplicationController
           render text: I18n.t('assignment.group.select_a_grader'), status: 400
         else
           if params[:skip_empty_submissions] == 'true'
-            @found_empty_submission = false;
+            @found_empty_submission = false
             grouping_ids = find_empty_submissions(grouping_ids)
           end
           randomly_assign_graders(grouping_ids, grader_ids)
           if @found_empty_submission == true
-            render text: I18n.t('assignment.group.group_submission_no_files'), status: 200
+            render text: I18n.t('assignment.group.group_submission_no_files'), 
+                   status: 200
           else
             head :ok
           end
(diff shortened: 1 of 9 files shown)


melneubert:
Sorry for all the hound stuff. I fixed what I could. The remaining ones I either cannot find or I can't avoid (like the 3 levels of block nesting - I believe that was already done in other places in that file)

app/controllers/graders_controller.rb
```diff
@@ -177,9 +177,19 @@ def global_actions
...
       when 'assign'
         if grader_ids.blank?
-          render text: I18n.t('assignment.group.select_a_grader'), status: 400
+          render text: I18n.t('assignment.group.select_a_grader'),
+                 status: 400
+        end        
+        if params[:skip_empty_submissions] == 'true'
+          # If the instructor wants to skip empty submissions, remove those groups
+          # from the list of grouping_ids to assign graders to
+          @found_empty_submission = false
```
david-yz-liu: Never use `@` to act as a global variable, this makes code much harder to reason about. In the next line, don't both mutate grouping_ids and assign to it. `find_empty_submissions` is not a very good name for what you're doing, which is filtering out submissions.

Perform two separate operations:
1) Obtain a list of non-empty submission grouping ids in a new variable. Don't mutate grouping_ids. Use `select`.
2) Compare the sizes of the two, and use this to set the value of `found_empty_submission`, which should be a local variable (no @).

app/controllers/graders_controller.rb
```diff
@@ -286,4 +305,16 @@ def unassign_graders(grader_membership_ids)
...
     Grouping.unassign_tas(grader_membership_ids, grouping_ids, @assignment)
   end
+
+  def find_empty_submissions(grouping_ids)
+    grouping_ids.each do |grouping_id|
+      submission = Submission.find_by_grouping_id(grouping_id)
+      if !submission
+        next
+      elsif !SubmissionFile.where(submission_id: submission.id).exists?
+        grouping_ids.delete(grouping_id)
```
david-yz-liu: Never delete from an iterable when you're iterating through it. But also, see above comment.

app/controllers/graders_controller.rb
```diff
@@ -286,4 +305,16 @@ def unassign_graders(grader_membership_ids)
     end
     Grouping.unassign_tas(grader_membership_ids, grouping_ids, @assignment)
   end
+
+  def find_empty_submissions(grouping_ids)
+    grouping_ids.each do |grouping_id|
+      submission = Submission.find_by_grouping_id(grouping_id)
```
david-yz-liu: Don't use `find_by_grouping_id`, but instead `find_by(grouping_id: grouping_id)`

app/models/grouping.rb
```diff
@@ -695,6 +695,17 @@ def last_commit_date
     end
   end
 
+  # Helper for populate_submission_table
+  # Returns boolean value based on if the submission has files or not
+  def has_files_in_submission?
```
david-yz-liu: Return `!has_submission || SubmissionFile.where(...).exists?` instead of using a conditional.

app/views/graders/_graders_manager.js.jsx.erb
```diff
@@ -53,6 +54,14 @@
         }.bind(this)
       });
     },
+    skipEmptySubmissionsClicked: function(event) {
+      var value = event.currentTarget.checked;
```
david-yz-liu: Just do `this.setState({skip_empty_submissions: event.currentTarget.checked});`

app/controllers/graders_controller.rb
```diff
@@ -286,4 +312,16 @@ def unassign_graders(grader_membership_ids)
     end
     Grouping.unassign_tas(grader_membership_ids, grouping_ids, @assignment)
   end
+
+  # Returns array of grouping ids with non empty submissions
+  def filter_empty_submissions(grouping_ids)
+    filtered_grouping_ids = Array.new
+    grouping_ids.each do |grouping_id|
```
david-yz-liu: Use `.select` instead of `.each`

app/controllers/graders_controller.rb
```diff
@@ -177,9 +177,19 @@ def global_actions
       case params[:global_actions]
       when 'assign'
         if grader_ids.blank?
-          render text: I18n.t('assignment.group.select_a_grader'), status: 400
+          render text: I18n.t('assignment.group.select_a_grader'),
+                 status: 400
+        end
```
david-yz-liu: @melneubert Have you not fixed this?

melneubert:
@david-yz-liu Sorry I missed that one hound message. Everything has been fixed now, I updated that each to a select.

david-yz-liu:
Note: I assume assigning graders works after assignments have been collected, but if the assignments aren't collected, there aren't any Submission objects, so no graders can be assigned.

It would be possible for now to just add this note as a warning to the checkbox label when assigning graders.

melneubert:
@david-yz-liu Do you mean to add a warning after all the text I have so that it looks like " Do not assign Graders to groups with empty submissions (requires that submissions are collected)", or something closer to a bit of an alert type warning that appears when the checkbox is clicked?

david-yz-liu:
@melneubert I think your first suggestion is better, so let's go with that.

reidka:
This still needs tests.  Working on that in a separate branch.

PR #2358, opened 2015-11-17
Fix penalty decay

reidka:
1. Changed interval from int to float.  This is primarily for easier testing, but it is possible that instructors might want to assign penalties on granularities less than an hour.
2. Fixed up some of the calculation following the change to a float.


app/models/grace_period_submission_rule.rb
```diff
@@ -109,7 +109,7 @@ def remove_deductions(submission)
           deduction.destroy
         end
       end
-      student_membership.user.grace_period_deductions.reload
+      deductions.reload
```
david-yz-liu: So this actually made a difference about what was cached?

reidka: No I don't think so.  I must not have done a pull request to get that into my code base while I was testing.  I thought the revised line was better style.

reidka:
I needed this one to test another PR

PR #2362, opened 2015-11-20
Test framework save testrun result

jeremygoh:
As discussed with @david-yz-liu , putting up this PR

- rewrote `process_result` method to save results of a testrun into the database
- saved it using the TestResult model. A number of the fields in this model don't make the most sense as it was originally intended to save the results of an individual unit test (e.g. completion status). In particular, I hacked a value for  `test_result.submission_id`
- saved `test_result.actual_output` as json so we can move away from XML @reidka 
- removed XML parser from `Gemfile` and recreated `Gemfile.lock`

### DIFF
--- a/app/helpers/automated_tests_helper.rb
+++ b/app/helpers/automated_tests_helper.rb
@@ -740,8 +740,8 @@ module AutomatedTestsHelper
   def self.process_result(raw_result)
     result = Hash.from_xml(raw_result)
     repo = @grouping.group.repo
-    @revision = repo.get_latest_revision
-    @revision_number = @revision.revision_number
+    revision = repo.get_latest_revision
+    revision_number = revision.revision_number
     raw_test_scripts = result['testrun']['test_script']
 
     # Hash.from_xml will yield a hash if only one test script
@@ -776,7 +776,7 @@ module AutomatedTestsHelper
     TestResult.create(grouping_id: @grouping.id,
                       test_script_id: test_script.id,
                       name: script_name,
-                      repo_revision: @revision_number,
+                      repo_revision: revision_number,
                       input_description: '',
                       actual_output: result.to_json,
                       expected_output: '',
--- a/app/helpers/automated_tests_helper.rb
+++ b/app/helpers/automated_tests_helper.rb
@@ -742,7 +742,7 @@ module AutomatedTestsHelper
     repo = @grouping.group.repo
     @revision = repo.get_latest_revision
     @revision_number = @revision.revision_number
-    raw_test_scripts = result["testrun"]["test_script"]
+    raw_test_scripts = result['testrun']['test_script']
 
     # Hash.from_xml will yield a hash if only one test script
     # and an array therwise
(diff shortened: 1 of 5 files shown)


app/helpers/automated_tests_helper.rb
```diff
@@ -738,9 +737,53 @@ def self.launch_test(assignment, repo_path, call_on)
...
+    if raw_test_scripts.nil?
+      return
+    elsif raw_test_scripts.is_a?(Array)
+      test_scripts = raw_test_scripts
+    else
+      test_scripts = [raw_test_scripts]
+    end
 
+    # For now, we just use the first test script for the association
+    test_result = TestResult.new
```
david-yz-liu: This is good, but can you pass the attributes directly to `new`? E.g. `TestResult.new(grouping_id: @grouping.id, ...)`

jeremygoh:
@david-yz-liu sure thing. I've made the change to use the .create method, passing in the relevant arguments.

app/helpers/automated_tests_helper.rb
```diff
@@ -738,9 +737,53 @@ def self.launch_test(assignment, repo_path, call_on)
     end
   end
 
-  def self.process_result(result, grouping_id, assignment_id)
-  end
+  def self.process_result(raw_result)
+    result = Hash.from_xml(raw_result)
+    repo = @grouping.group.repo
+    @revision = repo.get_latest_revision
+    @revision_number = @revision.revision_number
```
david-yz-liu: I'm a little skeptical that this and revision need to be instance variables... have you thought about this?

jeremygoh: I definitely don't think they should be instance variables but I followed
the pattern used by the previous implementation and other methods in the
helper file.

I'm more than happy to make that change. I just wasnt sure what impact it
would have on other methods if any - I guess I assumed there was a good
reason for it being instance variables previously but haven't fully
investigated.

david-yz-liu: Yeah, I don't think there's a good reason for it. I would take it out for now.

jeremygoh: Ok. I've made that change now.

Thanks for the quick review! :)

david-yz-liu:
Okay, I've looked more into actually running the tests. When I click on "Run tests" as the student, I see this in the resque log:

```
(Job{test_waiting_list} | AutomatedTestsHelper | [61, "request"]) failed: #<NoMethodError: undefined method `[]' for nil:NilClass>
```

I don't get a "test_runs" directory, either.

Also, do you need to type in vagrant's password every time? Pretty annoying. I would rather not see any ssh calls at all.

jeremygoh:
@david-yz-liu , There's an issue with the file upload mechanism at the moment. Did you rename the TestScript record after the upload of the file and then manually add the file to the repos folder, e.g. `data/dev/automated_tests/A3/rb_suite_1.rb`?

For the SSH password, I followed the "Setup Test Server" section in this document
https://docs.google.com/document/d/1IFfGQ3VIYq8p2ZSy13GJPkOAVoaAw-zRcrQOj9LyKuQ/edit#
 to add the SSH key. Should I change the commands to drop the SSH and just execute the commands locally?

david-yz-liu:
Yes, I renamed the script_name and it's in the right place.

And yes, drop the SSH.

jeremygoh:
Ok, I'll investigate the first issue and will address the SSH one with a PR.

PR #2363, opened 2015-11-22
Submission table fixes

melneubert:
Fixed search and filtering for submissions table. Also added a filter for submissions in the remark state.

### DIFF
--- a/app/views/submissions/_submissions_table.js.jsx.erb
+++ b/app/views/submissions/_submissions_table.js.jsx.erb
@@ -93,7 +93,7 @@
             name: 'complete',
             text: '<%= j raw I18n.t(:'browse_submissions.show_complete') %>',
             func: function(submission) {
-              return submission.state == 'complete';
+              return submission.state == 'completed';
             }
           },
           {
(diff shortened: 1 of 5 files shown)


app/views/submissions/_submissions_table.js.jsx.erb
```diff
@@ -102,6 +102,13 @@
             func: function(submission) {
               return submission.state == 'released';
             }
+          },
+          {
+            name: 'remark_requested',
+            text: '<%= j raw I18n.t(:'browse_submissions.show_remark_request') %>',
```
david-yz-liu: Use `t`, not `I18n.t`

app/views/submissions/_submissions_table.js.jsx.erb
```diff
@@ -188,9 +195,9 @@
         var s = {};
         s['id'] = submission.id;
         if (submission.name_url === '') {
-          s['group_name'] = <span>{submission.name}</span>;
+          s['group_name'] = <span dangerouslySetInnerHTML={{__html: submission.name}} />;
```
david-yz-liu: Don't use this. Find another way to solve this problem.

melneubert: That's the attribute it was looking for in javascripts/ReactComponents/table.js. I guess I should be changing what it's looking for there as well?

david-yz-liu: Yes, that's exactly right.

melneubert:
@david-yz-liu I made those changes you requested. I did some research and found that Reacts escapes all strings displayed, so I think that should fix the possibility that we were running into using dangerouslySetInnerHTML. 

I removed I18n.t in the place you requested, but I still see it in other places in this file, did you want me to change all instances or leave the other ones?

david-yz-liu:
@melneubert Well done, looks great. I'll worry about the other I18n's later.

PR #2364, opened 2015-11-23
Test upload form fixes

WilsonChiang:
Currently have fixed properly storing the script name for a test file.

### DIFF
--- a/app/helpers/automated_tests_helper.rb
+++ b/app/helpers/automated_tests_helper.rb
@@ -129,16 +129,16 @@ module AutomatedTestsHelper
     testscripts.each do |file_num, file|
       updated_script_files[file_num] = {}
 
-     if testscripts[file_num][:script_name].nil?
-       # Create new test script file
-       filename = new_script.original_filename
-       if TestScript.exists?(script_name: filename, assignment: assignment)
-         raise I18n.t('automated_tests.duplicate_filename') + filename
-       else
-         # Override filename from form
-         file[:script_name] = filename
-         updated_script_files[file_num] = file.clone
-       end
+      if testscripts[file_num][:script_name].nil?
+        # Create new test script file
+        filename = new_script.original_filename
+        if TestScript.exists?(script_name: filename, assignment: assignment)
+          raise I18n.t('automated_tests.duplicate_filename') + filename
+        else
+          # Override filename from form
+          file[:script_name] = filename
+          updated_script_files[file_num] = file.clone
+        end
       else
         # Edit existing test script file
         updated_script_files[file_num] = file.clone
--- a/app/helpers/automated_tests_helper.rb
+++ b/app/helpers/automated_tests_helper.rb
@@ -127,8 +127,6 @@ module AutomatedTestsHelper
 
     # Create/Update test scripts
     testscripts.each do |file_num, file|
-      updated_script_files[file_num] = {}
-
       # If no new_script then form is empty and skip
       next if testscripts[file_num][:seq_num].empty? && file[:script_name].nil?
 
--- a/app/helpers/automated_tests_helper.rb
+++ b/app/helpers/automated_tests_helper.rb
@@ -136,66 +136,21 @@ module AutomatedTestsHelper
     # Retrieve all test scripts
     testscripts = params[:test_scripts_attributes]
 
-    # First check for duplicate script names in test scripts
-    unless testscripts.nil?
-      testscripts.values.each do |tfile|
-        if tfile['script_name'].respond_to?(:original_filename)
-          fname = tfile['script_name'].original_filename
-          # If this is a duplicate script name, raise error and return
-          if !file_name_array.include?(fname)
-            file_name_array << fname
-          else
-            raise I18n.t('automated_tests.duplicate_filename') + fname
-          end
-        end
-      end
-    end
-
     # Retrieve all test support files
     testsupporters = params[:test_support_files_attributes]
 
-    # Now check for duplicate file names in test support files
-    unless testsupporters.nil?
-      testsupporters.values.each do |tfile|
-        if tfile['file_name'].respond_to?(:original_filename)
-          fname = tfile['file_name'].original_filename
-          # If this is a duplicate file name, raise error and return
-          if filename_array.include?(fname)
-            raise I18n.t('automated_tests.duplicate_filename') + fname
-          else
-            filename_array << fname
-          end
-        end
-      end
-    end
-
     # Filter out script files that need to be created and updated
     unless testscripts.nil?
-      testscripts.each_key do |key|
-
-        tfile = testscripts[key]
-
-        # Check to see if this is an update or a new file:
-        # - If 'id' exists, this is an update
-        # - If 'id' does not exist, this is a new test file
-        tf_id = tfile['id']
-
-        # If only the 'id' exists in the hash, other attributes were not updated
-        # so we skip this entry. Otherwise, this test file possibly requires an
-        # update
-        if !tf_id.nil? && tfile.size > 1
-
-          # Find existing test file to update
-          @existing_testscript = TestScript.find_by_id(tf_id)
-          if @existing_testscript
-            # Store test file for any possible updating
-            updated_script_files[key] = tfile
+      testscripts.each do |file_num, file|
+        updated_script_files[file_num] = {}
+        unless file_num.nil?
+          file.each do |key, value|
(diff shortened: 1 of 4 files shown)


WilsonChiang:
@david-yz-liu 

Having an issue right now with supporting "Editing" a file. I looked at the original implementation and am unsure how it was supposed to work. It seems they retrieve the test script file but don't ever actually apply anything to it. In my current implementation it seems to try to create a new one without properly providing the rest of the parameters. Tried various solutions without much luck, any advice?

Another issue is that the test support files don't seem to be passed into the form parameters properly.

app/views/automated_tests/_test_script_upload.html.erb
```diff
@@ -18,6 +18,8 @@
     <span class='file_name'>
       <% if saved %>
         <%= f.object.script_name %>
+        <%= f.hidden_field :script_name, :value => f.object.script_name,
```
david-yz-liu: Use the new hash syntax `value: f.object.script_name ...`. Also use single-quotes. Wait, why is there a class at all?

app/helpers/automated_tests_helper.rb
```diff
@@ -136,96 +136,60 @@ def process_test_form(assignment, params)
...
+        updated_script_files[file_num] = {}
+
+        if testscripts[file_num][:script_name].is_a? String
+          # Edit existing test script file
+          filename = file[:script_name]
+          file.each do |key, value|
+            updated_script_files[file_num][key] = value
+          end
+        else
+          if testscripts[file_num][:script_name].nil?
```
david-yz-liu: Use the pattern `next if <cond>` right near the top of the loop (after line 145, I suppose)

app/helpers/automated_tests_helper.rb
```diff
@@ -136,96 +136,59 @@ def process_test_form(assignment, params)
...
-    # Filter out script files that need to be created and updated
+    # Create/Update test scripts
     unless testscripts.nil?
-      testscripts.each_key do |key|
+      testscripts.each do |file_num, file|
+        updated_script_files[file_num] = {}
 
-        tfile = testscripts[key]
+        # Empty file submission, skip
+        next if testscripts[file_num][:script_name].nil?
```
david-yz-liu: @WilsonChiang This is going to be `nil` now for new test scripts.

Also, the new key (new_scripts or whatever) is being sent to the server, but it's being filtered out in the controller (see the "assignment_params") before this point. So why don't you add an extra argument to this function and actually pass in the "new_file" part of the params here, so that you can use it.

app/helpers/automated_tests_helper.rb
```diff
@@ -136,96 +136,59 @@ def process_test_form(assignment, params)
...
+          if TestScript.exists?(script_name: filename, assignment: assignment)
+            raise I18n.t('automated_tests.duplicate_filename') + filename
+          else
+            file.each do |key, value|
+              updated_script_files[file_num][key] = value
+            end
+            # Override filename from form
+            updated_script_files[file_num][:script_name] = filename
           end
         end
```
david-yz-liu: @WilsonChiang FYI I was able to quickly figure that out because I put a `debugger` statement on this line, and inspected `params` and `updated_script_files`.

app/helpers/automated_tests_helper.rb
WilsonChiang: Sorry did you mean just add it to the `assignment_params`?

```
def assignment_params
    params.require(:assignment)
        .permit(:enable_test,
                :assignment_id,
                :tokens_per_day,
                :unlimited_tokens,
                test_files_attributes:
                    [:id, :filename, :filetype, :is_private, :_destroy],
                test_scripts_attributes:
                    [:assignment_id, :seq_num, :script_name, :description,
                     :new_script,
                     :max_marks, :run_on_submission, :run_on_request,
                     :halts_testing, :display_description, :display_run_status,
                     :display_marks_earned, :display_input,
                     :display_expected_output, :display_actual_output])
  end
```

Tried that but it's still not being passed through params? Or do you mean add `new_script` as another argument to the `process_test_form` function itself?

app/helpers/automated_tests_helper.rb
```diff
@@ -119,114 +119,52 @@ def create_ant_test_files(assignment)
...
-          if !file_name_array.include?(fname)
-            file_name_array << fname
-          else
-            raise I18n.t('automated_tests.duplicate_filename') + fname
-          end
-        end
-      end
-    end
+      # Empty file submission, skip
+      next if testscripts[file_num][:script_name].nil?
```
david-yz-liu: @WilsonChiang I'm going to try to look at this code more carefully.

First comment: this has the potential to introduce empty hashes into `updated_script_files`. Why not do the `next if` check at the very top of the loop?

david-yz-liu: I can't remember if I pointed this out, but on second thought, I'm not sure this check is valid. The form won't have a `:script_name` field any more, since we renamed it.

app/helpers/automated_tests_helper.rb
```diff
@@ -119,114 +119,52 @@ def create_ant_test_files(assignment)
   # - Process new and updated test files (additional validation to be done at the model level)
   def process_test_form(assignment, params)
```
david-yz-liu: Did you leave a comment earlier? Add in an extra parameter here: `process_test_form(assignment, params, new_files)` where the last one is obtained directly from `params`.

WilsonChiang:
Still working on fixing editing the file. Similar issues as before as it's trying to create a new file.

app/helpers/automated_tests_helper.rb
```diff
@@ -117,116 +117,56 @@ def create_ant_test_files(assignment)
...
-            file_name_array << fname
-          else
-            raise I18n.t('automated_tests.duplicate_filename') + fname
-          end
+    testscripts = params[:test_scripts_attributes] || []
+    testsupporters = params[:test_support_files_attributes] || []
+
+    # Create/Update test scripts
+    testscripts.each do |file_num, file|
+      updated_script_files[file_num] = {}
```
david-yz-liu: Please remove this.

app/helpers/automated_tests_helper.rb
```diff
@@ -117,116 +117,56 @@ def create_ant_test_files(assignment)
...
+
+      # If no new_script then form is empty and skip
+      next if testscripts[file_num][:seq_num].empty? && file[:script_name].nil?
```
david-yz-liu: The second check should be `new_script.nil?`

app/views/automated_tests/_test_script_upload.html.erb
```diff
@@ -25,7 +26,7 @@
     <%= link_to_function (saved ? '[+]' : '[-]'), "toggleSettings(this);", :class => "collapse" %>
   </legend>
 
-  <%= f.hidden_field :seq_num, :value => f.object.seq_num || 1.0, :class => "seqnum" %>
+  <%= f.hidden_field :seq_num, :value => f.object.seq_num || nil, :class => "seqnum" %>
```
david-yz-liu: This isn't good enough, though I'm not sure why.

Use `value: saved ? f.object.seq_num : nil`.

app/helpers/automated_tests_helper.rb
WilsonChiang: `new_script` is nil when there is no file being uploaded, so would this not cause editing to always skip?

app/helpers/automated_tests_helper.rb
```diff
@@ -117,116 +117,54 @@ def create_ant_test_files(assignment)
...
-          else
-            raise I18n.t('automated_tests.duplicate_filename') + fname
-          end
+    testscripts = params[:test_scripts_attributes] || []
+    testsupporters = params[:test_support_files_attributes] || []
+
+    # Create/Update test scripts
+    testscripts.each do |file_num, file|
+      # If no new_script then form is empty and skip
+      next if testscripts[file_num][:seq_num].empty? && file[:script_name].nil?
```
david-yz-liu: @WilsonChiang I'm not sure where you're commenting. I got an alert but don't see a comment. If you're making comments, don't delete them until I respond.

WilsonChiang: @david-yz-liu Sorry! I realized what you meant and fixed it.

david-yz-liu:
@WilsonChiang good work!

PR #2366, opened 2015-11-23
#2365: Adding delete group button for group assignments

binuri:
For now, the function can only be performed by the inviter if there are no submissions

### DIFF
--- a/app/models/grouping.rb
+++ b/app/models/grouping.rb
@@ -474,7 +474,6 @@ class Grouping < ActiveRecord::Base
   def deletable_by?(user)
     return false unless self.inviter == user
     (!self.is_valid?) || (self.is_valid? &&
-                          self.accepted_students.size == 1 &&
                           self.number_of_submitted_files == 0 &&
                           self.assignment.group_assignment? &&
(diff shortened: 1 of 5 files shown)


app/models/grouping.rb
```diff
@@ -474,7 +474,6 @@ def decline_invitation(student)
   def deletable_by?(user)
     return false unless self.inviter == user
     (!self.is_valid?) || (self.is_valid? &&
-                          self.accepted_students.size == 1 &&
```
reidka: I think we should leave this in that students can only delete a group if they are the only member remaining in the group.  Otherwise a student could be blindsided if the inviter deletes the group but doesn't tell them.

reidka:
The repo permissions aren't being revoked when a group is deleted.  I believe this is because in the assignments_controller.rb (around line 386) the memberships are deleted before the permissions are updated.  I think we have to explicitly revoke permissions before removing the members.

The tricky part is that we should not revoke permissions on a repo if there is another grouping (for a different assignment) that the student is a member of.

david-yz-liu:
Thanks, @binuri! This is a very helpful addition.

PR #2370, opened 2015-11-28
Test framework tokens check

jeremygoh:
- Added check before students run a test that they have sufficient tokens
- Changed phrasing to indicate tokens are "per day"


david-yz-liu:
Sure, thanks @jeremygoh!

Issue #2297, opened 2015-09-29
One Group per Student


bruceiv:
I ran head-on into the CSV upload issue most recently reported in #2237 and am now entering groups for 225 students by hand. I would like a feature where you can have the system automatically set up one group per student, with one button push (it seems like the common case, and also the intent of everyone who's getting jammed up on CSV upload).

reidka:
That is the plan.  The workaround we have been using is to use a firefox add-on, imacros for creating groups with one student.  The group and repo get created when the student navigates to the assignment.   Using "switch role" and imacros, we can program the browser to login as each student and navigate to the relevant assignment page.

reidka:
Done in #2376 and #2414

PR #2376, opened 2015-12-01
#2297: Create groups for individual assignments

binuri:


app/controllers/groups_controller.rb
```diff
@@ -184,6 +184,58 @@ def csv_upload
     redirect_to action: 'index', id: params[:id]
   end
 
+  def create_groups_when_students_work_alone
+    @assignment = Assignment.find_by_id(params[:assignment_id])
+
+    if @assignment.group_max == 1
+      students.map do |student|
```
reidka: Where does students come from?  I get an error when I try to run this.

app/views/groups/_groups_manager.js.jsx.erb
```diff
@@ -319,6 +319,9 @@
         <div className={'col-right'}>
           <div className={'tab-pane ui-tabs ui-widget ui-widget-content ui-corner-all'} id='groups-tabs'>
             <AddGroupButton refresh={this.props.refresh} />
+            <% if @assignment.group_max == 1 %>
+              <%= button_to "Add all groups", {action: 'create_groups_when_students_work_alone', id: @assignment.id}, method: :get %>
```
reidka: "Add all groups" should be in the language files.

reidka:
The logic seems to make sense, but I'd like to test it further.  I'll test it more after the students variable is fixed. Thanks.

app/controllers/groups_controller.rb
binuri: Sorry must have deleted the variable by mistake. I fixed it now.

reidka:
Thanks!  It looks good.

app/controllers/groups_controller.rb
```diff
@@ -184,6 +184,58 @@ def csv_upload
...
+            grouping_id: grouping.id,
+            membership_status: StudentMembership::STATUSES[:inviter],
+            user_id: student.id)
+          member.save
+          # Update repo permissions if need be. This has to happen
+          # after memberships have been established.
+          grouping.update_repository_permissions
+          # Add permissions for TAs and admins.  This completely rerwrites the
+          # auth file but that shouldn't be a big deal in this case.
+          group.set_repo_permissions
```
david-yz-liu: @binuri @reidka This will rewrite the authz file once per student. We need to move this action outside the loop.

binuri: @david-yz-liu @reidka I took out these two repo permission statements and called Repository::SubversionRepository.__generate_authz_file.
I believe this method sets authorizations for all groups and I placed it after creating all the groups. 
The changes are in the pull request #2383

PR #2379, opened 2015-12-07
File selector

wkwan:
Now in nested menu form! Also added a bunch of directories to the repository seed data.

screen shot 2015-12-06 at 11 54 19 pm


david-yz-liu:
Great work, @wkwan!

PR #2382, opened 2015-12-07
Markdown Support in Annotation and Other Areas

### LINKED ISSUE #2305: Use markdown as language for annotations
Would be nice if markdown can be used as the language for writing annotations.
For example, the library [RedCarpet](https://github.com/vmg/redcarpet) can be used for rendering markdown to html (I assume this is a rails app, correct me if I'm wrong).

finnergizer:
Closes #2305. Switched simple_format everywhere to use new `markdown(text)` view helper. Started with annotations, but then made a global switch (i.e. remark requests, overall comments).

### DIFF
--- a/app/views/annotations/update_annotation.js.erb
+++ b/app/views/annotations/update_annotation.js.erb
@@ -9,6 +9,7 @@ document.getElementById('annotation_summary_list').innerHTML =
                                   locals: { annotation_category: @annotation_text.annotation_category }) %>';
 <% end %>
 
-update_annotation_text('<%= @annotation_text.id %>', '<%= escape_javascript(simple_format(@annotation_text.content)) %>');
+update_annotation_text('<%= @annotation_text.id %>', marked('<%= escape_javascript(simple_format(@annotation_text.content)) %>'));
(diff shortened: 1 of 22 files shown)


app/views/annotations/update_annotation.js.erb
```diff
@@ -9,6 +9,6 @@ document.getElementById('annotation_summary_list').innerHTML =
                                   locals: { annotation_category: @annotation_text.annotation_category }) %>';
 <% end %>
 
-update_annotation_text('<%= @annotation_text.id %>', '<%= escape_javascript(simple_format(@annotation_text.content)) %>');
+update_annotation_text('<%= @annotation_text.id %>', marked('<%= escape_javascript(simple_format(@annotation_text.content)) %>'));
```
david-yz-liu: @finnergizer This is the only one that catches my eye. There's an inner call to `simple_format`, why not use `markdown` here?

finnergizer: @david-yz-liu since this JS is rendered server-side, I used markdown helper instead and removed marked library use in this case. Good catch! Thanks!

PR #2383, opened 2015-12-08
Issue 2297: Creating a button to add groups for all students as background task

binuri:
Need to start a resque worker in a separate terminal using command VVERBOSE=1 QUEUE=default rake environment resque:work to run the job.


reidka:
This PR is closed because a new PR added a few additional fixes.

PR #2391, opened 2015-12-20
Currently opened file text now shows full path instead of just the name.

wkwan:
As @reidka requested, I switched the text showing the currently open file to show the full path.


david-yz-liu:
Thanks, @wkwan!

Issue #2388, opened 2015-12-18
MainController error when there are no assignments


david-yz-liu:
There's a nil error here: [https://github.com/MarkUsProject/Markus/blob/master/app/controllers/main_controller.rb#L163](https://github.com/MarkUsProject/Markus/blob/master/app/controllers/main_controller.rb#L163).

Just need to add a nil check there, and make sure it's properly checked in the view as well.

PR #2406, opened 2016-01-15
Added a check for nil when loading the current assignment in the main controller.

alexinman:
Fix for #2388

### DIFF
--- a/app/controllers/main_controller.rb
+++ b/app/controllers/main_controller.rb
@@ -160,7 +160,7 @@ class MainController < ApplicationController
     ]).order('id ASC')
 
     @current_assignment = Assignment.get_current_assignment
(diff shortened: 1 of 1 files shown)


app/controllers/main_controller.rb
```diff
@@ -160,7 +160,9 @@ def index
     ]).order('id ASC')
 
     @current_assignment = Assignment.get_current_assignment
-    @current_ta = @current_assignment.tas.first
+    if @current_assignment!=nil
```
aclar48: Should be:  
@current_ta = @current_assignment.tas.first if @current_assignment.present?
or 
 @current_ta = @current_assignment.tas.first unless @current_assignment.nil?

PR #2407, opened 2016-01-15
Added callback on remark comment update, fixed jQuery issue

alexvail:
Added callback to ensure user is notified on successful updates. Fixed jQuery issue.


app/views/results/marker/_overall_remark_comment.html.erb
```diff
@@ -10,12 +10,29 @@
...
       <%= f.submit t(:save_changes),
-      	           disable_with: t('working'),
-		   disabled: true,
-		   id: 'overall_remark_comment_submit' %>
+                   disabled: "disabled",
+                   id: 'overall_remark_comment_submit' %>
     </div>
   <% end -%>
 </div>
+<script>
```
david-yz-liu: @alexvail please move this Javascript code to the `_boot.js` file.

alexvail: There's is an almost identical piece of javascript code in _overall_comment.html.erb. Won't the function get wrongly overrided if I put it in _boot.js?

david-yz-liu: The jQuery objects should be different (different ids), so I don't think there should be any overriding. But this brings up another point - can you actually move both code blocks to _boot.js?

PR #2408, opened 2016-01-15
"Loading..." popup now appears when updating released grades

Sm1th:
"Loading..." popup now appears when releasing and unreleasing grades.

Fix for issue #2392


app/views/grade_entry_forms/_react_grades_table.js.jsx.erb
```diff
@@ -154,11 +154,7 @@
           this.setState({
             error: error
           });
-        }.bind(this),
-        // Close working indicator.
-        complete: function () {
-          document.getElementById('working').style.display = 'none';
-        }
```
Sm1th: Deleted the `complete` function since we don't want the "Loading" popup to go away just yet. This will happen on line 178 in the next ajax call.

david-yz-liu: @Sm1th We shouldn't assume that the second one will complete after the first one, so I think we will need this one.

Sm1th:
@david-yz-liu Is there anything else you suggest to fix for this issue? I'm not entirely pleased as it is right now, since the loading popup sometimes is hidden before the released icon actually appears/disappears.

david-yz-liu:
@Sm1th I suppose that even after the AJAX request finishes, the table still needs to be rendered. You could try putting the line which hides the "Loading..." in the render function, perhaps.

Issue #1999, opened 2015-01-17
Removing text annotation orphans annotation_texts record


spring-ryanc:
When a one-time code annotation is removed the record from annotation_texts is not removed. At this point the text record is orphaned and cannot be removed or accessed.

alexinman:
Can this be closed now? It was fixed in #2410

reidka:
Yes, thanks.

PR #2410, opened 2016-01-15
Fix for #1999

alexinman:
Added lines to destroy the text annotation while destroying the annotation that references is. Fix for #1999


reidka:
This would also remove the text from annotations that have been added to a category.

Issue #2291, opened 2015-09-25
Assignment rules doesn't line wrap correctly


tsimica:
In the student interface show under Assignment Rules, the description and message don't line wrap correctly (depending on your resolution).

tsimica:
Working on this.

alexinman:
@david-yz-liu Just to make sure I'm understanding the problem correctly:
![screen shot 2016-01-16 at 9 22 31 am](https://cloud.githubusercontent.com/assets/6303087/12372846/eadc64f8-bc32-11e5-9a1b-f97b357c5392.png)

david-yz-liu:
That's the issue I had. I'm not sure if @tsimica had a different issue though.

PR #2413, opened 2016-01-16
Fix for #2291. Added a new css class for the paragraph part so that i…

alexinman:
Added a new css class for the paragraph part so that it wraps correctly. Fix for #2291


app/assets/stylesheets/assignments.scss
```diff
@@ -189,6 +190,11 @@ input#assignment_repository_folder[readonly] {
     }
   }
 
+  .prop_paragraph {
```
alexinman: I used the same format as other selectors in the file.

Issue #2272, opened 2015-09-19
Confirmation for group member disinvite does not work


WilsonChiang:
Currently, there is code in place for confirming that you want to disinvite a group member in `assignments/_member.html.erb`:

```
     <%= link_to I18n.t(:cancel), '#',
        onclick: "
          new Ajax.Request('" +
             disinvite_member_assignment_path(id: @assignment.id, membership: member.id) + "');",

        confirm: I18n.t("student.cancel_invitation"),
        title: I18n.t("student.cancel_invitation_title"),
        class: "delete",
        remote: true
        %>
```
However, the confirmation window never happens.

WilsonChiang:
Looks like this is related to having both jQuery and Prototype.

See: http://stackoverflow.com/a/4988142

WilsonChiang:
Looks like the flash message that is apart of this does not load either in `assignments_controllers.rb`:

```
flash[:edit_notice] = I18n.t('student.member_disinvited')
```

alexinman:
@david-yz-liu Should the flash message look like this?
![screen shot 2016-01-16 at 1 07 47 pm](https://cloud.githubusercontent.com/assets/6303087/12373750/3c238476-bc52-11e5-8426-37a27e6cbcef.png)

david-yz-liu:
That looks great.

PR #2416, opened 2016-01-16
Fix for #2272

alexinman:
Fix for #2272. Changed the confirm to be a data-confirm so the confirm box actually shows up.


alexinman:
This fix isn't finished yet. Missed the last comment about the flash notice. Looking into that now.

PR #2427, opened 2016-01-17
Initial refactor of stylesheets

arkon:
This is the start of the work needed to use a single stylesheet instead of multiple different ones for each controller, to better leverage browser caching (download once; fetch from cache after that for every page).

Other misc. changes:
- Better organization for vendor styles (jQuery UI, pdf.js, etc.)
- Added a simple 12-column grid system (similar to Bootstrap's, but far simpler in terms of functionality)
- Automatically add the menu bar spacer for heading buttons instead of having the `<span class="menu_bar"></span>" elements

Future changes:
- General styling changes for everything -- since the stylesheets were separate before, there are conflicting styles now that they've "merged" together
- Cleaner layouts using the grid row/columns


arkon:
I don't think any of these Hound comments are actually worth dealing with at this point in time..

david-yz-liu:
Hi @arkon, I think this is a really great approach! However, I can't review and merge in a change this large. I would request that you not delete the controller-specific files yet, and instead work on refactoring to pull out common styles so that the files eventually are empty. You could also reorganize the vendor styles and introduce the 12-column layout in separate pull requests.

arkon:
@david-yz-liu Good idea. Could you also make a separate branch in upstream for me to make PRs to?

david-yz-liu:
@arkon yes, definitely - sorry, I meant to mention that as well and forgot. It's called `css-refactor`.

PR #2428, opened 2016-01-18
Change the grades view for grade_entry_form to use assignment_content layout

Sm1th:
Also add css to align the Release/Unrelease Marks buttons correctly, according to where they are in the browse view for submissions (http://0.0.0.0:3000/en/assignments/1/submissions/browse)

### DIFF
--- a/app/assets/stylesheets/grade_entry_forms.scss
+++ b/app/assets/stylesheets/grade_entry_forms.scss
@@ -43,6 +43,14 @@ fieldset {
   }
 }
 
+.react-release-marks {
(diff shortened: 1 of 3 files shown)


david-yz-liu:
Thanks, @Sm1th! This is exactly what I was looking for.

b2simms:
View it via: http://0.0.0.0:3000/en/grade_entry_forms/1/grades 
The grade_entry_form is rendered for a "Marks Spreadsheet" and does not seem to be for an assignment.

PR #2429, opened 2016-01-21
Remove repeated tags from page headers

arkon:


david-yz-liu:
Thanks, @arkon! I'll put this into the `css-refactor` branch, too.

Issue #2402, opened 2016-01-15
Rspec: Many faling tests


aclar48:
Currently we have 46 rspec tests failing

aclar48:
#2431

PR #2431, opened 2016-01-21
#2402 Fix all non csv related failing tests.

aclar48:
This fixes all failing rspec tests except for the ones in assignment related to csv files which @kevinmpm is taking care of along with this refactoring. #2402


david-yz-liu:
Thanks, @aclar48. I will soon submit a PR to fix the majority of failing rake tests, so please don't start working on those just yet.

lib/repo/git_repository.rb
```diff
@@ -30,15 +30,15 @@ class GitRepository < Repository::AbstractRepository
     # created using GitRespository.create(), it it is not yet existent
     def initialize(connect_string)
       # Check if configuration is in order
-      if Repository.conf[:IS_REPOSITORY_ADMIN].nil?
+      if IS_REPOSITORY_ADMIN.nil?
```
reidka: This has come up in another context.  There seem to be three ways to get this info:  Environment variable IS_REPOSITORY_ADMIN, Repository.conf[:IS_REPOSITORY_ADMIN], and MarkUsConfigurator.markus_config_repository_admin

I'd prefer to get rid of Repository.conf altogether because there is an issue with when it is created, and it really does seem like global configuration.  But, which of the other two is "better"?

aclar48: I don't have a strong preference between globals or MarkUsConfiguator, both are better than Repoistory.conf which doesn't work in some situations and never works with Travis, I'll update the file to use MarkUsConfiguator.

reidka: Thanks.  I'm getting rid of Repository.con altogether in another commit, but I'll wait to completely pull it out until after we merge your tests.

spec/models/assignment_spec.rb
```diff
@@ -1026,7 +1025,12 @@
           result = s.get_latest_result
           result.total_mark = total_mark
           result.marking_state = Result::MARKING_STATES[:complete]
-          result.save
+          @assignment.rubric_criteria.each do |cri|
+            result.marks.create!(markable_id: cri.id,
+                                 markable_type: RubricCriterion,
+                                 mark: (total_mark * 4.0 / 20).round)
```
david-yz-liu: The rounding invalidates the meaning of the `total_marks` array. On the other hand, the array could only be consistent with real marks if weights were assigned to the criteria. @aclar48, if you could clean up this test and create another one for flexible criteria, that would be much appreciated.

Issue #1935, opened 2014-12-09
PDF and image annotations do not work with premade annotations


david-yz-liu:
It seems the AnnotationTextManager class for both PDFs and images registers one annotation per text, which is problematic if the same text is used in multiple annotations (this is the whole point of premade annotations).

I think this is just a rendering problem, not a database problem: all annotations should appear in the annotations list.

wkwan:
Seems like its working. Added a bunch of a premade annotation and saw them all in the list.
![screen shot 2015-10-11 at 12 19 32 am](https://cloud.githubusercontent.com/assets/2907892/10414653/e9abc730-6fad-11e5-844e-1facb06bd07a.png)

reidka:
They all appear in the list but not on the pdf.

PR #2438, opened 2016-01-22
Fix for premade PDF and Image annotations not showing up

alexinman:
Changed image and PDF annotations to use the annotation id instead of the annotation text id (which was the same for all the premade annotations so they wouldn't show up because it was already there)
Fix for #1935


app/assets/javascripts/SourceCodeGlower/ImageAnnotationGrid.js
```diff
@@ -116,22 +116,26 @@ ImageAnnotationGrid.prototype.add_to_grid = function(extracted_coords) {
...
   }
 
   var annot_grid = this.get_annotation_grid();
+  var otherAnnotationsWithText = false;
   for (var i = 0; i < annot_grid.length; i++) {
-    if (annot_grid[i].id == annotation_text_id) {
+    if (annot_grid[i].annot_id == annotation_id) {
       annot_grid.splice(i, 1);
-      break;
+    }else if(annot_grid[i].id == annotation_text_id){
```
david-yz-liu: Please add spaces throughout to be consistent with the rest of the code here - e.g., after `}`, before `(` and `{`. Also make sure to use `===`.

app/models/image_annotation.rb
```diff
@@ -16,7 +16,7 @@ class ImageAnnotation < Annotation
   def extract_coords
     horiz_range = { start: [x1, x2].min, end: [x1, x2].max }
     vert_range = { start: [y1, y2].min, end: [y1, y2].max }
-    { id: annotation_text_id, x_range: horiz_range, y_range: vert_range }
+    { id: annotation_text_id, annot_id: self.id, x_range: horiz_range, y_range: vert_range }
```
david-yz-liu: This change breaks two tests in `test/unit/submission_file_test.rb`. Please update them and run the tests in this file (`rake test TEST=test/unit/submission_file_test.rb`) to make sure they're all okay.

Then repeat for `spec/models/image_annotation_spec.rb`.

david-yz-liu:
Thanks, @alexinman! It all looks good; I made a couple of minor comments for you to take a look at.

app/assets/javascripts/SourceCodeGlower/ImageAnnotationGrid.js
alexinman: Why should I use `===` over `==`? The original code used `==` and not `===`.

david-yz-liu: You're right, please change the one above as well.

We generally want to use `===` all throughout the code, since it is stricter (doesn't implicitly convert types). You can read more about this [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness) if you like.

PR #2440, opened 2016-01-23
Move core CSS files to subdirectories

arkon:
- Renamed "common" to "core"
- Moved mixins and variable partials into "utils" folder
- Cleaned up mixins and variables partials


app/assets/stylesheets/tas.scss
```diff
@@ -1,4 +1,4 @@
-@import 'common/core';
+@import 'core/core';
```
arkon: Weird, I thought the default was for `single_quotes`.....

david-yz-liu:
@arkon thanks. Do you know why the Github diffs for `_markus.scss` are so large? Was there a line ending issue before?

arkon:
@david-yz-liu That would be the only logical explanation, I guess. The contents weren't actually changed at all.

PR #2444, opened 2016-01-24
Changed the index view for assignments to use assignment_content layout

dmitrykhabarov:
Fix for issue 2442: Changed the index view for assignments to use assignment_content layout


david-yz-liu:
Thanks, @dmitrykhabarov!

PR #2445, opened 2016-01-24
CSS refactor: core styles file and components folder

arkon:
Moved parts of the core styling into a components folder, which needs to be cleaned up in a separate PR.

Also created a main `styles.scss` file, which is loaded in the head of each page for the core styling instead of being imported as a partial in every separate controller-specific file.


app/assets/stylesheets/styles.scss
```diff
@@ -0,0 +1,14 @@
+@charset "UTF-8";
```
arkon: Ignoring this because it has to be a double quoted string for `@charset`.

app/assets/stylesheets/core/_reset.scss
```diff
@@ -285,139 +174,117 @@ sub {
...
-/*
- * Corrects overflow displayed oddly in IE 9.
+/**
+ * Address differences between Firefox and other browsers.
  */
 
-svg:not(:root) {
-  overflow: hidden;
+hr {
+  -moz-box-sizing: content-box;
```
arkon: Ignoring since this is the standard reset file.

david-yz-liu:
@arkon this is exactly what I had in mind when I suggested incremental changes - thanks a lot.

One question - why don't we just use normalize.css? Or if we do, can we put it in `vendor`?

arkon:
@david-yz-liu The intention is so that if you include the single compiled CSS file (`styles.css`), it would have everything that you need, including the normalize reset.

david-yz-liu:
Okay, I suppose you are implying that you can't import a css file, only scss files. Can we still use the name `navigation.scss` and put it in vendor?

arkon:
Oh, I see what you're getting at now. Let me give that a shot later today.

arkon:
@david-yz-liu I moved the normalize file, since you can indeed import regular CSS files.

I'm also going to ignore the new Hound comments because none of those styles have actually been refactored yet.

david-yz-liu:
Cool, thanks @arkon!

PR #2453, opened 2016-01-27
Fix seeding error when using git backend

Sm1th:
Fixes a `Git::GitExecuteError (nothing to commit)` from the Git gem when seeding the database using a git backend, and cleans the code up a bit.

I'm not sure the reason for using both `Git` and `Rugged` gems. I would think we can only use one (If we can switch to only use one, I would suggest `Rugged`, as it's being actively maintained, unlike `Git`).

### DIFF
--- a/lib/repo/git_repository.rb
+++ b/lib/repo/git_repository.rb
@@ -27,7 +27,7 @@ module Repository
 
     # Constructor: Connects to an existing Git
     # repository, using Ruby bindings; Note: A repository has to be
-    # created using GitRespository.create(), it it is not yet existent
+    # created using GitRespository.create(), if it is not yet existent
     def initialize(connect_string)
       # Check if configuration is in order
       if MarkusConfigurator.markus_config_repository_admin?.nil?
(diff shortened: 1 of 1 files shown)


lib/repo/git_repository.rb
```diff
@@ -110,25 +110,10 @@ def self.create(connect_string)
...
       cloned_repo.reset
       cloned_repo.branch('master')
 
-      # Create the README.md on the filesystem
-      file_path_for_readme = File.join(repo.workdir, 'README.md')
-      File.open(file_path_for_readme, 'w+') do |readme|
-        readme.write('Initial commit.')
-      end
-
-      cloned_repo.add(all: true)
```
david-yz-liu: @Sm1th An initial commit is necessary for adding a folder for this assignment, for this group.

It seems like there is a file being created and staged in the code above, but somehow it's not getting committed. Please investigate this further to see why the file is not being staged (or committed).

reidka: It looks like the clone captures the file commits for the seeding.  So there is nothing to add and commit.  (Doing some more testing now.)

Sm1th: @reidka It's because `changed_filenames_at_path` is only defined for subversion repositories

reidka: As an admin, go to the Submissions view for A1.  Click on a link in the Repository column to go to the repository view.  This is where I get the error.

Yes, the changed_filenames_at_path needs to be implemented for git.  (It is something David added for svn to reduce the number of unnecessary revision entries in the drop down menu for selecting a revision.)

reidka:
I started with a clean gitolite setup and an empty repo directory, and this looks like it does fix the seeing problem.

Issue #82, opened 2010-09-29
Investigate Erronous Grader View File Truncation


benjaminvialle:
From Byron:

The grading view drops the last line or two of the file being displayed. It looks like there might be some kind of blank line suppression going on that may be interacting with it.

PR #2454, opened 2016-01-28
Fix error when creating new key-pair (git)

Sm1th:
Now you can upload public keys to your heart's desire.


reidka:
This looks good.  If I don't paste anything in or browse to a file, I get an error:
'' is not a valid SSH key string
Extracted source (around line #82):

    # Check to see if an individual repo exists for this user
    key = Gitolite::SSHKey.from_file(_path)

    ga_repo.add_key(key)

Also, the key file doesn't get deleted if I delete the key from the web interface.  Is that for another PR?

Sm1th:
@reidka I will fix these in two separate pull requests

Sm1th:
@reidka What should happen when a user inputs an invalid SSH key (such as an empty string)? Should there be an error message flashed?

reidka:
@Sm1th Yes, an error message would be helpful.

app/controllers/key_pairs_controller.rb
```diff
@@ -114,47 +115,53 @@ def create
...
-        format.html { render action: 'new' }
-        format.json do
-          render json: @key_pair.errors,
-                 status: :unprocessable_entity
+      end
+    else # if type and/or blob are nil
+      respond_to do |format|
+        format.html do
+          redirect_to :back,
+                      flash: { error: 'Key pair was not created: Invalid key.' }
```
reidka: Two comments:  First, please set up the message using flash_message (from helpers/application_helper.rb).  We are trying to move towards using this helper function more.  Second, the text of the message itself should go in the language files.  You'll need to add a key for it.

app/controllers/key_pairs_controller.rb
```diff
@@ -192,7 +199,10 @@ def destroy
     @key_pair.destroy
 
     respond_to do |format|
-      format.html { redirect_to key_pairs_url }
+      format.html do
+        redirect_to key_pairs_path,
+                    notice: 'Key pair was successfully removed.'
```
reidka: As above, the message should go in the language file.

Sm1th:
This pull request also includes fixes for issue #2455

reidka:
@Sm1th  Looks good.  Please add the Portuguese language keys, and then I will merge it.  (You can just use google translate for place holder strings.)

Sm1th:
@reidka Done :)

Issue #2458, opened 2016-01-29
NoMethodError in Assignments#edit


alexinman:
NoMethodError when I tried to view an assignment details as an admin. 
I created a bunch of sections and added students to them all. Except 2.
Then I created a new assignment and had different due dates for the sections.
Then I deleted the 2 sections that didn't have any students.
Then I went to view the assignment I had just created, which resulted in this error.
```
NoMethodError in Assignments#edit
Showing /home/vagrant/Markus/app/views/assignments/_form.html.erb where line #97 raised:

undefined method `name' for nil:NilClass
Extracted source (around line #97):
95              <p>
96              <%= due_date_f.label :section, t('section.legend') %>
97              <%= obj.section.name %>
98              <%= due_date_f.text_field :section_id, class: 'hidden' %>
99              <%= due_date_f.label :due_date, t(:due_date),
100                                  class: 'section_due_date_label' %>

Trace of template inclusion: app/views/assignments/edit.html.erb

Rails.root: /home/vagrant/Markus

Application Trace | Framework Trace | Full Trace
app/views/assignments/_form.html.erb:97:in `block (2 levels) in _app_views_assignments__form_html_erb___2348775664158113469_33868300'
app/views/assignments/_form.html.erb:92:in `block in _app_views_assignments__form_html_erb___2348775664158113469_33868300'
app/views/assignments/_form.html.erb:18:in `_app_views_assignments__form_html_erb___2348775664158113469_33868300'
app/views/assignments/edit.html.erb:5:in `_app_views_assignments_edit_html_erb__1715070396227624383_52934540'
Request

Parameters:

{"locale"=>"en",
 "id"=>"5"}
Toggle session dump
Toggle env dump
Response

Headers:

None
```
I will look into this.

alexinman:
Looks to be caused because the `section_due_date` record isn't being deleted when you delete a section. So there are orphaned records that are being retrieved for the assignment and then when it tries to find the section, it crashes.

PR #2459, opened 2016-01-29
Now destroys section_due_dates when destroying sections.

alexinman:
Fix for #2458. Now destroys section_due_dates when destroying sections.


david-yz-liu:
Perfect, thanks @alexinman.

PR #2461, opened 2016-01-31
Move view stylesheets to subdirectory

arkon:
Also use application.scss as the main file.


david-yz-liu:
Sorry, I should be looking at these more closely. Now I'm a little uneasy.

1) You have an `@param` but it should just be `param`
2) (in a previous PR) there's a variable called `$dimen-horizontal-mobile` which should actually be called `$dimen-hor-mobile`.

arkon:
@david-yz-liu: Whoops, that's the result of some bad merges on my part (which have now been fixed). I do intend on addressing issues that arise from the reorganization/refactor (i.e. the styling issues) later on though.

david-yz-liu:
Okay thanks @arkon!

PR #2462, opened 2016-02-01
One more performance fix for submissions page

aclar48:
Better formatted query, improves page load time also fixed n+1 regarding annotations


david-yz-liu:
Thanks, @aclar48!

PR #2466, opened 2016-02-03
Separate out component styles into partials

arkon:
Better separation of different common components into Sass partials.


arkon:
The other Hound comments will be addressed in later PRs that actual deal with the markup/styling. This PR is just for reorganizing things.

david-yz-liu:
Thanks, @arkon!

PR #2469, opened 2016-02-05
2468 - improve db:group performance

aclar48:
This seems to half the run time of db:seed when using git


reidka:
That is a definite improvement.  It seems to be creating 3-4 repos per minute rather than 2(ish).

PR #2472, opened 2016-02-09
pdf_storage/pdf_support removal

RFrailich:


david-yz-liu:
Hi @RFrailich, there are some changes in this commit which shouldn't be there (.ruby-version, Gemfile, dbout). You should also delete the keys from `markus_configurator.rb`

RFrailich:
.ruby-version and dbout can go in git ignore?

david-yz-liu:
No, you should not have staged them to be added to a commit.

However, at this point it would be easiest if you created another commit to revert those changes.

app/controllers/results_controller.rb
```diff
@@ -333,8 +333,7 @@ def codeviewer
     # if dealing with a pdf file, get the number of images to display
     if @file.is_pdf?
```
david-yz-liu: @RFrailich It seems this whole block can be removed, as it is calculating something which doesn't make any sense. Please also look for uses of the variable `@nb_pdf_files_to_download` and see if you can delete any of those.

app/models/submission_collector.rb
```diff
@@ -238,20 +238,9 @@ def manually_collect_submission(grouping, rev_num,
     self.safely_stop_child_exited = true
     self.save
 
-    #Let the child process handle conversion, as things go wrong when both
-    #parent and child do this.
-    start_collection_process do
```
david-yz-liu: The call to `start_collection_process` should be kept, as it is what... starts the collection process. I would keep the do block, but remove the checks inside which refer to the unused variables.

RFrailich: IS this still necessary? (from line 249)
    #setting is_collected here will prevent an sqlite lockout error when pdfs
    #aren't supported
      grouping.is_collected = true
      grouping.save

elec
```diff
@@ -0,0 +1,46 @@
+                  List of relations
```
david-yz-liu: @RFrailich can you remove this file as well?

david-yz-liu:
Hi @RFrailich, it seems there are references to removed code in the tests, some of which are now failing. Can you please fix the tests as well?

PR #2473, opened 2016-02-12
Added button to run tests as admin/grader

alexvail:
Added a button to run the tests.
Fixed the test interface for admins/graders as previous one was obsolete.
This is a first iteration obviously there are things that can be improved. I'll be waiting for feedback.

### DIFF
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -136,10 +136,12 @@ class ResultsController < ApplicationController
   def run_tests
     grouping_id = params[:grouping_id]
     begin
-      AutomatedTestsHelper.request_a_test_run(grouping_id, 'request', @current_user)
-    rescue Exception => e
-      #TODO: really shouldn't be leaking error if student.
-      return e.message
+      AutomatedTestsHelper.request_a_test_run(grouping_id,
+                                              'submission',
+                                              @current_user)
+    rescue => e
+      # TODO: really shouldn't be leaking error if student.
+      flash_message(:error, e.message)
     end
     redirect_to :back
   end
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -47,9 +47,8 @@ class ResultsController < ApplicationController
     @files = @submission.submission_files.sort do |a, b|
       File.join(a.path, a.filename) <=> File.join(b.path, b.filename)
     end
-    # This is outdated
-    #@test_result_files = @submission.test_results
-    # ----
+    @test_results = TestResult.where(grouping:
+      @grouping).order(created_at: :desc)
     @first_file = @files.first
     @extra_marks_points = @result.extra_marks.points
     @extra_marks_percentage = @result.extra_marks.percentage
@@ -132,6 +131,17 @@ class ResultsController < ApplicationController
         render nothing: true
       end
     end
+  end
+
+  def run_tests
+    grouping_id = params[:grouping_id]
+    begin
+      AutomatedTestsHelper.request_a_test_run(grouping_id, 'request', @current_user)
+    rescue Exception => e
+      #TODO: really shouldn't be leaking error if student.
+      return e.message
+    end
+    redirect_to :back
   end
 
   ##  Tag Methods  ##
--- a/app/views/automated_tests/_test_script_result.html.erb
+++ b/app/views/automated_tests/_test_script_result.html.erb
@@ -7,9 +7,9 @@
       <%= l(test_script_result.created_at, format: :long) %>
     </span>
   </h3>
-  <% tests_scripts_outputs = test_script_output['testrun']['test_script'] %>
-  <div class='table'>
 
+  <% tests = test_script_output['testrun']['test_script']['test'] %>
+  <div class='table'>
     <table>
       <thead>
         <tr>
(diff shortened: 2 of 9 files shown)


app/views/automated_tests/_test_script_result.html.erb
```diff
@@ -7,9 +7,9 @@
       <%= l(test_script_result.created_at, format: :long) %>
     </span>
   </h3>
-
-  <% tests = test_script_output['testrun']['test_script']['test'] %>
+  <% tests_scripts_outputs = test_script_output['testrun']['test_script'] %>
```
david-yz-liu: This change suggests that there are multiple outputs and/or test scripts here. I don't think that this is the proper use of this view right now. I now get an error on line 26 which didn't occur before.

Please revert this change. We will deal with multiple test scripts later.

app/views/results/common/_flexible_marks_summary_table.html.erb
```diff
@@ -26,7 +26,7 @@
     <tr class="<%= cycle('mark_summary_holderA', 'mark_summary_holderB') %>" id="<%= "test_script_summary" %>">
         <td><%=t("summary_table.test_script_title")%></td>
         <td><%= h(@assignment.total_test_script_marks * 1.0) %></td>
-        <td><span id="test_script_summary_mark"><%= h(@submission.result.get_total_test_script_marks * 1.0) %></span> / <%=  h(@assignment.total_test_script_marks * 1.0) %></td>
+        <td><span id="test_script_summary_mark"><%= h(@result.get_total_test_script_marks * 1.0) %></span> / <%=  h(@assignment.total_test_script_marks * 1.0) %></td>
```
david-yz-liu: Was there a reason for this change besides just cleaning up the code?

I approve of the change, but would have preferred to see it in a separate pull request to master if it is independent of the test-framework project.

app/views/results/marker/_marker_panes.html.erb
```diff
@@ -48,10 +48,31 @@
         <% if can_show_test_results_tab?(assignment, submission) %>
           <div id='test_result_tab' style='display:none;'>
             <div id='testviewer'>
-              <pre id='test_result_content' cols='80' rows='50'><%= t('marker.select_a_test') %></pre>
+              <%= link_to t("automated_tests.run_tests"),
+              run_tests_assignment_submission_result_path(
```
david-yz-liu: Please vertically align arguments for function calls.

app/views/results/marker/_marker_panes.html.erb
```diff
@@ -48,10 +48,31 @@
...
+              grouping_id: @grouping.id),
+              class: 'button' %>  
+            <% if !@test_results.nil? %>
```
david-yz-liu: Please use `unless` instead of `if !`, and indent the body of the if.

app/controllers/results_controller.rb
```diff
@@ -134,6 +133,19 @@ def edit
...
 
+  def run_tests
+    grouping_id = params[:grouping_id]
+    begin
+      AutomatedTestsHelper.request_a_test_run(grouping_id,
+                                              'request',
+                                              @current_user)
+    rescue => e
+      # TODO: really shouldn't be leaking error if student.
+      return e.message
```
david-yz-liu: This isn't the appropriate behaviour for a method which is directly responding to a request. Instead, do `flash_message(:error, e.message)`

app/controllers/results_controller.rb
```diff
@@ -134,6 +133,19 @@ def edit
     end
   end
 
+  def run_tests
+    grouping_id = params[:grouping_id]
+    begin
+      AutomatedTestsHelper.request_a_test_run(grouping_id,
+                                              'request',
```
david-yz-liu: Let's use `'submission'` instead. (Note that the tests will have to be enabled with "run on submission"). This will be something which differentiates a student and instructor running tests.

app/views/results/common/_flexible_marks_summary_table.html.erb
alexvail: Yes, I was getting "undefined method `result'" from _rubric_marks_summary_table.html.erb, so I changed this file too at the same time. If you look at view/results/marker/_marker_panes.erb.html at line 133, only result get passed in as local variables.

alexvail:
I encountered some issues with Git (I tried to revert one of the file to its previous state using git reset xxxx path, however that caused problem),had to do a git push -f, since I was getting "Updates were rejected because the tip of your current branch is behind hint: its remote counterpart. Integrate the remote changes". Before doing that I tried to pull the latest commit and then push again but for some reasons it didn't worked.
Let me know if I should've done something different.
Thank you

app/views/results/common/_flexible_marks_summary_table.html.erb
david-yz-liu: @alexvail the issue was not really what you said, but the fact that the submission model doesn't have a "result" method/attribute - see the `models/submission.rb` file. But anyway, the change is good.

app/views/automated_tests/_test_script_result.html.erb
david-yz-liu: @alexvail Okay I looked into this more. I got an error when you reverted the change, and I'm not sure why I wasn't getting that error before. In any case, I put in a simple fix, which you can see in @57edf2c15fa5eff458b61b966921ed20f2f3a460. We will talk later about handling the different levels of test results.

david-yz-liu:
@alexvail I think you wanted to do a `git revert` to add a new commit to undo some previous changes: https://www.atlassian.com/git/tutorials/undoing-changes/git-revert.

PR #2476, opened 2016-02-15
Fix for viewing git repositories

### LINKED ISSUE #2456: NoMethodError when trying to browse submission repo when admin
To reproduce:

1. Be logged in as admin
2. Go to submissions tab for an assignment  
    ex: http://0.0.0.0:3000/en/assignments/3/submissions/browse
3. Click on a **git** repository
4. Receive this error:
screen shot 2016-01-28 at 1 50 05 pm

I will look into this bug.

Sm1th:
This is still a work in progress... This fixes #2456, but contents of subfolders do not appear (for ex., no files/folders appear in 'A1/a')


lib/repo/git_repository.rb
```diff
@@ -1088,34 +1092,6 @@ def last_modified_date()
...
-        if type == :file
-          last_modified_date = @repo.__get_node_last_modified_date(File.join(path, file_name), @revision_number)
-          last_modified_revision = @repo.__get_history(File.join(path, file_name), nil, @revision_number).last
-
-          if(!only_changed || (last_modified_revision == @revision_number))
-            new_file = Repository::RevisionFile.new(@revision_number, {
-                                                      :name => file_name,
-                                                      :path => path,
-                                                      :last_modified_revision => last_modified_revision,
-                                                      :changed => (last_modified_revision == @revision_number),
```
reidka: I'm not sure whether this line is correct, but it looks like it was the only place that "changed" could be something other than true.

lib/repo/git_repository.rb
```diff
@@ -1077,9 +1077,13 @@ def path_exists?(path)
       true
     end
 
-    # Return changed files at 'path' (recursively)
+    # Return changed files at 'path'
     def changed_files_at_path(path)
-      return files_at_path_helper(path, true)
+      files = files_at_path(path)
```
reidka: files_at_path doesn't include directories.  So I think you need to explicitly check the directories_at_path too.

Sm1th:
Can now see immediate subfolders of assignments, but if you try to go deeper you get errors.
Ex: can see contents of A1/a fine, but can't see contents of A1/a/b. Problem is with the `revision.directories_at_path(previous_path)` call in the `get_exit_directory` function in `submissions_helper.rb`. `directories_at_path` needs to be called with a full path.

Sm1th:
All fixed! View contents of folders at any depth.

reidka:
@Sm1th I'm not sure that the selection by revision date works correctly.  The top level files seem to be okay.  In subdirectories, it seems to get the right files, but the date displayed doesn't look right.  I haven't experimented with different versions of files yet, but I will.
markus-file-browser

Sm1th:
@reidka I've made some bug fixes that I found.

My dates are not formatted like that though... I get this format: `3/11/2016, 3:09:00 PM`. Is that due to my computer settings? 

I think it's getting the right dates now, I'm just not exactly sure how the date is being formatted... It's suppose to be using the long_date time format in the language files, but the format in your screenshot and the format I see don't match that.

Sm1th:
The date is getting formatted again in `_repo_files_table.js.jsx.erb`, line 87:

```javascript
f.last_revised_date = new Date(file.last_revised_date).toLocaleString();
```

Shouldn't formatting be done only once? If so, which location should it be done in?

Sm1th:
@reidka This should be fixed now.

However, I think we should add seconds to the last revised date. What do you think?

reidka:
@Sm1th I completely agree that we should add seconds and timezone to last revised date.

reidka:
It would be good to fix the n+1 queries, but this now works well.  Thanks @Sm1th

PR #2478, opened 2016-02-20
Polling for Delayed Job Status/Completion

aclar48:


app/controllers/job_messages_controller.rb
```diff
@@ -0,0 +1,10 @@
+class JobMessagesController < ApplicationController
+
+  def show
```
david-yz-liu: This should be authorized only for admins.

app/views/groups/create_groups_when_students_work_alone.js.erb
```diff
@@ -0,0 +1,2 @@
+
+<%= render 'job_messages/job_poller', job_id: @current_job.job_id, processing_message: 'Creating groups for all students' %>
```
david-yz-liu: Please internationalize this and the other strings you used.

app/controllers/job_messages_controller.rb
```diff
@@ -0,0 +1,11 @@
+class JobMessagesController < ApplicationController
+  before_filter  :authorize_only_for_admin
+
+  def show
+    @job_message = Rails.cache.fetch(params[:job_id])
```
david-yz-liu: I wonder if there is a race condition here - the first time I tried out the button, I got an error because `@job_message` was nil. It seem this relies on the `before_enqueue` block in the job to execute first?

aclar48: I was under the impression that anything in before_enqueue ran synchronously which would ensure that a jobmessage existed. Were you able to replicate this issue?

PR #2479, opened 2016-02-22
Run tests when collecting all

alexvail:
When collecting all submissions from an assignment, a test run will also be run for every submission.

### DIFF
--- a/app/helpers/automated_tests_helper.rb
+++ b/app/helpers/automated_tests_helper.rb
@@ -378,22 +378,17 @@ module AutomatedTestsHelper
   def self.scripts_to_run(assignment, call_on)
     # Find all the test scripts of the current assignment
     all_scripts = TestScript.where(assignment_id: assignment.id)
-
     list_run_scripts = Array.new
 
     # If the test run is requested at collection (by Admin or TA),
     # All of the test scripts should be run.
-    if call_on == 'collection'
-      list_run_scripts = all_scripts
-    else
-      # If the test run is requested at submission or upon request,
-      # verify the script is allowed to run.
-      all_scripts.each do |script|
-        if call_on == 'submission' && script.run_on_submission
+    # If the test run is requested at submission or upon request,
+    # verify the script is allowed to run.
+    all_scripts.each do |script|
+      if (call_on == 'submission' && script.run_on_submission) ||
+         (call_on == 'request' && script.run_on_request) ||
+         (call_on == 'collection')
           list_run_scripts.insert(list_run_scripts.length, script)
-        elsif call_on == 'request' && script.run_on_request
-          list_run_scripts.insert(list_run_scripts.length, script)
-        end
       end
     end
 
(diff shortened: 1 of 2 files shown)


app/helpers/automated_tests_helper.rb
```diff
@@ -378,22 +378,17 @@ def self.delete_repo(repo_dir)
...
-    if call_on == 'collection'
-      list_run_scripts = all_scripts
-    else
-      # If the test run is requested at submission or upon request,
-      # verify the script is allowed to run.
-      all_scripts.each do |script|
-        if call_on == 'submission' && script.run_on_submission
+    # If the test run is requested at submission or upon request,
+    # verify the script is allowed to run.
+    all_scripts.each do |script|
```
david-yz-liu: @alexvail I refactored this code in a slightly different way. When you do a pull, you'll see a conflict here; please use my version.

app/helpers/automated_tests_helper.rb
```diff
@@ -404,7 +399,8 @@ def self.scripts_to_run(assignment, call_on)
 
   # Request an automated test. Ask Resque to enqueue a job.
   def self.async_test_request(grouping_id, call_on)
-    if files_available? && has_permission?
+    # when called on collection
+    if files_available? && (call_on == "collection" || has_permission?)
```
david-yz-liu: I don't know that `collection` should be checked here. If we look at how this is being called in submisison_collector, the `@current_user` is still an admin or TA, and so should have permission.

alexvail: Getting this error otherwise:
```
/home/vagrant/Markus/app/helpers/automated_tests_helper.rb:195:in `has_permission?': undefined method `admin?' for nil:NilClass (NoMethodError)
	from /home/vagrant/Markus/app/helpers/automated_tests_helper.rb:114:in `request_a_test_run'
```
which is caused from the call from submission_collector.rb

```
  if assignment.enable_test
      m_logger.log("Now requesting test run for #{assignment.short_identifier} \
                    on grouping: '#{grouping.id}'")
      AutomatedTestsHelper.request_a_test_run(new_submission.grouping.id,
                                              'collection',
                                              **@current_user**)
    end
```
Should we override the check, or otherwise what would be the prefered way to access/pass around @current_user?

alexvail:
Just a note: the error I discussed in an outdated diff is still relevant:
```
/home/vagrant/Markus/app/helpers/automated_tests_helper.rb:195:in `has_permission?': undefined method `admin?' for nil:NilClass (NoMethodError)
	from /home/vagrant/Markus/app/helpers/automated_tests_helper.rb:114:in `request_a_test_run'
```
which is caused from the call from submission_collector.rb

```
  if assignment.enable_test
      m_logger.log("Now requesting test run for #{assignment.short_identifier} \
                    on grouping: '#{grouping.id}'")
      AutomatedTestsHelper.request_a_test_run(new_submission.grouping.id,
                                              'collection',
                                             @current_user)
    end
```
Should we override the check, or otherwise what would be the prefered way to access/pass around @current_user?

david-yz-liu:
@alexvail well done. The other feature which is missing is running the tests automatically when manually collecting an individual submission (from the repo browser). If you could work on that too, that would be great.

PR #2483, opened 2016-03-03
Test framework multiple test files support

alexvail:


app/helpers/automated_tests_helper.rb
```diff
@@ -347,34 +347,40 @@ def self.process_result(raw_result)
...
-
-    completion_status = 'pass'
-    marks_earned = 0
     test_scripts.each do |script|
+      script_name = script['script_name']
+      test_script = TestScript.find_by(assignment_id: @assignment.id,
+                                       script_name: script_name)
+      completion_status = 'pass'
+      marks_earned = 0
+      submission_id = Submission.last.id
```
david-yz-liu: Keep the TODO - I'll have you fix this issue later.

app/models/test_result.rb
```diff
@@ -37,7 +37,8 @@ class TestResult < ActiveRecord::Base
   validates_inclusion_of :completion_status, :in => %w(pass fail error), :message => "%{value} is not a valid status"
   validates_numericality_of :marks_earned, :only_integer => true, :greater_than_or_equal_to => 0
 
-  validates_presence_of :input_description, :if => "input_description.nil?"
-  validates_presence_of :actual_output, :if => "actual_output.nil?"
-  validates_presence_of :expected_output, :if => "expected_output.nil?"
+  # input, actual_output and expected_output could be empty in some situations
```
david-yz-liu: This one is surprising - I would think the validation only occurs when the attribute is nil, as this is what the `:if` checks.

db/migrate/20160303184449_change_test_results_table.rb
```diff
@@ -0,0 +1,8 @@
+class ChangeTestResultsTable < ActiveRecord::Migration
+  def change
+    rename_column :test_results, :input_description, :input
+    change_column_null :test_results, :input, true
```
david-yz-liu: I think we should leave these columns as non-null, and set a default value of the empty string instead.

app/helpers/automated_tests_helper.rb
```diff
@@ -347,34 +347,38 @@ def self.process_result(raw_result)
...
-    test_script = TestScript.find_by(assignment_id: @assignment.id,
-                                     script_name: script_name)
-
-    completion_status = 'pass'
-    marks_earned = 0
     test_scripts.each do |script|
+      script_name = script['script_name']
+      test_script = TestScript.find_by(assignment_id: @assignment.id,
+                                       script_name: script_name)
+      completion_status = 'pass'
```
david-yz-liu: This variable doesn't need to be initialized here

david-yz-liu:
@alexvail I merged it in but left a comment. The other thing I remembered I wanted you to do was fix the association between test script results and submissions.

1) Submissions should be optional for test script results.
2) Students should only be shown test script results which aren't associated with a submission.
3) Collecting submissions, or pushing the test run button as an instructor/TA, should associate the created test script result with a particular submission.
4) The "Test Results" tab in the grading view should only show the test results associated with that particular submission.

I can't quite remember what the current state is, but there's definitely at least a little bit of work for you here.

PR #2486, opened 2016-03-06
Undo collections

aclar48:


reidka:
@aclar48 This is looking good.

The biggest problem is that there is no notification that collection or uncollection process is happening.  You have to know to refresh the page to see the results.  I think we have to wait until the collection process is moved over to ActiveJob to do this properly, but until then I'd like to go back to what we had before: the page is reloaded and a flash message saying that collection is in progress.

Also, I'd still like to see the popup warning on the "collect all submissions" button saying:
"This operation
 - Identifies the file version for each submission that meets the due date and/or late penalty set for this assignment.
- Creates the views for annotating and grading each submission.

Warning: If this operation is carried out after grading has begun, this grading work will be overwritten. This work can be recovered using the Undo Collection button.

-Submit-
-Cancel-
"

I would also like to see a popup on the Uncollect button (which I would like to rename to "Undo Previous Collection")
"This operation will undo the previous Collect All Submissions operation.  This should <b>only</b> be done if the previous Collect All Submissions operation was done in error. 

WARNING: All grading work done since the previous Collect All Submissions operation will be lost."

reidka:
I notice in the submissions table that if I collect and uncollect a few times, that we end up with multiple records with the same submission_version.  I have seen a case where the uncollect operation picks the wrong older version lead to a problem when looking for the previous version in the uncollect operation.  I think the simplest solution is to delete the old submission_version record when reverting.
(Let me know if this doesn't make sense)

Example:  In the file below, the uncollect operation picked id 33 with submission_version 2 as the one to go to, but the record with id 78 is newer.

[submission_table.txt](https://github.com/MarkUsProject/Markus/files/182965/submission_table.txt)

aclar48:
I can easily move the undo collection to a delayed job, the collection might a bit trickier as it seems for fork a bunch of processes for some reason.

reidka:
Sure, why don't you move the undo collection to an ActiveJob.  We have another student working on the collect all right now.

app/models/submission_collector.rb
```diff
@@ -274,18 +274,20 @@ def uncollect_submissions(assignment)
...
-        grouping.update_attributes(grouping_queue_id: nil)
+        grouping.update_attributes(grouping_queue_id: nil, is_collected: false)
+        # version = submission.submission_version
+        # grouping = submission.grouping
+        # if version == 1
+        #   grouping.assign_attributes(is_collected: false)
+        # else
+        #   prev_rev = submissions.where(submission_version: version - 1,
+        #                                grouping_id: grouping.id).first
+        #   prev_rev.update_attributes(submission_version_used: true)
```
reidka: This doesn't seem to have the correct behaviour.  If I collect all submissions, do some marking, and then uncollect, the previous collection is not restored. submission_version_used for the previously collected version is still false.

(I'm trying to debug it a bit and simply uncommenting this block doesn't seem to fix it.)

app/views/submissions/uncollect_all_submissions.js.erb
```diff
@@ -0,0 +1 @@
+<%= render 'job_messages/job_poller', job_id: @current_job.job_id, processing_message: I18n.t('collect_submissions.uncollection_processing_message') %>
```
reidka: I like how this works.

spec/models/submission_collector_spec.rb
```diff
@@ -10,29 +10,9 @@
...
-      last_version = versions[-1]
-      prev_version = versions[-2]
-      submissions.where(submission_version: last_version).each do |sub|
-        expect(sub.submission_version_used).to eq false
-      end
-      submissions.where(submission_version: prev_version).each do |sub|
-        expect(sub.grouping.is_collected).to eq true
-        expect(sub.submission_version_used).to eq false
-      end
+      expect(@assignment.submissions.size).to eq(0)
```
reidka: More tests?

PR #2492, opened 2016-03-12
2395 flexible criteria upload

kevinmpm:


reidka:
Here is a scenario where it doesn't seem to work.

1. create a new assignment with the flexible marking scheme
2. upload a valid csv file for the marking scheme  - This works fine
3. upload the same csv file again - Now I get the error "The following CSV lines were invalid:" for all lines of the file.

oddly enough, if I add a line to the file, I get both the error message and the success message and the new criterion appears.

kevinmpm:
@reidka 
I'm not sure which part is undesirable behaviour.
Should the entire transaction abort when there is an invalid row? For #3 on your list, each of the rows will already exist so they will all fail because they are duplicates. For the extra line, is it not possible to have both valid and invalid rows in a CSV? The way it was set up previously, if there were valid rows, they would be inserted into the DB even if there were invalid rows afterwards. The difference is now the user is notified of both.

reidka:
@kevinmpm then the error message should be "duplicate criterion detected" not invalid row. (For a while, I thought that no csv file would work at all.)
Thanks

david-yz-liu:
Thanks, @kevinmpm!

PR #2493, opened 2016-03-12
Orders Sections by due date

alexinman:
Fix for #1219.
On both Assignment>Edit and Assignment>Graders, the sections are now sorted by due date and then by name.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -368,3 +368,6 @@ DEPENDENCIES
   uglifier (>= 1.3.0)
   unicorn
   ya2yaml
+
+BUNDLED WITH
+   1.11.2
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -368,6 +368,3 @@ DEPENDENCIES
   uglifier (>= 1.3.0)
   unicorn
   ya2yaml
-
-BUNDLED WITH
-   1.11.2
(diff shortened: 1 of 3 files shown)


david-yz-liu:
@alexinman sorry I haven't looked at that issue in a while. The ordering on the Assignment form is good; but for the graders table, I'd prefer having the sections sorted alphabetically.

Gemfile.lock
```diff
@@ -368,6 +368,3 @@ DEPENDENCIES
   uglifier (>= 1.3.0)
   unicorn
   ya2yaml
-
```
david-yz-liu: Please don't include this change in the commit (applies to the other PR, too).

app/controllers/assignments_controller.rb
```diff
@@ -202,6 +202,9 @@ def edit
                                    .where.not(id: @assignment.id)
                                    .order(:id)
     @sections = Section.all
+    @section_due_dates = SectionDueDate.joins(:section)
```
david-yz-liu: It seems like this will be nil if the assignment has no section due dates (e.g., when the assignment is first created), leading this to be empty.

reidka:
Okay, I've stared at this for too long.  On the graders page the sections are ordered by name not due date.  I'm okay with this because visually I'd prefer seeing this list ordered by name.  It just doesn't match what is stated above.

[sections1.pdf](https://github.com/MarkUsProject/Markus/files/184763/sections1.pdf)
[sections2.pdf](https://github.com/MarkUsProject/Markus/files/184764/sections2.pdf)

david-yz-liu:
@reidka I asked @alexinman to sort the sections by name on the graders page.

reidka:
@david-yz-liu and @alexinman +1

Issue #2448, opened 2016-01-25
Sections don't appear in  Marks Spreadsheet


reidka:
The section column remains empty even when students are assigned to a section.

PR #2494, opened 2016-03-12
Fix for #2448 - Sections don't appear in Marks Spreadsheet

alexinman:
Fix for #2448. Added the student's section name to the mark spreadsheets. Also adds the group's section name to the summaries spreadsheet.


app/controllers/grade_entry_forms_controller.rb
```diff
@@ -162,6 +162,7 @@ def populate_grades_table
       s = student.attributes
       student_grade_entry = @grade_entry_form.grade_entry_students
                                              .find_by_user_id(student.id)
+      s[:section] = student.section.name
```
david-yz-liu: This is only good if the student is enrolled in a section; otherwise, it'll raise an error.

david-yz-liu:
Thanks, @alexinman!

PR #2499, opened 2016-03-18
2395 grade entry forms upload

kevinmpm:


app/controllers/grade_entry_forms_controller.rb
```diff
@@ -290,63 +290,48 @@ def csv_download
...
-                                    num_updates: num_updates)
+          if totals.empty?
+            totals = row
+            # Create/update the grade entry items
+            GradeEntryItem
+              .create_or_update_from_csv_rows(names, totals, @grade_entry_form)
+            next
           end
+          GradeEntryStudent
+            .create_or_update_from_csv_row(row,
```
reidka: I'd like to see more error checking here.  If the row with marks in it is longer than the names or totals row, there should be an error message.

PR #2501, opened 2016-03-19
2395-rubrics-upload

kevinmpm:


reidka:
I tried this out, but immediately got the following error.  Here are the steps I took:
1. Create a new assignment
2. Download the rubric from A1 as CSV
3. Upload the rubric into the new assignment


ActiveRecord::RecordNotUnique in RubricsController#csv_upload
PG::UniqueViolation: ERROR: duplicate key value violates unique constraint "rubric_criteria_pkey" DETAIL: Key (id)=(10) already exists. : INSERT INTO "rubric_criteria" ("rubric_criterion_name", "assignment_id", "weight", "position", "level_0_name", "level_1_name", "level_2_name", "level_3_name", "level_4_name", "level_0_description", "level_1_description", "level_2_description", "level_3_description", "level_4_description", "created_at", "updated_at") VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16) RETURNING "id"
Extracted source (around line #155):

153
154
155
156
157
158
              

	

      criterion['level_' + i.to_s + '_description'] = working_row.shift
    end
    unless criterion.save
      raise CSVInvalidLineError
    end
    criterion

Rails.root: /home/vagrant/Markus
Application Trace | Framework Trace | Full Trace

app/models/rubric_criterion.rb:155:in `create_or_update_from_csv_row'
app/controllers/rubrics_controller.rb:82:in `block (2 levels) in csv_upload'
lib/classes/markus_csv.rb:34:in `call'
lib/classes/markus_csv.rb:34:in `block in parse'
lib/classes/markus_csv.rb:32:in `parse'
app/controllers/rubrics_controller.rb:80:in `block in csv_upload'
app/controllers/rubrics_controller.rb:79:in `csv_upload'

kevinmpm:
@reidka 
Interesting. I tried doing the same thing with the old code, and the same error occurs, so this issue wasn't introduced by the refactoring. Nevertheless, it should be fixed. I looked into the code and the original functionality includes this:
# If a RubricCriterion of the same name exits, load it up.  Otherwise,
    # create a new one.
    criterion = assignment.rubric_criteria.find_or_create_by(
      rubric_criterion_name: rubric_criterion_name)

I'm assuming this should only find criterion in the the assignment that is being modified. However, it seems to be looking across all assignments. Can you confirm that it should only look at the current assignment? I can change that.

david-yz-liu:
@kevinmpm Yes, the name uniqueness should be scoped to the current assignment.

kevinmpm:
I spent too long trying to figure out this one.. Turns out it wasn't looking across all assignments for entries, it only looks at the current one. The reason @reidka  and I were having issues is because the DB wasn't being seeded properly, the id's were being set manually causing the DB to be unstable

david-yz-liu:
@kevinmpm this saved me a lot of time, so thanks for figuring out that odd bug. :)

Issue #2449, opened 2016-01-25
Add the ability to assign a grader to a section


reidka:
This will take some thought about what the UI should look like.

alexinman:
Hey, just read this one now. This is a brand new feature, correct?
Where would the UI for this be and where/how would it affect everything?
We can talk on Monday's meeting for more details, if that would be easier than explaining everything here. :)

alexinman:
@david-yz-liu So, just clarifying what you said yesterday.
Currently this is what the Assignment>Graders Tab looks like, with the ability to filter by sections:
![screen shot 2016-03-15 at 9 33 02 am](https://cloud.githubusercontent.com/assets/6303087/13777863/5b5d7280-ea91-11e5-9b82-2b8f4860e4ac.png)
Do you want me to add that functionality to the Quiz>Graders Tab? Which currently looks like this:
![screen shot 2016-03-15 at 9 32 35 am](https://cloud.githubusercontent.com/assets/6303087/13777879/74faabd6-ea91-11e5-8f85-e1f6c0100095.png)
And would you also like me to add it to the Marks Spreadsheets? Which currently looks like this:
![screen shot 2016-03-15 at 9 32 27 am](https://cloud.githubusercontent.com/assets/6303087/13777888/87d60f48-ea91-11e5-9794-ce2d29d7bd52.png)

david-yz-liu:
1) The sections should appear in the graders tab.
2) The "Grades" table should not have filters, but it should be possible to sort by the Section column.

alexinman:
Alright :+1:

PR #2503, opened 2016-03-19
Ability to Add a Grader to a Section

alexinman:
#2449 

Adds the ability to filter students by section in the Quiz>Graders tab. 
Also adds the ability to sort the Quiz Marks Spreadsheets by section.

screen shot 2016-03-19 at 1 07 48 pm
screen shot 2016-03-19 at 1 07 12 pm
screen shot 2016-03-19 at 1 07 01 pm

@david-yz-liu Is there anything else you would like done for #2449?


reidka:
What I want to be able to do is to select a section to filter by, and the click on the select all checkbox, the grader I want to grade that section, and add the grader only to students in that section.

E.g. in the picture above, for Manage Graders, 
- select section c5 to display students in the section c5.
- check the checkbox next to UserName (so all the visible checkboxes turn blue), 
- check a grader, and add the grader to the students in the section.

Right now, this sequence of operations will add that grader to everyone.

reidka:
The sorting by section in the spreadsheets doesn't work for me, even though in firebug I can see that the GradesTable has sortable set to true for the sections column.

alexinman:
@reidka I can't replicate your issue with sorting by section in the spreadsheets

alexinman:
The issue with clicking the select all box is due to this line (and lines like it):
https://github.com/MarkUsProject/Markus/pull/2503/files#diff-2e11ca15fd9b8b5c85d7958ab4a499d1R18

It just selects or deselects all groups/students for the table and doesn't check the current filter or the visible_rows.

I don't know how to access the current filter or the visible_rows because they are handled by `app/views/assets/javascripts/ReactComponents/table.js` and the select all checkbox is in the `app/views/graders/_graders_manager.js.jsx.erb` and the `app/views/marks_graders/_marks_graders_student_table.js.jsx.erb` files.

@reidka or @david-yz-liu do you know how I can access the visible_rows from the `table.js` class in the graders manager and marks graders student tables?

app/views/graders/_graders_manager.js.jsx.erb
```diff
@@ -553,7 +558,8 @@
 
       // need to filter for each section
       var filters = [];
-      filters.push({name: '-', text: '-', func: function(g) { return g; } });
+      var all = '<%= j raw I18n.t('all') %>';
```
david-yz-liu: Let's not use `j` and `raw` here

app/views/graders/_graders_manager.js.jsx.erb
```diff
@@ -589,9 +595,9 @@
       }.bind(this));
 
       return (
-       <div className='tab-pane ui-tabs ui-widget ui-widget-content ui-corner-all' id='groups-tabs'>
-        <Table data={groups_data} columns={columns}
-               filters={filters} filter_type={false}
+        <div className='tab-pane ui-tabs ui-widget ui-widget-content ui-corner-all' id='groups-tabs'>
+          <Table data={groups_data} columns={columns} ref="myTable"
+               filters={filters} filter_type={filters.length>3}
```
david-yz-liu: I'm not sure what the significance of the "3" is.

app/views/graders/_graders_manager.js.jsx.erb
alexinman: Alright.. What does j and raw do? I just copied that from somewhere else..

app/views/graders/_graders_manager.js.jsx.erb
alexinman: In the table.js file is suggest to use the select filter type when there's more than 3 choices to filter by.

alexinman: SetFilter_type to false for simple filter type and true for the select (or fancy, I think its called) filter type

app/views/graders/_graders_manager.js.jsx.erb
david-yz-liu: http://apidock.com/rails/ActionView/Helpers/JavaScriptHelper/escape_javascript and http://apidock.com/rails/ActionView/Helpers/OutputSafetyHelper/raw.

Neither apply here.

app/views/graders/_graders_manager.js.jsx.erb
david-yz-liu: Hmm alright. Please put spaces around the `>`

PR #2504, opened 2016-03-19
2395-tags-controller-upload

kevinmpm:


app/controllers/tags_controller.rb
```diff
@@ -123,29 +123,21 @@ def csv_upload
...
-                            invalid_lines.join(', ')
-          end
-          if nb_updates > 0
-            flash[:success] = I18n.t('tags.upload.upload_success',
-                                     nb_updates: nb_updates)
-          end
+    if file
+      Tag.transaction do
+        result = MarkusCSV.parse(file.read, encoding: encoding) do |row|
+          next if CSV.generate_line(row).strip.empty?
```
david-yz-liu: I don't like using `next` here, the block should really just deal line by line. Why not keep an `if` check and put the following line in the "then" part?

PR #2507, opened 2016-03-20
2395-annotations-categories-download

kevinmpm:
Created spec tests for the assignments csv downloads


spec/controllers/assignments_controller_spec.rb
```diff
@@ -108,4 +107,73 @@
...
+      expect(@controller).to receive(:send_data)
+                               .with(csv_data.join(',') + "\n", csv_options) {
+        # to prevent a 'missing template' error
+        @controller.render nothing: true
+      }
+      get :download_assignment_list, file_format: 'csv'
+    end
+
+    # parse header object to check for the right content type
+    it 'returns vnd.ms-excel type' do
```
david-yz-liu: It seems like this description is not accurate?

PR #2508, opened 2016-03-21
Test framework restful api support

alexvail:


app/controllers/api/test_results_controller.rb
```diff
@@ -3,81 +3,72 @@ module Api
   # Allows for pushing and downloading of TestResults
   # Uses Rails' RESTful routes (check 'rake routes' for the configured routes)
   class TestResultsController < MainApiController
-    # Define default fields for index method
-    @@default_fields = [:id, :filename]
 
     # Returns a list of TesResults associated with a group's assignment submission
-    # Requires: assignment_id, group_id
+    # Requires: assignment_id, group_id, test script result id
```
david-yz-liu: Please add underscores here

app/controllers/api/test_results_controller.rb
```diff
@@ -3,81 +3,72 @@ module Api
...
       submission = get_submission(params[:assignment_id], params[:group_id])
       return if submission.nil?
 
-      collection = submission.test_results
-
-      test_results = get_collection(TestResult, collection)
-      fields = fields_to_render(@@default_fields)
+      test_results = submission.test_script_results
+                  .includes(:test_results)
+                  .find_by_id(params[:test_script_result_id])
```
david-yz-liu: Use `find` instead of `find_by_id`.

app/controllers/api/test_results_controller.rb
```diff
@@ -3,81 +3,72 @@ module Api
...
       # get_submission renders appropriate error if the submission isn't found
       submission = get_submission(params[:assignment_id], params[:group_id])
       return if submission.nil?
 
-      collection = submission.test_results
-
-      test_results = get_collection(TestResult, collection)
-      fields = fields_to_render(@@default_fields)
+      test_results = submission.test_script_results
+                  .includes(:test_results)
```
david-yz-liu: Please align the `.` across multiple lines

app/controllers/api/test_results_controller.rb
```diff
@@ -120,31 +113,25 @@ def destroy
     #  - filename: New name for the file
     #  - file_content: New contents of the test results file
     def update
-      # get_submission renders appropriate error if the submission isn't found
+            # get_submission renders appropriate error if the submission isn't found
```
david-yz-liu: Revert the space here

app/controllers/api/test_results_controller.rb
```diff
@@ -155,6 +142,13 @@ def update
       end
     end
 
+    # User params for create & update
+    def user_params
```
david-yz-liu: We generally try to name this based on the model, so `test_result_params` rather than `user_params`.

app/controllers/api/test_script_results_controller.rb
```diff
@@ -0,0 +1,190 @@
...
+      if test_script_result.nil?
+        render 'shared/http_status', locals: { code: '404', message:
+          'Test script result was not found'}, status: 404
+        return
+      end
+
+      grouping = submission.grouping
+      assignment = submission.assignment
+
+      if test_script_result.destroy &&
```
david-yz-liu: I'm not sure this is the best approach. If there's an error processing the new params, won't the existing test script result be removed?

I suppose you could process the params first, and then if that succeeds to the replacement.

app/controllers/api/test_script_results_controller.rb
```diff
@@ -0,0 +1,190 @@
...
+      else
+        # Some other error occurred
+        render 'shared/http_status', locals: { code: '500', message:
+          HttpStatusHelper::ERROR_CODE['message']['500'] }, status: 500
+      end
+    end
+
+    # Given assignment and group id's, returns the submission if found, or nil
+    # otherwise. Also renders appropriate responses on error.
+    def get_submission(assignment_id, group_id)
```
david-yz-liu: I'd prefer using an existing method in either the assignment or group model (can't remember which). While this does lose the specificity of where the error occurred (assignment id vs group id lookup, for example), I don't really care.

Look into something like `Grouping#current_submission_used`.

app/models/test_script_result.rb
```diff
@@ -19,13 +19,12 @@
 #############################################################
 
 class TestScriptResult < ActiveRecord::Base
-  belongs_to :submission
+  belongs_to :submission, required: false
   belongs_to :test_script
   belongs_to :grouping
 
-  has_many :test_results
+  has_many :test_results, :dependent => :destroy
```
david-yz-liu: Please use the new hash syntax, `dependent: :destroy`

app/helpers/automated_tests_helper.rb
```diff
@@ -435,4 +443,5 @@ def self.process_result(raw_result)
     end
 
   end
+
```
david-yz-liu: Please delete this space.

PR #2510, opened 2016-03-22
added token regeneration period with start of availability date

dmitrykhabarov:
Added token regeneration period with start of availability date


app/helpers/automated_tests_helper.rb
```diff
@@ -45,7 +51,7 @@ def process_test_form(assignment, params, assignment_params,
       # If no new_script then form is empty and skip
       next if testscripts[file_num][:seq_num].empty? && new_script.nil?
       # Seq_num only exists if it is a file being edited
-      if testscripts[file_num][:seq_num].empty?
+      if new_script
```
david-yz-liu: What is the purpose of this change? You should update the comment, as well.

app/models/grouping.rb
```diff
@@ -382,7 +382,7 @@ def invalidate_grouping
 
   # Token Credit Query
   def give_tokens
-    Token.create(grouping_id: self.id, tokens: self.assignment.tokens_per_day) if self.assignment.enable_test
+    Token.create(grouping_id: self.id, tokens: DateTime.now >= self.assignment.tokens_start_of_availability_date ? self.assignment.tokens_per_day : nil) if self.assignment.enable_test
```
david-yz-liu: Let's split this into multiple lines.

app/models/test_script.rb
```diff
@@ -74,23 +74,23 @@ class TestScript < ActiveRecord::Base
...
-  # validates the uniqueness of seq_num for the same assignment
-  validates_each :seq_num do |record, attr, value|
-    # FIXME: create a loop to loop through all dup_file
-    # dup_files = TestScript.find_all_by_assignment_id_and_seq_num(record.assignment_id, value)
-    dup_file = TestScript.find_by_assignment_id_and_seq_num(record.assignment_id, value)
-    if dup_file && dup_file.id != record.id
-      # FIXME: fix the error message: this is not filename
-      record.errors.add attr, ' ' + value.to_s + ' ' + I18n.t("automated_tests.filename_exists")
-    end
+  #   # FIXME: create a loop to loop through all dup_file
```
david-yz-liu: What happened with this one?

app/models/token.rb
```diff
@@ -25,18 +25,22 @@ def last_used_date
...
+      self.last_token_used_date = Time.now
     end
     self.save
   end
 
-  def reassign_tokens_if_new_day
-    if self.last_token_used_date and self.last_token_used_date < Date.today
+  def reassign_tokens_if_after_regen_period(regeneration_period)
+    if self.last_token_used_date
+      if (self.last_token_used_date.to_i + regeneration_period*60*60) <= Time.now.to_i
```
david-yz-liu: This logic seems a bit off. First, it's the "token regenerates X hours after using it" rather than "all tokens regenerate every X hours, offset from a fixed time." I don't mind that *so* much, except it seems from above that each time a test is run, the "last_token_used_date" attribute is updated. This seems to mean that the tokens won't regenerate until "X hours after the *last* token is used", which I definitely don't want.

app/views/automated_tests/_boot.js.erb
```diff
@@ -1,8 +1,36 @@
...
-  });
+
+  function check_start_of_availability_date(new_start_date) {
+    if (moment(new_start_date, '<%= t('date.formats.js_old_date') %>').isBefore(
+    moment())) {
+      alert('<%= t(:past_start_date_edit_warning) %>');
+    }
+    }
+
+  jQuery(window).load(function() {
```
david-yz-liu: I'd prefer using the `.ready` function.

app/views/automated_tests/_form.html.erb
```diff
@@ -1,67 +1,92 @@
...
+               formats: [:js],
+               handlers: [:erb] %>
 <% end %>
 
 <div class="wrapLeft">
-<% if flash[:success] %>
-  <div class="success">
-    <span><%=flash[:success]%></span>
-  </div>
+  <% if flash[:success] %>
```
david-yz-liu: Please revert all the indentation changes in this file.

app/controllers/automated_tests_controller.rb
```diff
@@ -163,7 +165,7 @@ def assignment_params
                      :halts_testing, :display_description, :display_run_status,
                      :display_marks_earned, :display_input,
                      :display_expected_output, :display_actual_output,
-                     :_destroy],
+                     :associated_criterion, :_destroy],
```
david-yz-liu: I don't want to see changes related to criterion associations in this PR.

app/helpers/automated_tests_helper.rb
dmitrykhabarov: Uploading new script files wasn't working

app/views/automated_tests/_boot.js.erb
dmitrykhabarov: I was only able to get it to work with .load function. I could not get timepicker on Test Framework to work with .ready function.

app/models/token.rb
dmitrykhabarov: My mistake, I will update with correct logic.

app/models/test_script.rb
dmitrykhabarov: Forgot to uncomment.

app/models/token.rb
```diff
@@ -44,6 +47,7 @@ def reassign_tokens
       self.tokens = 0
     else
       self.tokens = assignment.tokens_per_day
+      assignment.last_token_regeneration_date = DateTime.now
```
david-yz-liu: This doesn't support regeneration at fixed times.

```
while assignment.last_token_regeneration_date + period < DateTime.now
  assignment.last_token_regeneration_date += period
```

app/views/automated_tests/_test_script_upload.html.erb
```diff
@@ -21,10 +27,10 @@
                    assignment_id: @assignment.id,
                    type: 'script',
                    filename: f.object.script_name) %>
-      <%= link_to saved ? '[-]' : '[+]',
+      <%= link_to(saved ? '[-]' : '[+]',
```
david-yz-liu: Remove the `(` and `)`

app/views/automated_tests/_test_script_upload.html.erb
```diff
@@ -96,10 +102,15 @@
       <%= file_field_tag (saved ? ("new_update_script_"+f.object.script_name).to_sym : :new_script), size: "30", class: 'upload_file' %><br/>
 
       <% if saved %>
-        <%= f.label :_destroy, t('automated_tests.remove_file') %>
```
david-yz-liu: Revert this change

app/views/automated_tests/_test_script_upload.html.erb
```diff
@@ -96,10 +102,15 @@
...
-        <%= f.check_box :_destroy, class: "remove_test_script_file" %>
+          <%= f.label :_destroy, t('automated_tests.remove_file') %>
+          <%= f.check_box :_destroy, class: "remove_test_script_file" %>
+          <br/><br/>
       <% end %>
 
+      <%= f.label :associated_criterion, t('automated_tests.script_option.associated_criterion') %>
+      <%= f.select :associated_criterion, options_for_select(criterion_options, f.object.associated_criterion) %>
+
```
david-yz-liu: Delete these lines

app/models/token.rb
david-yz-liu: Okay so it seems this code is called from `execute_test_run`, which can happen at arbitrary times. This means that using `DateTime.now` is not correct, since future checks will use `last_token_regeneration_date` and lose track of the scheduled regeneration times.

`assignment.last_token_generation_date` should always be equal to `assignment.tokens_start_of_availability_date + n * assignment.regeneration_period`, where `n` is some number.

PR #2515, opened 2016-03-25
 2395-grade-entry-forms-download

kevinmpm:


kevinmpm:
I need to modify the spec tests, don't merge this yet

app/controllers/grade_entry_forms_controller.rb
```diff
@@ -282,9 +282,61 @@ def update_grade_entry_students
...
+            .grades
+            .where(grade_entry_item_id: grade_entry_item.id)
+            .first
+          if grade.nil?
+            row.push('')
+          else
+            row.push(grade.grade || '')
+          end
+        end
+        total_percent = grade_entry_form
```
reidka: I think this should only be added to the download if it is checked in the properties of the spreadsheet.

app/controllers/grade_entry_forms_controller.rb
```diff
@@ -282,9 +282,61 @@ def update_grade_entry_students
...
+    grade_entry_items = grade_entry_form.grade_entry_items
+    csv_rows = []
+    # prepare first two csv rows
+    # The first row in the CSV file will contain the question names
+    row = ['']
+    grade_entry_items.each do |grade_entry_item|
+      row.push(grade_entry_item.name)
+    end
+    csv_rows.push(row)
+    # The second row in the CSV file will contain the question totals
```
reidka: If the total is to be included, we should at a column header for it, with the "out of" in the second row.
(But as below, I only want to include this if the "show total" property is selected in the spreadsheet.)

reidka:
@kevinmpm Looking good.  I'm finding tweaks that I'd like you to make that will improve the experience for the end user, but weren't really part of the original problem.

david-yz-liu:
@kevinmpm `MarkusCSV` doesn't handle header row(s), which is a problem for this PR. Can you please add an optional param to `generate` called `headers`, which is a list of lists specifying header rows for the output csv. (Default: empty).

PR #2516, opened 2016-03-26
Fix #2513: Add .gitkeep file when creating empty group assignment folder

### LINKED ISSUE #2513: Assignment folder doesn't get committed when assignment is created [git]
Because the assignment folder is empty when it is first created, it doesn't get added to the git repository. (Git doesn't track empty directories).

An easy way to fix this issue is to commit some sort of README file when an assignment is created, maybe containing the assignment name and the contents of the message field if it was provided. Or maybe the assignment creator could upload a file containing the assignment description.

Sm1th:
Fix for #2513

### DIFF
--- a/app/models/grouping.rb
+++ b/app/models/grouping.rb
@@ -580,11 +580,6 @@ class Grouping < ActiveRecord::Base
         else
           txn = repo.get_transaction('markus')
           txn.add_path(assignment_folder)
-          # If it is a git repo, we have to add a keep file so
-          # the assignment folder gets committed.
-          if repo.instance_of?(Repository::GitRepository)
-            txn.add(File.join(assignment_folder, '.gitkeep'), '')
-          end
           return repo.commit(txn)
         end
       end
--- a/app/models/grouping.rb
+++ b/app/models/grouping.rb
@@ -578,9 +578,14 @@ class Grouping < ActiveRecord::Base
         if revision.path_exists?(assignment_folder)
           return true
         else
-          txn = self.group.repo.get_transaction('markus')
+          txn = repo.get_transaction('markus')
           txn.add_path(assignment_folder)
-          return self.group.repo.commit(txn)
+          # If it is a git repo, we have to add a keep file so
+          # the assignment folder gets committed.
+          if repo.instance_of?(Repository::GitRepository)
+            txn.add(File.join(assignment_folder, '.gitkeep'), '')
+          end
+          return repo.commit(txn)
         end
       end
     end
--- a/lib/repo/git_repository.rb
+++ b/lib/repo/git_repository.rb
@@ -804,7 +804,14 @@ module Repository
     end
 
     # Removes a file from a transaction and eventually from repository
-    def remove_file(path, author, _expected_revision_number = 0)
+    def remove_file(path, author, expected_revision_number = 0)
+      if latest_revision_number != expected_revision_number
+        raise Repository::FileOutOfSyncConflict.new(path)
+      end
+      if !path_exists_for_latest_revision?(path)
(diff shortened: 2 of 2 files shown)


app/models/grouping.rb
```diff
@@ -578,9 +578,14 @@ def create_grouping_repository_folder
         if revision.path_exists?(assignment_folder)
           return true
         else
-          txn = self.group.repo.get_transaction('markus')
+          txn = repo.get_transaction('markus')
           txn.add_path(assignment_folder)
-          return self.group.repo.commit(txn)
+          # If it is a git repo, we have to add a keep file so
+          # the assignment folder gets committed.
+          if repo.instance_of?(Repository::GitRepository)
```
reidka: Can this be done in the git_repository.rb file instead of here?  It would be better to keep repo specific code out of the models.  Maybe in make_directory?

Sm1th: @reidka Yes... That would be better.

lib/repo/git_repository.rb
```diff
@@ -805,8 +803,8 @@ def add_file(path, file_data = nil, author)
       write_file(path, file_data, author)
     end
 
-    # removes a file from a transaction and eventually from repository
-    def remove_file(txn, path, author, _expected_revision_number = 0)
+    # Removes a file from a transaction and eventually from repository
+    def remove_file(path, author, _expected_revision_number = 0)
```
Sm1th: Can we remove the argument for the `_expected_revision_number`?

lib/repo/git_repository.rb
```diff
@@ -818,29 +816,30 @@ def remove_file(txn, path, author, _expected_revision_number = 0)
...
     end
 
-    # replaces file at provided path with file_data
-    def replace_file(txn, path, file_data = nil,
-                     mime_type = nil, _expected_revision_number = 0)
-      txn = write_file(path, file_data, mime_type)
-      return txn
+    # Replaces file at provided path with file_data
+    def replace_file(path, file_data, author, _expected_revision_number = 0)
+      write_file(path, file_data, author) if path_exists_for_latest_revision?(path)
```
Sm1th: Can we remove the argument for the `_expected_revision_number`?

lib/repo/git_repository.rb
```diff
@@ -860,23 +859,12 @@ def make_file(path, file_data, author)
...
-
-      # Recursively create parent folders (if doesn't exist)
-      parent_path = File.dirname(path)
-      make_directory(parent_path)
-
-      # Now that the parent folder has been created,
-      # create the current folder
-      FileUtils.mkdir_p(path)
+      gitkeep_filename = File.join(path, '.gitkeep')
+      add_file(gitkeep_filename, '', 'markus')
```
Sm1th: Author argument should not be hardcoded to `'markus'` here. Should an argument be added to make_directory to pass down the author name? i.e.: `make_directory(path, author)`

reidka: I can't figure out how to exercise this code.  It doesn't get called when a new group is created.  I thought it might be called in the rake tasks when we see the database.   The clue is that I can't find a case where a file called .gitkeep is created.  I see .git and .keep.

Sm1th: Very sorry @reidka, I've been busy moving to Toronto and starting my new co-op job.

This solution seemed to be functioning as it should for me. Is it still not working for you?

reidka: @Sm1th I figured it out.  It was my fault.  I had some commits in my local repo that never got pushed to master that interacted.  (Hope you are enjoying your co-op job!)

PR #2517, opened 2016-03-27
2395-groups-controller-download 

kevinmpm:


app/controllers/groups_controller.rb
```diff
@@ -194,23 +194,24 @@ def create_groups_when_students_work_alone
   end
 
   def download_grouplist
-    assignment = Assignment.find(params[:assignment_id])
+    assignment = Assignment.where(id: params[:assignment_id]).first
```
david-yz-liu: I prefer the original (using find to lookup one entry by id).

app/controllers/groups_controller.rb
```diff
@@ -194,23 +194,24 @@ def create_groups_when_students_work_alone
...
-         end
-         csv << group_array
-       end
-     end
-
-    send_data(file_out, type: 'text/csv', disposition: 'inline')
+    file_out = MarkusCSV.generate(groupings) do |grouping|
+      group_array = [grouping.group.group_name, grouping.group.repo_name]
+      # csv format is group_name, repo_name, user1_name, user2_name, ... etc
+      grouping.student_memberships.includes(:user).each do |member|
```
david-yz-liu: Please see my note in the other downloads PR. This may affect future PRs in the downloads.

kevinmpm:
I changed it back to using .find, I was having trouble with it initially in the rspec class but I figured out that the issue was due to the .find being stubbed across the entire test class. I'm not sure how to use pluck on this one as student_memberships is a join

PR #2518, opened 2016-03-27
 2395-marks-graders-download

kevinmpm:


app/controllers/marks_graders_controller.rb
```diff
@@ -60,24 +60,21 @@ def download_grader_students_mapping
...
-          entry.grade_entry_form_id == grade_entry_form.id
-        end
-        unless grade_entry_student.nil?
-          grade_entry_student.tas.order(:user_name).each do |ta|
-            student_array.push(ta.user_name)
-          end
+    file_out = MarkusCSV.generate(students) do |student|
+      # csv format is student_name, ta1_name, ta2_name, ... etc
+      student_array = [student.user_name]
+      grade_entry_student = student.grade_entry_students.find do |entry|
```
david-yz-liu: Passing a block to `find` is a little unusual. I guess you can do `find(grade_entry_form_id: grade_entry_form.id)`

app/controllers/marks_graders_controller.rb
```diff
@@ -60,24 +60,21 @@ def download_grader_students_mapping
...
+        entry.grade_entry_form_id == grade_entry_form.id
+      end
+      unless grade_entry_student.nil?
+        grade_entry_student.tas.order(:user_name).each do |ta|
```
david-yz-liu: 1. It's cleaner to do `student_array.concat(...)`, if possible.
2. Since you're only using usernames, you should be able to use `pluck` here effectively.

david-yz-liu:
@kevinmpm You converted the code appropriately; however, I pointed out some simplifications I'd like you to do while you're at it.

app/controllers/marks_graders_controller.rb
```diff
@@ -60,24 +60,21 @@ def download_grader_students_mapping
...
-    file_out = CSV.generate do |csv|
-      students.each do |student|
-        # csv format is student_name, ta1_name, ta2_name, ... etc
-        student_array = [student.user_name]
-        grade_entry_student = student.grade_entry_students.find do |entry|
+    file_out = MarkusCSV.generate(students) do |student|
+      # csv format is student_name, ta1_name, ta2_name, ... etc
+      student_array = [student.user_name]
+      grade_entry_student = student.grade_entry_students
+        .find(grade_entry_form_id: grade_entry_form.id) do |entry|
```
david-yz-liu: You don't actually need the `do` block any more.

PR #2520, opened 2016-04-01
 2395-rubrics-download

kevinmpm:


app/controllers/rubrics_controller.rb
```diff
@@ -61,8 +63,20 @@ def destroy
...
+      criterion_array = [criterion.rubric_criterion_name,criterion.weight]
+      (0..RUBRIC_LEVELS - 1).each do |i|
+        criterion_array.push(criterion['level_' + i.to_s + '_name'])
+      end
+      (0..RUBRIC_LEVELS - 1).each do |i|
+        criterion_array.push(criterion['level_' + i.to_s + '_description'])
+      end
+      criterion_array
+    end
+    send_data(file_out,
```
reidka: This is more of a question for @david-yz-liu  
Should the send_data go in the MarkusCSV class?  It seems that we are unlikely to do anything with generated csv output other than send it.

david-yz-liu: Sure, I can go for that.

app/controllers/rubrics_controller.rb
```diff
@@ -2,6 +2,8 @@ class RubricsController < ApplicationController
 
   before_filter :authorize_only_for_admin
 
+  RUBRIC_LEVELS = 5
```
david-yz-liu: Rather than redefine it here, you can use `RubricCriterion::RUBRIC_LEVELS`

PR #2524, opened 2016-04-02
2395-submissions-download (

kevinmpm:


david-yz-liu:
@kevinmpm Since we're refactoring, I'm thinking about ways of removing code duplication for the detailed reports between the rubric and flexible criteria cases. Can you please do the following refactor: add an instance method `get_mark_and_value` to both `RubricCriterion` and `FlexibleCriterion`, which takes a submission and returns a list `[x, y]`, where `x` is the mark the submission got on the criterion, and `y` is how much the criterion is worth (either `:weight` or `:max`, depending on the criterion.

I believe that if you do this, it will be possible to merge the two methods into one.

PR #2527, opened 2016-04-05
Added Section Filters to the Marks Spreadsheets

alexinman:
Added Section Filters to the Marks Spreadsheets
screen shot 2016-04-05 at 10 00 55 am
screen shot 2016-04-05 at 10 01 05 am

### DIFF
--- a/app/views/grade_entry_forms/_react_grades_table.js.jsx.erb
+++ b/app/views/grade_entry_forms/_react_grades_table.js.jsx.erb
@@ -193,7 +193,7 @@
             name: section.name,
             text: section.name,
             func: function(group) {
-              return (group.section == this.name);
+              return group.section === section.name;
             }.bind(section)
           }
           return filter;
--- a/app/views/grade_entry_forms/_react_grades_table.js.jsx.erb
+++ b/app/views/grade_entry_forms/_react_grades_table.js.jsx.erb
@@ -188,19 +188,18 @@
       // need to filter for each section
       var sections = <%= raw @sections.to_json %>;
       if (sections.length > 0) {
-        var section_filters = [];
-        var all = '<%= I18n.t('browse_submissions.sections.all') %>';
-        section_filters.push({name: all, text: all, func: function(g) { return g; } });
-        for (var i = 0; i < sections.length; i++) {
+        var section_filters = sections.map(function(section) {
           var filter = {
-            name: sections[i].name,
-            text: sections[i].name,
+            name: section.name,
+            text: section.name,
             func: function(group) {
               return (group.section == this.name);
-            }.bind(sections[i])
-          };
-          section_filters.push(filter);
-        }
+            }.bind(section)
+          }
+          return filter;
+        });
+        var all = '<%= I18n.t('browse_submissions.sections.all') %>';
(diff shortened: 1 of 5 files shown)


app/views/grade_entry_forms/_react_grades_table.js.jsx.erb
```diff
@@ -185,6 +185,24 @@
       });
     },
     render: function() {
+      // need to filter for each section
+      var sections = <%= raw @sections.to_json %>;
+      if (sections.length > 0) {
+        var section_filters = [];
+        var all = '<%= I18n.t('browse_submissions.sections.all') %>';
+        section_filters.push({name: all, text: all, func: function(g) { return g; } });
+        for (var i = 0; i < sections.length; i++) {
```
david-yz-liu: Please use a `map` here.

alexinman: This is how it's done for all the other section filters...
Do you want me to use a `map` for all of them?

david-yz-liu: Sure.

app/views/grade_entry_forms/_react_grades_table.js.jsx.erb
```diff
@@ -185,6 +185,23 @@
...
     render: function() {
+      // need to filter for each section
+      var sections = <%= raw @sections.to_json %>;
+      if (sections.length > 0) {
+        var section_filters = sections.map(function(section) {
+          var filter = {
+            name: section.name,
+            text: section.name,
+            func: function(group) {
+              return (group.section == this.name);
```
david-yz-liu: Maybe I'm missing something, but why can't you simply do `return group.section === section.name`?

alexinman: I don't see a (significant) difference. Why does that matter?

david-yz-liu: It's simpler because it avoids an unnecessary function call. Also, we should generally use variables when they are in scope.

alexinman: So, just to clarify, switch this line:
`return (group.section == this.name);`
with this:
`return (group.section === section.name);`
?

david-yz-liu: Yep, but you can remove the `()` as well.

app/views/grade_entry_forms/_react_grades_table.js.jsx.erb
```diff
@@ -185,6 +185,23 @@
...
+      // need to filter for each section
+      var sections = <%= raw @sections.to_json %>;
+      if (sections.length > 0) {
+        var section_filters = sections.map(function(section) {
+          var filter = {
+            name: section.name,
+            text: section.name,
+            func: function(group) {
+              return group.section === section.name;
+            }.bind(section)
```
david-yz-liu: Okay, I guess I wasn't clear. If you use `section` instead of `this`, then you don't need to call `bind` on the function.

Issue #2522, opened 2016-04-02
Sort by section doesn't work in Spreadsheet tables


david-yz-liu:
It seems like the column is sortable (has the little arrows), but the order of rows is unaffected when the column is clicked.

alexinman:
I am unable to reproduce this issue. Sorting by section works fine for me.
What browser are you using?

david-yz-liu:
It's possible you haven't tried to sort when some students are in sections, and others are not.

PR #2531, opened 2016-04-11
Fix for #2522

alexinman:
Problem was section was null when student's weren't in a section which caused problems when sorting. Changed it to empty string and now sorting works fine.


alexinman:
#2522

PR #2533, opened 2016-04-21
Merge test framework

david-yz-liu:
This merges in the test-framework branch, which implements an autotesting framework in MarkUs.

@reidka we'll need some good testing and documentation for this pull request.


reidka:
In Settings->Test Framework there is no reason to limit the number of characters the "regenerate every __ hours" field

PR #2536, opened 2016-04-28
Add gemfile group for testing a production-like deployment

adisandro:
* Add production_test gemfile group, to cover gems that are not strictly
   needed in production, but still required to test a production-like
   deployment; an example is 'faker', to seed the database with test data
 * Fix rubric rake task to avoid stympy/faker#278 when run in a production
   environment


david-yz-liu:
@adisandro I redid the bundle install and did indeed get that dependency removed.

PR #2542, opened 2016-05-04
Migration criteria names

GitGeminiano:
Hello everyone,

I changed flexible_criterion_name and rubric_criterion_name for name. Please have a look at the changes.

Thanks


app/models/flexible_criterion.rb
```diff
@@ -125,7 +125,7 @@ def add_tas(ta_array)
   end
 
   def get_name
```
david-yz-liu: @GitGeminiano It seems this method may have been used earlier to abstract the different attribute names, but is not redundant. Replace usages of `get_name` with `name`, and delete this method from the two criterion classes.

GitGeminiano: Thanks for noticing that. I will make the changes.

PR #2545, opened 2016-05-04
New Result Mark Creation

christinem:
This pull request fixes the error where marks were not created when a result is initially created, which prevented us from assigning marks automatically through the API.

### DIFF
--- a/app/models/result.rb
+++ b/app/models/result.rb
@@ -10,6 +10,7 @@ class Result < ActiveRecord::Base
   has_many :marks
   has_many :extra_marks
 
+  after_create :create_marks
   validates_presence_of :marking_state
   validates_inclusion_of :marking_state, in: MARKING_STATES.values
 
@@ -119,5 +120,17 @@ class Result < ActiveRecord::Base
       return false
     end
     true
+  end
+
+  def create_marks
+    @assignment = self.submission.assignment
+    @marks_map = Hash.new
+    @assignment.get_criteria.each do |criterion|
+      mark = criterion.marks.find_or_create_by(result_id: self.id)
+      @marks_map[criterion.id] = mark
+
+      mark.save(validate: false)
+      self.update_total_mark
+    end
   end
 end
--- a/app/models/result.rb
+++ b/app/models/result.rb
@@ -123,14 +123,10 @@ class Result < ActiveRecord::Base
   end
 
   def create_marks
-    @assignment = self.submission.assignment
-    @marks_map = Hash.new
-    @assignment.get_criteria.each do |criterion|
-      mark = criterion.marks.find_or_create_by(result_id: self.id)
-      @marks_map[criterion.id] = mark
-
-      mark.save(validate: false)
-      self.update_total_mark
+    assignment = self.submission.assignment
+    assignment.get_criteria.each do |criterion|
+      mark = criterion.marks.create(result_id: id)
     end
+    self.update_total_mark
   end
 end
--- a/spec/models/assignment_spec.rb
+++ b/spec/models/assignment_spec.rb
@@ -1030,10 +1030,9 @@ describe Assignment do
           result = s.get_latest_result
           result.total_mark = total_mark
           result.marking_state = Result::MARKING_STATES[:complete]
-          @assignment.rubric_criteria.each do |cri|
-            result.marks.create!(markable_id: cri.id,
-                                 markable_type: RubricCriterion,
-                                 mark: (total_mark * 4.0 / 20).round)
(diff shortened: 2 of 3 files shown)


app/models/result.rb
```diff
@@ -120,4 +121,16 @@ def check_for_nil_marks
     end
     true
   end
+
+  def create_marks
+    @assignment = self.submission.assignment
```
david-yz-liu: Don't use `@` here, all these variables should be local, not instance variables.

app/models/result.rb
```diff
@@ -120,4 +121,16 @@ def check_for_nil_marks
...
+    @marks_map = Hash.new
```
david-yz-liu: I don't think this is actually used.

app/models/result.rb
```diff
@@ -120,4 +121,16 @@ def check_for_nil_marks
...
+    @assignment.get_criteria.each do |criterion|
+      mark = criterion.marks.find_or_create_by(result_id: self.id)
```
david-yz-liu: Just create - there won't be anything to find. Also use `id` instead of `self.id`; self is redundant in this case.

app/models/result.rb
```diff
@@ -120,4 +121,16 @@ def check_for_nil_marks
...
+      @marks_map[criterion.id] = mark
+
+      mark.save(validate: false)
+      self.update_total_mark
```
david-yz-liu: I'm guessing you can put this outside the loop to just call it once.

app/models/result.rb
```diff
@@ -120,4 +121,16 @@ def check_for_nil_marks
...
   end
+
+  def create_marks
+    @assignment = self.submission.assignment
+    @marks_map = Hash.new
+    @assignment.get_criteria.each do |criterion|
+      mark = criterion.marks.find_or_create_by(result_id: self.id)
+      @marks_map[criterion.id] = mark
+
+      mark.save(validate: false)
```
david-yz-liu: No need to save right after a `create`.

spec/models/assignment_spec.rb
```diff
@@ -1030,10 +1030,9 @@
           result = s.get_latest_result
           result.total_mark = total_mark
           result.marking_state = Result::MARKING_STATES[:complete]
-          @assignment.rubric_criteria.each do |cri|
-            result.marks.create!(markable_id: cri.id,
-                                 markable_type: RubricCriterion,
-                                 mark: (total_mark * 4.0 / 20).round)
+          result.marks.each do |m|
+            m.mark = (total_mark * 4.0 / 20).round
```
david-yz-liu: Use the `update!` method here.

spec/models/result_spec.rb
```diff
@@ -99,9 +128,21 @@
                 let(:marks) { [3, 0, 9] }
                 let!(:results) do
                   Array.new(3) do |i|
-                    create(:complete_result,
-                           total_mark: marks[i],
-                           submission: submissions[i])
+                    create(:result,
+                           # total_mark: marks[i],
```
david-yz-liu: Remove this line.

spec/models/result_spec.rb
```diff
@@ -99,9 +128,21 @@
...
+                           submission: submissions[i],
+                           marking_state: Result::MARKING_STATES[:partial])
+                  end
+                end
+                before do
+                  results.each_with_index do |result, i|
+                    result.marks.each do |m|
+                      m.mark = 0.0
```
david-yz-liu: use `update` here as well

christinem:
@david-yz-liu Fixed the comments you made.

spec/models/assignment_spec.rb
```diff
@@ -1031,8 +1031,7 @@
           result.total_mark = total_mark
           result.marking_state = Result::MARKING_STATES[:complete]
           result.marks.each do |m|
-            m.mark = (total_mark * 4.0 / 20).round
-            m.save!
+            m.update!(:mark => (total_mark * 4.0 / 20).round)
```
david-yz-liu: Use the new hash syntax `mark: ...`

spec/models/result_spec.rb
```diff
@@ -129,16 +129,14 @@
...
                            submission: submissions[i],
                            marking_state: Result::MARKING_STATES[:partial])
                   end
                 end
                 before do
                   results.each_with_index do |result, i|
                     result.marks.each do |m|
-                      m.mark = 0.0
-                      m.save
+                      m.update(:mark => 0.0)
```
david-yz-liu: Same here - new hash syntax

PR #2552, opened 2016-05-05
Issue 2547 Decimal Fix

### LINKED ISSUE #2547: Marks for criteria randomly alternate between integer and float representations in rubric
A user reports:

"How to reproduce: Happens unpredictably. In the TA view, click the name of an assignment that has been collected and has a marking scheme set up and click the name of a group. Expand the rubric panel on the right-hand side. Usually the numbers are represented as floats (e.g. 4.0), but occasionally the decimal points will be truncated and choosing a number when this happens will save it in this form until the page is refreshed."

This happens in the rubric marking scheme and I was able to reproduce it in the development environment. (I wonder if it might be related to remarks, but this is a total guess.)

christinem:
Fixes #2547

### DIFF
--- a/app/models/result.rb
+++ b/app/models/result.rb
@@ -63,22 +63,22 @@ class Result < ActiveRecord::Base
 
   # The sum of the bonuses and deductions, other than late penalty
   def get_total_extra_points
-    extra_marks.points.map(&:extra_mark).reduce(0, :+)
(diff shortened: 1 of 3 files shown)


david-yz-liu:
Don't forget to change the rounding in the tab where graders actually give marks.

Also make sure to double-check the grade total and "out of". Test this out by making some unusual weights (e.g., 3.1) and make sure that the math works out.

PR #2556, opened 2016-05-06
Fixes #2549

### LINKED ISSUE #2549: Entering an incorrect password clears the username field during login
A user reports:
"How to reproduce: When attempting to log into MarkUs, enter a username and an incorrect password.
Expected result: The entered username is persisted in the username field on the resulting incorrect password page.
Actual result: The username field is cleared so that the username has to be entered again."

I notice this too.  I don't think it is a major issue, but it is an irritant.

WChrisK:


david-yz-liu:
Let's not change the session. Instead, why don't you pass a user name as a local variable to the `login.html.erb` template?

app/controllers/main_controller.rb
```diff
@@ -86,8 +86,7 @@ def login
     validation_result = validate_user(params[:user_login], params[:user_login], params[:user_password])
     unless validation_result[:error].nil?
       flash_message(:error, validation_result[:error])
-      redirect_to action: 'login'
-      return
+      render :login, locals: {user_login: params[:user_login]} and return
```
david-yz-liu: In general, put spaces inside `{` and `}`, and don't use `and` (so keep the return on a separate line).

PR #2557, opened 2016-05-06
Simplify result status

david-yz-liu:
Changes the result status from `{:unmarked, :partial, :complete}` to `{:incomplete, :complete}`. Simplifies the grader UI to use buttons for marking a submission as complete and releasing a submission.

Builts on #2337.


reidka:
This looks good.  I tested all the variations I could think of, and I couldn't break it.

PR #2563, opened 2016-05-11
Make submission collection a background task

david-yz-liu:
Work of @RFrailich.


app/jobs/submissions_job.rb
```diff
@@ -0,0 +1,87 @@
...
+      end
+
+      m_logger.log('Submission collection process established database' +
+                   ' connection successfully')
+
+      groupings.each do |grouping|
+			  m_logger.log("Now collecting: #{assignment.short_identifier} for grouping: " +
+                     "#{grouping.id}")
+
+        if options[:revision_number].nil?
```
reidka: probably should fix indentation throughout this method.

app/models/submission.rb
```diff
@@ -30,6 +30,27 @@ def self.create_by_timestamp(grouping, timestamp)
      repo.close
      submission
   end
+  
+  def apply_penalty_or_add_grace_credits(grouping,
```
reidka: Does this function need to be duplicated in both submission.rb and submissions_job.rb?

app/models/submission.rb
```diff
@@ -30,6 +30,27 @@ def self.create_by_timestamp(grouping, timestamp)
...
+    if grouping.assignment.submission_rule.is_a? GracePeriodSubmissionRule
+      # Return any grace credits previously deducted for this grouping.
+      grouping.assignment.submission_rule.remove_deductions(self)
+    end
+    if apply_late_penalty
+      grouping.assignment.submission_rule.apply_submission_rule(self)
+    end
+  end
+  
+  def collect_single(grouping, rev_num, apply_late_penalty)
```
reidka: I think this can be deleted.  I don't think it is called from anywhere.

PR #2566, opened 2016-05-12
Remove some N+1 queries on submission table load

david-yz-liu:


david-yz-liu:
@GitGeminiano I fixed a bunch of other N+1 queries. You were probably putting the `includes` in the wrong place yesterday. Note that the includes *always* should go where the original objects are being queries. Study the file changes that I made here.

PR #2568, opened 2016-05-12
Assignment association

WChrisK:


app/models/assignment.rb
```diff
@@ -896,6 +900,24 @@ def can_upload_starter_code?
...
+  # Returns true if this is a peer review, meaning it has a parent assignment,
+  # false otherwise.
+  def is_peer_review?
+    not self.parent_assignment_id.nil?
+  end
+
+  # Checks if this assignment has any children which reference this assignment
+  def has_peer_review_assignments?
+    not Assignment.where(:parent_assignment_id => self.parent_assignment_id).empty?
+  end
```
WChrisK: On second thought, this method is redundant... remove?

app/models/assignment.rb
```diff
@@ -42,6 +42,10 @@ class Assignment < ActiveRecord::Base
   # Because of app/views/main/_grade_distribution_graph.html.erb:25
   validates_presence_of :assignment_stat
 
+  # Assignments can now refer to themselves, where this is null if there
+  # is no parent (the same holds for the child peer reviews)
+  belongs_to :parent_assignment, :class_name => "Assignment"
```
david-yz-liu: Use the new hash syntax

app/models/assignment.rb
```diff
@@ -896,6 +900,24 @@ def can_upload_starter_code?
     false
   end
 
+  # Returns true if this is a peer review, meaning it has a parent assignment,
+  # false otherwise.
+  def is_peer_review?
+    not self.parent_assignment_id.nil?
```
david-yz-liu: Don't use `self`

app/models/assignment.rb
david-yz-liu: That's pretty interesting. I don't think it's redundant, but I don't really like the approach. (Also I think the id should be passed in, not parent_assignment_id)

Why don't we create another association `has_one :peer_review` which is the inverse of `parent_assignment`, and check whether that's nil.

app/models/assignment.rb
```diff
@@ -896,6 +900,24 @@ def can_upload_starter_code?
...
+  # Checks if this assignment has any children which reference this assignment
+  def has_peer_review_assignments?
+    not Assignment.where(:parent_assignment_id => self.parent_assignment_id).empty?
+  end
+
+  # Gets a relation object that contains all of the child assignments that
+  # are to be peer reviews for this assignment (which means .empty? may be
+  # true)
+  def get_peer_review_assignments
+    Assignment.where(:parent_assignment_id => self.parent_assignment_id)
```
david-yz-liu: See comment above. There will only be one, and it should be an association, so you can delete this method.

db/migrate/20160511224541_add_peer_review_assignment_to_assignments.rb
```diff
@@ -0,0 +1,5 @@
+class AddPeerReviewAssignmentToAssignments < ActiveRecord::Migration
+  def change
+    add_column :assignments, :parent_assignment_id, :integer, default: nil
```
david-yz-liu: I'd like to see an `add_reference` here

spec/models/assignment_spec.rb
```diff
@@ -65,6 +65,30 @@
     it 'should require case sensitive unique value for short_identifier' do
       assignment = create(:assignment)
       expect(assignment).to validate_uniqueness_of(:short_identifier)
+      end
+    it 'should have a nil parent_assignment_id by default' do
+      assignment = build(:assignment)
+      expect(assignment.parent_assignment_id).to be_nil
+    end
+    it 'should not be a peer review if there is no parrent_assignment_id' do
```
david-yz-liu: typo on `parrent`

app/models/assignment.rb
david-yz-liu: Oh, and this would be good to validate as well.

spec/models/assignment_spec.rb
```diff
@@ -65,6 +65,30 @@
...
+      assignment = build(:assignment)
+      expect(assignment.parent_assignment_id).to be_nil
+      expect(assignment.is_peer_review?).to be false
+    end
+    it 'should be a peer review if it has a parent_assignement_id' do
+      assignment = build(:assignment)
+      assignment.parent_assignment_id = 1  # Arbitrary number
```
david-yz-liu: Will fail if you add a validation (see above). Create a second assignment.

app/models/assignment.rb
```diff
@@ -42,6 +42,11 @@ class Assignment < ActiveRecord::Base
   # Because of app/views/main/_grade_distribution_graph.html.erb:25
   validates_presence_of :assignment_stat
 
+  # Assignments can now refer to themselves, where this is null if there
+  # is no parent (the same holds for the child peer reviews)
+  belongs_to :parent_assignment, class_name: 'Assignment', inverse_of: :peer_review
+  has_one :peer_review, class_name: 'Assignment', foreign_key: :parent_assignment_id, inverse_of: :parent_assignment
```
david-yz-liu: Sorry, I meant to make a comment on this last time. I don't think "peer_review" is a good name for this association, as it conflicts with the other new model you've added. Maybe just refer to these as `pr_assignment` to keep it short.

spec/models/assignment_spec.rb
```diff
@@ -65,6 +67,37 @@
...
+      expect(parent_assignment.is_peer_review?).to be false
+    end
+    it 'should give a true has_peer_review_assignment result if it does' do
+      parent_assignment = create(:assignment)
+      assignment = create(:assignment, parent_assignment: parent_assignment)
+      expect(parent_assignment.has_peer_review_assignment?).to be true
+      expect(assignment.has_peer_review_assignment?).to be false
+    end
+    it 'should find children assignments when they reference the parent' do
+      parent_assignment = create(:assignment, id: 1)
```
david-yz-liu: Don't hard code the ids in test objects (it can mess up some later validations, even though it doesn't in this case).

db/migrate/20160511224541_add_peer_review_assignment_to_assignments.rb
```diff
@@ -0,0 +1,6 @@
+class AddPeerReviewAssignmentToAssignments < ActiveRecord::Migration
+  def change
+    add_reference :assignments, :parent_assignment
+    add_reference :assignments, :pr_assignment
```
david-yz-liu: This column isn't necessary.

db/schema.rb
```diff
@@ -115,6 +115,8 @@
     t.boolean  "vcs_submit",                       default: false
     t.datetime "token_start_date"
     t.float    "token_period"
+    t.integer  "parent_assignment_id"
+    t.integer  "pr_assignment_id"
```
david-yz-liu: Okay so this line shouldn't be there any more.

PR #2575, opened 2016-05-15
fixing the nested directory file browsing problem

### LINKED ISSUE #2573: Broken view for student file manager for nested directories
If a student selects a subdirectory in the file manager view it leads to a 404.

reidka:
fixes #2573 

Note:  There is an awful lot of duplicated code in the assignments controller and the submissions controller regarding file_manager.  It looks like the code in the assignments controller is meant to support the addition of starter code to student's repos through MarkUs, but I'm betting we really ought to refactor it to minimize duplication and having to fix bugs in multiple places.

### DIFF
--- a/app/views/submissions/_react_file_manager_boot_table.js.jsx.erb
+++ b/app/views/submissions/_react_file_manager_boot_table.js.jsx.erb
@@ -59,7 +59,8 @@
       this.setState({ selected_files: [] });
 
       jQuery.ajax({
-        url: '<%= populate_file_manager_react_assignment_submissions_path %>',
+        url: '<%= populate_file_manager_react_assignment_submissions_path(
+          path: @path) %>',
         method: 'GET',
         dataType: 'json',
         success: function(data) {
--- a/app/views/submissions/_react_file_manager_boot_table.js.jsx.erb
+++ b/app/views/submissions/_react_file_manager_boot_table.js.jsx.erb
@@ -59,9 +59,9 @@
       this.setState({ selected_files: [] });
 
       jQuery.ajax({
-        url: '<%= populate_file_manager_react_assignment_submissions_path(
-          path: @path) %>',
+        url: '<%= populate_file_manager_react_assignment_submissions_path %>',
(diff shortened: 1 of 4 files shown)


app/views/submissions/_react_file_manager_boot_table.js.jsx.erb
```diff
@@ -59,7 +59,8 @@
       this.setState({ selected_files: [] });
 
       jQuery.ajax({
-        url: '<%= populate_file_manager_react_assignment_submissions_path %>',
+        url: '<%= populate_file_manager_react_assignment_submissions_path(
+          path: @path) %>',
```
david-yz-liu: I get what you were doing here, but it's more standard to pass this as part of the `data` option in the ajax call. Something like

```javascript
method: 'GET',
data: { path: '<%= @path %>' },
...
```

reidka: Okay.  I'm happy to change it.  Here is what I was using as a model from submissions/_repo_files_table.js.jsx.erb.  (Oddly enough, the form you recommend isn't used as much
`jQuery.ajax({
        url: '<%= repo_browser_assignment_submission_path(
          @assignment,
          id: @grouping.id,
          path: @path,
          revision_timestamp: @revision_timestamp,
          revision_number: @revision_number) %>',`

reidka:
I think I should also add a link for the parent directory when we are in the subdirectory.  So don't merge this just yet.

reidka:
Okay, I think I got the parent directory link implemented correctly now.  (This area of the code base could be refactored.  There is a fair bit of code duplication between the repo_browser and file_manager.)

david-yz-liu:
I think the parent link should have the "folder" icon (which looks quite nice).

david-yz-liu:
@reidka I accidentally merged this into master, so I'll look into the failing test and see if I can get it soon.

reidka:
@david-yz-liu I'll see if can finish the test tonight or first thing tomorrow morning.

Issue #2561, opened 2016-05-10
Annotation completion


reidka:
Annotation completion was mentioned as a useful feature.

david-yz-liu:
@christinem this is the "big picture" of what I had you working on today

david-yz-liu:
Note: #2577 seems to search for all annotations, not the ones for the current assignment. That should be fixed (@christinem if you have time).

PR #2577, opened 2016-05-16
Annotation text search

christinem:
Beginning work for #2561, this adds a function which searches for an annotation text that matches the first part of a string, and renders the text if exactly one match is found, renders an empty string otherwise.

### DIFF
--- a/app/controllers/annotation_categories_controller.rb
+++ b/app/controllers/annotation_categories_controller.rb
@@ -81,11 +81,11 @@ class AnnotationCategoriesController < ApplicationController
 
   def find_annotation_text
     string = params[:string]
-    @annotation_texts = AnnotationText.where("content LIKE ?", "#{string}%")
-    if @annotation_texts.size == 1
-      render json: "#{@annotation_texts.first.content}".html_safe
+    annotation_texts = AnnotationText.where("content LIKE ?", "#{string}%")
+    if annotation_texts.size == 1
+      render json: "#{annotation_texts.first.content}".html_safe
     else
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -238,6 +238,7 @@
           get 'add_annotation_text'
           post 'add_annotation_text'
           put 'update_annotation'
+          get 'find_annotation_text'
```
david-yz-liu: This isn't necessary.

app/controllers/annotation_categories_controller.rb
```diff
@@ -79,6 +79,16 @@ def delete_annotation_text
     @annotation_text.destroy
   end
 
+  def find_annotation_text
+    string = params[:string]
+    @annotation_texts = AnnotationText.where("content LIKE ?", "#{string}%")
+    if @annotation_texts.size == 1
+      render json: "#{@annotation_texts.first.content}".html_safe
+    else
+      render json: "".html_safe
```
david-yz-liu: Use single quotes when not doing string interpolation

app/controllers/annotation_categories_controller.rb
```diff
@@ -79,6 +79,16 @@ def delete_annotation_text
     @annotation_text.destroy
   end
 
+  def find_annotation_text
+    string = params[:string]
+    @annotation_texts = AnnotationText.where("content LIKE ?", "#{string}%")
```
david-yz-liu: Don't make this an instance variable (no `@`)

PR #2578, opened 2016-05-17
Migrated tests from result_test to result_spec

GitGeminiano:
Hello everyone,

I migrated the tests from result_test into result_spec. Please have a look at it.

Regards,
Christian


david-yz-liu:
@GitGeminiano if you have correct replicated all of the old tests, you should delete the old ones.

spec/models/result_spec.rb
```diff
@@ -115,6 +126,17 @@
                 it_returns 'empty'
               end
 
+              context 'when an invalid marking state is found' do
+                let!(:invalid_result) { create(:result, marking_state: Result::MARKING_STATES[:incomplete]) }
+                before do
+                  invalid_result.marking_state = 'wrong'
+                end
+
+                it 'catches an invalid result' do
```
david-yz-liu: "catches" isn't a good verb here. Maybe "considers the result invalid". (Same below.)

GitGeminiano: Thank you for the suggestion. I made the necessary changes and deleted the old tests.

PR #2583, opened 2016-05-17
Added support for peer review flag in GUI, testing, and database

WChrisK:


WChrisK:
Test failures will be fixed tomorrow (not ready for integration yet)

db/migrate/20160518004516_add_peer_review_to_assignments.rb
```diff
@@ -0,0 +1,5 @@
+class AddPeerReviewToAssignments < ActiveRecord::Migration
+  def change
+    add_column :assignments, :peer_review, :boolean, default: false
```
david-yz-liu: Again, I don't want to re-use the name "peer_review" too much. Maybe change the column name to "has_peer_review". Also, put in a null: false.

WChrisK:
I assume it's implied that I did this, but to confirm: It does save to the database (and checking it w/ selecting assignments from psql) and saves/loads properly.

david-yz-liu:
@WChrisK please add translations to the other files (you can just use Google Translate here).

PR #2586, opened 2016-05-18
Fix failing test in submissions_controller_spec.rb

david-yz-liu:
The memory repository version of `files_at_path_helper` didn't handle a the root directory, with path value `.`


david-yz-liu:
@reidka I fixed the test yesterday, but I forgot to tag you. I'm not entirely sure about my fix though, so if you have another idea please go for it.

PR #2592, opened 2016-05-20
Added link for peer review checkbox

WChrisK:


app/views/assignments/_form.html.erb
```diff
@@ -160,6 +160,19 @@
...
       <%= f.label :has_peer_review,
                   t('assignment.has_peer_review') %>
+      <% if @assignment.has_peer_review %>
+          <% peer_review_assignment = Assignment.where(parent_assignment_id: @assignment.id).first %>
+          <% if peer_review_assignment.nil? %>
+            <% m_logger = MarkusLogger.instance
+               a_short_id = @assignment.short_identifier
+               a_id = @assignment.id
+               m_logger.log("Assignment #{a_short_id} (id: #{a_id}) has no peer review assignment.",
+                           MarkusLogger::ERROR) %>
```
WChrisK: Should the logger be included?

WChrisK:
Two things:

1) Is there better usage of `<% ... %>` without me accidentally creating some malformed command that runs unexpectedly? For example, having one big `<% ... stuff ... %>` whereby some statements always get evaluated even when we don't want them to?
The current stuff works but it looks messy, which is bad -- unless there's no choice.

2) One thing I'm trying to avoid is for an error to be thrown on database corruption, since they couldn't do any more editing until someone intervened and fixed the database. At least in the current case, they'd still be able to access the assignments page without an error being thrown.
On second thought, we wouldn't know what the assignment is, it would just have the `has_peer_review` column set to true, which we could easily toggle if no such assignment exists.

Not sure what you want me to do here... keep it, change it, etc.

app/views/assignments/_form.html.erb
```diff
@@ -160,6 +160,19 @@
                       onclick: 'set_onbeforeunload(true);' %>
       <%= f.label :has_peer_review,
                   t('assignment.has_peer_review') %>
+      <% if @assignment.has_peer_review %>
+          <% peer_review_assignment = Assignment.where(parent_assignment_id: @assignment.id).first %>
```
WChrisK: Better to have: `peer_review_assignment = @assignment.pr_assignment`?

app/views/assignments/_form.html.erb
```diff
@@ -160,6 +160,19 @@
...
+          <% if peer_review_assignment.nil? %>
+            <% m_logger = MarkusLogger.instance
+               a_short_id = @assignment.short_identifier
+               a_id = @assignment.id
+               m_logger.log("Assignment #{a_short_id} (id: #{a_id}) has no peer review assignment.",
+                           MarkusLogger::ERROR) %>
+            <div class="info red">[<%= t('assignment.peer_review_record_missing') %>]</div>
```
WChrisK: Thoughts on this line?
It appears on the line below it because of the <div> so maybe the square brackets should be removed.

app/views/assignments/_form.html.erb
david-yz-liu: Yes!

app/views/assignments/_form.html.erb
david-yz-liu: No, don't put it here. (Generally, logger calls should go in the controller or model, but I don't think this is important to log right now anyway.)

app/views/assignments/_form.html.erb
david-yz-liu: This is a nice thought, but I'm not sure it belongs here. Perhaps you could make this is a flash message instead, so you decide when to display it in the controller.

PR #2590, opened 2016-05-19
Assignment tab

WChrisK:

PR #2593, opened 2016-05-20
Updated assignment form to handle peer review form views

WChrisK:


WChrisK:
Let me know if this is what you had in mind. I tested this with a peer review assignment. I also tried to use common sense and guess at what the admins would want but there might be stuff that I missed from judgement errors.

If you want a screenshot of this so you can easily look over it, let me know and I'll upload one.

david-yz-liu:
@WChrisK On the one hand, you do have the right idea and used good judgement on which fields to remove. On the other, your other pull request #2590 introduces a new form to use for peer review assignments. I happened to make comments there as well about what fields to remove.

So I'd prefer you build on top of the previous pull request for customizing the form.

PR #2595, opened 2016-05-24
Peer review upgrade

WChrisK:
This is a collection of the two previous branches, because I realized after trying to debug them that they should have gone together. It was taking way too long to debug one without the other, so I merged them into a new branch and confirmed all is working well.


app/views/assignments/_form.html.erb
```diff
@@ -160,6 +160,12 @@
                       onclick: 'set_onbeforeunload(true);' %>
       <%= f.label :has_peer_review,
                   t('assignment.has_peer_review') %>
+      <% if @assignment.has_peer_review %>
+          <% peer_review_assignment = @assignment.pr_assignment %>
+          <% if not peer_review_assignment.nil? %>
+             (<%= t('assignment.configure_peer_review') %>)
```
david-yz-liu: @WChrisK This should be a link, right?

PR #2589, opened 2016-05-19
Move file selector

christinem:
This PR moved the file selector into the annotations menu. It currently removes the "download files" functionality, as I removed the original file selector bar. It will be added back in the next pull request.

PR #2596, opened 2016-05-25
Download files Modal

christinem:
This PR is based off of #2589 (as it needed the file selector menu changes from that PR). This adds the download file functionality back to the results view, in the form of a modal that can be accessed through a "Download" link in the top right corner.


app/assets/javascripts/Results/index.js
```diff
@@ -0,0 +1,3 @@
+jQuery(document).ready(function () {
+    window.modal_download_files   = new ModalMarkus('#download_files_dialog');
```
david-yz-liu: remove the unnecessary whitespace

david-yz-liu: @christinem this is my only comment.

david-yz-liu:
@christinem this is an interesting problem - if the dropdown list is too long, the bottom is inaccessible. (Can't even scroll)

Here is a simple workaround for now: adjust the CSS of this (and only this) modal so that appears at the top, rather than the middle, of the screen.

christinem:
@david-yz-liu Interesting, I never had enough files to notice that. Made that fix, along with fixing merge conflicts and re-arranging the elements on the modal.

david-yz-liu:
Thanks, @christinem.

I suspect there is quite a bit of duplicated code in the different results views, so that's something we should look into at some point.

christinem:
@david-yz-liu Thanks! Yeah, I can see a lot of places where we can put things into partials, definitely would be good to look at.

PR #2597, opened 2016-05-25
Marking scheme design

GitGeminiano:
Hello everyone,

I modified both the front end and back end to be able to choose the criteria visibility among the different users. Please have a look at it.

Regards,
Christian Garcia


app/controllers/graders_controller.rb
```diff
@@ -280,7 +280,7 @@ def criteria_with_assoc(assignment, options = {})
     criterion_ids = options[:criterion_ids]
     includes = options[:includes] || CRITERION_ASSOC
 
-    criteria = assignment.get_criteria.includes(includes)
+    criteria = assignment.get_criteria.empty? ? assignment.get_criteria : assignment.get_criteria.includes(includes)
```
david-yz-liu: What is the purpose of this change?

app/models/assignment.rb
```diff
@@ -723,10 +723,22 @@ def criterion_class
   end
 
   def get_criteria
+    is_peer_review? ? get_peer_visible_criteria : get_ta_visible_criteria
```
david-yz-liu: The `is_peer_review` check is not correct. Add a parameter which takes a symbol, either `:all`, `:ta`, and `:peer`, and returns the correct criteria in each of these three cases.

app/models/result.rb
```diff
@@ -126,8 +126,9 @@ def unrelease_partial_results
   end
 
   def check_for_nil_marks
-    num_criteria = submission.assignment.rubric_criteria.count +
-                   submission.assignment.flexible_criteria.count
+    peer_review = submission.assignment.is_peer_review?
+    num_criteria = filter_criteria(submission.assignment.rubric_criteria, peer_review).count +
```
david-yz-liu: Use the `get_criteria` method in the Assignment model rather than this "filter_criteria".

app/views/rubrics/_criterion_editor.html.erb
```diff
@@ -38,6 +38,20 @@
      <%= t('criteria_edit_total_mark', criterion_total_mark: criterion.weight*4) %>
     </span>
     <br>
+      <h4><span><%= I18n.t(:visibility) %></span></h4>
```
david-yz-liu: In views, use `t` instead of `I18n.t`

spec/models/assignment_spec.rb
```diff
@@ -136,6 +136,71 @@
...
+                                                   assignment: @assignment,
+                                                   ta_visible: false,
+                                                   peer_visible: true) }
+            @ta_and_peer_criteria = Array.new(2) { create(:rubric_criterion,
+                                                          assignment: @assignment,
+                                                          peer_visible: true) }
+            end
+
+          it 'shows the criteria visible to tas only' do
+            expect(@assignment.get_criteria.length).to eq 4
```
david-yz-liu: Check the actual criteria ids

spec/models/assignment_spec.rb
```diff
@@ -1035,7 +1100,7 @@
   describe '#grade_distribution_as_percentage' do
     before :each do
       @assignment = create(:assignment)
-      5.times { create(:rubric_criterion, assignment: @assignment) }
+      5.times { create(:rubric_criterion, assignment: @assignment, peer_visible: true) }
```
david-yz-liu: Calculating Assignment grades should not require `peer_visible`criteria to be successful. Does this method `grade_distribution_as_percentage` need to be updated as well?

spec/models/result_spec.rb
```diff
@@ -14,7 +14,7 @@
 
   describe '.student_marks_by_assignment' do
     let(:assignment) { create(:assignment) }
-    let!(:criteria) { Array.new(2) { create(:rubric_criterion, assignment: assignment) } }
+    let!(:criteria) { Array.new(2) { create(:rubric_criterion, assignment: assignment, peer_visible: true) } }
```
david-yz-liu: Same comment as previous one.

app/controllers/graders_controller.rb
GitGeminiano: One of the test cases causes assignment.get_criteria to be empty. The includes method raises an error when applied to an empty array. That is why I added this check.

app/views/rubrics/_criterion_editor.html.erb
GitGeminiano: I will change it. Thank you for noticing it.

test/functional/results_controller_test.rb
```diff
@@ -465,7 +465,11 @@ def teardown
 
           should 'and the result is available' do
             SubmissionFile.make(:submission => @submission)
-            Mark.make(:result => @result)
+            if @assignment.marking_scheme_type == Assignment::MARKING_SCHEME_TYPE[:rubric]
+              Mark.make(:rubric, :result => @result)
```
david-yz-liu: Use the new hash syntax `result: @result`

app/views/flexible_criteria/_criterion_editor.html.erb
```diff
@@ -41,9 +41,21 @@
...
     <span><%= t('flexible_criteria.description_hint') %></span>
+    <br>
+    <br>
+    <h4><span><%= I18n.t(:visibility) %></span></h4>
+    <h5>
+      <%= f.check_box :ta_visible,
+                      onclick: 'set_onbeforeunload(true);' %>
+      <%= f.label :ta_visible,
+                  t('flexible_criteria.ta_visible') %>
+      <%= f.check_box :peer_visible,
```
david-yz-liu: Put a line break between the two checkboxes.

app/models/rubric_criterion.rb
```diff
@@ -2,6 +2,7 @@
 
 class RubricCriterion < Criterion
   self.table_name = 'rubric_criteria' # set table name correctly
+  include ActiveModel::Validations
```
david-yz-liu: I don't think this is still necessary, is it?

david-yz-liu:
@GitGeminiano I noticed that the **flexible criteria** "mark as complete" button bug is still there (it seems to check all criteria, rather than all ta-visible criteria.

test/functional/results_controller_test.rb
GitGeminiano: I changed it. Thanks. I noticed many other instances in this file in which we use the old syntax. Is this something I should change? Is it worth making an issue for?

app/models/rubric_criterion.rb
GitGeminiano: Not anymore. Thanks.

david-yz-liu:
@GitGeminiano Thanks! I made a few minor changes, which you can see in the latest commit.

PR #2608, opened 2016-06-08
Flexible Criteria Path Fix

christinem:
Fixes some out-dated path helpers for flexible criteria which were breaking the Marking Scheme page in the settings.


david-yz-liu:
This was fixed in #2597

PR #2612, opened 2016-06-09
Added peer review seeding

WChrisK:


lib/tasks/peer_review.rake
```diff
@@ -0,0 +1,36 @@
+# NOTE: The assignment cannot be generated in the assignments rake task since
```
david-yz-liu: @WChrisK Investigate this error further. This type of comment is too vague to belong.

lib/tasks/peer_review.rake
```diff
@@ -0,0 +1,36 @@
...
+# it causes errors to occur. I guess this is due to it not being populated
+# properly and something errors out. Therefore, all the peer review stuff is
+# done here, including but not limited to creating the PR assignment.
+
+namespace :db do
+  desc 'Create peer reviews and related data'
+  task :peer_reviews => :environment do
+    puts 'Creating A1 Peer Review and updating A1 to have the peer review'
+    a1 = Assignment.find_by(short_identifier: 'A1')
+    a1.has_peer_review = true  # Creates 'a1pr' via callback.
```
david-yz-liu: It's the saving which runs the callback. You should use the `update` instead.

lib/tasks/peer_review.rake
```diff
@@ -0,0 +1,36 @@
...
+    a1.save
+
+    a1pr = Assignment.find_by(short_identifier: a1.short_identifier + '_pr')
```
david-yz-liu: Better to do `a1pr = a1.peer_review_assignment`, or whatever the association is called.

lib/tasks/peer_review.rake
```diff
@@ -0,0 +1,36 @@
...
+    a1pr.clone_groupings_from(a1.id)
+
+    # TODO - Make a proper class for this so the code is clearer? Rather than some arbitrary list index?
+    # Populate a list of 'list of [group_id, result]'.
+    list_of_list_of_results_group_id = []
```
david-yz-liu: It's not necessary to make this list.

Keep the second `each`, and inside that pick a random grouping from `a1.groupings`, and make the PeerReview.

WChrisK:
Looks much nicer now IMO

david-yz-liu:
Indeed it does.

PR #2614, opened 2016-06-10
Moved the 'after save' to a more reasonable location in the source file

WChrisK:
Is this what you had in mind?

There's probably more that should be moved down to proper areas so everything is together, the assignment file header area w/ all the validations/after-before saves/etc looks messy.


david-yz-liu:
Yes, this is good. Structuring the model headers requires some thought, I suppose.

PR #2617, opened 2016-06-10
Added uniqueness pair for peer review table, removed timestamps

WChrisK:
For removal of timestamps, they made it very annoying to add/remove any new rows manually, and I don't believe we use them at all (or may never use it). If you want to keep it I'll drop the remove timestamp changeset.


david-yz-liu:
I'm not sure why timestamps make adding/removing rows challenging. Perhaps because you're doing this in SQL, rather than in the rails console (which is what you should do)?

david-yz-liu:
Oh right, also remember to check in the new schema file.

PR #2618, opened 2016-06-10
Changed rubric_criteria and flexible_criteria for get_criteria

GitGeminiano:
Hello everyone,

I changed instances of rubric_criteria and flexible_criteria for get_criteria, in order to optimize our code. Please have a look at the changes.

Regards,
Christian

### DIFF
--- a/app/controllers/api/groups_controller.rb
+++ b/app/controllers/api/groups_controller.rb
@@ -112,7 +112,7 @@ module Api
           'Marking for that submission is already completed' }, status: 404
         return
       end
-      matched_criteria = assignment.get_criteria.select{|criterion| criterion.name == param.keys}
+      matched_criteria = assignment.get_criteria.select{ |criterion| criterion.name == param.keys }
       if matched_criteria.empty?
         render 'shared/http_status', locals: { code: '404', message:
           'No criteria were found that match that request.' }, status: 404
--- a/app/controllers/api/groups_controller.rb
+++ b/app/controllers/api/groups_controller.rb
@@ -112,13 +112,7 @@ module Api
           'Marking for that submission is already completed' }, status: 404
         return
       end
-
-      matched_criteria = RubricCriterion
-                           .where(name: params.keys,
-                                  assignment_id: params[:assignment_id])
-      matched_criteria.concat(FlexibleCriterion
-                                .where(name: params.keys,
-                                       assignment_id: params[:assignment_id]))
+      matched_criteria = assignment.get_criteria.select{|criterion| criterion.name == param.keys}
       if matched_criteria.empty?
         render 'shared/http_status', locals: { code: '404', message:
           'No criteria were found that match that request.' }, status: 404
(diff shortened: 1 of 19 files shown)


GitGeminiano:
I am still working on this.

app/controllers/api/groups_controller.rb
```diff
@@ -112,13 +112,7 @@ def update_marks
...
         return
       end
-
-      matched_criteria = RubricCriterion
-                           .where(name: params.keys,
-                                  assignment_id: params[:assignment_id])
-      matched_criteria.concat(FlexibleCriterion
-                                .where(name: params.keys,
-                                       assignment_id: params[:assignment_id]))
+      matched_criteria = assignment.get_criteria.select{|criterion| criterion.name == param.keys}
```
david-yz-liu: Put spaces inside `{ ... }`

app/controllers/rubrics_controller.rb
```diff
@@ -4,7 +4,7 @@ class RubricsController < ApplicationController
 
   def index
     @assignment = Assignment.find(params[:assignment_id])
-    @criteria = @assignment.rubric_criteria(order: 'position')
+    @criteria = @assignment.get_criteria.order(:'position')
```
david-yz-liu: Probably don't need a string here

app/helpers/course_summaries_helper.rb
```diff
@@ -77,13 +77,16 @@ def get_mark_for_all_assignments_for_student(student, assignments)
 
   # Get max mark for assignment with id a_id
   def get_max_mark_for_assignment(a_id)
+    assignment = Assignment.find_by_id(a_id)
```
david-yz-liu: Use `find`

app/models/assignment.rb
```diff
@@ -661,7 +661,7 @@ def get_detailed_csv_report
     students = Student.all
     # determine whether to use flexible criterion or rubric
     is_flexible = self.marking_scheme_type == MARKING_SCHEME_TYPE[:flexible]
```
david-yz-liu: Can this line be removed?

app/models/flexible_criterion.rb
```diff
@@ -69,8 +69,7 @@ def self.new_from_csv_row(row, assignment)
     criterion.assignment = assignment
     criterion.name = row[0]
     # assert that no other criterion uses the same name for the same assignment.
-    unless FlexibleCriterion.where(assignment_id: assignment.id,
-                                   name: row[0]).size.zero?
+    unless assignment.get_criteria.select{|criterion| criterion.name == row[0]}.size.zero?
```
david-yz-liu: Put spaces inside `{ ... }`, and use `.empty?`

app/models/assignment.rb
GitGeminiano: @david-yz-liu I will try to delete it in my second pull request. The reason why I do not delete it here is that 'is_flexible' is used in other places in this method, and deleting it will need some method refactoring instead of using get_criteria.

app/models/flexible_criterion.rb
GitGeminiano: Thanks for pointing this out. I added spaces in the other select calls, too.

reidka:
I tested this somewhat.  (Probably should have tested it a bit more thoroughly, but I think I hit the main code paths.)

PR #2619, opened 2016-06-13
Remark Result creation fix

christinem:
This fixes the issue where initial remark results would be null, because marks are now created automatically when a new result is created - so the marks need to be updated vs being created in the controller.

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -240,6 +240,7 @@ class Submission < ActiveRecord::Base
 
     # populate remark result with old marks
     original_result = get_original_result
+    original_marks = original_result.marks
 
     original_result.extra_marks.each do |extra_mark|
       remark.extra_marks.create(result: remark,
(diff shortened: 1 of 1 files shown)


app/models/submission.rb
```diff
@@ -249,13 +250,12 @@ def make_remark_result
...
     end
 
-    original_result.marks.each do |mark|
-      remark.marks.create(result: remark,
-                          created_at: Time.zone.now,
-                          markable_id: mark.markable_id,
-                          mark: mark.mark,
-                          markable_type: mark.markable_type)
+    remark.marks.each_with_index do |mark, index|
+      mark.mark = original_marks[index].mark
```
david-yz-liu: Use `mark.update`

app/models/submission.rb
```diff
@@ -249,13 +250,12 @@ def make_remark_result
                                 markable_type: extra_mark.markable_type)
     end
 
-    original_result.marks.each do |mark|
-      remark.marks.create(result: remark,
-                          created_at: Time.zone.now,
-                          markable_id: mark.markable_id,
-                          mark: mark.mark,
-                          markable_type: mark.markable_type)
+    remark.marks.each_with_index do |mark, index|
```
christinem: I realize this is going to take a little more than this, as the marks are still not associated with a criterion yet, and updating the markable_type and markable_id attributes isn't working as I am expecting it to

christinem:
@david-yz-liu Fixed. My problem was validation errors with markable_id since, because the results are automatically created, they are all automatically assigned to a criterion, so I had to go through each criterion and update the mark associated with it to avoid validation errors.

PR #2620, opened 2016-06-13
Changed to using objects instead of IDs

WChrisK:


lib/tasks/rubric.rake
```diff
@@ -51,15 +51,14 @@ namespace :db do
...
-                                  position: 1,
-                                  annotation_category_name: random_words(3))
+        ac = AnnotationCategory.create(assignment: assignment,
+                                       position: 1,
+                                       annotation_category_name: random_words(3))
 
-        (rand(10) + 3).times do
-          AnnotationText.create(annotation_category_id: index + assignment.id * 5,
-                                content: random_sentences(3))
+        3.times do
```
david-yz-liu: It would be nice to keep the random number of annotations.

WChrisK:
Changed back to random amounts of annotations.

PR #2621, opened 2016-06-13
Automarking

adisandro:
This adds support to run https://github.com/ProjectAT/uam 's Python autotester using Markus' automated test framework. It will need more work, but it's stable enough to be previewed.


.ruby-version
```diff
@@ -1 +1 @@
-2.1.3
+2.3.1
```
david-yz-liu: The ruby version on master is tied to the version of Ruby which is already installed on the vagrant VM the developers use. Please don't change this file now.

david-yz-liu:
@adisandro thanks! The script looks alright, but I can't try it out until next week. Would be nice to have a Python version (greater knowledge base among the faculty and TAs here).

.ruby-version
adisandro: I see. Well the problem is that 2.1.3 does not compile with rvm on Debian (and presumably on the latest Ubuntu too), the other developers that don't use the VM need to install a different version and remember to switch to it every time, which is error prone (e.g. if I run bundle install from the wrong ruby version). I'm not familiar with vagrant, is it hard to update the ruby version there?

adisandro:
That would introduce two different pieces to be kept in sync every time the autotesting framework change. @reidka and I were discussing something that could be related, in short we were thinking that maybe the users should not modify this script at all or even be aware of it, but just provide the test suite. Feel free to drop by my office to talk.

.ruby-version
david-yz-liu: Ah yes. It's not hard to update the virtual machine, but I would wait until that update has happened before updating the ruby version in this file. It should not be the case that other developers need to keep switching back and forth, though; this file really doesn't change, so it can be ignored for git purposes.

We can talk after I get back next week about updating the virtual machine.

david-yz-liu:
I am hesitant to rule out the possibility of customizing the script, although it is true that I don't know how much need there really would be to customize. It is certainly possible to provide the script as a default option for a MarkUs assignment's "test script".

I'm certainly in favour of maintaining just one script in one language. Given that, the language of choice should be python, not ruby.

david-yz-liu:
Sure, that's fine.

PR #2623, opened 2016-06-20
Automarking

adisandro:
This adds support to run https://github.com/ProjectAT/uam 's Python autotester using Markus' automated test framework, fixes a bunch of problem in the test framework itself, fixes the api to upload feedback files and marks.


lib/tools/example_script.py
```diff
@@ -19,9 +19,9 @@
 Macros:
 API_KEY       -- an Admin API key. It can be found on the admin dashboard page.
 ROOT_URL      -- the root domain of the MarkUs instance.
-ROOT_DIR      -- the directory containing the group repos.
+ROOT_DIR      -- the directory containing the submission repos.
```
david-yz-liu: @adisandro do you have time to revert changes to this file and markusapi.py? The only change I want to make to the existing scripts at this point is adding the cookie part. Anything else requires a bit more thought.

adisandro: Do you mean all changes (i.e. leave the script as broken), or just the changes about the repo name (real svn name vs the name exported via the markus' script)?

david-yz-liu: Fair enough. Either option is fine for me, whatever you have time to do now.

app/models/grouping.rb
```diff
@@ -749,17 +749,16 @@ def marking_state(result)
...
 
     #add the latest result from each of our test scripts
     test_script_ids.sum do |test_script_id|
-      last_mark = self.test_script_results
-                      .where(test_script_id: test_script_id)
-                      .last
-      last_mark.nil? ? 0 : last_mark.marks_earned
+      last_result = self.test_script_results
+                        .where(test_script_id: test_script_id)
+                        .first
```
david-yz-liu: There are two interesting changes here.

1. `first` instead of `last`. Is my understanding of the default order wrong? The mark should be based on the most recent test script result. Maybe just an explicit order in the association is necessary.

2. No `nil` check. I don't know that we can assume at least one test script result exists when this method is called.

app/controllers/results_controller.rb
```diff
@@ -138,12 +138,11 @@ def edit
 
   def run_tests
     grouping_id = params[:grouping_id]
-    # TODO: The submission id is set incorrectly as the grouping_id
     submission_id = Result.find(params[:id]).submission.id
 
     begin
       AutomatedTestsHelper.request_a_test_run(grouping_id,
-                                              'submission',
+                                              'request',
```
david-yz-liu: I'm surprised by this change; if this is changed, then it doesn't seem like the marks will be updated (see last part of the `process_result` method)

app/models/grouping.rb
adisandro: 1. There is an explicit order set for the association and it's DESC, so we need to use first to get the latest. I don't know why it's that way though, I can change it but I didn't want to risk affecting other queries.
2. Right, I don't think I got rid of it on purpose, I'll put it back.

app/controllers/results_controller.rb
adisandro: True, but shouldn't it be exactly that way since it is triggered here by a student pressing the "run tests" button based on tokens?

app/models/grouping.rb
david-yz-liu: 1. Ah, nice one. That seems like an odd order, but it's fine to leave it.

app/controllers/results_controller.rb
david-yz-liu: The "run tests" button is meant to be pushed by TAs and Admins, not students. (If the button shows for students that's a bug. They should only see that whole page once the assignment has been marked and released.)

adisandro: I think we are talking about two different "run tests" buttons. There is one in the admin/ta view for a collected submission, and one in the student view, Automated testing tab.
The student button is the one enabled by the token mechanism, and from my understanding is a way for students to test their code prior to submission, if enabled.

david-yz-liu: This is true, but I believe the student button doesn't use this method, but instead `execute_test_run` in `AutomatedTestsController`.

I looked into it more, and that method calls the `AutomatedTestsController`'s `run_tests` method, so there is a name collision, which may be contributing to the confusion.

adisandro: You're totally right, I was basically looking at it the opposite way.
Ok so the admin/ta button should trigger a new marking, I'll fix that.

adisandro: Making the button do the correct thing discovered some errors in the marking phase that happen only with multiple test runs. I'll make sure to fix that too before recreating this pull request with the changes we discussed.

PR #2624, opened 2016-06-20
Refactoring criteria methods

GitGeminiano:
Hello everyone,

I created a common interface for criteria which involved creation and refactoring of methods, and deletion of checks for marking_scheme.

Regards,
Christian


app/controllers/flexible_criteria_controller.rb
```diff
@@ -32,15 +32,10 @@ def new
...
-    if @criteria.length > 0
-      new_position = @criteria.last.position + 1
-    else
-      new_position = 1
-    end
     @criterion = FlexibleCriterion.new
     @criterion.assignment = @assignment
     @criterion.max = FlexibleCriterion::DEFAULT_MAX
-    @criterion.position = new_position
+    @criterion.position = @assignment.next_criterion_position
```
david-yz-liu: The behaviour for `Assignment#next_criterion_position` currently doesn't actually find the maximum position, which isn't good (when criteria are deleted and then new criteria are added). You should modify that method to actually compute the max position.

app/models/assignment.rb
```diff
@@ -264,15 +264,12 @@ def display_for_note
...
-    else
-      total = get_criteria(:ta).map(&:max).sum()
-    end
-    total.round(2)
+    get_criteria.first.nil? ? 0.round(2) : get_criteria.first.total_mark
+  end
+
+  # Returns the maximum possible mark for a particular assignment
+  def get_max_mark
+    get_criteria.first.nil? ? 0.round(2) : get_criteria.first.get_max_mark
```
david-yz-liu: 1. You can probably use `.empty?`
2. Seems like this doesn't add the marks for individual criteria together.

app/models/assignment.rb
```diff
@@ -695,6 +655,18 @@ def get_detailed_csv_report
...
+  # Returns an array of arrays of the form [mark, max] for the corresponding criteria
+  def get_marks_list(submission)
+    get_criteria.first.nil? ? [] : get_criteria.first.get_marks_list(submission)
+  end
+
+  # Returns an array of the form ['', max] or ['', weight] depending on the criteria
+  # This is not a good name right now because we either return weight or max
+  # TODO This may change when unifying flexible and rubric
+  def get_max_mark_for_criteria
+    get_criteria.first.nil? ? ['', 0] : get_criteria.first.get_max_mark_for_criteria
```
david-yz-liu: Calling `get_criteria.first.get_max_mark_for_criteria` makes no sense since it looks at all the criteria.

Generate the list here, not inside `get_criteria.first.get_max_mark_for_criteria``

app/models/assignment.rb
```diff
@@ -734,27 +707,15 @@ def get_criteria(user_visibility = :all)
   end
 
   def get_all_criteria
-    if marking_scheme_type == 'rubric'
-      rubric_criteria
-    else
-      flexible_criteria
-    end
+    criterion_class.where(assignment_id: self.id).order(:position)
```
david-yz-liu: `self` is redundant

app/models/assignment.rb
```diff
@@ -734,27 +707,15 @@ def get_criteria(user_visibility = :all)
...
   end
 
   def get_ta_visible_criteria
-    if marking_scheme_type == 'rubric'
-      rubric_criteria.where(ta_visible: true)
-    else
-      flexible_criteria.where(ta_visible: true)
-    end
+    get_all_criteria.where(ta_visible:true)
```
david-yz-liu: Put a space after the `:`

app/models/flexible_criterion.rb
```diff
@@ -58,49 +62,40 @@ def update_assigned_groups_count
...
-
-    criterion.description = row[2] if !row[2].nil?
-    criterion.position = next_criterion_position(assignment)
-
+    # Only set the position if this is a new record.
+    if criterion.new_record?
+      criterion.position = assignment.next_criterion_position
+    end
+    # Set description to the one cloned only if the original description is valid
+    criterion.description = working_row.shift if !row[2].nil?
```
david-yz-liu: Use `unless`

app/models/flexible_criterion.rb
```diff
@@ -163,4 +158,42 @@ def visible?
...
+      mark_to_change.mark = nil
+    else
+      mark_to_change.mark = criterion_name.to_f
+    end
+    mark_to_change.save
+  end
+
+  # Calculates the maximum mark that could be achieved considering all criteria marked
+  def total_mark
+    self.assignment.get_criteria(:ta).map(&:max).sum().round(2)
```
david-yz-liu: Again, this calculation doesn't belong here. Do the sum at the point where you already have access to all the criteria.

app/models/flexible_criterion.rb
```diff
@@ -58,49 +62,40 @@ def update_assigned_groups_count
...
+    working_row = row.clone
+    name = working_row.shift
+    # If a FlexibleCriterion with the same name exits, load it up.  Otherwise,
+    # create a new one.
+    criterion = assignment.get_criteria.find_or_create_by(name: name)
+    # Check that max is not a string.
+    begin
+      criterion.max = Float(working_row.shift)
+    rescue ArgumentError
+      raise CSVInvalidLineError, I18n.t('csv.invalid_row.invalid_format')
```
GitGeminiano: I will change instances of I18n.t for t, too.

app/controllers/flexible_criteria_controller.rb
GitGeminiano: @david-yz-liu Thank you. It is interesting to know that different criteria may have the same position.

david-yz-liu: Well, the point is to make sure that criteria *don't* have the same position.

app/models/assignment.rb
```diff
@@ -655,16 +652,25 @@ def get_detailed_csv_report
...
 
-  # Returns an array of arrays of the form [mark, max] for the corresponding criteria
+  # Returns an array of either [mark, max] or [mark, weight].
   def get_marks_list(submission)
-    get_criteria.first.nil? ? [] : get_criteria.first.get_marks_list(submission)
+    marks_list = []
+    get_criteria.each do |criterion|
+      mark = submission.get_latest_result
+                 .marks
+                 .where(markable_id: criterion.id)
```
david-yz-liu: use `find_by`

app/models/assignment.rb
```diff
@@ -655,16 +652,25 @@ def get_detailed_csv_report
     end
   end
 
-  # Returns an array of arrays of the form [mark, max] for the corresponding criteria
+  # Returns an array of either [mark, max] or [mark, weight].
   def get_marks_list(submission)
-    get_criteria.first.nil? ? [] : get_criteria.first.get_marks_list(submission)
+    marks_list = []
+    get_criteria.each do |criterion|
```
david-yz-liu: Use `map`

app/models/flexible_criterion.rb
```diff
@@ -100,6 +100,16 @@ def get_weight
     1
   end
 
+  # Returns the value entered when the criterion is set up.
+  def mark_set_up_value
+    max_mark
+  end
+
+  # Returns the maximum mark for a particular criterion.
+  def max_mark
```
david-yz-liu: It's not clear why there are two of these methods.

GitGeminiano: One name may be better than the other depending on the context, so I think it could improve code readability. I was thinking that the 'mark_set_up_value' method may be used in contexts where we either get the weight of a rubric criterion or the max of a flexible criterion. While, the 'max_mark' method may be used when we want to get the maximum mark that could be achieved for either a flexible or a rubric criterion.

app/controllers/flexible_criteria_controller.rb
GitGeminiano: I found out that all criteria for A1 have the same position, which I did not expect. I guess this is for testing purposes only.

app/models/flexible_criterion.rb
david-yz-liu: The first situation you described is not a good reason to use a different name. Please make them the same.

app/controllers/flexible_criteria_controller.rb
david-yz-liu: That's an error in the seeding, and you can make an issue for that.

app/models/flexible_criterion.rb
GitGeminiano: For now, the value returned by mark_set_up_value and max_mark is the same here in Flexible Criterion, but it is different in Rubric Criterion. I will change the name, add a TODO comment and get rid of the method in my next pull request, because I am using both methods in assignment.rb right now. I hope this is ok with you.

app/models/assignment.rb
```diff
@@ -654,23 +654,18 @@ def get_detailed_csv_report
...
     end
-    marks_list
   end
 
   # Returns an array of the form ['', max] or ['', weight] depending on the criteria.
   # This may not be a good name to account for the fact that we either return weight or max.
   # TODO This may change when unifying flexible and rubric.
-  def get_set_up_value_for_criteria
-    ([""] * get_criteria.count).zip(get_criteria.map(&:mark_set_up_value)).flatten
+  def get_max_mark_for_criteria
```
david-yz-liu: Change name to `get_max_marks_for_criteria`

PR #2626, opened 2016-06-21
Num reviews scroller

WChrisK:
Has some familiar stuff (was removed from previous commit chunk).


david-yz-liu:
I actually don't think it is necessary to add a database field for this. Why don't you make this an input element in the peer review page.

PR #2627, opened 2016-06-22
Added number scroller for randomization

WChrisK:


app/models/assignment.rb
```diff
@@ -7,6 +7,9 @@ class Assignment < ActiveRecord::Base
     rubric: 'rubric'
   }
 
+  MIN_PEER_REVIEWS_PER_PERSON = 1
+  MAX_PEER_REVIEWS_PER_PERSON = 5
```
WChrisK: Do you think these are needed? I don't know if in the future we want to constrain the admins (though currently they can just enter whatever value in the input field regardless of the above) and remove this, or keep it and use it for something later on?

app/assets/stylesheets/peer_reviews.scss
```diff
@@ -41,3 +41,11 @@
   padding-left: 19em;
   padding-right: 33em;
 }
+
+.peer-review-scrollbar-div {
+  padding-bottom: 8px;
+}
```
WChrisK: Is the `-div` redundant? (should be removed?)

app/models/assignment.rb
david-yz-liu: Yeah, I don't think we need a max. Keep the min, though.

app/assets/stylesheets/peer_reviews.scss
david-yz-liu: It's alright

app/views/peer_reviews/_peer_reviews_new_manager.js.jsx.erb
```diff
@@ -96,6 +96,11 @@
       return (
         <div>
           <ErrorDiv error={this.state.error} />
+          <div className="peer-review-scrollbar-div">
+            <%= j raw t('peer_review.number_per_group') %>: <input type="number"
```
david-yz-liu: I think you want to give the input a name so that it's passed in `params`

WChrisK: Currently it's not part of any form that's submitted, I was hoping to use JavaScript to pull out the value when a user presses the random button and transmit that via ajax.
Assuming there's not a better way and the above is to be done, can the name be left off then?

david-yz-liu: You can give it an id then

app/assets/stylesheets/peer_reviews.scss
david-yz-liu: Looking at it again though, I don't know if `scrollbar` is the best choice of noun

app/assets/stylesheets/peer_reviews.scss
```diff
@@ -41,3 +41,11 @@
   padding-left: 19em;
   padding-right: 33em;
 }
+
+.peer-review-amount-spinner {
+  padding-bottom: 8px;
+}
```
WChrisK: Looks like spinner was the name I was looking for in the end.

david-yz-liu:
Thanks, @WChrisK!

PR #2629, opened 2016-06-23
Operational new random peer review assignment algorithm

WChrisK:
~~**Note:** The backward assignment is not added yet, this is a request for a review of the code.~~


app/helpers/random_assign_helper.rb
```diff
@@ -0,0 +1,138 @@
...
+  def eligible_to_be_assigned(reviewer, reviewee)
+    # If they already have an assignment to the group, it's not viable.
+    if @reviewer_ids_assigned_reviewee_ids_map.has_key?(reviewer.id) and
+          @reviewer_ids_assigned_reviewee_ids_map[reviewer.id].member?(reviewee.id)
+      return false
+    end
+
+    # Since they're not assigned, the returned boolean is if they don't share students.
+    reviewer.does_not_share_any_students?(reviewee)
+  end
```
WChrisK: Can the above be refactored in any better way?

Ex:

```return not (@reviewer_ids_assigned_reviewee_ids_map.has_key?(reviewer.id) and @reviewer_ids_assigned_reviewee_ids_map[reviewer.id].member?(reviewee.id)) ? false : reviewer.does_not_share_any_students?(reviewee)```

```

app/helpers/random_assign_helper.rb
```diff
@@ -0,0 +1,138 @@
...
+  # - pr_assignment: The peer review assignment.
+  # - num_groups_for_reviewers: How many PR's each group in reviewer_groups should have.
+  def perform_random_assignment(pr_assignment, num_groups_for_reviewers)
+    reviewer_groups_relation = pr_assignment.valid_groupings
+    reviewee_groups_relation = pr_assignment.parent_assignment.valid_groupings
+    @shuffled_reviewees = []
+    @num_groups_for_reviewers = num_groups_for_reviewers
+    @reviewer_ids_assigned_reviewee_ids_map = Hash.new { |h, k| h[k] = Set.new }
+    @eligible_reviewers = []
+    reviewer_groups_relation.each { |reviewer| @eligible_reviewers.push(reviewer) }
```
WChrisK: Should I move the above into its own method to clean this part of the function up?

app/models/assignment.rb
```diff
@@ -46,6 +46,7 @@ class Assignment < ActiveRecord::Base
   # is no parent (the same holds for the child peer reviews)
   belongs_to :parent_assignment, class_name: 'Assignment', inverse_of: :pr_assignment
   has_one :pr_assignment, class_name: 'Assignment', foreign_key: :parent_assignment_id, inverse_of: :parent_assignment
+  has_many :peer_reviews, through: :groupings
```
WChrisK: This isn't doing what I think it should. I assume we want to call this on the PR assignment, instead of on the parent? This one only works for the parent.

david-yz-liu: This is interesting. Groupings do have to different named associations to peer reviews, right? (as_reviewer and as_reviewee, or something)

I'm not sure exactly what you want, but I think you'd want to specify a `:source` option for the `has_many` - http://guides.rubyonrails.org/association_basics.html#options-for-has-many-source

app/helpers/random_assign_helper.rb
```diff
@@ -0,0 +1,158 @@
...
+        return potential_reviewee
+      end
+    end
+
+    nil
+  end
+
+  def eligible_to_be_assigned(reviewer, reviewee)
+    # If they already have an assignment to the group, it's not viable.
+    if @reviewer_ids_assigned_reviewee_ids_map.has_key?(reviewer.id) and
```
david-yz-liu: You could initialize this dictionary to contain all the ids at the beginning.

app/helpers/random_assign_helper.rb
david-yz-liu: The name should be shortened anyway.

app/helpers/random_assign_helper.rb
```diff
@@ -0,0 +1,158 @@
...
+    @shuffled_reviewees[first_index], @shuffled_reviewees[second_index] =
+        @shuffled_reviewees[second_index], @shuffled_reviewees[first_index]
+  end
+
+  def add_peer_review_to_db_and_remember_assignment(reviewer, reviewee, shuffle_index)
+    result = reviewee.current_submission_used.get_latest_result
+    PeerReview.create!(reviewer: reviewer, result: result)
+
+    @reviewer_ids_assigned_reviewee_ids_map[reviewer.id].add(reviewee.id)
+    @shuffled_reviewees_assigned_reviewers[shuffle_index] = reviewer
```
david-yz-liu: Please change this to storing the `PeerReview` object.

app/helpers/random_assign_helper.rb
```diff
@@ -0,0 +1,158 @@
...
+
+    @shuffled_reviewees_assigned_reviewers = @shuffled_reviewees.map { |dummy| nil }
+  end
+
+  # We need to get all the existing peer reviews for assignments so we can
+  # skip some assignments if they exist.
+  def populate_from_existing_peer_reviews(pr_assignment)
+    # TODO : Remove? --- pr_assignment.get_peer_reviews().each do |peer_review|
+    # TODO - Is this okay? the 'peer reviews' forces us to go through the parent... doesn't seem right (do we have a choice?)
+    pr_assignment.parent_assignment.peer_reviews.each do |peer_review|
```
david-yz-liu: It should be possible to do this by defining a properly named `:through` association

app/helpers/random_assign_helper.rb
```diff
@@ -0,0 +1,158 @@
...
+  # - pr_assignment: The peer review assignment.
+  # - num_groups_for_reviewers: How many PR's each group in reviewer_groups should have.
+  def perform_random_assignment(pr_assignment, num_groups_for_reviewers)
+    @eligible_reviewers = []
+    @shuffled_reviewees = []
+    @shuffled_reviewees_assigned_reviewers = []
+    @num_groups_for_reviewers = num_groups_for_reviewers
+    @reviewer_ids_assigned_reviewee_ids_map = Hash.new { |h, k| h[k] = Set.new }
+
+    reviewer_groups_relation = pr_assignment.valid_groupings
```
david-yz-liu: Why don't you do `@eligible_reviewers = pr_assignment.valid_groupings`?

app/helpers/random_assign_helper.rb
```diff
@@ -0,0 +1,158 @@
...
+    while @eligible_reviewers.any?
+      @eligible_reviewers.each do |reviewer|
+        reviewee = get_next_reviewee_from_forward_search(reviewer, shuffle_index)
+
+        # A nil reviewee means we could not find one going forwards, and if so,
+        # then try going backwards through assignments (failure will throw).
+        if not reviewee.nil?
+          add_peer_review_to_db_and_remember_assignment(reviewer, reviewee, shuffle_index)
+        else
+          assign_backwards_or_throw(reviewer, shuffle_index)
```
david-yz-liu: It occurred to me that it should be possible to just do one big loop (`@shuffled_reviewees.length` times) and use mod to find the next "good" assignment for the current reviewer, rather than go forwards and then backwards. This should help simplify the algorithm further.

WChrisK:
Requesting final comments before changing the names.

app/helpers/random_assign_helper.rb
```diff
@@ -0,0 +1,174 @@
...
+    @eligible_reviewers = pr_assignment.valid_groupings
+    @shuffled_reviewees = []
+    @shuffled_reviewees_pr = []  # A list of peer reviews that match the indices for @shuffled_reviewees
+    @num_groups_for_reviewers = num_groups_for_reviewers
+    @reviewers_assigned_to = Hash.new { |h, k| h[k] = Set.new }  # reviewer_id => set(reviewee_ids)
+
+    reviewer_groups_relation = pr_assignment.valid_groupings
+    reviewee_groups_relation = pr_assignment.parent_assignment.valid_groupings
+
+    @eligible_reviewers.each { |reviewer|  @reviewers_assigned_to[reviewer.id] = Set.new }
```
WChrisK: On second thought, let me see if this pointless due to line 15...

david-yz-liu: Yeah, I wonder if it is.

app/helpers/random_assign_helper.rb
```diff
@@ -0,0 +1,174 @@
...
+
+    generate_shuffled_reviewees(reviewer_groups_relation, reviewee_groups_relation)
+    remove_groups_from_shuffle_having_peer_review(pr_assignment)
+    perform_assignments()
```
david-yz-liu: Don't put `()` at the end of a method call with no arguments.

app/helpers/random_assign_helper.rb
```diff
@@ -0,0 +1,174 @@
...
+  private
+  # Make sure there's enough reviewee groups in the following list such that
+  # we have enough repeats of 'reviewee_groups' that we can assign every
+  # reviewer group at least 'num_groups_for_reviewers' groups.
+  def generate_shuffled_reviewees(reviewer_groups, reviewee_groups)
+    num_times_to_add_reviewee = (reviewer_groups.size.to_f * @num_groups_for_reviewers / reviewer_groups.size).ceil
+    num_times_to_add_reviewee.times { @shuffled_reviewees += reviewee_groups }
+    @shuffled_reviewees.shuffle
+
+    @shuffled_reviewees_pr = @shuffled_reviewees.map { nil }
```
david-yz-liu: Put this in `perform_random_assignments` after you've removed the existing peer reviews. It'll save the call to `pop` below

app/helpers/random_assign_helper.rb
```diff
@@ -0,0 +1,174 @@
...
+    # are not in the list. Since we've removed all occurences, ignoring is okay.
+    unless index.nil?
+      @shuffled_reviewees.delete_at(index)
+      @shuffled_reviewees_pr.pop()
+    end
+  end
+
+  def perform_assignments
+    shuffle_index = 0
+    remove_ineligible_reviewers()
```
david-yz-liu: Move this to `perform_random_assignment` as well

app/helpers/random_assign_helper.rb
```diff
@@ -0,0 +1,174 @@
...
+  def get_next_reviewee_from_forward_search(reviewer, shuffle_index)
+    # First, check if the current index provided works.
+    potential_reviewee = @shuffled_reviewees[shuffle_index]
+    if eligible_to_be_assigned(reviewer, potential_reviewee)
+      return potential_reviewee
+    end
+
+    # Since the index doesnt work, look at the next element until the end of
+    # the list, and swap if we find a working group.
+    ((shuffle_index + 1)...@shuffled_reviewees.size).each do |new_shuffle_index|
```
david-yz-liu: Alrighty. Let's change this loop so that it effectively merges in `find_and_swap_with_group_or_throw` too:

```ruby
@shuffled_reviewees.size.times do |i|
  new_shuffle_index = shuffle_index + i % @shuffled_reviewees.size
  ...
  if eligible_to_be_assigned(reviewer, potential_reviewee)
    if new_shuffle_index == shuffle_index  # keep same position
      ...
    elsif new_shuffle_index > shuffle_index  # swap with empty reviewee
      ...
    else  # logic from find_and_swap_with_group_or_throw
      ...
```

If you find this makes the method too long, you can use helpers where appropriate. But only this method should be looping over `@shuffled_reviewers`

app/helpers/random_assign_helper.rb
WChrisK: The above has been shortened (I guess git only shows the diff of the code at the time of the comment, rather than what it currently is... which is nicer).

david-yz-liu: The comment is tied to a particular line; since this line didn't change (`end`), the comment still shows.

The purpose is to make the comment on the line that (is likely to)/should change.

david-yz-liu:
Well done, @WChrisK.

PR #2631, opened 2016-06-28
Added a missing translation

WChrisK:


WChrisK:
This is sort of interesting because I realized that JavaScript code I've written will not allow someone to trigger the error (as in submitting an empty list of reviewe[e/r]s), so technically the translation should never be called, but I don't know if someone would ever change it in the future and it'd handle this exceptional case.

So... should I instead leave it as in this PR, or remove the begin/rescue?

david-yz-liu:
This is fine, thanks @WChrisK.

PR #2632, opened 2016-06-29
Peer Result Student Editing View

christinem:


app/controllers/assignments_controller.rb
```diff
@@ -135,7 +135,7 @@ def student_interface
   def peer_review
     assignment = Assignment.find(params[:id])
     @assignment = assignment.is_peer_review? ? assignment : assignment.pr_assignment
-    if @assignment.is_hidden
+    if @assignment.nil?
```
david-yz-liu: Check both nil and is_hidden

app/controllers/peer_reviews_controller.rb
```diff
@@ -83,9 +85,10 @@ def assign(reviewer_groups, reviewee_groups)
   def unassign(reviewers_to_remove_from_reviewees_map)
     reviewers_to_remove_from_reviewees_map.each do |reviewee_id, reviewer_id_to_bool|
       reviewer_id_to_bool.each do |reviewer_id, dummy_value|
+        # find the PR that this reviewer made on this reviewee's submission
         reviewee_group = Grouping.find_by_id(reviewee_id)
-        result_id = reviewee_group.current_submission_used.get_latest_result.id
-        pr = PeerReview.find_by(result_id: result_id, reviewer_id: reviewer_id)
+        reviewee_prs = reviewee_group.peer_reviews
```
david-yz-liu: You can add a `.find(reviewer_id: reviewer_id)` to this line instead of making it a separate check below

app/controllers/results_controller.rb
```diff
@@ -56,7 +60,14 @@ def edit
     @extra_marks_percentage = @result.extra_marks.percentage
     @marks_map = Hash.new
     @old_marks_map = Hash.new
-    @mark_criteria = @assignment.get_criteria(:ta)
+
+    if !@assignment.pr_assignment.nil?
```
david-yz-liu: I believe there's a helper in `Assignment` which checks if an assignment has a peer review

app/controllers/results_controller.rb
```diff
@@ -56,7 +60,14 @@ def edit
...
+      @mark_criteria = @current_user.is_reviewer_for?(@assignment.pr_assignment.id, @result.id) ?
```
david-yz-liu: Don't use multi-line ternary operations, just use an `if`

app/controllers/results_controller.rb
```diff
@@ -336,13 +347,22 @@ def codeviewer
     @assignment = Assignment.find(params[:assignment_id])
     @submission_file_id = params[:submission_file_id]
     @focus_line = params[:focus_line]
-
+    @grouping = @current_user.grouping_for(Integer(params[:assignment_id]))
```
david-yz-liu: These `Integer` conversions are almost never necessary, so you should check to see if there's another way of writing this code if you're encountering an error.

app/controllers/results_controller.rb
```diff
@@ -336,13 +347,22 @@ def codeviewer
...
     @file = SubmissionFile.find(@submission_file_id)
     @result = @file.submission.get_latest_result
-    # Is the current user a student?
+
+    if @current_user.is_a_reviewer?(@assignment)
+      prs = PeerReview.where(reviewer_id: @grouping.id)
```
david-yz-liu: In general, you should be using associations a lot more. It's very rare to need to do a query of the form `Class.where...`, but instead something like `@grouping.peer_reviews`.

app/helpers/submissions_helper.rb
```diff
@@ -44,25 +44,38 @@ def set_release_on_results(groupings, release)
   end
 
   def get_submissions_table_info(assignment, groupings)
-    parts = groupings.select &:has_submission?
-    results = Result.where(submission_id:
-                             parts.map(&:current_submission_used))
+    if current_user.is_a_reviewer?(assignment)
+
```
david-yz-liu: I'm not sure what is supposed to go here (if anything).

app/helpers/submissions_helper.rb
```diff
@@ -44,25 +44,38 @@ def set_release_on_results(groupings, release)
...
+              r.submission_id == submission.id
+            end).first
+          else
+            result = submission.remark_result
+          end
         end
+
         g[:name] = grouping.get_group_name
-        unless current_user.student?
+        unless (current_user.is_a?(Student) && !current_user.is_a_reviewer?(assignment))
```
david-yz-liu: I don't know why this check changed

app/models/grouping.rb
```diff
@@ -669,24 +669,31 @@ def self.get_groupings_for_assignment(assignment, user)
...
-                                                    grouping: :group] },
-                        { accepted_student_memberships: :user },
-                        { inviter: :section }
-                        )
-              .where(memberships: { membership_status:
-                                   [StudentMembership::STATUSES[:inviter],
-                                    StudentMembership::STATUSES[:pending],
-                                    StudentMembership::STATUSES[:accepted]] })
-              .distinct
+      groupings = assignment.groupings
```
david-yz-liu: I don't know why this block changed.

app/models/grouping.rb
```diff
@@ -754,6 +761,11 @@ def get_total_test_script_marks
     end
   end
 
+  def review_for(reviewee_group)
+    prs = reviewee_group.peer_reviews
```
david-yz-liu: Again, you can do `.find` on this line.

app/views/results/marker/_marker_panes.html.erb
```diff
@@ -29,13 +29,15 @@
...
 
-            <button id='new_annotation_button' onclick='make_new_annotation(); return false;'>
-              <%= t('marker.new_annot')%>
-            </button>
-
-            <%= render partial: 'annotations/annotation_categories',
-                       locals: { annotation_categories: annotation_categories,
-                                 result_id: result.id } %>
+            <% if !@result.is_a_review? %>
+              <button id='new_annotation_button' onclick='make_new_annotation(); return false;'>
```
david-yz-liu: 1. Use `unless`.
2. I don't know why this button is not show to peer reviewers.

app/helpers/submissions_helper.rb
christinem: It was an attempt to avoid the Result query when it's not needed, but looking at it I realize I could just flip the `is_a_reviewer` check to a "not" check and avoid the empty block.

app/helpers/submissions_helper.rb
christinem: It changed because now we want to grab submissions info for student reviewers (for the reviewee submissions table on the landing page). So the old check won't allow us to do that because reviewers can be students. So now we needed the extra check to make sure the `unless` is only triggered if the user is a student but is not a reviewer. The first part of the check is redundant though (its included in `is_a_reviewer?`) and can be removed.

app/views/results/marker/_marker_panes.html.erb
christinem: My mistake, the button should show to peer reviewers.

app/models/grouping.rb
christinem: If you mean the extra `elsif` block, it's because I just had to filter out the groupings that the reviewer is responsible for, instead of all the assignments groupings.

david-yz-liu: It looks like this block has the form

```
groupings = assignment.groupings
groupings.something
```
and before it had
```
assignment.groupings.something
```

So unless the large query on the next line is meant to be outside of the block, I'm still not sure why this change was made.

app/helpers/submissions_helper.rb
david-yz-liu: None or almost none of the information provided in this block is necessary for the reviews table.

app/models/grouping.rb
christinem: Ah yes, you're right, that change was unnecessary, when I was working with this function I must have separated `groupings` onto it's own line to use it for something and forgot to change it back.

app/helpers/submissions_helper.rb
christinem: `g[:name_url]` is used to generate the link, but otherwise that's true. I could move that outside of the block instead then.

christinem:
@david-yz-liu This PR is ready for another review.

david-yz-liu:
@christinem I'm going to merge this in because it's all good work. There are a couple of other changes that need to be made for the student view:

1. Do not show the "Notes" segment beside the total mark.
2. Do not show the "Tags" tab in the right pane.
3. Do not show that Submission Rule or Bonus/Deductions in the Summary pane (the only thing shown should be the criteria).
4. Do not show the "Annotation Category" header or dropdown in the New Annotation modal.
5. I'm not sure jpgs (or pdfs) are being rendered properly.
6. The "Previous Submission" and "Next Submission" links aren't correct.

christinem:
Thanks @david-yz-liu, I'll work on these in a new pull request

PR #2633, opened 2016-06-29
Csv upgrade

WChrisK:
Now does: `reviewerGroup, revieweeGroup [, revieweeGroup...]` and fixed a bug.


app/controllers/peer_reviews_controller.rb
```diff
@@ -107,18 +115,21 @@ def download_reviewer_reviewee_mapping
...
-        raise CSVInvalidLineError if row.empty?
-        result = Result.find(row.first)
-        reviewer = Grouping.joins(:group).find_by(
-                                groups: { group_name: row.second },
-                                assignment_id: assignment_id)
-        PeerReview.create!(result: result, reviewer: reviewer)
+      result = MarkusCSV.parse(pr_mapping.read, encoding: encoding) do |row|
+        raise CSVInvalidLineError if row.size < 2
+        reviewer = Grouping.joins(:group).find_by(groups: { group_name: row.first })
+        row.shift()  # Drop the reviewer, the rest are reviewees and makes iteration easier.
```
david-yz-liu: You don't need the `()`

app/controllers/peer_reviews_controller.rb
```diff
@@ -107,18 +115,21 @@ def download_reviewer_reviewee_mapping
...
-                               encoding: params[:encoding]) do |row|
-        raise CSVInvalidLineError if row.empty?
-        result = Result.find(row.first)
-        reviewer = Grouping.joins(:group).find_by(
-                                groups: { group_name: row.second },
-                                assignment_id: assignment_id)
-        PeerReview.create!(result: result, reviewer: reviewer)
+      result = MarkusCSV.parse(pr_mapping.read, encoding: encoding) do |row|
+        raise CSVInvalidLineError if row.size < 2
+        reviewer = Grouping.joins(:group).find_by(groups: { group_name: row.first })
```
david-yz-liu: This logic isn't quite right if the same group has groupings for multiple assignments.

First get the `Group`, then use the `grouping_for_assignment` method to get the right grouping. Applies below as well.

lib/tasks/peer_review.rake
```diff
@@ -8,8 +8,17 @@ namespace :db do
     a1 = Assignment.find_by(short_identifier: 'A1')
     a1.update(has_peer_review: true)  # Creates 'a1pr' via callback.
 
+    # The group names must not be closed, doing a1pr.clone_groupings_from(a1.id)
```
david-yz-liu: I think you mean "cloned," but I also think that this is not a problem if you resolve my above comment.

WChrisK:
After doing the change for using group names, it now works with the old clone method.

PR #2634, opened 2016-06-29
Added mass unassigning convenience

WChrisK:


app/controllers/peer_reviews_controller.rb
```diff
@@ -89,6 +90,13 @@ def unassign(reviewers_to_remove_from_reviewees_map)
         pr.destroy
```
WChrisK: Should I also do `.map(&:destroy)` here? If so, another PR or do it in this PR?

david-yz-liu: I don't know that it is easy to get a list of all the PeerReviews to remove here, so I wouldn't worry about it.

app/controllers/peer_reviews_controller.rb
```diff
@@ -89,6 +90,13 @@ def unassign(reviewers_to_remove_from_reviewees_map)
...
       end
     end
+
+    # Mass unassigning comes last, reason being that the above code does not
+    # have to worry about find failing and returning nil.
+    selected_reviewee_group_ids.each do |reviewee_id|
+      result = Grouping.find(reviewee_id).current_submission_used.get_latest_result()
```
david-yz-liu: This line (which is similar to one above in the old code) should be able to be simplified using an association off of groupings for peer reviews.

PR #2636, opened 2016-06-30
Max mark refactoring

GitGeminiano:
Hello everyone,

I edited the criteria columns so that they have a max_mark common column that stores the maximum mark for a criterion. I also repurpose the mark column in the marks table to store a weighted mark, included some helpers to aid the use of internationalization in some modules, and changed the module name total_mark to max_mark.

Regards,
Christian


GitGeminiano:
I am still working on this

app/models/assignment.rb
```diff
@@ -338,7 +337,7 @@ def update_remark_request_count
 
   def total_criteria_weight
     factor = 10.0 ** 2
-    (get_criteria.sum('weight') * factor).floor / factor
+    (get_criteria.map(&:weight).sum() * factor).floor / factor
```
david-yz-liu: I'm not sure why you made this change

app/models/assignment.rb
```diff
@@ -338,7 +337,7 @@ def update_remark_request_count
 
   def total_criteria_weight
```
david-yz-liu: Can this method be deleted?

app/models/flexible_criterion.rb
```diff
@@ -3,7 +3,8 @@
 # Represents a flexible criterion used to mark an assignment that
 # has the marking_scheme_type attribute set to 'flexible'.
 class FlexibleCriterion < Criterion
-
+  extend ActionView::Helpers
```
david-yz-liu: I doubt both of these are necessary

app/models/flexible_criterion.rb
```diff
@@ -162,17 +153,17 @@ def add_tas_by_user_name_array(ta_user_name_array)
...
   end
 
-  def set_mark_by_criteria(mark_to_change, criterion_name)
-    if criterion_name == 'nil'
+  def set_mark_by_criterion(mark_to_change, mark_value)
+    if mark_value == 'nil'
       mark_to_change.mark = nil
     else
-      mark_to_change.mark = criterion_name.to_f
+      mark_to_change.mark = mark_value.to_f * weight
```
david-yz-liu: There's no need to multiply by `weight` for flexible criteria

app/models/flexible_criterion.rb
david-yz-liu: Actually, please revert the internationalization changes. You can do it separately later.

app/models/result.rb
```diff
@@ -90,7 +90,7 @@ def get_total_extra_percentage
 
   # Point deduction for late penalty
   def get_total_extra_percentage_as_points
-    get_total_extra_percentage * submission.assignment.total_mark / 100
+    get_total_extra_percentage * submission.assignment.max_mark / 100
```
david-yz-liu: Please call `round` here

app/views/results/common/_submission_selector.html.erb
```diff
@@ -24,8 +24,8 @@
     |
     (
     <em><%= t('marker.marks.total_mark') %></em>
-    <span id='current_mark_div'><%= result.total_mark %></span> /
-    <span id='total_mark_div'><%= assignment.total_mark %></span>
+    <span id='current_mark_div'><%= result.total_mark %></span>
+    <span> / <%= assignment.max_mark %></span>
```
david-yz-liu: I don't know why you changed the `/`

app/models/flexible_criterion.rb
GitGeminiano: Some instances of the t helper work by using extend and some others by using include. I am guessing sometimes we use instance methods, and sometimes we use class methods under the hood. I will revert the Internationalization changes of this pull request.

app/models/flexible_criterion.rb
GitGeminiano: Yes, you are right. The weight is always 1.

app/views/results/common/_submission_selector.html.erb
GitGeminiano: I changed it to keep consistency across files, since whenever we show the total we do it this way. However, I can revert this change.

app/models/assignment.rb
GitGeminiano: I was thinking about deleting it in my next pull request because until this point we are showing a total weight in the table that includes the list of criteria, but this will change in the next pull request when we merge the UI for rubric and flexible criteria.

app/models/assignment.rb
GitGeminiano: The sum method (when given an argument) can be applied to add up values in a particular column in the database. 'weight' used to be a column, but we replaced it with 'max mark'.

david-yz-liu: Ah, makes sense, thanks.

app/models/assignment.rb
david-yz-liu: Okay that's fine.

app/views/results/common/_submission_selector.html.erb
david-yz-liu: All the other ones should be changed (but don't do that here).

PR #2643, opened 2016-07-08
New criteria

GitGeminiano:
Hello everyone,

I moved the 'new' action to the 'criteria' controller as another step for merging flexible and rubric criteria. I also added a criterion type parameter that will likely appear more in my future pull requests, which will help if we were to allow the creation of different criteria for the same assignment. Finally, I added spec tests.

Regards,
Christian Garcia


app/controllers/criteria_controller.rb
```diff
@@ -0,0 +1,32 @@
+class CriteriaController < ApplicationController
+
+  def new
+    @assignment = Assignment.find(params[:assignment_id])
+    @criterion_type = params[:criterion_type]
+    @criterion = get_class(@criterion_type).new
```
david-yz-liu: Rather than define the helper, like this, first check whether the param is `"FlexibleCriterion"` or `"RubricCriterion"` (make a constant list in this controller which contains these two strings), and then use the `constantize` method.

app/models/flexible_criterion.rb
```diff
@@ -166,4 +166,9 @@ def set_mark_by_criterion(mark_to_change, mark_value)
     mark_to_change.save
   end
 
+  # Returns the criterion type.
+  def type
```
david-yz-liu: You don't need to do this to convert `FlexibleCriterion` to `flexible_criteria`, Rails has helpers for automatically doing this sort of thing as long as the naming conventions are followed: http://api.rubyonrails.org/classes/ActiveSupport/Inflector.html

Look for `pluralize` and `underscore`

app/views/flexible_criteria/index.html.erb
```diff
@@ -9,16 +9,17 @@
...
-          document.getElementById('new_criterion_prompt').focus();
-        } else {
-          jQuery.ajax({
-            url: '" +new_assignment_flexible_criterion_path(@assignment.id) + "',
-            type: 'GET'
-          });
-        }"%>
+    <%= link_to t('criteria.add_criterion'),
+      'javascript:void(0);',
+      onclick: "if(document.getElementById('new_criterion') != null) {
```
david-yz-liu: Please make this a separate function in one of the erb files while you're changing this.

Make the criterion type a parameter to this function, so when you call it here with the onclick, you pass in either "FlexibleCriterion" or "RubricCriterion"

GitGeminiano: How are we going to handle the creation of either a flexible criterion or a rubric criterion for the same assignment? Are we going to create two different links that the user can choose from?

app/models/flexible_criterion.rb
GitGeminiano: Thank you. I will look this up.

app/views/flexible_criteria/index.html.erb
david-yz-liu: Eventually, something like that.

app/controllers/criteria_controller.rb
GitGeminiano: I could avoid doing the check and creating the constants if I use:
@criterion = (@criterion_type.camelize + 'Criterion').constantize.new

david-yz-liu: I'm saying to pass in the full class name as the param, not just "flexible" or "rubric".

GitGeminiano: Yes, I was thinking that would interfere with the string polarization we use for controllers, but I just saw the tableize method. Thank you for your advice.

PR #2644, opened 2016-07-08
Edit criteria

GitGeminiano:
Hello everyone,

I moved the 'edit' action to the 'criteria' controller, and added the necessary spec tests.

Regards,
Christian


app/views/rubric_criteria/_criterion_editor.html.erb
```diff
@@ -55,7 +57,7 @@
   </div>
 
   <div class='rubric_criteria_levels'>
-    <%= render partial: 'rubric_criterion_levels',
+    <%= render partial: 'rubric_criteria/rubric_criterion_levels',
```
david-yz-liu: Are these changes because now the view search is through `criteria/`?

spec/controllers/criteria_controller_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 RSpec.describe CriteriaController, type: :controller do
 
-  describe 'Using Flexible criteria' do
+  describe 'Using Rubric criteria' do
```
david-yz-liu: All I can tell about the changes here is that you swapped the order from Flexible -> Rubric to Rubric -> Flexible. This seems strange.

GitGeminiano: I have two blocks, one for rubric criteria and another one for flexible criteria. Rubric Criteria is usually first, so I just swapped the tests. However, they are very similar, so git interpreted the swap this way. I ran the tests and it is working fine.

app/views/rubric_criteria/_criterion_editor.html.erb
GitGeminiano: Something just happened with GitHub. It is behaving weirdly. Yes, I got an error because it was looking for criteria/rubric_criterion_levels.

spec/controllers/criteria_controller_spec.rb
GitGeminiano: OK, I will revert the swap.

PR #2646, opened 2016-07-10
Update positions transaction

GitGeminiano:
Hello everyone,

I updated the update_positions action so that it includes a transaction that protects the database integrity, since multiple updates are performed.

Regards,
Christian


app/controllers/criteria_controller.rb
```diff
@@ -1,17 +1,14 @@
 class CriteriaController < ApplicationController
 
-  # This method handles the drag/drop criteria sorting
+  # This method handles the drag/drop criteria sorting.
   def update_positions
     @assignment = Assignment.find(params[:assignment_id])
     @criteria = @assignment.get_criteria
-    position = 0
+    position = 1
```
david-yz-liu: There's no need for this variable.

PR #2648, opened 2016-07-12
Peer Review Table Fix

christinem:
Fixes some errors with the rendering of the peer review table, and added a flash error for when submissions for the original assignment aren't collected before assigning reviewers.

### DIFF
--- a/app/models/result.rb
+++ b/app/models/result.rb
@@ -148,6 +148,11 @@ class Result < ActiveRecord::Base
     # want the peer-visible criteria
     visibility = is_a_review? ? :peer : user_visibility
 
+    if submission == nil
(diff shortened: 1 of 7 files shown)


app/models/result.rb
```diff
@@ -148,6 +148,11 @@ def check_for_nil_marks(user_visibility = :ta)
     # want the peer-visible criteria
     visibility = is_a_review? ? :peer : user_visibility
 
+    if submission == nil
```
david-yz-liu: This is the wrong place for this check. You should be able to do it before creating the result.

christinem: @david-yz-liu thanks, I changed the check to before it's created in the assign function.

PR #2650, opened 2016-07-12
Made the loading bar work properly for Peer Review view

WChrisK:
This was missing some calls, and I figured I'd convert them all to jQuery while I was at it.

---

Question, is a function like this overkill?
```javascript
void setMarkusLoadingDisplay() {
  jQuery('#working').css('display', 'none');`
}
```

### DIFF
--- a/app/views/peer_reviews/_peer_reviews_new_manager.js.jsx.erb
+++ b/app/views/peer_reviews/_peer_reviews_new_manager.js.jsx.erb
@@ -76,14 +76,14 @@
         this.refs.revieweeTable.clearCheckboxes();
       }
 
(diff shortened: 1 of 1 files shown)


david-yz-liu:
1. Creating a general function to wrap around the AJAX calls is good. It is actually possible to do this globally for all AJAX requests, which I am inclined to do. See http://api.jquery.com/Ajax_Events/. (But don't worry about doing this right now.)
2. In general an effort was made to not use jQuery for just DOM element lookup by id (and in other places). Follow what has already been done.
3. You should learn about the `show` and `hide` jQuery methods.

PR #2649, opened 2016-07-12
Static pr creator

WChrisK:

PR #2655, opened 2016-07-13
Peer Review CSV spec test (+ upgrades)

WChrisK:


WChrisK:
*NOTE*: This also contains the pull request from here: https://github.com/MarkUsProject/Markus/pull/2649

The only file that needs to be focused on is `peer_reviews_controller_spec.rb`

Overall I think it's a bit messy, but I figured I could get some comments on it and see what is the best way to clean this up (if possible)

spec/fixtures/files/peer_reviews/.gitignore
```diff
@@ -0,0 +1 @@
+temp_peer_review.csv
```
david-yz-liu: Why don't we make the test remove the file if it exists after it's done executing.

david-yz-liu:
@WChrisK the test is a little on the long side, but I do like that it's checking that the upload/download formats are inverses of each other. I'll merge this in after you've addressed my comment.

spec/fixtures/files/peer_reviews/.gitignore
david-yz-liu: Yeah, I think it's safe to assume no one will use the name, but maybe prefix it with an underscore just to be safe.

I'm surprised it's automatically deleted, but I'll go with it.

PR #2657, opened 2016-07-13
Criteria routes

GitGeminiano:
Hello everyone,

I changed the routing for criteria, flexible criteria and rubric criteria, so that we do not keep unnecessary routes.

Regards,
Christian

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -6,6 +6,18 @@ Markus::Application.routes.draw do
 
   # optional path scope (denoted by the parentheses)
   scope "(:locale)", locale: /en|fr|pt/  do
+    # Regular routes for rubric criteria
+    get 'assignments/:assignment_id/rubric_criteria' => 'rubric_criteria#index', as: :assignment_rubric_criteria
+    post 'assignments/:assignment_id/rubric_criteria/csv_upload' => 'rubric_criteria#csv_upload', as: :csv_upload_assignment_rubric_criteria
+    post 'assignments/:assignment_id/rubric_criteria/yml_upload' => 'rubric_criteria#yml_upload', as: :yml_upload_assignment_rubric_criteria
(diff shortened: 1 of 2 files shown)


GitGeminiano:
This pull request fails the tests for now, but I expect(it).to pass after all the other pull requests are merged.

config/routes.rb
```diff
@@ -74,36 +86,12 @@
...
       end
 
-      resources :criteria do
+      resources :criteria, except: [:index] do
         collection do
           post 'update_positions'
         end
       end
 
-      resources :rubric_criteria do
```
david-yz-liu: If you want to eliminate the RESTful routes, you don't need to delete this block entirely (and shouldn't because of the other routes). Instead, do `only: []` to get rid of all the automatically generated routes. This way you won't need to change the `collection` part.

GitGeminiano: That sounds reasonable. I will follow your approach. Thanks.

config/routes.rb
```diff
@@ -74,17 +75,13 @@
...
-        member do
-          delete 'destroy'
-        end
-
+      resources :rubric_criteria, only: [:index] do
```
david-yz-liu: I doubt `:index` is necessary.

GitGeminiano: I can double check, but we have an index action in the rubric and flexible controllers, so it seems  reasonable to have a route to display the marking scheme page.

GitGeminiano: I checked and we need to include index, at least until we migrate the index action into the criteria controller.

david-yz-liu: What does the `index` action do? Do we need it at all?

GitGeminiano: The controller grabs the criteria for a particular assignment. We use this info in the marking scheme page to display  all criteria with their respective maximum mark and visibility. We were not able to merge both index pages because they have different upload/download links. If you want I can try migrating the index action into the criteria controller and create partials for the links.

david-yz-liu: Ah, of course. That's fine, you can leave that.

PR #2658, opened 2016-07-14
PR Codeviewer Fix

christinem:
Fixes bug where codeviewer could not load files for a PR result.


david-yz-liu:
@christinem seems good, but the failing tests are troublesom. Might be a case of a hard-coded id instead of using an object's `id` attribute.

christinem:
@david-yz-liu ah yes, these tests are due to hard-coded result ids in the tests from back when these tests were written, it looks like. I had to fix quite a few of these already in another pull request. The reason they fail now and never did before is because my new authentication filter actually uses that id to try and find the result.

PR #2659, opened 2016-07-14
Pr section dropdown

WChrisK:


app/views/peer_reviews/_peer_reviews_new_manager.js.jsx.erb
```diff
@@ -1,6 +1,9 @@
 <script type='text/jsx'>
   /** @jsx React.DOM */
 
+  // A constant that represents all sections
+  var ALL_SECTIONS = '';
```
WChrisK: I set this to be a constant that is used to identify an "all selections selected" from the dropdown menu, since the dropdown menu has a non-empty string for the section name being selected. This would require a change to the `Section` model where:

```rails
validates_presence_of :name
```

would become

``rails
validates :name, presence: true, allow_blank: false
```

So that way this would be correct. The form may prevent an empty section name from being set when creating them though.

File: app/assets/stylesheets/peer_reviews.scss [hunk 2]
```diff
@@ -49,3 +49,7 @@
 .peer-review-amount-spinner input {
   width: 75px;
 }
+
+.white-background {
+  background-color: #FFFFFF;
```
david-yz-liu: You can use `white` here

app/views/peer_reviews/_peer_reviews_new_manager.js.jsx.erb
```diff
@@ -218,6 +268,13 @@
 
     clearCheckboxes: function() {
       this.refs.table.clearCheckboxes();
+      this.props.clearReviewerToRevieweeCheckboxData();
+    },
+
+    changeSection: function(event) {
+      this.clearCheckboxes();
+      this.setState({ sectionName: event.target.value });
+      this.props.refresh();
```
david-yz-liu: Is this call necessary? Aren' the tables just being hidden?

app/views/peer_reviews/_peer_reviews_new_manager.js.jsx.erb
```diff
@@ -266,17 +326,21 @@
...
+          section: group.section
         };
       }.bind(this));
 
       return (
         <div className={'col-right'}>
           <div className={'tab-pane ui-tabs ui-widget ui-widget-content ui-corner-all'} id='groups-tabs'>
+            <SectionSelectionMenu
+              changeSection={this.changeSection}
+              selectId={"peer-review-reviewee-section"} />
```
david-yz-liu: Interesting. You don't need curly braces here, and you can use single-quotes.

app/views/peer_reviews/_peer_reviews_new_manager.js.jsx.erb
```diff
@@ -147,22 +180,31 @@
...
+
     render: function() {
       var filters = [
         {
           name: 'all',
           text: '<%= j raw t('groups.all_grouping_counts') %>',
           func: function(group) {
-            return true;
-          }
+            if (this.state.sectionName == ALL_SECTIONS) {
```
david-yz-liu: Use `===` to avoid typecasting

PR #2661, opened 2016-07-18
Peer Review Student View Fixes

christinem:
Addresses and fixes comments from #2632, along with another issue where you could peer reviewer annotations when viewing an original site (as annotations are grabbed from submission files and submissions).

### DIFF
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -206,7 +206,7 @@ class ResultsController < ApplicationController
   def next_grouping
     grouping = Grouping.find(params[:grouping_id])
     assignment = Assignment.find(params[:assignment_id])
-    result = Result.find(params[:result_id])
+    result = Result.find(params[:id])
 
     if grouping.has_submission? && grouping.is_collected?
       if @current_user.is_reviewer_for?(assignment.pr_assignment, result)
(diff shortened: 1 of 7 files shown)


app/controllers/results_controller.rb
```diff
@@ -369,8 +397,15 @@ def codeviewer
       end
     end
 
-    @annots = @file.annotations
-    @all_annots = @file.submission.annotations
+    # if we are viewing an original result, we don't want
+    # to view annotations from any peer reviewers on those files
+    if !@result.is_a_review?
+      @annots = @file.annotations.select{|a| !User.find(a.creator_id).student?}
```
david-yz-liu: This doesn't need to be this complex. Select annotations based on their associated result.

app/views/results/common/_submission_selector.html.erb
```diff
@@ -3,25 +3,29 @@
...
+                                             noteable_id: @grouping.id,
+                                             noteable_type: 'Grouping',
+                                             action_to: 'note_message',
+                                             controller_to: 'results',
+                                             highlight_field: 'notes_dialog_link',
+                                             number_of_notes_field: 'number_of_notes'),
+                      id: 'notes_dialog_link',
+                      remote: true %>
     |
+    <%end%>
```
david-yz-liu: put spaces around end

PR #2662, opened 2016-07-19
Pr url anonymous

WChrisK:
The idea was to have it so you get:

```.../peer_reviews/42/test``` which would redirect you to the appopriate submission. After everything is working, then `test` would be renamed to something appropriate (edit).

### CODE REVIEW

File: app/controllers/peer_reviews_controller.rb [hunk 1]
```diff
@@ -157,6 +158,43 @@ def csv_upload_handler
...
+  # Forwarders
+  # ===========================================================================
+
+  def test
+    assignment = Assignment.find(params[:assignment_id])
+    parent_assignment = assignment.parent_assignment
+    pr = PeerReview.find(params[:id])
+    path = edit_assignment_submission_result_path(parent_assignment, pr.reviewee, pr.result)
+
+    #redirect_to path
```
WChrisK: This was proof it works (which it does when uncommented

app/controllers/peer_reviews_controller.rb
```diff
@@ -157,6 +158,43 @@ def csv_upload_handler
...
+    # controller_you_want.request.params["controller"] = "results"
+    # controller_you_want.request.params["action"] = "edit"
+    # controller_you_want.request.params["assignment_id"] = parent_assignment.id.to_s
+    # controller_you_want.request.params["submission_id"] = params[:submission_id].to_s
+    # controller_you_want.request.params["id"] = pr.result.id.to_s
+    #
+    # controller_you_want.response = response
+    #
+    # controller_you_want.edit
+  end
```
WChrisK: All of the above error out in some way either by failing with having the `html` field set so the controller calling `format.html` fails, or some variable is not set since a controller isnt instantiated (which is what I tried to do last but it didn't work)

app/helpers/submissions_helper.rb
```diff
@@ -59,6 +59,7 @@ def get_submissions_table_info(assignment, groupings)
           # Get the respective reviewee's result from grouping
           result_pr = current_user.grouping_for(assignment.id).review_for(grouping)
           result = Result.find(result_pr.result_id)
+          g[:name_url] = edit_assignment_result_path(assignment.parent_assignment.id, result_pr.result_id)
```
WChrisK: Note: Needed to use the parent ID for this to work properly with the view.

app/helpers/submissions_helper.rb
```diff
@@ -86,7 +87,9 @@ def get_submissions_table_info(assignment, groupings)
           g[:grace_credits_used] = grouping.grace_period_deduction_single
           g[:section] = grouping.section
         end
-        g[:name_url] = get_grouping_name_url(grouping, result)
+        unless assignment.is_peer_review?
+          g[:name_url] = get_grouping_name_url(grouping, result)
+        end
```
WChrisK: On second thought, should this just be a ternary statement with the line from the diff above @ line 62?

app/views/results/common/_annotations.js.erb
```diff
@@ -380,8 +380,9 @@
     positions = null;
     annotation_manager = null;
 
+    // TODO - Need to see if submission_id is required here, or if it breaks something unexpectedly?
     jQuery.ajax({
-      url:  '<%= codeviewer_assignment_submission_result_path(uid: @uid) %>',
+      url:  '<%= codeviewer_assignment_submission_result_path(uid: @uid, submission_id: @submission.id) %>',
```
WChrisK: I marked this because I had to add this in for it to work. I've checked it in the other view for the admin to make sure the view doesn't some drastically break, and I had no problem loading it as an admin under the submissions viewer.

I noted this because rails passes `submission_id` in from the controller normally, but in our case this is absent now, the question then is it potentially dangerous to use this? It assumes that the edit part of the controller will never change the `@submission` to be something else.

app/views/results/marker/_extra_marks_table.html.erb
```diff
@@ -12,13 +12,14 @@
...
        <td colspan="3">
+         <!-- TODO - Need to check if the submission_id breaks any other view? -->
          <%= link_to I18n.t("marker.marks.add_extra_mark"),
                'javascript:void(0);',
                onclick: "if (document.getElementById('add_extra_mark_prompt') != null) {
                  document.getElementById('add_extra_mark_prompt').down('input').focus();
                } else {
                  jQuery.ajax({
-                   url: '" + add_extra_mark_assignment_submission_result_path + "',
+                   url: '" + add_extra_mark_assignment_submission_result_path(submission_id: @submission.id) + "',
```
WChrisK: Highlighted for the same reason as the previous.

config/routes.rb
```diff
@@ -207,7 +207,17 @@
         end
       end
 
-      resources :peer_reviews, only: :index do
+      resources :results, only: [:edit], path: '/peer_reviews' do
```
WChrisK: Scoping was giving me issues, and at best it was putting it beforehand (ex: `assignments/peer_reviews/1/results/5`) whereas this way does it as per requested on slack (meaning: `assignments/1/peer_reviews/results/5/edit`)

app/helpers/submissions_helper.rb
```diff
@@ -86,7 +86,9 @@ def get_submissions_table_info(assignment, groupings)
           g[:grace_credits_used] = grouping.grace_period_deduction_single
           g[:section] = grouping.section
         end
-        g[:name_url] = get_grouping_name_url(grouping, result)
+        g[:name_url] = assignment.is_peer_review? && current_user.is_a_reviewer?(assignment) ?
+            edit_assignment_result_path(assignment.parent_assignment.id, result_pr.result_id) :
+            get_grouping_name_url(grouping, result)
```
WChrisK: Note: Needed to use the parent ID for this to work properly with the view. Don't know if this should be changed to using the PR assignment ID and updating everything else? Unsure how messy that might be.

If we drop use parent_assignment to just using assignment.id, a 404 error occurs when using the peer review id instead of the parent assignment's ID.

david-yz-liu: We will always use the parent assignment id.

app/views/results/common/_annotations.js.erb
david-yz-liu: This is a place where the path helper should be different depending on whether it's student or not.

I.e., there needs to be a `codeviewer_assignment_result_path` helper, corresponding to a codeviewer route under `assignments/../results/`

config/routes.rb
```diff
@@ -207,7 +207,17 @@
...
+        member do
+          get 'edit'
```
david-yz-liu: You don't need this one (it's generated automatically from the `resources`)

config/routes.rb
```diff
@@ -207,7 +207,17 @@
...
+          get 'next_grouping'
+          post 'toggle_marking_state'
+          post 'update_overall_comment'
+          post 'add_extra_mark'
+        end
+      end
+
+      resources :peer_reviews do
```
david-yz-liu: Don't change the `only: :index` part

app/views/results/common/_annotations.js.erb
```diff
@@ -381,7 +381,8 @@
     annotation_manager = null;
 
     jQuery.ajax({
-      url:  '<%= codeviewer_assignment_submission_result_path(uid: @uid) %>',
+      url: '<%= @current_user.student? ? codeviewer_assignment_result_path(uid: @uid) :
+                                         codeviewer_assignment_submission_result_path(uid: @uid) %>',
```
WChrisK: I assume that the path update also applied to this.

david-yz-liu: Yes, that's right.

david-yz-liu:
@WChrisK there seems to be a problem with `codeviewer` for images (and possibly PDFs). Try updating from master and seeing if you can replicate, and fix, this problem. @christinem would have done this before, so it would be good to consult her.

WChrisK:
Images and PDFs are now fully functional (special thanks to Christine as well)

david-yz-liu:
@christinem this has been merged in now.

PR #2663, opened 2016-07-19
Made table resize properly for all browser sizes

WChrisK:
This was needed since my changes caused it to fail on any different window size than my resolution. If I shrunk my window it would break everything, now the tables are a bit stretched but work great for any resolution (even insanely compressed screens too).

### DIFF
--- a/app/assets/stylesheets/peer_reviews.scss
+++ b/app/assets/stylesheets/peer_reviews.scss
@@ -2,44 +2,15 @@
 @import 'common/columns';
 
 .col-left {
-  width: 30%;
+  width: 40%;
+}
+
+.col-right {
+  width: 50%;
 }
 
 .col-center #icons {
   left: 45%;
-}
-
-.col-right {
-  width: 30%;
-}
-
-.new_group {
-  float: right;
-}
-
-.graders-to-criteria {
-  padding-bottom: 5px;
-}
-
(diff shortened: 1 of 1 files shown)


WChrisK:
One way to make this better might be to just drop the hacky floating middle piece and structure it like a table whereby we have
```[left space] [table] [fixed width space] [table] [right space]```

Where the tables are a fixed size, and the padding from the single column takes up the rest.

The floating buttons that go with you down the page is a cool idea, and maybe I'm not good enough at CSS yet to know how to tweak the file to do what I imagine (I can ask Christine tomorrow as well).

arkon:
You're right -- the current stuff is pretty janky. [Here's a potential replacement](http://codepen.io/arkon94/pen/YWZokQ) for it.

WChrisK:
I like the improvements, there's only one thing to fix and it should be good (I'm not sure how to fix it): http://i.imgur.com/i3TOCff.png
If the elements can be nudged right a bit, I think it'd be a good to go replacement.
This one works better than the old one including corner cases of empty tables.

david-yz-liu:
Thanks, @WChrisK and @arkon.

I don't have strong preferences here, and this does seem to be an improvements. Whatever is done for these tables, though, should also be done for the groups and graders table (and pull this out into `common.scss` or something).

PR #2664, opened 2016-07-19
Fixes #2601

### LINKED ISSUE #2601: No such translation element for 'graders.problem'
Cannot do `t('graders.problem')` as it doesn't appear to exist. It would appear on line 35 of the graders view, as:

```<h3><%= t('graders.problem') %></h3>```

Easiest solution is to remove it, the proper solution probably is take groups.problem (the only one with a problem field defined) and make it a global 'problem', since the text for 'groups.problem' is:

```problem: "There was a problem..."```

WChrisK:
#2601


WChrisK:
I assume the failing test is not a by-product of these changes since it was just adding definitions to an unrelated file.

PR #2665, opened 2016-07-19
Fixed a few n+1 queries

WChrisK:
Sadly the last query is giving me a lot more trouble to do the 'rails way' than I thought, but the other n+1 queries have all been fixed that are caused by this controller.


app/helpers/peer_review_helper.rb
```diff
@@ -45,10 +43,13 @@ def create_map_group_id_to_name(reviewer_groups, reviewee_groups)
   # Returns a map of reviewer_id => num_of_reviews
   def create_map_number_of_reviews_for_reviewer(reviewer_groups)
     number_of_reviews_for_reviewer = {}
+
+    # TODO - [N+1 query] Optimize with: SELECT reviewer_id, COUNT(reviewer_id) FROM peer_reviews GROUP BY reviewer_id
```
WChrisK: Bothers me that I didn't get this one, I figured I'd leave a comment and come back to it in the future, or maybe implement it in raw SQL and then put a nice TODO next to it to be fixed later. I tried a lot of things to attempt to make this work, and rails fought me too hard... I'm not giving up on this but rather putting it as a 'to do later', though if you know a way to get the above query to work that would be awesome.

app/helpers/peer_review_helper.rb
```diff
@@ -8,8 +8,7 @@ def create_map_reviewee_to_reviewers(reviewer_groups, reviewee_groups)
     peer_review_map = Hash.new { |hash, key| hash[key] = [] }
     reviewee_groups.each { |reviewee| peer_review_map[reviewee['id']] }
 
-    peer_reviews = PeerReview.where(reviewer_id: reviewer_ids)
-    peer_reviews.each do |peer_review|
+    PeerReview.preload(:result).where(reviewer_id: reviewer_ids).each do |peer_review|
```
david-yz-liu: Generally use `includes`, not `preload`

app/helpers/peer_review_helper.rb
david-yz-liu: Do something like `number_of_reviews_for_reviewer = PeerReview.group(:reviewer_id).having(reviewer_id: reviewer_groups.map { |g| g['id'] }).count` (note that `count` returns a hash, I believe).

I've lost track of the type of `reviewer_groups`. If it's a plain dictionary you have to do `['id']`, otherwise you might be able to do `&:id` if it's an array of objects.

app/helpers/peer_review_helper.rb
```diff
@@ -8,8 +8,7 @@ def create_map_reviewee_to_reviewers(reviewer_groups, reviewee_groups)
     peer_review_map = Hash.new { |hash, key| hash[key] = [] }
     reviewee_groups.each { |reviewee| peer_review_map[reviewee['id']] }
 
-    peer_reviews = PeerReview.where(reviewer_id: reviewer_ids)
-    peer_reviews.each do |peer_review|
+    PeerReview.includes(:result).where(reviewer_id: reviewer_ids).each do |peer_review|
```
david-yz-liu: Hmmm, do you not also have to include the associatied submission and grouping? (Nested include.)

WChrisK: Added the required includes, and _wow_ this is insanely fast (probably as expected though).

app/helpers/peer_review_helper.rb
```diff
@@ -8,8 +8,9 @@ def create_map_reviewee_to_reviewers(reviewer_groups, reviewee_groups)
     peer_review_map = Hash.new { |hash, key| hash[key] = [] }
     reviewee_groups.each { |reviewee| peer_review_map[reviewee['id']] }
 
-    peer_reviews = PeerReview.where(reviewer_id: reviewer_ids)
-    peer_reviews.each do |peer_review|
+    PeerReview.includes(:reviewer)
```
david-yz-liu: I don't think you need the `reviewer` includes; you could actually do `peer_review.reviewer_id` below

PR #2666, opened 2016-07-20
Added criteria dropdown selection

WChrisK:


app/views/rubric_criteria/index.html.erb
```diff
@@ -18,15 +18,23 @@
   <div class='heading_buttons'>
     <%= link_to t('criteria.add_criterion'),
       'javascript:void(0);',
-      onclick: "if(document.getElementById('new_criterion') != null) {
+      onclick: "var criteriaSelectionText = jQuery('#marking-scheme-criteria-dropdown :selected').val();
+      if (criteriaSelectionText !== 'RubricCriterion' && criteriaSelectionText !== 'FlexibleCriterion') {
+        alert('Please select a criterion from the dropdown menu.');
```
WChrisK: Better way of doing this? Ex: Using a flash message (if so, is there an example from javascript somewhere I could use? I was looking for one but was unsuccessful so far).

WChrisK:
Since Christian is doing the work for making these work, I'm assuming FlexibleCriteria's index file will end up getting changed, and the code listed here will just become a single determiner for criteria. Therefore is there a need to also add this to the other index file?

GitGeminiano:
Chris, I will add it to both flexible and rubric :)

app/views/rubric_criteria/index.html.erb
```diff
@@ -18,15 +18,23 @@
...
       } else {
         jQuery.ajax({
           url: '" + new_assignment_criterion_path(@assignment) + "',
-          data: 'criterion_type=RubricCriterion',
+          data: 'criterion_type=' + criteriaSelectionText,
           type: 'GET'
         });
       } "%>
+    <select id="marking-scheme-criteria-dropdown">
+      <option value="">Select type:</option>
```
david-yz-liu: These strings should all be internationalized.

app/views/rubric_criteria/index.html.erb
```diff
@@ -18,15 +18,23 @@
...
+      onclick: "var criteriaSelectionText = jQuery('#marking-scheme-criteria-dropdown :selected').val();
+      if (criteriaSelectionText !== 'RubricCriterion' && criteriaSelectionText !== 'FlexibleCriterion') {
+        alert('Please select a criterion from the dropdown menu.');
+      } else if (document.getElementById('new_criterion') != null) {
         document.getElementById('new_criterion_prompt').focus();
       } else {
         jQuery.ajax({
           url: '" + new_assignment_criterion_path(@assignment) + "',
-          data: 'criterion_type=RubricCriterion',
+          data: 'criterion_type=' + criteriaSelectionText,
```
david-yz-liu: Please change the value of `data` to be an object (`{...}`) rather than a string.

app/views/rubric_criteria/index.html.erb
david-yz-liu: Eh, an alert is fine for now. This message should also be internationalized, though.

PR #2667, opened 2016-07-20
Unify results

GitGeminiano:
Hello everyone,

I deleted marking scheme checks  related to the results view and merged some of the views that shared common information between markers and students. Please have a look at it.

Regards,
Christian


app/controllers/results_controller.rb
```diff
@@ -59,8 +59,10 @@ def edit
     @first_file = @files.first
     @extra_marks_points = @result.extra_marks.points
     @extra_marks_percentage = @result.extra_marks.percentage
-    @marks_map = Hash.new
-    @old_marks_map = Hash.new
+    @flexible_marks_map = Hash.new
```
david-yz-liu: Don't make two different sets of variables. Anywhere you need to distinguish between the types in the views, just use the type of the first mark's criterion.

app/views/annotations/update_annotation.js.erb
```diff
@@ -1,7 +1,8 @@
 document.getElementById('annotation_summary_list').innerHTML =
-  '<%= escape_javascript(render partial: "results/marker/annotation_summary",
+  '<%= escape_javascript(render partial: "results/common/annotation_summary",
                                 locals: { annots: @annotations,
-                                          submission_file_id: @submission_file_id }) %>';
+                                          submission_file_id: @submission_file_id,
+                                          marker: true}) %>';
```
david-yz-liu: I only wanted you to combine the views related to the criteria types. Please revert all other changes you made.

david-yz-liu:
You need to merge the summary table into one view as well. Keep the headings "Score", "Weight", and "Grade" for now. Give the flexible criteria a "weight" of 1.

app/controllers/results_controller.rb
GitGeminiano: I made two sets because looking ahead we will need to show the marks in the summary pane for both flexible and rubric criteria. Flexible and rubric have different formats so we need to pass a list of flexible criteria to one partial and a list of rubric criteria to the other partial. Before we used to check for the marking_scheme type and either render the partial for flexible criteria or render the partial for rubric_criteria.

GitGeminiano: Also, if it seems like I am passing a lot of variables, I could use the fact that instance variables can be accessed directly anywhere down the road, but it is good practice to pass variables to partials.

app/views/annotations/update_annotation.js.erb
GitGeminiano: I will revert the changes where needed. Thanks.

app/controllers/results_controller.rb
GitGeminiano: I looked at your last comment. If we will make no distinction between flexible and rubric criteria in the summary pane, then I think I will no longer need these sets of variables. Thank you.

app/views/results/common/_annotations.js.erb
```diff
@@ -5,219 +5,186 @@
   var modal = null;
```
GitGeminiano: I will revert these changes.

app/assets/javascripts/Grader/marking.js
```diff
@@ -99,23 +101,23 @@ jQuery(document).ready(function() {
   // Handle the expand/collapse buttons
   jQuery('#expand_all').click(function() {
     jQuery('.mark_criterion_level_container').each(function() {
-      show_criterion(parseInt(this.getAttribute('data-id'), 10), jQuery(this).attr('data-scheme'));
```
david-yz-liu: @GitGeminiano I realized that `id` does not uniquely identify criteria, if we allow multiple types of criteria to appear. This means that `show_criterion` and `hide_criterion` need to have both an id and class passed in.

However, you do *not* need to have a new data attribute (`criterion-class`) to store this information. You should be able to use the existing (HTML) classes of the div themselves (either `mark_criterion_level_container` or the one for flexible criteria), and simply pass those into these functions, and do the corresponding search in `show_criterion` and `hide_criterion`.

app/views/results/_result_summary.html.erb
```diff
@@ -5,6 +5,11 @@
                        result: @result,
                        mark_criteria: @mark_criteria,
                        marks_map: @marks_map,
+                       old_marks_map: @old_marks_map,
+                       rubric_marks_map: @rubric_marks_map,
```
david-yz-liu: Didn't you remove these?

GitGeminiano: I did. I will delete these ones, too.

GitGeminiano:
I will add one more change to this pull request.

GitGeminiano:
One more styling change seems to be needed, so I will be pushing one more commit.

PR #2673, opened 2016-07-25
Random optimize

WChrisK:


Gemfile
```diff
@@ -15,7 +15,7 @@ gem 'ya2yaml'
 gem 'i18n'
 gem 'dynamic_form'
 gem 'exception_notification'
-gem 'activerecord-import'
+gem 'activerecord-import', '>= 0.2.0'
```
WChrisK: Do you think this is needed? The wiki said I should be using the latest version, however upon doing a gem upgrade... I don't recall it upgrading and it appeared to have used 0.10.0
Probably minor since this is working well regardless of the version currently.

app/helpers/random_assign_helper.rb
```diff
@@ -5,186 +5,234 @@ class UnableToRandomlyAssignGroupException < Exception
...
+    # Peer reviews require IDs to have been made, so they come last.
+    peer_reviews_reviewer_result.each do |prdata|
+      peer_review = PeerReview.new(reviewer: prdata[0], result: prdata[1])
+      peer_reviews << peer_review
+    end
 
-  def unlink_and_delete_previous_peer_review_assignment(peer_review, reviewee)
-    @reviewers_assigned_to[peer_review.reviewer.id].delete(reviewee.id)
-    peer_review.destroy
+    PeerReview.import peer_reviews, :validate => false
```
WChrisK: The good stuff is in the lines above. I think it looks a bit messy (since I'm making a pair out of an array, which I don't like doing as its cryptic, but this PR is to prove that it's working).

Gemfile
david-yz-liu: You can probably revert this change. We're using activerecord-import in a pretty limited way.

WChrisK:
In summary:

Working great, the speed is phenomenally faster, even faster than the original way it was done.
The algorithm I'm pretty sure is not right, it appears to be working fine now but I'm _very sure_ I saw a case where it didn't assign the distribution of students properly, and I think I know why. However, it's working 99.9% of the time (or rather, I've been unable to make the corner case that I swear I saw happened...)

app/helpers/random_assign_helper.rb
```diff
@@ -5,186 +5,234 @@ class UnableToRandomlyAssignGroupException < Exception
...
+
+      result = Result.new(submission: reviewee.current_submission_used,
+                          marking_state: Result::MARKING_STATES[:incomplete])
+      results << result
+      assignment_criteria.each { |criterion| marks << criterion.marks.new(result: result) }
+      peer_reviews_reviewer_result << [reviewer, result]
+    end
+
+    Result.import results, :validate => false
+    Mark.import marks, :validate => false
```
WChrisK: Note: I will change the `=>` symbols to something more newer if I can do something like `validate: false`

app/helpers/random_assign_helper.rb
```diff
@@ -5,186 +5,234 @@ class UnableToRandomlyAssignGroupException < Exception
...
-      end
+    unmark_reviewer_assignment(prev_index)
+    swap_indices(@shuffled_reviewees, prev_index, shuffle_index)
+    mark_reviewer_assigned_to(prev_reviewer_id, @shuffled_reviewees[prev_index], prev_index)
+    mark_reviewer_assigned_to(reviewer_id, @shuffled_reviewees[shuffle_index], shuffle_index)
+  end
 
-      perform_group_exchange(reviewer, shuffle_index, next_index)
-      return
+  def attempt_assign_backwards(reviewer_id, shuffle_index)
```
david-yz-liu: I don't know why you went back to doing this.

WChrisK: That was me trying to skip out on using the modulus and seeing if there was a performance boost (I don't think there was or it was negligible, this is probably what I get for spending a large chunk of my free time reading assembly optimizations).
This would be converted back, but was just done as a test to see if it helped at all.

app/helpers/random_assign_helper.rb
```diff
@@ -92,38 +96,32 @@ def generate_group_to_students_map(group_id_list)
...
-
-    if groups_share_students?(reviewer_id, reviewee_id)
-      # If they do overlap with students, then we know they are never eligible
-      # again, so we should then remember these cannot be assigned to each other.
-      @reviewers_already_assigned_to[reviewer_id].add(reviewee_id)
-      return false
-    end
-
-    true
+    !(is_reviewer_assigned_to?(reviewer_id, reviewee_id) || groups_share_students?(reviewer_id, reviewee_id))
```
WChrisK: I decided to change the above since after going over it, I realized that it's technically misleading to store it in `@reviewers_already_assigned_to` when it's not actually assigned to. I also figure that the amount of times `group_share_students(...)` is called should not be that taxing (or if it is, then a change can be made to either revert this, or put it in a set of groups that do not match and update the logic accordingly which probably would be the more clearer option).
If you want, I could add in the last option.

PR #2675, opened 2016-07-26
Get criteria both

GitGeminiano:
Hello everyone,

I refactored the get_criteria method and its appearances so that it supports having two types of criteria for the same assignment .

Regards,
Christian


app/controllers/criteria_controller.rb
```diff
@@ -51,14 +50,20 @@ def update
...
   def update_positions
     @assignment = Assignment.find(params[:assignment_id])
     @criteria = @assignment.get_criteria
 
     ActiveRecord::Base.transaction do
-      params[:criterion].
-        each_with_index { |id, index| @assignment.criterion_class.update(id, position: index + 1) if id != '' }
+      params[:criterion].each_with_index do |id, index|
+        begin
+          criterion_to_update = FlexibleCriterion.find(id)
```
david-yz-liu: This isn't good (same comment as other pull request).

Need to modify the view to send not just id and index, but criterion type as well.

app/controllers/graders_controller.rb
```diff
@@ -280,8 +280,9 @@ def criteria_with_assoc(assignment, options = {})
     criterion_ids = options[:criterion_ids]
     includes = options[:includes] || CRITERION_ASSOC
 
-    criteria = assignment.get_criteria.includes(includes)
-    criterion_ids ? criteria.where(id: criterion_ids) : criteria
+    criteria = assignment.get_criteria(:all, :rubric).includes(includes) +
```
david-yz-liu: Pass in `includes` as another argument to `get_criteria`

app/models/assignment.rb
```diff
@@ -628,7 +628,8 @@ def get_detailed_csv_report
         # total percentage, total_grade
         result.concat(['','0'])
         # mark, max_mark
-        result.concat(get_criteria.pluck("''", :max_mark).flatten)
+        result.concat(Array.new(criteria_count, '').
+          zip(get_criteria(:all, :rubric).pluck(:max_mark) + get_criteria(:all, :flexible).pluck(:max_mark)).flatten)
```
david-yz-liu: Alright, you don't need to use `pluck` here. Just map over the result of `get_criteria`

app/models/assignment.rb
```diff
@@ -688,26 +689,32 @@ def criterion_class
...
-      get_peer_visible_criteria
+      get_peer_visible_criteria(type)
     end
   end
 
-  def get_all_criteria
-    criterion_class.where(assignment_id: id).order(:position)
+  def get_all_criteria(type)
+    if type == :all
+      (RubricCriterion.where(assignment_id: id) + FlexibleCriterion.where(assignment_id: id)).sort_by(&:position)
```
david-yz-liu: You don't need to use `where` here. Assignment already has an association for both rubric and flexible criterion. (Should have caught this one earlier, but oh well.)

app/models/criterion.rb
```diff
@@ -37,28 +37,37 @@ def self.assign_all_tas(criterion_ids, ta_ids, assignment)
   def self.assign_tas(criterion_ids, ta_ids, assignment)
```
david-yz-liu: It's the same comment as the first one. Need `criterion_ids` to now have both ids and type.

GitGeminiano:
There is still one more comment to be addressed. However, I pushed my changes so that the method get_criteria is up to date. @WChrisK You can use this code to guide your changes.

PR #2679, opened 2016-07-29
Christian unity criteria merge

WChrisK:
~~There are a lot of minor things I likely will have to check, and the CSS styling is the last battle to overcome at the end of this -- which could be easy if luck is on my side.~~

### DIFF
--- a/app/assets/javascripts/Grader/marking.js
+++ b/app/assets/javascripts/Grader/marking.js
@@ -47,7 +47,7 @@ jQuery(document).ready(function() {
       var params = {
         'mark': this.value || '',
         'authenticity_token': AUTH_TOKEN
-      }
+      };
 
       jQuery.ajax({
         url:  this.getAttribute('data-action'),
@@ -169,7 +169,7 @@ jQuery(document).ready(function() {
     jQuery('.mark_grade_input').each(function () {
       expand_unmarked(this, 'FlexibleCriterion');
     });
-    jQuery('.mark_grade_input_checkbox').each(function () {
+    jQuery('.mark_grade_input_checkbox_container').each(function () {
       expand_unmarked(this, 'CheckboxCriterion');
     });
   });
@@ -197,15 +197,28 @@ function expand_unmarked(elem, criterion_class) {
     } else {
       hide_criterion(parseInt(elem.getAttribute('data-id'), 10), criterion_class);
     }
-  } else {
+  } else if (criterion_class == 'FlexibleCriterion') {
     if (elem.value == '') {
       show_criterion(parseInt(elem.getAttribute('data-id'), 10), criterion_class);
     } else {
       hide_criterion(parseInt(elem.getAttribute('data-id'), 10), criterion_class);
     }
-  }
-};
+  } else {
+    // We have two radio buttons nested in this block that we must evaluate.
+    var anyRadioButtonSet = false;
+    jQuery(elem).find('.mark_grade_input_checkbox').each(function(index, element) {
+      if (element.checked) {
+          anyRadioButtonSet = true;
+      }
+    });
 
+    var hideOrShowCriterionFunc = anyRadioButtonSet ? hide_criterion : show_criterion;
+    hideOrShowCriterionFunc(parseInt(elem.getAttribute('data-id'), 10), criterion_class);
+  }
+}
+
+// NOTE: This function is only called by rubric/flexible, not checkbox.
+// It should be upgraded to focus_mark_criterion_type() in the future.
 function focus_mark_criterion(id) {
   if (jQuery('#mark_criterion_' + id).length != 0) {
     if (jQuery('#mark_criterion_' + id).hasClass('expanded')) {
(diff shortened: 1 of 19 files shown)


app/assets/stylesheets/grader.scss
```diff
@@ -649,6 +649,118 @@ ul.annotation_text_list {
...
+    float: right;
+
+    .mark_grade_input{
+      font-size: 1em;
+      margin: 1px 1px 4px;
+      min-width: 0;
+      padding: 2px;
+    }
+  }
+
```
WChrisK: **Note to self:** The above may not be needed, I will check this as I'm doing the CSS handling at the end.

app/assets/stylesheets/grader.scss
```diff
@@ -671,6 +783,8 @@ ul.annotation_text_list {
 
 }
 
+/* ============= */
+
```
WChrisK: **Note to self:** Remove

app/views/results/common/_marks_summary_table.html.erb
```diff
@@ -7,7 +7,7 @@
   </thead>
   <% @mark_criteria.each do |criterion| -%>
     <%
-      mark = marks_map[criterion.id]
+      mark = marks_map[[criterion.class.to_s, criterion.id]]
```
WChrisK: This was a really cool idea Christine had which made the code changes significantly smaller. Since its a string/int pair, that makes key-lookup stable.

app/assets/javascripts/Grader/marking.js
```diff
@@ -175,12 +215,23 @@ function focus_mark_criterion(id) {
...
+      case 'RubricCriterion':
+          nodeToHide = document.getElementById('mark_criterion_' + id);
+          break;
+      case 'FlexibleCriterion':
+          nodeToHide = document.getElementById('flexible_criterion_' + id);
+          break;
+      default:
+          nodeToHide = document.getElementById('checkbox_criterion_' + id);
+          break;
+  }
```
WChrisK: I could condense this down to:

```var leadingText = (criterion_class == 'RubricCriterion' ? 'mark' : 
                                        criterion_class == 'FlexibleCriterion' ? 'flexible' : 'checkbox');
nodeToHide = document.getElementById(leadingText + '_criterion_' + id);
```

david-yz-liu: Use `if` statements, not `switch` nor nested ternary operators.

app/assets/javascripts/Grader/marking.js
```diff
@@ -197,15 +197,28 @@ function expand_unmarked(elem, criterion_class) {
...
     } else {
       hide_criterion(parseInt(elem.getAttribute('data-id'), 10), criterion_class);
     }
+  } else {
+    // We have two radio buttons nested in this block that we must evaluate.
+    var anyRadioButtonSet = false;
+    jQuery(elem).find('.mark_grade_input_checkbox').each(function(index, element) {
+      if (element.checked) {
+          anyRadioButtonSet = true;
+      }
```
WChrisK: Is this a good or bad idea? Turning the above .each from

```javascript jQuery(elem).find('.mark_grade_input_checkbox').each(function(index, element) {
  if (element.checked) {
    anyRadioButtonSet = true;
  }
```

to

```javascript jQuery(elem).find('.mark_grade_input_checkbox').each(function(index, element) {
  anyRadioButtonSet |= element.checked;
}
```

david-yz-liu: Eh, you should be able to put the `checked` into the selector itself and then see if the `find` returns anything or not.

app/models/result.rb
```diff
@@ -59,7 +59,7 @@ def get_subtotal(user_visibility = :ta)
     unless marks.empty?
       assignment = submission.grouping.assignment
       assignment.get_criteria(user_visibility).each do |criterion|
-        mark = marks.find_by(markable_id: criterion.id)
+        mark = marks.find_by(markable_id: criterion.id, markable_type: criterion.class.to_s)
```
GitGeminiano: This brings to light other issues. Since it seems possible to have different types of criteria with the same id, we should create marks with the appropriate markable_type, which may require changes in other areas in this file such as in the callback method create_marks, and in check_for_nil_marks. Other files that deal with the creation or checking of marks may also need to be modified so that they are not only associated to a particular markable_id, but also to a markable_type. Line 514 in the results_controller for example.

Also I looked at the test that started failing with this change and it fails because the mark that we get now has a different markable_type, so I guess we have to double check both the tests and the code. Unless it is feasible to make sure we have different ids across criteria.

I hope this helps.

david-yz-liu: Your first paragraph is correct.

The test that was failing is because the created flexible criterion isn't associated with the proper assignment.

spec/models/result_spec.rb
```diff
@@ -199,13 +199,15 @@
                   context 'when marks are created for this incomplete result' do
                     let!(:incomp_result) { results[2] }
                     let!(:prev_subtotal) { incomp_result.get_subtotal }
+                    let!(:flex_criteria_first) { create(:flexible_criterion, assignment: assignment) }
+                    let!(:flex_criteria_second) { create(:flexible_criterion, max_mark: 2.0, assignment: assignment) }
                     before do
-                      create(:flexible_mark, result: incomp_result, mark: 1)
-                      create(:flexible_mark, result: incomp_result, mark: 1)
+                      create(:flexible_mark, result: incomp_result, mark: 1, markable: flex_criteria_first)
```
david-yz-liu: @WChrisK I'm wondering if you tried just doing `assignment: assignment` right here. It might not work because it's really a nested association, but I thought it would be worth a try.

WChrisK:
In retrospect, my commit named `View now works for the student and when released` should also apply to remark requests, since I tested it by making a remark request and performing it as a TA, etc...

app/views/results/student/_student_checkbox_criterion_li.html.erb
```diff
@@ -0,0 +1,15 @@
+<li id='<%= "checkbox_criterion_#{mark_criterion.id}" %>' class='checkbox_criterion'>
+  <%= render partial: 'results/common/checkbox_criterion',
+             locals: { mark_criterion: mark_criterion,
+                       mark: mark,
+                       old_mark: old_mark,
+                       view: 'student'} %>
+
+  <% unless mark_criterion.description.blank? %>
+      <div id='<%="criterion_info_#{mark_criterion.id}"%>'
```
david-yz-liu: Why doesn't this have the same "duplicate id" problem?

PR #2682, opened 2016-08-04
Unify criteria, deletes csv routes and uses upload_yml and download_yml

GitGeminiano:
Hello everyone,

I merged the marking scheme pages into one common page, and implemented the download_yml and upload_yml options for all criteria.

Regards,
Christian Garcia


test/functional/criteria_controller_test.rb
```diff
@@ -0,0 +1,133 @@
+# encoding: utf-8
```
GitGeminiano: I know we are migrating to rspec, but I do not know if migrating these tests is a priority. I will be happy to migrate them if needed.

config/routes.rb
```diff
@@ -75,22 +75,24 @@
...
-      resources :criteria, except: [:index] do
+      resources :criteria do
         collection do
           post 'update_positions'
+          post 'upload_yml'
+          get  'download_yml'
         end
       end
 
-      resources :rubric_criteria, only: [:index] do
```
GitGeminiano: I can get rid of this route and the one for flexible_criteria because they are not used at the moment. However, If I deleted them, I would have to delete the tests for the routes, which will have to be created back again if we decide to use csv later on. Should I delete them or comment them out?

david-yz-liu:
@GitGeminiano I merged this in, with a few changes. There are still a couple things to fix:

1. Should be able to upload visibility settings, but make them optional so that the default is applied.
2. For download: it is essential to put a "to_yml" type method in each of the criterion classes, much like you did for upload. There should be a single call to `assignment.get_criteria`, which gets all the criteria. This is the only way the criteria will have their positions preserved.
3. The download format should be the same as the upload format. In particular, it shouldn't generate a top-level `omap` string. To fix this, you might need to convert the `get_criteria` array to a hash (again, using the "to_yml" methods).

GitGeminiano:
@david-yz-liu Thank you for your comments. I will work on this on Monday. Regarding point number 4,  I think that the fact that we give a weight is one of the things that makes rubric somehow different from the other types. However, you have more experience using this feature, so if uploading a max mark is more convenient I will use max mark.

PR #2684, opened 2016-08-08
Upload download yml

GitGeminiano:
Hello everyone,

I made some improvements on the upload and download options. Among which are having the same format for uploading and downloading, including visibility options and using max_mark when uploading criteria of rubric type.  

Regards,
Christian

### DIFF
--- a/app/helpers/criteria_helper.rb
+++ b/app/helpers/criteria_helper.rb
@@ -18,11 +18,11 @@ module CriteriaHelper
       else
         begin
           type = criterion_yml[1]['type']
-          if type.nil? || type == 'Rubric'
+          if type.nil? || type.casecmp('rubric') == 0
             criterion = RubricCriterion.load_from_yml(criterion_yml)
-          elsif type == 'Flexible'
+          elsif type.casecmp('flexible') == 0
             criterion = FlexibleCriterion.load_from_yml(criterion_yml)
-          elsif type == 'Checkbox'
+          elsif type.casecmp('checkbox') == 0
             criterion = CheckboxCriterion.load_from_yml(criterion_yml)
           else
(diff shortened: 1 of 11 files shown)


app/helpers/criteria_helper.rb
```diff
@@ -18,11 +18,11 @@ def load_criteria(criteria, assignment)
       else
         begin
           type = criterion_yml[1]['type']
-          if type.nil? || type == 'rubric'
+          if type.nil? || type == 'Rubric'
```
david-yz-liu: Do a case-insensitive string comparison

app/models/checkbox_criterion.rb
```diff
@@ -177,6 +177,20 @@ def self.load_from_yml(criterion_yml)
...
       criterion_yml[1]['description'].nil? ? '' : criterion_yml[1]['description']
+    # Visibility options
+    criterion.ta_visible = criterion_yml[1]['ta_visible'] unless criterion_yml[1]['ta_visible'].nil?
+    criterion.peer_visible = criterion_yml[1]['peer_visible'] unless criterion_yml[1]['peer_visible'].nil?
     criterion
   end
+
+  # Returns a hash containing the information of a single checkbox criterion.
+  def self.to_yml(criterion)
+    criterion_info = {}
```
david-yz-liu: Make this a single expression, which is a hash literal:

```
{ "#{criterion.name}" => {
  type: 'checkbox',
  ...
}}
```

If symbol keys don't work, you can keep using strings.

GitGeminiano: I tried using symbol keys, but the format changed. I will make it a single expression.

PR #2686, opened 2016-08-08
assignment-marking-scheme-type: Deleting marking_scheme_type for assignments.

GitGeminiano:
Hello everyone,

I deleted the option to have assignments strictly associated with one type of criteria; since an assignment can have multiple criteria now.

Regards,
Christian


GitGeminiano:
The work on this pull request is not complete yet.

app/models/assignment.rb
```diff
@@ -697,17 +691,6 @@ def next_criterion_position
     get_criteria.count > 0 ? get_criteria.last.position + 1 : 1
   end
 
-  # Returns the class of the criteria that belong to this assignment.
-  def criterion_class
```
GitGeminiano: @david-yz-liu We do not need this class anymore. However, there is one use in automated_tests_helper in line 93 that I would like to delete. I was thinking that you may have a better understanding of why this line was added. I did see  that an index was created to have a proper association and faster queries, but I would like to hear your thoughts before I try to delete the above line tomorrow.

david-yz-liu: @GitGeminiano this indeed should be removed. The test script attributes should contain a `criterion_id` field that you can use to set the `criterion_type`.

GitGeminiano: Thank you. I was thinking about using criterion_id, but I could get another criterion with a different type. I am trying to pass both the "criterion_id" and the "criterion_type" through the form.

GitGeminiano:
There is one more thing to fix in this pull request.

PR #2688, opened 2016-08-09
Create mark seed

GitGeminiano:
Hello everyone,

I changed the create_mark function in marks.rake so that we set an appropriate mark and so that we do not pass the criterion type.

Regards,
Christian


lib/tasks/marks.rake
```diff
@@ -4,12 +4,11 @@ namespace :db do
...
 
-    #Function used to create marks for both criterias
-    def create_mark(result_id, markable_type, markable)
+    # Creates marks for all criteria
+    def create_mark(result_id, markable)
       Mark.create(
         result_id: result_id,
-        mark: rand(0..4),
-        markable_type: markable_type,
+        mark: rand(0..markable.max_mark).floor,
```
david-yz-liu: This isn't good for checkbox criteria.

GitGeminiano: I changed this. I have not uploaded my latest work because I thought there was a bug, but the bug seems to be in master. When I run db:reset in master and check assignment 3 or 4, I see a red table.

GitGeminiano: Update: I did a pull from master and the table is no longer red.

lib/tasks/marks.rake
```diff
@@ -21,9 +13,14 @@ namespace :db do
...
       #Automate marks for assignment using appropriate criteria
-      grouping.assignment.get_criteria.each do |criterion|
-        mark = create_mark(result.id, criterion)
-        result.marks.push(mark)
+      grouping.assignment.get_criteria(:all, :all, includes: :marks).each do |criterion|
+        mark = Mark.find_by(result_id:     result.id,
+                            markable_id:   criterion.id,
+                            markable_type: criterion.class.to_s)
+        mark.update({ result_id:     result.id,
+                      mark:          rand(0..criterion.max_mark.to_i),
```
david-yz-liu: As far as I can tell this still has the same problem I mentioned earlier. The different criterion types can have different allowed marks.

GitGeminiano: The marks got added successfully, when I tested this. Is there any particular case you would like me to check?

david-yz-liu: You don't understand what I mean. Rubric criteria can have exactly 4 possible mark values. Checkbox criteria can have only two possible mark values. Doing `rand` allows for many more values than are permitted.

GitGeminiano: I see what you mean. criterion.max_mark would work for checkbox and flexible because criterion.max_mark would actually be the maximum mark a marker can give. But in the rubric case criterion.max_mark, the maximum mark a marker can give is 4. I will change this.

david-yz-liu: Keep it random, but do different random generation depending on the criterion type.

lib/tasks/marks.rake
```diff
@@ -14,13 +14,17 @@ namespace :db do
...
-                            markable_id:   criterion.id,
-                            markable_type: criterion.class.to_s)
-        mark.update({ result_id:     result.id,
-                      mark:          rand(0..criterion.max_mark.to_i),
-                      markable_id:   criterion.id,
-                      markable_type: criterion.class.to_s })
+        if criterion.class == RubricCriterion
+          random_mark = criterion.max_mark / 4 * rand(0..4)
+        elsif criterion.class == FlexibleCriterion
+          random_mark = criterion.max_mark.to_i
```
david-yz-liu: This one should also be random.

PR #2692, opened 2016-08-11
Optimized peer review n+1 call away

WChrisK:


WChrisK:
**NOTE:** This is not complete yet since I still want to solve an n+1 query for submissions that are closely related (this way I get tests being checked w/ Travis)

PR #2694, opened 2016-08-11
Marks now update if not reverting to incomplete and changed

WChrisK:


WChrisK:
This occurred when a submission was viewed, and reverted from released, had the mark changed, and then re-releasing. The average would not change since it assumed that it was reverted back to 'incomplete' first.

PR #2699, opened 2016-08-15
Pr assign fix

WChrisK:


app/helpers/random_assign_helper.rb
```diff
@@ -27,7 +27,7 @@ def perform_random_assignment(pr_assignment, num_groups,
 
   def initialize_data_structures(num_groups, reviewer_ids, reviewee_ids)
     # A dictionary mapping reviewer id to a set of reviewee ids.
-    @reviewer_to_reviewee_sets = Hash.new { Set.new }
+    @reviewer_to_reviewee_sets = Hash.new { |h, k| h[k] = Set.new }
```
david-yz-liu: Mmmm. I see now what my mistake was: the same `Set.new` object was being used for all keys before. Interestingly, same issue as Python default arguments. Good catch.

PR #2701, opened 2016-08-15
Pr testing extended

WChrisK:


spec/controllers/peer_reviews_controller_spec.rb
```diff
@@ -23,6 +23,10 @@
     it 'random assigns properly' do
       expect(PeerReview.all.size).to eq 3
       expect(PeerReview.where(result: Result.all).size).to eq 3
+      PeerReview.all.each do |pr|
+        expect(pr.reviewer.id).not_to eq pr.reviewee.id
+        expect(pr.reviewer.does_not_share_any_students?(pr.reviewee)).to be_truthy
+      end
```
WChrisK: This is the only stuff to look at for this commit.

WChrisK:
In good news: Apparently all along seeding has been using the random assigner, but it needed these extra checks to make sure things properly assign (since before it didn't check for assignment to the same group, which was a big flaw as it was the reason other issues weren't detected). Now it covers that.

PR #2702, opened 2016-08-15
Fixed an n+1 query for the populate controller action (get groupings …

WChrisK:
…too)

### DIFF
--- a/app/helpers/random_assign_helper.rb
+++ b/app/helpers/random_assign_helper.rb
@@ -59,7 +59,8 @@ module RandomAssignHelper
   #
   # Also add existing peer reviews to @reviewer_to_reviewee_sets.
   def process_existing_peer_reviews(reviewer_ids)
-    PeerReview.includes(:reviewer)
+    PeerReview.includes(:reviewer,
+                        { result: { submission: :grouping } })
               .where(reviewer_id: reviewer_ids)
               .each do |peer_review|
       reviewer_id = peer_review.reviewer_id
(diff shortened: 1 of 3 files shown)


WChrisK:
This appears to solve the last n+1 query when done from zero peer reviews.

app/models/peer_review.rb
```diff
@@ -13,8 +13,7 @@ class PeerReview < ActiveRecord::Base
   validate :no_students_should_be_reviewer_and_reviewee
 
   def reviewee
-    # TODO - Research optimizing or see if rails can do better
-    Grouping.joins({ submissions: { results: :peer_reviews }}).where('peer_reviews.id = ?', self.id).first
+    result.submission.grouping
```
WChrisK: I changed this line as it allowed me to do a nested includes, and it also looks cleaner. Thoughts?
I remember making this on probably my third or fourth day before I knew how to properly use rails, so this might be some ugly stack-overflow hacky thing I did.

app/helpers/random_assign_helper.rb
```diff
@@ -56,7 +59,8 @@ def initialize_data_structures(num_groups, reviewer_ids, reviewee_ids)
   #
   # Also add existing peer reviews to @reviewer_to_reviewee_sets.
   def process_existing_peer_reviews(reviewer_ids)
-    PeerReview.includes(:reviewee)
+    PeerReview.includes(:reviewer,
+                        { result: { submission: :grouping } })
```
WChrisK: This is the only part of this commit that is not including the other changes (I forked off of the PR fix so I could test this stuff worked).

app/models/peer_review.rb
david-yz-liu: Yup, this is best.

PR #2691, opened 2016-08-10
Assignment Table Styling

christinem:
This PR restyles the assignment and dashboard table on both the student and admin/ta side so peer review assignments are now grouped together with their original assignments in the table.

PR #2705, opened 2016-08-16
Peer Review to Results Page fix

christinem:
Fixes an error when switching from peer review to results page, where it redirects to the wrong assignment ID.

### DIFF
--- a/app/views/layouts/_sub_menu.html.erb
+++ b/app/views/layouts/_sub_menu.html.erb
@@ -222,7 +222,9 @@
             (!@result.nil? && !@result.is_a_review?) %>'>
           <%# FIXME should add the real submission %>
           <%= link_to t('results.results_name'),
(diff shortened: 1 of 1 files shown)


christinem:
@david-yz-liu This failing test confuses me. I don't see how my code change could have caused this errors, and when running the rspec test myself, I do not get any errors.

david-yz-liu:
@christinem yeah don't worry about this test.

david-yz-liu:
@christinem I don't know how to test this until the peer review links work again. (Note: maybe that problem is actually a seeding problem?)

christinem:
@david-yz-liu since it seems that one failed test also appeared in #2691, I will check to see if it is a seeding issue. 

And even though you merged this in already, the problem was actually just if you were switching from the peer review tab (so the peer review 'landing page') to the results tab, not from a particular review.

PR #2710, opened 2016-08-19
Refactor test framework into client and server components

adisandro:
Refactors the test framework to run tests either locally or on a remote server for which we have previously set up ssh passwordless authentication.

The framework uses two resque queues, the first to process the test preparation and file transmission from the client side, and the second to execute the test on the server side, using an api callback to send back results. When run locally, a single resque worker can process both queues and no file transfer over the network is necessary. When run remotely, a copy of all Markus' code needs to be on the test server to be able to start the remote resque worker, but no web server is needed (i.e. a Markus instance does not need to run on the test server).


app/helpers/automated_tests_client_helper.rb
```diff
@@ -252,13 +251,11 @@ def self.request_a_test_run(host_with_port, grouping_id, call_on, current_user,
   def self.get_test_scripts_to_run(assignment, call_on)
 
     all_scripts = TestScript.where(assignment_id: assignment.id)
-    # If the test run is requested at collection (by Admin or TA),
+    # If the test run is requested after collection (by Admin or TA),
     # All of the test scripts should be run.
-    if call_on == 'collection'
+    if call_on == 'after_collection'
```
david-yz-liu: @adisandro maybe I wasn't clear about this.

It is true that we want to differentiate between scripts run for students vs. ones run for instructors. However, it should be possible to mark scripts which are run *only* by students, and not for instructors. This is why we need the `run_on_submission` attribute; it and `run_on_request` aren't strictly negations of each other.

adisandro: I see your point. It was not clear because it would be somewhat a new feature, up until now run_on_submission and run_on_request were used to identify "triggers" rather than "ownership". E.g. run_on_submission identified scripts running at every student submission, run_on_request scripts running explicitly when pressing a button (both from the student and admin side), and there was an implicit run_on_collection which would run all the scripts at collection time.

PR #2715, opened 2016-08-26
Refactor test framework into client and server components

adisandro:
Refactors the test framework to run tests either locally or on a remote server for which we have previously set up ssh passwordless authentication.

The framework uses two resque queues, the first to process the test preparation and file transmission from the client side, and the second to execute the test on the server side, using an api callback to send back results. When run locally, a single resque worker can process both queues and no file transfer over the network is necessary. When run remotely, a copy of all Markus' code needs to be on the test server to be able to start the remote resque worker, but no web server is needed (i.e. a Markus instance does not need to run on the test server).


david-yz-liu:
@adisandro two tests failed. Maybe this is my fault; it seems validations are not running as expected.

I'll look into the revision number thing.

PR #2734, opened 2016-09-24
Fix for issue 2729: Implemented a wrapper class to ensure the footer is near the bottom of the page.

b2simms:
Put div wrapper in:
- content.html.erb
- assignment_content.html.erb
Added styling in:
- _core.scss to ids: loggedIn and wrapper
- footer: placed top border for footer here


david-yz-liu:
@b2simms this appears to be the wrong commit.

Issue #2724, opened 2016-09-23
Rubric criterion replaces flexible criterion


MBAustin:
After seeding a new server, creating an assignment and adding first a flexible criterion, then a rubric criterion, the rubric criterion appears to replace the flexible criterion (though it does not affect the total mark for the assignment). This can happen more than once in the same assignment as more flexible and rubric criteria are added, but adding a checkbox criterion will cause it to stop happening. Leaving and coming back to the assignment causes the criteria to display normally.

david-yz-liu:
Fixed in #2735

Issue #2683, opened 2016-08-08
Class average rounding different to 'your mark' rounding


ghost:
After seeding and making a new server, if you mark one single student submission and release only that one, the averages are off by 0.01 sometimes:

`Your mark: 28.12% (Class Average: 28.13%)`

MBAustin:
I couldn't reproduce this (I set up an assignment whose criteria added up to 2500, gave a student 703/2500 which is 28.12%, and the system shows: Your mark: 28.12% (Class Average: 28.12%)), could you provide more details about how to cause the bug?

PR #2735, opened 2016-09-24
Fix for criterion overwrite

MBAustin:
Changed element IDs to include the criterion’s class, resolving the
problem where multiple criteria of different classes could have the
same element ID.


MBAustin:
The branch name says it's for issue #2683, but this is actually a fix for issue #2724

Issue #2728, opened 2016-09-23
No distinction between "Flexible" and "Checkbox" criteria on the Marking Scheme page


MBAustin:
In the Marking Scheme view, there's no way to tell whether a given criterion is Flexible or Checkbox, there should be a field for what kind of criterion it is.

PR #2737, opened 2016-09-24
Issue #2728

MBAustin:
Made it easier to distinguish the different types of criteria in the “Marking Scheme” view by adding icons and extra title information.


reidka:
Please add the attributes so that hovering brings up the alt tag.  (I think it needs a title attribute.)

PR #2740, opened 2016-09-24
Fix Issue-2558: fix required field check for filename

bzhou26:
The issue opened by David was:
Instructor reports: "I had a typo in one of 10 required files (“q10,sql” with a comma instead of a period). MarkUs flagged every one of the required files as invalid, with a red star."

When create a new assignment.
Old:
1. After a new row for adding required filename, there is no required field symbol.
2. The input of required filename can be blank (submit button is not disabled), even if the input is a required input.

New:
1. All required filename are required fields and have red star symbol to remind user to input or remove the row.
2. If any required field is empty, submit button is disabled.


david-yz-liu:
Good job, @bzhou26!

PR #2749, opened 2016-09-25
Fix for issue 2729: Formatting the footer to drop to the bottom of th…

b2simms:
…e page so it does not overlap the other elements being displayed. Added a wrapper class to content.html and assignment_content.html and some styling to core and navigation

### DIFF
--- a/app/assets/stylesheets/common/_core.scss
+++ b/app/assets/stylesheets/common/_core.scss
@@ -69,13 +69,21 @@ label.required::after {
 
 #content {
   background: $white;
-  border-bottom: 1px solid $blue;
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/common/_core.scss
```diff
@@ -78,6 +77,15 @@ label.required::after {
   }
 }
 
+#loggedIn {
+  height: 100%;
+}
+
+#wrapper {
+  background: $white;
```
david-yz-liu: This changes the background color of the nav bar, which should remain the colour it was before.

PR #2750, opened 2016-09-25
Fixing the N + 1 errors

b2simms:
Should not occur on:
Dashboard
Assignments
-annotations
-groups
-graders
Users


app/models/annotation_text.rb
```diff
@@ -1,12 +1,12 @@
 class AnnotationText < ActiveRecord::Base
 
+  belongs_to :annotation_category, :counter_cache => true
```
david-yz-liu: Use the new hash syntax `counter_cache: true`

david-yz-liu: Also, please move this back down to line 9.

app/models/assignment.rb
```diff
@@ -280,7 +280,7 @@ def median(marks)
   def self.get_current_assignment
     # start showing (or "featuring") the assignment 3 days before it's due
     # query uses Date.today + 4 because results from db seems to be off by 1
-    current_assignment = Assignment.where('due_date <= ?', Date.today + 4)
+    current_assignment = Assignment.where('due_date <= ?', Date.today + 4).include
```
david-yz-liu: There's a test failing because of this change I think, and this does look kind of weird. Is there supposed to be something after the `.include` that's missing?

app/views/assignments/student_interface.html.erb
```diff
@@ -210,7 +210,7 @@
 
           <%# For each member of this group, display name and status %>
           <ul>
-            <% @studentmemberships.each do |member|%>
+            <% @studentmemberships.includes([:user]).each do |member|%>
```
david-yz-liu: You shouldn't need to use a list here.

app/views/shared/_assignments_dropdown_menu.html.erb
```diff
@@ -13,7 +13,7 @@
 <% end %>
 
 
-<% @assignments = Assignment.order(short_identifier: :asc) %>
+<% @assignments = Assignment.includes([:pr_assignment]).order(short_identifier: :asc) %>
```
david-yz-liu: You shouldn't need a list here, either.

b2simms: @david-yz-liu I am looking into these issues now. Thanks for the feedback.

app/models/annotation_text.rb
```diff
@@ -6,7 +6,7 @@ class AnnotationText < ActiveRecord::Base
   # AnnotationText is destroyed.
   has_many :annotations, dependent: :destroy
 
-  belongs_to :annotation_category
+  belongs_to :annotation_category, :counter_cache::true
```
david-yz-liu: Er, so the correct syntax is actually `counter_cache: true`. What you have now is an error.

b2simms: Sorry David, it is breaking on mine now too. I will fix it right away.

b2simms: Just pushed the proper syntax.

b2simms:
Hello David, these changes should be implemented now. The problem was that I was out of sync with the git repository on my local version. I think I have that sorted out now.

david-yz-liu:
@b2simms well done, this is a nice improvement.

PR #2752, opened 2016-09-26
Set max mark in "new criterion" box

MBAustin:
Implemented #2744
I also removed the “criterion type” line because I feel like it’s
self-evident what the radio buttons are for
Changed the language files to make “max_mark” a property of “criteria” instead of each criterion type since all 3 types have it now


app/models/flexible_criterion.rb
```diff
@@ -35,7 +35,7 @@ class FlexibleCriterion < Criterion
 
   validate :visible?
 
-  DEFAULT_MAX_MARK = 1
+  DEFAULT_MAX_MARK = 2
```
david-yz-liu: Please change this back to 1.

MBAustin: Changed

app/views/criteria/_new_criterion.html.erb
```diff
@@ -7,9 +7,12 @@
 		              t('criteria.name') %><br>
     <%= text_field_tag(:new_criterion_prompt,
                        '',
-                       size: 15) %><br>
-    <%= label_tag 'new_criterion_type_prompt',
-                  t('criteria.type') %><br><br>
+                       size: 5) %><br>
+    <%= label_tag 'max_mark_prompt',
+                  t('criteria.max_mark')%><br>
+    <%= text_field_tag(:max_mark_prompt,
```
david-yz-liu: This should probably be a `number_field_tag`

david-yz-liu:
It seems like some of the tests need to be updated. See the Travis CI build for details.
You can run this file on your Vagrant machine with the command

```
$ rspec spec/controllers/path/to/test/file
```

(I think there's only one file that needs to be changed.

david-yz-liu:
Great, thanks @MBAustin.

PR #2759, opened 2016-09-28
PDF Viewer Fix

christinem:
This fixes a bug where students could not view pdf files and pdf annotations were missing.


david-yz-liu:
@christinem thanks for your work. I will take it from here.

PR #2762, opened 2016-09-29
Issue Fixed: Issue 2550 Add a column for bonus and deductions on the Summaries view

bzhou26:
Added a column to show bonus/deductions other than late penalty. 
If it is needed to add late penalty to this column or to show the late penalty in another column, please let me know.


app/views/summaries/_summaries_table.js.jsx.erb
```diff
@@ -176,23 +184,23 @@
...
+        s['total_extra_points'] = summary.total_extra_points;
         <% for criterion in @criteria %>
           var key = 'criterion_' + '<%= criterion.class.to_s %>' + '_' + '<%= criterion.id %>';
           s[key] = summary.criteria[key];
         <% end %>
         return s;
       }.bind(this));
-
-      return (
+        return (
```
david-yz-liu: There are some whitespace changes here and below that should be reverted.

app/models/grouping.rb
```diff
@@ -740,6 +740,16 @@ def final_grade(result)
   end
 
   # Helper for populate_submissions_table.
+  # Returns the total bonus/deductions for this grouping other than late penalty.
+  def total_extra_points(result)
+    if !result.nil? && result.marking_state == Result::MARKING_STATES[:complete]
```
david-yz-liu: Two comments here.

1. This seems like it belongs in the `Result` model, not `Grouping`.
2. Let's always show the points, even if the result is not marked as complete. This way, late penalties will show immediately (once they're added).

bzhou26: I get the extra points the same way as the final grade which makes the least change. The controller gets all data by model grouping.

bzhou26:
screen shot 2016-10-03 at 16 45 47

I updated this pull request. Right now, final grade and total extra points will be always shown if they are available. The sum includes all bonus, deductions and late penalties and if late penalty is applied, the percentage of the late penalty will show.

david-yz-liu:
@bzhou26 looks good. I made one more minor comment.

PR #2768, opened 2016-10-03
fix for issue-2049: widened annotation pop-up menu from 200px to 300px

maplecomdori:
changed the width of the annotation pop-up menu in markus.scss in `ul.tags > ul { ... width: 300px ...}`


david-yz-liu:
@maplecomdori this is the correct fix, but can you please also increase the length of the annotation text that's actually shown? Right now, they seem to be cut off well before the width of the dropdown.

PR #2777, opened 2016-10-07
Automarking output fixes

adisandro:
This pull request makes error/output handling better, and fixes some other problems:

- Try to catch all test errors and display them on the client side, either by flashing a message or creating a test result in the db with the appropriate description and mark = 0.
- Handle timeouts on the server robustly with threads.
- Fix mark calculation for rubric criteria to match nearest level.
- Set marking to complete if tests generate them all.
- Fixes to the python api.


lib/tools/markusapi.py
```diff
@@ -120,24 +120,25 @@ def upload_feedback_file(self, assignment_id, group_id, title, contents):
         path = self.get_path(assignment_id, group_id) + 'feedback_files'
         return self.submit_request(params, path, 'POST')
 
-    def upload_test_script_result(self, assignment_id, group_id, results):
-        """ (Markus, int, str, dict) -> list of str """
```
david-yz-liu: @adisandro I don't think you realize that the part after the `->` is the return type of the funtion, so this shouldn't be changed.

Also you should use the type `list` not `array`.

PR #2785, opened 2016-10-18
Ta pr bugfix

MBAustin:
Prevented the assignments table from trying to create links that call submissions_controller.browse() on pr_assignments, this should be removed once browse() is updated to handle them


.gitignore
```diff
@@ -53,3 +53,5 @@
 
 # Ignore redis dump file
 .dump.rdb
+markus_rsa
```
david-yz-liu: Please revert the changes to `.gitignore`.

MBAustin: Sorry, didn't realize I'd committed those. Fixed.

PR #2789, opened 2016-10-21
Issue 2779 one try: add one flash message after successfully 'add annotation'

bzhou26:
screen shot 2016-10-20 at 22 44 28

After click the save, it will show like this

screen shot 2016-10-20 at 22 44 36


david-yz-liu:
@bzhou26 yeah, this is on the right track. I believe there should be some harder issues in general, but you did it right here.

PR #2790, opened 2016-10-21
Automarking

adisandro:
Refactor the server component to be installable separately from MarkUs.


app/helpers/automated_tests_client_helper.rb
```diff
@@ -1,6 +1,4 @@
-require 'net/ssh'
-require 'net/scp'
-require 'json'
+require File.join(Rails.root, 'lib', 'automated_tests', 'server', 'automated_tests_server')
```
david-yz-liu: This should probably be guarded by a `if MarkusConfigurator.markus_ate_server_host == 'localhost'`; this would make it possible, in principle, to omit the `AutomatedTestsServer` file from production, I think.

app/helpers/automated_tests_client_helper.rb
```diff
@@ -260,9 +258,12 @@ def self.get_test_scripts(assignment, user)
 
   def self.request_a_test_run(host_with_port, grouping_id, current_user, submission_id = nil)
 
-    test_server_user = get_test_server_user
     grouping = Grouping.find(grouping_id)
     assignment = grouping.assignment
+    unless assignment.enable_test
```
david-yz-liu: Please do this check at the top of the method

adisandro: Isn't it? I need to get the assignment db object first.

app/helpers/automated_tests_client_helper.rb
adisandro: I struggled with this for a bit, let me try to recap the critical points:
1) We want a developer instance to work out of the box locally, without having to install the server separately.
2) A production instance with ate_server_host == 'localhost' is perfectly viable, it's just not our production instance.
3) Rails 5 will want to eagerly load all the code to detect potential concurrency problems, i.e. all the relevant code needs to be loaded at boot time.

What you are suggesting is basically a dynamic require, right? That won't be possible/supported with Rails5 as far as I understand (http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#autoloading-is-disabled-after-booting-in-the-production-environment), I can't figure out a way to follow Rails5's philosophy and drop the file in some production environments (note the 'some', it's not in all, which would be somewhat easier). This made me think that after all the server is still dependent on Markus for the communication and not an entirely separate project with a specified communication protocol, so I went for the compromise where the con is having the server ruby file loaded by clients even if they don't need it.

app/helpers/automated_tests_client_helper.rb
david-yz-liu: You're right. My point is that maybe this shouldn't be a database column, but rather a computed value based on the "no tests flag" and the presence of runnable test scripts.

PR #2795, opened 2016-10-25
Fix "no criteria" test scripts

MBAustin:
Added check on the fields of criterion_id to make JSON.parse doesn’t throw a fit about them being nil


app/helpers/automated_tests_client_helper.rb
```diff
@@ -92,8 +92,10 @@ def process_test_form(assignment, params, assignment_params,
       end
       # Always make sure the criterion type is correct.
       # The :criterion_id parameter contains a list of the form
-      # [criterion_id, criterion_type]
-      if testscripts[file_num][:criterion_id].nil?
+      # [criterion_id, criterion_type], which JSON.parse cannot parse unless
+      # neither element is nil
+      if testscripts[file_num][:criterion_id].nil? || testscripts[file_num][:criterion_id][0].nil?
+        || testscripts[file_num][:criterion_id][1].nil?
```
david-yz-liu: This line seems to be a syntax error. Wrap the condition in parentheses, and line break as follows:

```
if (... ||
    ... ||
    ...)
```

MBAustin: Sorry about that! RubyMine complains about wrapping a conditional in parentheses, is this a stylistic convention that can be ignored?

PR #2803, opened 2016-10-28
Fix Issue 2792: No update_tag action in tags_controller

bzhou26:
The issue is that there is no update action to update a particular tag. Now I add a such action for updating a tag.


david-yz-liu:
Thanks, @bzhou26!

PR #2806, opened 2016-10-30
Add flash messages for AJAX requests (create new group, delete group)

bzhou26:
Add an after_filter in applications controller to listen to all AJAX requests.
Then copy messages from flash[{key}] to the response header.
Use jquery to add messages to div.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -63,12 +63,12 @@ Element.prototype.hasClass = function(className) {
 }
 
 jQuery(document).ajaxComplete(function(event, request) {
-    var keys = ["notice", "warning", "success", "error"];
+    var keys = ['notice', 'warning', 'success', 'error'];
     var keysLength = keys.length;
-    for (var i = 0; i < keysLength; i++){
-        if (request.getResponseHeader('X-Message-'+keys[i])) {
-            jQuery(".flash-"+keys[i]).show();
-            jQuery(".flash-"+keys[i]).text(request.getResponseHeader('X-Message-'+keys[i]));
+    for (var i = 0; i < keysLength; i++) {
+        if (request.getResponseHeader('X-Message-' + keys[i])) {
+            jQuery('.flash-' + keys[i]).show();
+            jQuery('.flash-' + keys[i]).text(request.getResponseHeader('X-Message-' + keys[i]));
         }
     }
 });
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -61,3 +61,14 @@ Element.prototype.hasClass = function(className) {
   else
     return new RegExp('(^| )' + className + '( |$)', 'gi').test(this.className);
 }
+
+jQuery(document).ajaxComplete(function(event, request) {
+    var keys = ["notice", "warning", "success", "error"];
+    var keysLength = keys.length;
+    for (var i = 0; i < keysLength; i++){
+        if (request.getResponseHeader('X-Message-'+keys[i])) {
+            jQuery(".flash-"+keys[i]).show();
+            jQuery(".flash-"+keys[i]).text(request.getResponseHeader('X-Message-'+keys[i]));
+        }
+    }
+});
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -63,12 +63,28 @@ Element.prototype.hasClass = function(className) {
 }
 
 jQuery(document).ajaxComplete(function(event, request) {
-    var keys = ['notice', 'warning', 'success', 'error'];
-    var keysLength = keys.length;
-    for (var i = 0; i < keysLength; i++) {
-        if (request.getResponseHeader('X-Message-' + keys[i])) {
-            jQuery('.flash-' + keys[i]).show();
-            jQuery('.flash-' + keys[i]).text(request.getResponseHeader('X-Message-' + keys[i]));
-        }
+  var keys = ['notice', 'warning', 'success', 'error'];
+  var keysLength = keys.length;
+  var flashMessageList = [];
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/application.js
```diff
@@ -61,3 +61,14 @@ Element.prototype.hasClass = function(className) {
   else
     return new RegExp('(^| )' + className + '( |$)', 'gi').test(this.className);
 }
+
+jQuery(document).ajaxComplete(function(event, request) {
+    var keys = ["notice", "warning", "success", "error"];
```
david-yz-liu: Use single-quotes for strings

app/assets/javascripts/application.js
```diff
@@ -61,3 +61,14 @@ Element.prototype.hasClass = function(className) {
...
+    var keysLength = keys.length;
+    for (var i = 0; i < keysLength; i++){
+        if (request.getResponseHeader('X-Message-'+keys[i])) {
```
david-yz-liu: Put spaces around operators, and before `{` (in the previous line)

app/assets/javascripts/application.js
```diff
@@ -61,3 +61,14 @@ Element.prototype.hasClass = function(className) {
...
+            jQuery(".flash-"+keys[i]).show();
```
david-yz-liu: Do `show` after calling `text`

app/controllers/groups_controller.rb
```diff
@@ -30,6 +30,7 @@ def new
     assignment = Assignment.find(params[:assignment_id])
     begin
       assignment.add_group(params[:new_group_name])
+      flash.now[:success] = I18n.t('groups.rename_group.success')
```
david-yz-liu: Please use the helpers `flash_now` and `flash_message` instead of accessing the flash hashes directly.

app/views/shared/_flash_message.html.erb
```diff
@@ -1,13 +1,17 @@
 <% [:error, :success, :warning, :notice].each do |key| -%>
-  <% unless flash[key].nil? -%>
-    <% if flash[key].is_a?(Array) -%>
-      <% flash[key].each do |msg| -%>
-        <%= content_tag :div, msg, class: key %>
-        <% response.header['X-message ' + key.to_s] = msg %>
+  <% if flash[key].is_a?(Array) -%>
+    <% flash[key].each do |msg| -%>
+      <% unless msg -%>
+        <%= content_tag :div, msg, class: "#{key} flash-#{key}", style: "display: none" %>
```
david-yz-liu: I doubt that a new class is necessary.

david-yz-liu: Also, I would much prefer having one div per flash type, and if there are multiple messages, they appear as an unordered list within this one div. If you could fix this while you're at it, that would be much appreciated.

app/views/shared/_flash_message.html.erb
```diff
@@ -1,13 +1,17 @@
 <% [:error, :success, :warning, :notice].each do |key| -%>
-  <% unless flash[key].nil? -%>
-    <% if flash[key].is_a?(Array) -%>
-      <% flash[key].each do |msg| -%>
-        <%= content_tag :div, msg, class: key %>
-        <% response.header['X-message ' + key.to_s] = msg %>
+  <% if flash[key].is_a?(Array) -%>
```
david-yz-liu: Please delete the `-` (in `-%>`) from every Ruby tag in this file.

bzhou26:
This should be working well for all cases flash message if it uses `flash_message()` or `flash_now()` and all flash messages in "annotations" should work.

app/views/shared/_flash_message.html.erb
```diff
@@ -1,13 +1,17 @@
...
-    <% end -%>
-  <% end -%>
-<% end -%>
+<% [:error, :success, :warning, :notice].each do |key| %>
+  <% if flash[key].is_a?(Array) %>
+    <% flash[key].each do |msg| %>
+      <% unless msg -%>
+        <%= content_tag :div, msg, class: "#{key}", style: "display: none" %>
+      <% else %>
+        <%= content_tag :div, msg, class: "#{key}" %>
```
david-yz-liu: This part needs to be updated so that only one `div` gets created per key.

david-yz-liu:
@bzhou26 I asked you to not make any more changes to different routes on this branch. Please only work on the comments I've made here about the overall structure, and work on other routes on a separate branch. I would appreciate it if you reverted the changes you made to the other routes on this branch as well.

bzhou26:
Sorry for messing up the branch. I finished the `_flash_message.html` and `application.js` as a global function for showing AJAX flash messages. The changes in groups_controller was made before. I will change other things in another branch from now.

bzhou26:
@david-yz-liu I fixed a bug in my code that I forgot to put nil as content in `_flash_message.html`. You can test it now.

david-yz-liu:
Thanks, @bzhou26!

app/assets/javascripts/application.js
```diff
@@ -61,3 +61,30 @@ Element.prototype.hasClass = function(className) {
...
+      jQuery('.' + keys[i]).empty();
+      for (var j = 0; j < messages.length; j++) {
+        jQuery('.' + keys[i]).append('<p>' + messages[j] + '</p>');
+      }
+      jQuery('.' + keys[i]).show();
+    } else if (receive) {
+      jQuery('.' + keys[i]).empty();
+    }
+  }
+  jQuery(function() {
```
david-yz-liu: @bzhou26 I just realized that this can mess with other parts of the application. In general, do not make this type of general jQuery call, it's way too broad. Please remove this.

PR #2819, opened 2016-11-21
Automarking

adisandro:
This pull request reorganizes how testers are installed in a Markus autotesting server, adds a SQL tester, and paves the way to separate the Markus autotesting server into its own repository.
Other fixes:
- Handle null character in xml.
- The upload feedback file api now overwrites a file with the same name.


lib/automated_tests/server/python_apis/markusapi.py
```diff
@@ -127,7 +138,11 @@ def upload_feedback_file(self, assignment_id, group_id, title, contents):
             'mime_type': mimetypes.guess_type(title)[0]
         }
         path = self.get_path(assignment_id, group_id) + 'feedback_files'
-        return self.submit_request(params, path, 'POST')
+        request_type = 'POST'
+        if feedback_file_id:
```
david-yz-liu: @adisandro I didn't like this part very much. I did some more reading about the appropriate HTTP verbs (<https://jcalcote.wordpress.com/2008/10/16/put-or-post-the-rest-of-the-story/>). upsert (update or insert) seems like it might not map nicely to Rails' configuration, but we should try.

It would be good to

1) Always use a `PUT` here
2) Change the FeedbackFilesController in the API to do an update or create in the `update` method.

PR #2821, opened 2016-11-22
Standardize flash messages for all controllers

bzhou26:
Standardize all controllers to use `flash_message()` or `flash_now()`


app/views/groups/index.html.erb
```diff
@@ -49,33 +49,7 @@
...
-  </div>
-
-  <div class="warning" id="global_action_warning" style="display: none; ">
-    <h3><%= t("groups.warning") %></h3>
-    <div id="global_action_warning_msg"></div>
-  </div>
-  <div id="job-messenger"></div>
-
-</div>
+<%= render 'shared/flash_message' %>
```
david-yz-liu: Can you double-check that this partial isn't already included in the template? Seems odd.

app/views/groups/index.html.erb
```diff
@@ -49,33 +49,7 @@
     <%= link_to t('upload'), 'javascript:void(0);', onclick:'modal_upload.open();' %>
 <% end %>
 
-<div id="notice">
-  <div class="success" id="add_success" style="display: none;"></div>
-  <% # This is for error display injected by JavaScript %>
-
-  <div class="error" id="errors" style="display: none; ">
```
david-yz-liu: ids here suggest there was some related Javascript. Can you please do a search, and if it can be removed (likely), remove it.

david-yz-liu:
@bzhou26 this is great, thank you. Please take a look at the details of the failed tests on the Travis CI build - it's some minor formatting stuff in the tests themselves that need to be updated.

app/views/groups/index.html.erb
bzhou26: I go through all the JS file in the project but I can't find any code related to this div tag. I think the CSS is combined with classes which is used for new flash messages as well so I didn't delete it.

app/views/groups/index.html.erb
bzhou26: This is true, this render make the page has two duplicated flash message div. I just delete this one. But the one already included in the template has layout issue which will show the flash message on the bottom of the page

bzhou26:
Updated all functional test code for flash messages.

david-yz-liu:
@bzhou26 this is good. Please remove the `error` and `success` state from all of the React components you changed. (These are no longer used.)

PR #2826, opened 2016-11-29
Add Non-Regenerating Tokens Option

MBAustin:


david-yz-liu:
Hi @MBAustin - you should not edit the schema.rb file directly. Instead:

1. Create a migration <http://edgeguides.rubyonrails.org/active_record_migrations.html>
2. Run the migration on the Vagrant machine
3. Download the updated schema.rb from the virtual machine
4. Commit both the migration and the schema.rb, and push.

david-yz-liu:
@MBAustin This is the one that I was referring to in our meeting.

david-yz-liu:
Thanks, @MBAustin!

PR #2838, opened 2016-12-04
Fix student view 'Run tests'

MBAustin:
Local variable was being referenced without ‘@‘ which caused resolution failure


david-yz-liu:
@MBAustin a partial's local variables are different from instance variables (the latter use the `@`). I believe the real problem is that the local variable `show_output` is not passed when this partial is rendered in the `automated_tests/student_interface` view. (Should be explicitly passed in as false.)

MBAustin:
@david-yz-liu Ok, I've changed it to do that instead

david-yz-liu:
Thanks, @MBAustin.

PR #2839, opened 2016-12-08
Fix duplicate file upload

MBAustin:
Resolved server error page when attempting to upload a test script file with the same name as an existing one.


david-yz-liu:
@MBAustin while this is a good thought, I'd prefer if we actually fixed this properly so that uploading a file with the same name *replaces* the existing file (this is what everyone wants). This is a bit more challenging, but do you think you could tackle that?

PR #2845, opened 2016-12-13
Turn off jQuery noConflict mode

arkon:
This ought to be done now that Prototype has been removed, since (almost) everyone expects jQuery to be used with its default `$` variable. This lull between semesters is probably the best time to get this out of the way.


david-yz-liu:
Thanks, @arkon!

PR #2847, opened 2016-12-15
Matt test result styling

MBAustin:
Test results from the most recent run are shown by default, while the others are hidden. Also fixed script trying to operate on page element that did not exist in the current view.


MBAustin:
This contains a couple of what you may or may not to consider to be hacks, so feedback on better ways of implementing things is welcomed.

david-yz-liu:
Thanks, @MBAustin! I didn't know how much time you still had to work on this, so I worked a bit myself on making the styling consistent for both instructors and students.

Regarding the table styling, the `!important` is undesirable, but really I want to eventually remove the alternating table row colours anyway, so I wouldn't worry too much about it now. Great work!

PR #2865, opened 2016-12-16
Create git permission file

adisandro:
First step towards removing gitolite

### DIFF
--- a/lib/repo/repository.rb
+++ b/lib/repo/repository.rb
@@ -204,6 +204,47 @@ module Repository
       raise NotImplementedError, "Repository.delete_bulk_permissions: Not yet implemented"
     end
 
+    # Builds a hash of all repositories and users allowed to rw to them
+    def self.get_all_permissions
+
+      permissions = {}
+      admins = Admin.all
+      admins = admins.map(&:user_name)
+      tas = Ta.all
+      tas = tas.map(&:user_name)
+      non_student_repos = Group.all
+      non_student_repos = non_student_repos.map(&:repository_name)
+      # Permission subtleties:
+      # 1) a repository is associated with a Group, but..
+      # 2) ..students are associated with a Grouping (an "instance" of Group for a specific Assignment)
+      # That creates a problem since authentication in svn/git is at the repository level, while Markus handles it at
+      # the assignment level, allowing the same Group repo to have different students according to the assignment.
+      # The two extremes to implement it are union (permissive) or intersection (restrictive), but we are going to take
+      # a last-deadline approach instead, where we assume that the valid students at any point in time are the ones
+      # valid for the last assignment due.
+      # (Basically, it's nice for a group to share a repo among assignments, but at a certain point during the course
+      # we may want to add or [more frequently] remove some students from it)
+      assignments = Assignment.order(due_date: :desc)
+      assignments.each do |assignment|
+        valid_groupings = assignment.valid_groupings
+        valid_groupings.each do |valid_grouping|
+          repo_name = valid_grouping.group.repo_name
+          if permissions.key?(repo_name)
+            next
+          end
+          accepted_students = valid_grouping.accepted_students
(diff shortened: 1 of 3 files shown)


david-yz-liu:
@adisandro we want this to be a separate method in the Repository superclass, and we also want this to write to stdout. I would say we shouldn't touch the GitRepository's method here.

adisandro:
I was asking @reidka about this, come talk to me because I don't understand the stdout requirement.

adisandro:
@david-yz-liu: Alan is going to run a script that reads from the auth file at every http request to the git repo path, which is exactly what happens with the svn case.
I'll refactor the code to be in Repository, but then I'll keep the file creation in GitRepository to be invoked appropriately by markus. Is __set_all_permission the only function involved? (i.e. is the file always recalculated in its entirety?)

david-yz-liu:
@adisandro okay, that sounds fine. And yes, the file is always recomputed (at least for now).

adisandro:
@david-yz-liu @reidka Unfortunately this won't be enough, because it turns out that we DO try to modify individual permissions in the auth file and NOT rewriting it every time (see e.g. usages of the overcomplicated remove_user function)

lib/repo/repository.rb
```diff
@@ -204,6 +204,47 @@ def self.delete_bulk_permissions(groups, user_ids)
       raise NotImplementedError, "Repository.delete_bulk_permissions: Not yet implemented"
     end
 
+    # Builds a hash of all repositories and users allowed to rw to them
+    def self.get_all_permissions
+
+      permissions = {}
+      admins = Admin.all
```
david-yz-liu: Use `.pluck(:user_name)` here and below for single-attribute access.

lib/repo/repository.rb
```diff
@@ -204,6 +204,47 @@ def self.delete_bulk_permissions(groups, user_ids)
...
+      # 1) a repository is associated with a Group, but..
+      # 2) ..students are associated with a Grouping (an "instance" of Group for a specific Assignment)
+      # That creates a problem since authentication in svn/git is at the repository level, while Markus handles it at
+      # the assignment level, allowing the same Group repo to have different students according to the assignment.
+      # The two extremes to implement it are union (permissive) or intersection (restrictive), but we are going to take
+      # a last-deadline approach instead, where we assume that the valid students at any point in time are the ones
+      # valid for the last assignment due.
+      # (Basically, it's nice for a group to share a repo among assignments, but at a certain point during the course
+      # we may want to add or [more frequently] remove some students from it)
+      assignments = Assignment.order(due_date: :desc)
```
david-yz-liu: We can get this to go a lot faster if we use includes properly. Look up http://guides.rubyonrails.org/active_record_querying.html#eager-loading-associations.

lib/repo/repository.rb
```diff
@@ -204,6 +204,47 @@ def self.delete_bulk_permissions(groups, user_ids)
...
+          accepted_students = accepted_students.map(&:user_name)
+          permissions[repo_name] = admins + tas + accepted_students
+          non_student_repos.delete(repo_name)
+        end
+      end
+      non_student_repos.each do |repo_name|
+        permissions[repo_name] = admins + tas
+      end
+
+      return permissions
```
david-yz-liu: Please switch to not using return for the last expression.

lib/repo/repository.rb
adisandro: Ok I get it.
If I go as deep as possible, does it interact well with the various subqueries, e.g. assignment.valid_groupings and valid_grouping.accepted_students?

lib/repo/repository.rb
adisandro: Didn't do much sql in rails yet, sorry.

lib/repo/repository.rb
adisandro: I'll write this down for the next time we have a nasty authentication bug due to the implicit return :)

PR #2868, opened 2016-12-17
Fix pane dragging issues

david-yz-liu:
The initial problem was introduced by @christinem in adding an `!important` term to css rules. The commit message said this was to fix a window resizing problem, although I can't remember exactly what that was.

Possibly related: I also fixed the problem of the divider not appearing in the correct location when the window is resized. Tested this out on Chrome and Microsoft Edge (yup). @arkon you may be interested in this as well.

### DIFF
--- a/app/assets/javascripts/panes.js
+++ b/app/assets/javascripts/panes.js
@@ -75,8 +75,11 @@ jQuery(document).ready(function() {
 });
 
 /* Handle window resizing */
-window.onresize = function(event) {
+// NOTE: Don't manually override window.onresize. This will conflict with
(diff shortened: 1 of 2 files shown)


arkon:
The logic seems sound.

I really ought to come up with a more elegant version of this though, that also scales to height better...

christinem:
If I recall correctly, I believe the issue was when you made the window smaller, the rubric pane would jump down to be directly under the code viewer instead of remaining to the right of the codeviewer like usual, and it was really disorienting. So I had to force it to always be a certain percentage of the window so it wouldn't automatically jump down when there wasn't enough window space anymore. My mistake, I didn't realize it broke the dragging.

PR #2869, opened 2016-12-17
Add .editorconfig

arkon:
Adding an [EditorConfig file](http://editorconfig.org/) so that some basic editor settings will be consistent across developers' environments.


david-yz-liu:
Thanks, @arkon!

PR #2873, opened 2016-12-21
Added autocapitalize: 'off' to username field on login page

jwebbed:
Something I noticed particularly with the switch to utorid as the login, is that the username is case sensitive. On phones, the first letter of your username will get capitalized which results in failing to log in until you manually fix it. This is just a minor ux improvement.


david-yz-liu:
Thanks, @jwebbed!

PR #2875, opened 2017-01-07
Fix some missed jQuery references

arkon:
A followup to #2845, since I think some files were modified after that PR.


david-yz-liu:
You're right, thanks @arkon!

PR #2877, opened 2017-01-19
Fix feedback file upload api

adisandro:
When a textual feedback file is uploaded without a text/plain mime type (e.g. a json file), it is mistakenly treated as binary.


david-yz-liu:
@adisandro Actually if it's all the same to you, since we don't need this for this term, I'd rather you rebase this onto master and make a PR there (so I don't need to bother Alan).

PR #2885, opened 2017-02-02
Fix file manager path anchors displaying as plaintext

arkon:
The "Current Path" in the file manager for students was being displayed as plain text instead of actual HTML.

**Before:**
screen shot 2017-02-02 at 16 48 40

**After:**
screen shot 2017-02-02 at 16 48 12


david-yz-liu:
Thanks, @arkon!

PR #2886, opened 2017-02-09
Make student test tokens work

adisandro:
This pull request makes the rate limiting mechanism for student-initiated tests (aka tokens) reliable.
It also introduces a way to detect enqueued tests being executed, so that only one test per grouping exists in the queue, regardless of tokens availability.
(This works nicely with #2883, although some feedback on who requested a test is missing and can be added later)


app/models/token.rb
```diff
@@ -18,32 +18,56 @@ def last_used_date
...
+    if Time.zone.now < assignment.token_start_date || !grouping.is_valid?
       self.remaining = 0
     elsif assignment.unlimited_tokens
-      # grouping has  1 token that is never consumed
+      # grouping has 1 token that is never consumed
       self.remaining = 1
-    elsif last_used.nil? ||
-          (last_used.to_time.to_i + assignment.token_period * 60 * 60 <=
-            DateTime.now.to_time.to_i)
+    elsif self.last_used.nil? || (self.last_used + assignment.token_period.hours) < Time.zone.now
```
david-yz-liu: The semantics of this isn't great. Tokens will only regenerate when there is a large enough gap between the *last* token used and the current time; if multiple tokens are used in sequence, the first one can't be recovered before the last one.

I don't really care right now, but it's worth mentioning that the simplest way to use tokens unambiguously is to only grant students a single token.

david-yz-liu:
@adisandro given the `requested_by` is a new field, how does the code correctly display only student-requested tests to students, and only instructor-requested tests in the grading panel (for all three of instructors, TAs, and students)?

app/models/token.rb
adisandro: Yes, I kept the existing semantics, which basically is "X tests every token_period from your last test", I agree it's confusing.
I still like the idea of assigning multiple tokens at once, especially from a group perspective, or simply because you want to allow students to run X tests every day.
Let me think if there's a way to make it work in a nicer way..

adisandro:
@david-yz-liu I think I completely ignored that, damn it. For some reason it works and I don't really know how it manages to do that, which means that it is doing it subtly wrong and needs to be fixed, or worse that requested_by is useless.
I'll investigate.

adisandro:
Turns out the database queries were subtly wrong, they were based on the assumption that a student always requests a test with submission_id = nil, but in theory that is possible for an instructor too (and in the future we'll hopefully allow such a thing without collecting first).

app/models/token.rb
adisandro: I don't think I have time to look at an alternative, would you be ok with this for now?

david-yz-liu: Yes, this is fine.

app/helpers/automated_tests_client_helper.rb
```diff
@@ -415,38 +407,33 @@ def self.perform(host_with_port, test_scripts, user_api_key, server_api_key, gro
...
     result = nil
     begin
       result = Hash.from_xml(raw_result)
     rescue => e
-      create_all_test_scripts_error_result(test_scripts_ran, assignment, grouping, submission,
+      create_all_test_scripts_error_result(test_scripts_ran, assignment, grouping, submission, requested_by,
                                            I18n.t('automated_tests.test_result.all_tests'),
                                            I18n.t('automated_tests.test_result.bad_results', {xml: e.message}))
-      unless submission.nil?
-        submission.set_marks_for_tests
```
david-yz-liu: @adisandro what was the reason you removed this?

adisandro: In case of errors it's directly in the function create_all_test_scripts_error_result, so it was a duplicate. It's still at the end of process_test_result if no errors happen.
(the diff makes it hard to see it)

PR #2887, opened 2017-02-10
Feature: annotation context menu - "Common Annotations", "Copy", and "Download File" menu items added

jasonmai:
## Ready for Review

This pull request adds additional menu items to the annotation context menu (which currently, as of this PR only has "New Annotation" and "Delete Annotation").

 - Added "Common Annotations" menu item: all of the "pre-canned" annotations can now be accessed through the context menu
 - Added "Copy" menu item: behaves in the same manner as the original "Copy" item in the default browser context menu
 - Added "Download File" and "Download Annotated File" menu items: when right-clicking and going to either of these 2 download menu items, the file currently being viewed is downloaded. Behaves the same as clicking the "Download" file link at the top right - and then selecting the file you're currently viewing.
 - Moved the bulk of the context menu javascript source code to `context_menu.js` in dir `app/assets/javascript/Results/`

### Relevant remaining TODOs for future PR:
 - Add "Edit" annotation menu item
 - Add internationalization/localization support for the annotation context menu (currently all texts are ENG hard-coded)
 - Add support for using "Delete" or "Edit" when clicking on layered annotations (currently just accessing the most recently added annotation)
 - Move on to adding hotkey support

#### Common Annotations (pre-canned annotations) automagically set in the context menu:
common_annotations

#### Copy is back!:
copy

#### Download current files with ease:
download

### DIFF
--- a/app/assets/javascripts/Results/context_menu.js
+++ b/app/assets/javascripts/Results/context_menu.js
@@ -48,19 +48,16 @@ var annotation_context_menu = {
 
           if (annot_id !== null && annot_id.length !== 0 &&
               sub_file_id !== null && sub_file_id.length !== 0) {
-            var query_str_params = [
-              ['id=' + annot_id],
-              ['submission_file_id=' + sub_file_id],
-              ['result_id=' + result_id],
-              ['assignment_id=' + assignment_id]
-            ];
             $.ajax({
-              url: annot_path + '?' + query_str_params.join('&'),
-              method: 'POST',
-              data: { _method: 'delete' },
+              url: annot_path,
+              method: 'DELETE',
+              data: { _method: 'delete',
+                      id: annot_id,
+                      submission_file_id: sub_file_id,
+                      result_id: result_id,
+                      assignment_id: assignment_id },
               dataType: 'script'
             });
-
           }
           return;
         },
@@ -92,7 +89,7 @@ var annotation_context_menu = {
     function download_func(include_annot) {
       var sub_file_id = $('#select_file_id').val();
       if (sub_file_id !== null && sub_file_id.length !== 0) {
-        window.open(file_dl_path + '?utf8=?&include_annotations=' +
+        window.open(file_dl_path + '?utf8=✓&include_annotations=' +
                     include_annot + '&select_file_id=' + sub_file_id);
       }
     }
--- a/app/assets/javascripts/Results/context_menu.js
+++ b/app/assets/javascripts/Results/context_menu.js
@@ -1,0 +1,172 @@
+/*
+  This file contains the Javascript code necessary for the functionality
+  of the annotation context menu. For the annotation context menu to
+  function properly, these dependencies are necessary:
+
+  Context-Menu and library related:
+    * jquery.ui-contextmenu.min.js (..\app\assets\javascripts\)
+      * Depends on jQuery and jQuery-ui
+    * context_menu.css (..\app\assets\stylesheets\)
+
+  Context-Menu button functionality:
+    * _annotations.js.erb (..\app\views\results\common\)
+
+   */
+var annotation_context_menu = {
+  setup: function(annot_path, result_id, assignment_id, file_dl_path) {
+    var menu_items = {
+      new_annotation: {
+        title: 'New Annotation',
+        cmd: 'new_annotation',
+        action: make_new_annotation,
+        disabled: true
+      },
+      common_annotations: {
+        title: 'Common Annotations <kbd>></kbd>',
+        cmd: 'common_annotations',
+        disabled: true
+      },
+      delete_annotation: {
+        title: 'Delete Annotation',
+        cmd: "delete_annotation",
+        action: function(event, ui) {
+          var clicked_element = $(ui.target);
+          var annot_id = '';
+          var sub_file_id = $('#select_file_id').val();
+          if (annotation_type === ANNOTATION_TYPES.CODE) {
+            $.each(clicked_element[0].attributes, function(index, attr) {
+              if (attr.nodeName.toLowerCase()
+                      .indexOf('data-annotationid') != -1) {
+                annot_id = attr.value;
+                // Continue iteration in case of multiple annotations
+              }
+            });
+          } else {
+            annot_id = clicked_element.attr('id')
+                                      .replace('annotation_holder_', '');
+          }
+
+          if (annot_id !== null && annot_id.length !== 0 &&
+              sub_file_id !== null && sub_file_id.length !== 0) {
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/Results/context_menu.js
```diff
@@ -0,0 +1,172 @@
...
+            var query_str_params = [
+              ['id=' + annot_id],
+              ['submission_file_id=' + sub_file_id],
+              ['result_id=' + result_id],
+              ['assignment_id=' + assignment_id]
+            ];
+            $.ajax({
+              url: annot_path + '?' + query_str_params.join('&'),
+              method: 'POST',
+              data: { _method: 'delete' },
```
david-yz-liu: This would be a good time to move the `query_str_params` into `data` here.

app/assets/javascripts/Results/context_menu.js
```diff
@@ -0,0 +1,172 @@
...
+              sub_file_id !== null && sub_file_id.length !== 0) {
+            var query_str_params = [
+              ['id=' + annot_id],
+              ['submission_file_id=' + sub_file_id],
+              ['result_id=' + result_id],
+              ['assignment_id=' + assignment_id]
+            ];
+            $.ajax({
+              url: annot_path + '?' + query_str_params.join('&'),
+              method: 'POST',
```
david-yz-liu: You should be able to make this `DELETE`, I think.

app/assets/javascripts/Results/context_menu.js
```diff
@@ -0,0 +1,172 @@
...
+        cmd: 'download_annotated',
+        action: function() { download_func('true'); },
+        disabled: false
+      }
+    };
+
+    function download_func(include_annot) {
+      var sub_file_id = $('#select_file_id').val();
+      if (sub_file_id !== null && sub_file_id.length !== 0) {
+        window.open(file_dl_path + '?utf8=?&include_annotations=' +
```
david-yz-liu: See my comment below about using the rails path helpers in this file. (You should be able to pass the query parameters as options to the helper as well.)

app/views/results/common/_context_menu.js.erb
```diff
@@ -1,113 +1,46 @@
...
+    <% annotation_categories.each do |annotation_category| %>
+      {
+        title: '<%= truncate(annotation_category.annotation_category_name, length: 40) %> <kbd>></kbd>',
+        cmd: 'annotation_category_<%= annotation_category.id %>',
+        action: function(event, ui){ return false; },
+        children: [
+        <% if annotation_category.annotation_texts.empty? %>
+          {
+            title: '<%= t('marker.annotation.no_annotation_in_this_category') %>',
+            action: function(event, ui){ return false; }
```
jasonmai: The default behaviour of the context menu, when clicking on the parent of the submenu, is to close the entire context menu. This action, which has a function that returns false, causes the behaviour to instead open the first item of the submenu of the clicked parent.

app/assets/javascripts/Results/context_menu.js
jasonmai: Hmm this is strange. When I first implemented this I initially passing the necessary parameters through the post body with the `data` property, but it did not work properly. This was the reason why I was passing them as query string parameters instead (since this was how it was through the delete button). 

However, it seems to be working now when I use method `DELETE`, and passing the arguments through `data`. Thanks for prompting me to check this.

app/assets/javascripts/Results/context_menu.js
jasonmai: (see comment [above](https://github.com/MarkUsProject/Markus/pull/2887/files/a238396fd350d9993935cc344199891fe8838d05#diff-8d92e671f9c2e0bbb28130742dad5407))

app/views/results/common/_context_menu.js.erb
```diff
@@ -1,113 +1,46 @@
...
-              url: '<%= annotations_path %>?' + query_str_params.join('&'),
-              method: 'POST',
-              data: { _method: 'delete' },
-              dataType: 'script'
-            });
 
+    var annot_path = '<%= annotations_path %>';
+    var result_id = '<%= result.id %>';
+    var assignment_id = '<%= result.submission.grouping.assignment.id %>';
+    var file_dl_path = '<%= download_assignment_submission_result_path(
```
jasonmai: I was under the impression that the `javascript_include_tag` and the asset pipeline are for static assets. If I am able to append `.erb` to the asset `context_menu.js` file and move the ruby call to `download_assignment_submission_result_path` there, I should also be able to move all the remaining code in `_context_menu.js.erb` there. This does not feel right to me. `download_assignment_submission_result_path` will produce a different path for each student/group/assignment, thus causing the combined/compressed js resulting pipeline file to constantly change. Sorry, I might be misunderstanding something but I feel this is correct.

david-yz-liu: Thanks for pointing this out, you're correct that we can't use the path helper here. (We *can* append `.erb` to the extension for arbitrary static Ruby code though, like internationalization.)

I was distracted by the fact that you're already passing in ids here, which is logically redundant. It reminds me that one of my wishlist is using [js-routes](https://github.com/railsware/js-routes).

Anyway, you don't have to change anything here.

app/views/results/common/_context_menu.js.erb
david-yz-liu: Right, I didn't mean the `action`, I meant adding a `disabled: true` here (after this line).

app/assets/javascripts/Results/context_menu.js
david-yz-liu: If you use `method: 'DELETE'` then I don't think you should need the `_method: 'delete'` as part of the data any more.

app/views/results/common/_context_menu.js.erb
jasonmai: Since these items are going to be submenus/children of `Common Annotations`, having `disabled: true` should not be necessary here since its parent will be disabled unless the conditions correct for the children as well as the parent.

david-yz-liu: When the parent is enabled (i.e., when some text/region is highlighted), this particular option (saying that this annotation category has no annotations) should appear disabled so make it more obvious you can't click on it.

jasonmai: Ah I did not realize this was under the condition that the annotation category is empty.

app/assets/javascripts/Results/context_menu.js
jasonmai: It wasn't needed, thanks for catching this.

PR #2896, opened 2017-03-02
Feature: annotation context menu "edit" item and localization

jasonmai:
## Ready for Review

This pull request contains some major refactoring of the workings of the "Create new annotation" modal. It also contains the implementation for the context menu "edit" feature and ground work for localization of the context menu text items (since we were holding this off during initial implementation). The modal window used for "Create new annotation" initially had its markup fixed on the `results/edit/` page, tied together with various javascript around the codebase. This has now been changed and the markup for the modal window is sent back as a response from `/annotations/new` endpoint. This follows with the Ruby on Rails idioms and it allowed us to more easily re-use this model for the "Edit annotation" context menu feature as well.

 - BUGFIX (pre-existing): Pre-canned annotations were never working with image files [(see description and fix here)](https://github.com/MarkUsProject/Markus/pull/2896/commits/5d75678d4c71e2c06f854ae014ad537fee9805d9)
 - Major refactoring of "Create new annotation" modal window content/markup
 - Context menu "Edit Annotation" item added
 - API end-point `/annotation/new` added, under `annotations_controller.rb`
  - This end-point is hit when a user clicks on the "Create new annotation" button or "New Annotation" on the annotation context menu
 - This endpoint responds with the markup and javascript necessary to have a functional modal that allows for creating new annotations, or editing existing ones.
 - API end-point `/annotation/edit` added, under `annotations_controller.rb`
   - Requires an `annotation_id` (the annotation end point requires `result_id` and `assignment_id` as well)
   - Returns the raw annotation text content corresponding to the `annotation_id`
 - Original modal window element that was made for "Create new annotation" has been modified to also support "Edit Annotation"
 - Hard coded English text elements relating to the annotation context menu has now been changed to utilize localization (however, the non-English translations still need to be added).

### Relevant remaining TODOs for future PR:
 - Add support for using "Delete" or "Edit" when clicking on layered annotations (currently just accessing the most recently added annotation)
 - Move on to adding hotkey support

#### Edit existing annotations with ease:
edit

#### Re-uses existing familiar modal window:
edit_modal
edit_modal_2

#### Annotations updated dynamically, without needing to refresh the page:
result


config/routes.rb
```diff
@@ -351,6 +351,7 @@
       collection do
         post 'add_existing_annotation'
         patch 'update_annotation'
+        post 'edit'
```
david-yz-liu: You shouldn't need this. `resources :annotations` automatically creates an `edit` route, except it's a `GET` (which is the standard).

app/views/results/common/_annotations.js.erb
```diff
@@ -437,8 +490,19 @@
    * @param  {[type]} result_id   [description]
```
david-yz-liu: You can remove these `@param` lines from the docstring.

app/assets/javascripts/Results/context_menu.js.erb
```diff
@@ -16,35 +16,39 @@ var annotation_context_menu = {
   setup: function(annot_path, result_id, assignment_id, file_dl_path) {
     var menu_items = {
       new_annotation: {
-        title: 'New Annotation',
+        title: '<%= I18n.t('marker.annotation.context_menu.new') %>',
```
david-yz-liu: Please use just `t`, not `I18n.t`.

david-yz-liu: It also would have been better to submit these translations in a separate branch.

app/controllers/annotations_controller.rb
```diff
@@ -144,7 +152,11 @@ def update_annotation
     @content = params[:annotation_text][:content]
     @id = params[:annotation_text][:id]
     @submission_file_id = params[:submission_file_id]
```
david-yz-liu: Please remove this line. `@submission_file` should be populated by `@annotation.submission_file`. You can make this change in `destroy` as well.

By doing so, you can eliminate the need to pass in the `submission_file_id` in the AJAX request.

config/routes.rb
jasonmai: I realized that some default routes (including `edit`) were created for `resources: annotations`. However, I thought maybe we'd like to explicitly state the routes we make use of so someone working on the project may use the routes file as a reference to see all list of all implemented routes. I can remove this if it is deemed unnecessary.

app/assets/javascripts/Results/context_menu.js.erb
jasonmai: I initially tried using `t`, but I was getting  a server error upon visiting the page:
```
undefined method `t' for #<#<Class:0x000000029bd320>:0x00000007abb6c8>
```

app/controllers/annotations_controller.rb
jasonmai: The current implementation of editing an existing annotation (from the annotations tab), only passes the `annotation_text_id` rather than the `annotation_id`. I don't think I am able to obtain the `@annotation.submission_file` from the `annotation_text_id`, unless I am able to look up the `annotation_id` from the `annotation_text_id`. With that being said, another way I may be able to get rid of `submission_file` altogether in `update_annotation` and `destroy` is if I have the original implementation for editing an annotation pass in `annotation_id` instead of `annotation_text_id`. Should I proceed to make this change?

config/routes.rb
david-yz-liu: Use of `resources` is a common enough convention that it is not worth restating. It's also not the convention we follow for the rest of the routes, and not something that should be deviated from.

You're having trouble because you didn't actually replicate the route that `resources` generates; instead, you added a new one that looks similar, but is actually different. The correct path helper does exist: it is `edit_annotation_path`. See http://guides.rubyonrails.org/routing.html#path-and-url-helpers for details.

app/assets/javascripts/Results/context_menu.js.erb
david-yz-liu: Ah yes, you're right. I'll make a note of this, but you don't have to change these for now.

app/controllers/annotations_controller.rb
david-yz-liu: This is interesting. Your code change below adds a check for `params[:id]`, which is the annotation id. Is this new? Does this not always get passed? Certainly the ideal behaviour would have the annotation_id, rather than the text id, passed in. (For example, this would allow us to specify whether to change the text of a single annotation or all annotations with that text.)

config/routes.rb
jasonmai: Ahh I see. I did not notice they drop the `s`.

PR #2905, opened 2017-03-13
Guard student tests execution with assignment due date check

adisandro:
This adds a check to prevent student running tests after the due date.
An equal check existed to show the button in the user interface, but not to prevent the actual test function from running.

### DIFF
--- a/app/helpers/automated_tests_client_helper.rb
+++ b/app/helpers/automated_tests_client_helper.rb
@@ -196,7 +196,7 @@ module AutomatedTestsClientHelper
   # Verify the user has the permission to run the tests - admins
   # always have the permission, while student has to
   # belong to the group, and have at least one token.
-  def self.check_user_permission(user, grouping, assignment)
+  def self.check_user_permission(user, grouping)
 
     # admins are always ok
     if user.admin?
@@ -217,7 +217,7 @@ module AutomatedTestsClientHelper
       raise I18n.t('automated_tests.error.bad_group')
     end
     # deadline has not passed
-    if assignment.submission_rule.can_collect_now?
+    if grouping.assignment.submission_rule.can_collect_now?
       raise I18n.t('automated_tests.error.after_due_date')
     end
     token = grouping.prepare_tokens_to_use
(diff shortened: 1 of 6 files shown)


app/helpers/automated_tests_client_helper.rb
```diff
@@ -196,7 +196,7 @@ def self.get_test_server_user
   # Verify the user has the permission to run the tests - admins
   # always have the permission, while student has to
   # belong to the group, and have at least one token.
-  def self.check_user_permission(user, grouping)
+  def self.check_user_permission(user, grouping, assignment)
```
david-yz-liu: You shouldn't need to pass in the assignment; there's an association from `grouping`.

app/helpers/automated_tests_client_helper.rb
```diff
@@ -216,6 +216,10 @@ def self.check_user_permission(user, grouping)
     unless user.accepted_groupings.include?(grouping)
       raise I18n.t('automated_tests.error.bad_group')
     end
+    # deadline has not passed
+    if assignment.submission_rule.can_collect_now?
+      raise I18n.t('automated_tests.error.after_due_date')
```
david-yz-liu: I'm not sure why you didn't use the existing key (the one that you changed) instead of adding a new one.

app/helpers/automated_tests_client_helper.rb
adisandro: I can change it.
I just had it already available in the outer scope and don't know Rails' caching enough to be sure it wasn't triggering a query.

app/helpers/automated_tests_client_helper.rb
adisandro: There are two different pieces of user interface, the autotest "box" with tokens and button, and the flash message for errors. I created it in case they need to diverge.

david-yz-liu: That's fine, you can just use the same key.

app/helpers/automated_tests_client_helper.rb
david-yz-liu: I'm not sure which outer scope you're referring to, but the correct approach seems to me to be to access `grouping.assignment`.

PR #2907, opened 2017-03-19
Markus colors

brendahuang93:
## Ready for Review

This pull request includes up to 40 colours to be used for the ta grade distribution graph. It currently has only four colours, which is insufficient for assignments with more than four TAs.

- Removed list of colours from ta_grade_distribution_graph and moved the colours to markus_colors.js
- Added more colours to a total of 40 colours in markus_colors.js


david-yz-liu:
@brendahuang93 I'm not sure how this change can work properly - it doesn't look like the new Javascript file `markus_colors` is actually included in any pages.

brendahuang93:
@david-yz-liu Hi David, I added references to markus_colors in index and grader_summary. For some reason the graphs were rendering correctly before, but once I updated vagrant it stopped working.

PR #2917, opened 2017-04-10
Use alternative CDN provider for MathJax

arkon:
[The MathJax CDN is shutting down at the end of this month.](https://www.mathjax.org/cdn-shutting-down/)


david-yz-liu:
Thanks, @arkon!

PR #2919, opened 2017-04-10
Added graphs for grade entry forms

brendahuang93:
## Ready for Review

This pull request adds two graphs for grade entry forms - one with the grade distribution for the entire class, the other with the grade distributions for each grade entry item. It also includes the info summary for the grade entry form. The toggle feature for the grade distribution graph for each grade entry item has been modified so only the grade entry item that was clicked on would show while the other ones are disabled. The original path to grade entry form's grades is now linked to the grade entry form's title in view_summary.

- Created two new graphs - grade_distribution_graph and grade_entry_item_distribution_graph for grade entry forms
- Changed Chart.js toggle configuration in grade_entry_item_distribution_graph so graph displays only the dataset that was clicked on and disables the other datasets
- Expanded graph intervals to include 95-100 and fixed issue where grades of 0 would not show up on the graphs
- Updated controller and model for grade entry forms, updated path to view_summary for grade entry form's url link, and created new view_summary, grade_entry_form_summary, grade_entry_form_info_summary, and refresh_graph files to allow the graphs to be rendered and refreshed in main
- Moved path of grade entry form's grades to be linked to grade entry form's title in view_summary
- Updated Chart.js global configuration so the y-axis of the default graph starts at 0
- Updated grade_distribution_graph for assignments and ta_grade_distribution_graph to expand the graphs' intervals to include 95-100

#### Grade distribution graphs, info summary, and refresh graph for grade entry forms:
image

#### Fixed issue where 0 would not show up on graph, expanded intervals to include 95-100, updated toggle function of grade_entry_item_distribution_graph:
image

#### Moved path of grade entry form's grades to be linked to title in view_summary
image
image


app/models/grade_entry_form.rb
```diff
@@ -56,6 +56,32 @@ def calculate_total_percent(grade_entry_student)
     percent
   end
 
+  def grade_distribution_array(intervals = 20)
+    distribution = Array.new(intervals, 0)
+    grade_entry_students.each do |grade_entry_student|
+      if grade_entry_student.total_grade.nil?
```
david-yz-liu: Do not include nil marks in the distribution.

PR #2938, opened 2017-05-02
GradeEntryItemSpec tests added

DecathectZero:


spec/models/grade_entry_item_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    it { is_expected.to allow_value(0).for(:position) }
+    it { is_expected.to allow_value(1).for(:position) }
+    it { is_expected.to allow_value(2).for(:position) }
+    it { is_expected.to allow_value(100).for(:position) }
+    it { is_expected.not_to allow_value(-1).for(:position) }
+    it { is_expected.not_to allow_value(-100).for(:position) }
+  end
+
+  context 'A good Grade Entry Item model' do
```
david-yz-liu: Seems like this block can be removed (it's a duplicate of the test on line 11).

spec/models/grade_entry_item_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    before { GradeEntryItem.create(:position => 1) }
+
+    it { is_expected.to validate_uniqueness_of(:name).scoped_to(
+      :grade_entry_form_id).with_message(
+      I18n.t('grade_entry_forms.invalid_name')) }
+  end
+
+  # Make sure different grade entry forms can have grade entry items
+  # with the same name
+  it 'allow same column name for different grade entry forms' do
```
david-yz-liu: Use present tense here I think ("allows" instead of "allow")

spec/models/grade_entry_item_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    grade_entry_form_1 = GradeEntryForm.create!(:short_identifier => 'a', :is_hidden => false)
```
david-yz-liu: 1. In general, we are switching to use the new object syntax `short_identifier: 'a'` instead, so please change this. Also, I think we can get rid of the `is_hidden` option altogether.

PR #2940, opened 2017-05-02
added StudentMembershipSpec with equivalent tests to previous test file

DecathectZero:


david-yz-liu:
@DecathectZero You don't seem to have committed the change of removing the old file.

PR #2942, opened 2017-05-02
converted test for section to spec

jenachoi:


jenachoi:
@david-yz-liu Ah I see! Do you mean test/unit/section_test.rb?

PR #2945, opened 2017-05-03
added in additional assignment with grace tokens

DecathectZero:
Students are given 5 grace tokens and 1 should be deduced on first assignment.

Note: current version of this branch will cause N+1 query errors to appear as inbrowser prompts on runtime.


lib/tasks/marks.rake
```diff
@@ -4,14 +4,18 @@ namespace :db do
...
-    Grouping.where(assignment_id: [1, 2]).each do |grouping|
+    #Right now, only generate marks for three assignments
+    Grouping.where(assignment_id: [1, 2, 3]).each do |grouping|
       time = grouping.assignment.submission_rule.calculate_collection_time.localtime
       new_submission = Submission.create_by_timestamp(grouping, time)
       result = new_submission.results.first
       grouping.is_collected = true
       grouping.save
 
+      if grouping.assignment_id == 1
```
david-yz-liu: Using ids is quite brittle. I recommend instead checking `grouping.assignment.short_identifier` or something.

lib/tasks/assignments.rake
```diff
@@ -71,7 +98,7 @@ namespace :db do
         group_name_autogenerated: true,
         group_name_displayed: false,
         repository_folder: 'A3',
-        due_date: 2.months.from_now,
+        due_date: 1.months.from_now,
```
david-yz-liu: I'm not sure why you changed this or the other due dates for these assignments; I'd prefer if you revert these changes and keep your changes localized to the new assignment.

lib/tasks/assignments.rake
```diff
@@ -3,6 +3,33 @@ namespace :db do
...
+      short_identifier: 'A0',
+      description: 'Variables and Simple Operations',
+      message: 'using basic operators and assigning variables',
+      group_min: 1,
+      group_max: 1,
+      student_form_groups: true,
+      group_name_autogenerated: true,
+      group_name_displayed: false,
+      repository_folder: 'A0',
+      due_date: 12.hour.ago,
```
david-yz-liu: I believe ruby allows you to use both singular and plural forms for dates and times; please try `12.hours.ago` and see how it goes. (Similar comment below.)

lib/tasks/marks.rake
DecathectZero: This was part of the original code, but I'll modify it to the specifications

david-yz-liu: You can change this if it appears elsewhere too.

lib/tasks/assignments.rake
```diff
@@ -3,6 +3,37 @@ namespace :db do
...
+      submission_rule: rule,
+      assignment_stat: assignment_stat,
+      allow_remarks: false,
+      enable_test: false,
+      tokens_per_period: 0,
+      token_start_date: DateTime.now,
+      token_period: 1
+    )
+    p = Period.create(submission_rule: rule, hours: 3)
+    (1..3).each do |item|
```
david-yz-liu: I'm not sure why this block was added?

DecathectZero: because there are groups now, there are now multiple periods to deduct more than one grace credit per assignment. (This is relative to the submission time)

So now there's 4 periods of 3 hours each as opposed to the original single period of 10 hours

PR #2946, opened 2017-05-04
added new spec/deleted old test for test_script & added more assignment files for seeding

jenachoi:


spec/models/test_script_spec.rb
```diff
@@ -0,0 +1,196 @@
...
+    end
+
+    context 'script file expected to be invalid when the script_name already exists in the same assignment' do
+      it 'return false when the script_name already exists' do
+        @invalidscriptfile.script_name = 'validscript.sh'
+        expect(@invalidscriptfile).not_to be_valid
+      end
+    end
+
+    context 'script file expected to be invalid when the seq_num already exists in the same assignment' do
```
adisandro: Just a typo, it's supposed to be 'valid' instead of 'invalid'

jenachoi: Thanks! I will fix them!!

PR #2949, opened 2017-05-04
feedback files for all assignments

DecathectZero:


lib/tasks/marks.rake
```diff
@@ -4,6 +4,11 @@ namespace :db do
   task :marks => :environment do
     puts 'Assign Marks for Assignments'
 
+    mfile = File.open("db/data/feedback_files/machinefb.txt", "rb")
+    hfile = File.open("db/data/feedback_files/humanfb.txt", "rb")
+    mcont = mfile.read
+    hcont = hfile.read
+
```
david-yz-liu: You should probably close the files after opening them.

PR #2950, opened 2017-05-04
this should work in principle

DecathectZero:
simply deletes everything in the repos folder every time database is generated. 

The commit name was bad, I'm sorry

### DIFF
--- a/lib/tasks/groups.rake
+++ b/lib/tasks/groups.rake
@@ -3,10 +3,13 @@ namespace :db do
   desc 'Create groups for assignments'
   task :groups => :environment do
     puts 'Assign Groups/Students for Assignments'
+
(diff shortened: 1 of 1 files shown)


lib/tasks/groups.rake
```diff
@@ -3,10 +3,13 @@ namespace :db do
   desc 'Create groups for assignments'
   task :groups => :environment do
     puts 'Assign Groups/Students for Assignments'
+
+    # remove previously existing group repos to create room for new ones
+    FileUtils.rm_rf(Dir.glob('data/dev/repos/*'))
```
adisandro: Just be sure that this does not delete the .gitkeep file in the repos dir

PR #2951, opened 2017-05-04
Added seeding for marking scheme

rathukugan:


lib/tasks/marking_scheme.rake
```diff
@@ -0,0 +1,25 @@
...
+    )
+
+    #for each assignment, add a marking weight to marking_scheme
+    Assignment.find_each do |a|
+      random_weight = 1 + rand(0...10)
+      marking_scheme.marking_weights << MarkingWeight.new(
+        gradable_item_id: a.id, weight: random_weight, is_assignment: true)
+    end
+
+    GradeEntryForm.all.each do |grade_entry_form|
```
david-yz-liu: You can use `find_each` here as well, to be consistent.

PR #2957, opened 2017-05-04
fixed group initialization, and having double inviters

DecathectZero:


lib/tasks/groups.rake
```diff
@@ -10,32 +10,26 @@ namespace :db do
       num_groups.times do |time|
         student = students[time]
         if assignment.short_identifier == 'A1' || assignment.short_identifier == 'A3'
+          student.create_group_for_working_alone_student(assignment.id)
+          group = Group.where(group_name: student.user_name).first
```
david-yz-liu: Use `find_by` when only searching for one record.

PR #2960, opened 2017-05-05
Merged submission files in two different directories and changed the specified path in results.rake

jenachoi:


lib/tasks/results.rake
```diff
@@ -64,11 +64,12 @@ namespace :markus do
           grouping = student.accepted_grouping_for(a1.id)
           group = grouping.group
           #commit some files into the group repository
-          file_dir = File.join(File.dirname(__FILE__), '..', '..', 'db', 'sample_submission_files')
+          file_dir = File.join(File.dirname(__FILE__), '..', '..', 'db', 'data', 'submission_files')
           Dir.foreach(file_dir) do |filename|
             unless File.directory?(File.join(file_dir, filename))
               file_contents = File.open(File.join(file_dir, filename))
               file_contents.rewind
+              file_contents.close
```
david-yz-liu: This should only be added after the file contents have been read (in the next block).

jenachoi: whoops, I will move the statement

PR #2962, opened 2017-05-05
deleted files that are not in use

jenachoi:
Cleaned up files that are not in use now that we created seperate folder for submission files and feedback files.


david-yz-liu:
This is great, except the file `long_students.csv` can be used when seeding the students in the database, and so it shouldn't be removed.

jenachoi:
Ah I see! I will get back log_students.csv

jenachoi:
long_students.csv was located under db/data/submission_files so I moved it to the location where students.csv and tas.csv are

PR #2967, opened 2017-05-08
Spreadsheet add import

rathukugan:
I fixed that error I was getting with uploading csv's, turns out it was the blank lines at the end of the .csv file.


app/controllers/grade_entry_forms_controller.rb
```diff
@@ -382,7 +383,7 @@ def csv_upload
         if totals.empty?
           totals = row
           # Create/update the grade entry items
-          GradeEntryItem.create_or_update_from_csv_rows(
+          grades << GradeEntryItem.create_or_update_from_csv_rows(
```
david-yz-liu: You can just do `grades = ...` here, since `create_or_update_from_csv_rows` returns a list.

app/controllers/grade_entry_forms_controller.rb
```diff
@@ -415,6 +416,7 @@ def csv_upload
           end
         end
       end
+      GradeEntryItem.import grades.flatten
```
david-yz-liu: You should move this to right after Line 392 (you want this to be before the block that creates the grades).

app/models/grade_entry_item.rb
```diff
@@ -88,13 +88,17 @@ def self.create_or_update_from_csv_rows(names, totals, grade_entry_form, overwri
     names.shift
     totals.shift
 
+    grade_entrys = []
```
david-yz-liu: Use the correct plural form `entries`

app/models/grade_entry_item.rb
```diff
@@ -88,13 +88,17 @@ def self.create_or_update_from_csv_rows(names, totals, grade_entry_form, overwri
...
+
     (names.size).times do |i|
-      grade_entry_item = grade_entry_form.grade_entry_items.find_or_create_by(name: names[i])
+      grade_entry_item = grade_entry_form.grade_entry_items.find_by(name: names[i])
```
david-yz-liu: This is okay, but you can use [first_or_initialize](http://www.mitchcrowe.com/10-most-underused-activerecord-relation-methods/) instead.

app/controllers/grade_entry_forms_controller.rb
```diff
@@ -383,13 +383,14 @@ def csv_upload
...
           # Create/update the grade entry items
-          grades << GradeEntryItem.create_or_update_from_csv_rows(
+          grades = GradeEntryItem.create_or_update_from_csv_rows(
             names,
             totals,
             @grade_entry_form,
             overwrite)
           next
         end
+        GradeEntryItem.import grades.flatten
```
david-yz-liu: Good, except you shouldn't need to extra "flatten" any more.

david-yz-liu:
@rathukugan looks like there's a failing test in `./spec/controllers/grade_entry_forms_controller_spec.rb:74`, if you could take a look.

app/controllers/grade_entry_forms_controller.rb
```diff
@@ -382,13 +383,14 @@ def csv_upload
...
           # Create/update the grade entry items
-          GradeEntryItem.create_or_update_from_csv_rows(
+          grades = GradeEntryItem.create_or_update_from_csv_rows(
             names,
             totals,
             @grade_entry_form,
             overwrite)
           next
         end
+        GradeEntryItem.import grades
```
david-yz-liu: @rathukugan this line should be moved to inside the previous block (before the `next`).

PR #2970, opened 2017-05-09
converted GradeEntryForm to spec

jenachoi:


spec/models/grade_entry_form_spec.rb
```diff
@@ -0,0 +1,241 @@
+require 'spec_helper'
+require File.expand_path(File.join(File.dirname(__FILE__), '..', '..', 'test', 'test_helper'))
```
david-yz-liu: You can delete these two lines. And code that was necessary from the old test_helper should be moved into a new helper under `spec/`.

spec/models/grade_entry_form_spec.rb
```diff
@@ -0,0 +1,241 @@
...
+
+  context ' A good Grade entry model' do
+    it do
+      is_expected.to validate_uniqueness_of(:short_identifier)
+                       .with_message(I18n.t('grade_entry_forms.invalid_identifier'))
+    end
+  end
+
+  # Make sure validate works appropriately when the date is valid
+  def test_validate_valid_date
```
david-yz-liu: These functions should be turned into one-line tests as in the ones at the top of the file.

spec/models/grade_entry_form_spec.rb
```diff
@@ -0,0 +1,241 @@
...
+  it { is_expected.to have_many(:grades) }
+  it { is_expected.to validate_presence_of(:short_identifier) }
+
+  # Dates in the past should also be allowed
+  it { is_expected.to allow_value(1.day.ago).for(:date) }
+  it { is_expected.to allow_value(1.day.from_now).for(:date) }
+  it { is_expected.not_to allow_value('100-10').for(:date) }
+  it { is_expected.not_to allow_value('abcd').for(:date) }
+
+  context ' A good Grade entry model' do
```
david-yz-liu: This `context` block is unnecessary. The test can be rewritten in the same style as the ones above.

spec/models/grade_entry_form_spec.rb
```diff
@@ -0,0 +1,241 @@
...
+
+  # Tests for out_of_total
+  context 'A grade entry form object: ' do
+    before(:each) do
+      @grade_entry_form = make_grade_entry_form_with_multiple_grade_entry_items
+    end
+
+    # Need at least one GradeEntryForm object created for this
+    # test to pass.
+    it do
```
david-yz-liu: This test is redundant, and can be removed.

spec/models/grade_entry_form_spec.rb
```diff
@@ -0,0 +1,241 @@
...
+          grade_entry_student.released_to_student = true
+        else
+          grade_entry_student.released_to_student = false
+        end
+        grade_entry_student.save
+      end
+    end
+
+    it 'verify the correct value is returned when there are no marks released' do
+      expect(@grade_entry_form_none_released.calculate_released_average()).to eq 0
```
david-yz-liu: Omit the `()` for method calls that have no parameters. (This occurs in other places as well.)

spec/models/grade_entry_form_spec.rb
```diff
@@ -0,0 +1,241 @@
...
+    (1..3).each do |i|
+      grade_entry_items << GradeEntryItem.create(grade_entry_form: @grade_entry_form,
+                                                 out_of: 10,
+                                                 name: 'Q' + i.to_s,
+                                                 position: i)
+    end
+    grade_entry_form.grade_entry_items = grade_entry_items
+    return grade_entry_form
+  end
+
```
david-yz-liu: Delete these blank lines.

jenachoi:
I changed them! blueprints.rb has to be added in order to run rspec on this file. I will add blueprints.rb on the separate branch after this!

david-yz-liu:
@jenachoi This test is a little more interesting - there are already *factories* that are under `rspec/` that you should use instead. (So you can remove the blueprints import.)

jenachoi:
fixed it so that it doesn't require blueprints!

PR #2974, opened 2017-05-10
Fix bug from submission_rule model

jenachoi:


app/models/submission_rule.rb
```diff
@@ -49,13 +49,13 @@ def get_collection_time(section=nil)
...
   end
-  
+
   def calculate_collection_time(section=nil)
     assignment.section_due_date(section) + hours_sum.hours
   end
 
   def calculate_grouping_collection_time(grouping)
-    if grouping.inviter.section
+    if (!grouping.inviter.nil? && grouping.inviter.section)
```
david-yz-liu: Good, but you don't need parentheses around this expression.

PR #2985, opened 2017-05-11
removed repetition of code in criterion.rb

jenachoi:


app/models/criterion.rb
```diff
@@ -101,62 +101,36 @@ def self.unassign_tas(criterion_ta_ids, criterion_ids_by_type, assignment)
...
-    if criterion_ids_by_type.nil?  or criterion_ids_by_type['CheckboxCriterion'].nil?
-      checkbox_criterion_ids_str = ''
-    else
-      checkbox_criterion_ids_str = Array(criterion_ids_by_type['CheckboxCriterion'])
-        .map { |criterion_id| connection.quote(criterion_id) }
-        .join(',')
+    criteria = ['RubricCriterion', 'FlexibleCriterion', 'CheckboxCriterion']
+    criteria.each do |criterion|
+      if criterion_ids_by_type.nil? or criterion_ids_by_type[criterion].nil?
+        rubric_criterion_ids_str = ''
```
david-yz-liu: This loop is re-writing the same variable at each iteration. A loop is not the right way to abstract away the repetition. A helper function is probably more appropriate.

PR #2986, opened 2017-05-11
Remove header rows from valid_line_count, so entry count is accurate

rathukugan:


app/controllers/grade_entry_forms_controller.rb
```diff
@@ -381,14 +381,14 @@ def csv_upload
...
-          # Create/update the grade entry items
-          GradeEntryItem.create_or_update_from_csv_rows(
-            names,
-            totals,
-            @grade_entry_form,
-            overwrite)
           next
         end
+        # Create/update the grade entry items
+        GradeEntryItem.create_or_update_from_csv_rows(
```
david-yz-liu: You should revert this change. This part should only happen once (when processing the first row).

rathukugan:
I changed it to use `.blank?` instead so the `""` case is also caught, should I still raise the `CSV::MalformedCSVError`for when the first entry is not blank?

david-yz-liu:
If you want to subtract 2, then you'll need to add a parameter to the Markus csv parsing, with a default value of 0. But I suppose that's actually fine. The hard-coding is a symptom of the larger problem that header lines aren't handled in a generic way, but that's okay for now.

lib/classes/markus_csv.rb
```diff
@@ -15,14 +15,16 @@ def self.generate(objects, headers=[], &gen_csv)
...
+  # number of header lines to remove from the displayed valid_line_count
+  # and parse_obj is a block which takes a line and performs an action,
+  # or raises CSVInvalidLineError.
   # Returns a result hash, containing a success message with the number of
   # successful rows parsed, as well as an error message, consisting of one
   # of the following:
   #   1) A string listing all erroneous lines.
   #   2) A more generic error message for invalid files.
-  def self.parse(input, options = {}, &parse_obj)
+  def self.parse(input, options = {}, header_count = 0, &parse_obj)
```
david-yz-liu: Ahh I forgot there was an options hash already. In this case, you shouldn't add an extra parameter, but instead pass in `header_count: 2`.

Then in the body of this method, you can do `header_count = options[:header_count] || 0` to access the argument, or use a default value.

rathukugan: Hey David, I'm getting an `ArgumentError` here when adding header_count to the options hash
`Unknown options:  header_count.` on line 34 `CSV.parse(input, options) do |row|`
From what I understood, I thought the options hash could take in any parameters?

david-yz-liu: Oh geez, it looks like it's the `CSV.parse` method specifically that is rejecting options that aren't recognized. I forgot that we passed the options hash to this method. In this case, use `options.delete(:header_count)` to both extract the option and remove it from `options`. (This returns `nil` if the key is not present, so you still need to specify the default value.)

PR #2987, opened 2017-05-11
Removed old download buttons and replaced them

DecathectZero:


app/views/results/marker/_marker_panes.html.erb
```diff
@@ -25,15 +25,20 @@
...
               <% unless @current_user.is_reviewer_for?(@assignment.pr_assignment, @result) %>
-                <%= render partial: 'annotations/annotation_categories',
-                           locals: { annotation_categories: annotation_categories,
-                                     result_id: result.id } %>
+                <button id='file_download_button' onclick='modal_download.open(); return false;'>
+                  <%= t(:download)%>
+                </button>
+
+                <div class="annotation_options">
+                  <button id='new_annotation_button' onclick='make_new_annotation(); return false;'>
```
david-yz-liu: This is a little tricky. For peer reviews, we want the button to be displayed, but *not* the annotation categories, nor the "file download button."

DecathectZero: so the only button needed for peer reviews is the new annotation button?

david-yz-liu: Yes, that's right.

PR #2988, opened 2017-05-11
Check if row array is all nil to skip line

rathukugan:
Example of a csv with blank lines that caused an error
,Q1,Q2,Q3
,3,4,5
c8godows,2,2,2
,,,
",",,,

`CSV.generate_line(row).strip.empty?` was false for a blank line (therefore it didnt get skipped) since ",,," isnt empty so instead I checked the row array for a blank line instead.

### DIFF
--- a/app/controllers/grade_entry_forms_controller.rb
+++ b/app/controllers/grade_entry_forms_controller.rb
@@ -373,7 +373,7 @@ class GradeEntryFormsController < ApplicationController
 
       # Parse the grades
       result = MarkusCSV.parse(grades_file.read, encoding: encoding) do |row|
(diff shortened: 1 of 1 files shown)


app/controllers/grade_entry_forms_controller.rb
```diff
@@ -373,7 +373,7 @@ def csv_upload
 
       # Parse the grades
       result = MarkusCSV.parse(grades_file.read, encoding: encoding) do |row|
-        next if CSV.generate_line(row).strip.empty?
+        next if !row.any?
```
david-yz-liu: Good! But please use `unless` instead of `if !` here.

PR #2995, opened 2017-05-15
making sure we only have one flash message in a page

jenachoi:
Making sure that when we update criteria or marks spreadsheet for quiz, we get only one flash message at a page.


david-yz-liu:
@jenachoi the `assignment_content` inclusion of the flash messages should be the one that's kept (all other ones are the ones to remove).

PR #2996, opened 2017-05-15
make flash message appear when criterion gets created

jenachoi:
When we update or delete criterion, flash messages appear. I added flash message for when we create new criterion.

### DIFF
--- a/app/controllers/criteria_controller.rb
+++ b/app/controllers/criteria_controller.rb
@@ -31,8 +31,8 @@ class CriteriaController < ApplicationController
       return
     end
     @criteria = @assignment.get_criteria
-    render :create_and_edit
     flash_now(:success, t('criterion_created_success'))
+    render :create_and_edit
   end
 
   def edit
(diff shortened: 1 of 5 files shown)


app/controllers/criteria_controller.rb
```diff
@@ -32,6 +32,7 @@ def create
     end
     @criteria = @assignment.get_criteria
     render :create_and_edit
+    flash_now(:success, t('criterion_created_success'))
```
david-yz-liu: Good; in general, put this before the `render.

config/locales/en.yml
```diff
@@ -1333,6 +1333,8 @@ en:
         help: "Notes can be created for students, groups, or assignments. They are visible to admins and graders, but not students."
 
     criteria_edit_total_mark:     "This criterion will be marked out of %{criterion_total_mark}"
+    criterion_created_success:   "Criterion was successfully created"
+    criterion_created_failture:   "Criterion failed to be created"
```
david-yz-liu: This is spelled wrong. Also, you should add these two keys for the other locale files (you can use Google translate if you like).

PR #2999, opened 2017-05-16
Markingstatus

DecathectZero:


app/views/layouts/result_content.erb
```diff
@@ -58,11 +58,40 @@
         </div>
       </div>
     </div>
+    <!-- Progress Bar -->
     <section id='content'>
+      <div class="progress_meter">
+        <span id="progress_span"><p>
```
david-yz-liu: I'm not convinced the `p` tag is necessary (seems like we can just use the span). If you can remove this, that would be great (and from the css as well).

app/views/layouts/result_content.erb
```diff
@@ -58,11 +58,40 @@
...
+        </p></span>
+      </div>
       <%= render partial: 'layouts/noscript'%>
       <%= content_for?(:content) ? yield(:content) : yield %>
     </section>
-
+    <%
+      if @current_user.admin?
+        num_marked = @assignment.get_num_marked()
```
david-yz-liu: Don't include parentheses when there are no arguments (i.e., just do `@assignment.get_num_marked`)

app/views/layouts/result_content.erb
```diff
@@ -58,11 +58,40 @@
...
+        num_assigned = @assignment.get_num_assigned()
+      else
+        num_marked = @assignment.get_num_marked(@current_user.id)
+        num_assigned = @assignment.get_num_assigned(@current_user.id)
+      end
+    %>
+    <script type="text/javascript">
+        var marked = <%= num_marked %>
+        var total = <%= num_assigned %>
+        function update_bar(score, outof){
```
david-yz-liu: Please put a space before `{`

app/views/layouts/result_content.erb
```diff
@@ -58,11 +58,40 @@
...
+      if @current_user.admin?
+        num_marked = @assignment.get_num_marked()
+        num_assigned = @assignment.get_num_assigned()
+      else
+        num_marked = @assignment.get_num_marked(@current_user.id)
+        num_assigned = @assignment.get_num_assigned(@current_user.id)
+      end
+    %>
+    <script type="text/javascript">
+        var marked = <%= num_marked %>
```
david-yz-liu: I'd rather not introduce global variables here. The easiest thing to do is perform the function call by passing in `num_marked` and `num_assigned` directly, and then in the AJAX response, call `get_num_marked` and `get_num_assigned` again to recompute the numbers and pass them to `update_bar`.

app/views/layouts/result_content.erb
```diff
@@ -58,11 +58,40 @@
...
+        var total = <%= num_assigned %>
+        function update_bar(score, outof){
+            var width = (score/outof) * 100;
+            var $bar = $('#progress_span');
+            $bar.children().text(score + '/' + outof + ' Completed');
+            $bar.css('width', width + '%');
+            if(width > 75){
```
david-yz-liu: Please put a space before the `(` and before the `{`.

app/views/layouts/result_content.erb
```diff
@@ -58,11 +58,40 @@
...
+                $bar.css('background-color', 'green')
```
david-yz-liu: Please don't modify the css directly here. You can set a data attribute here and match it in the css file.

app/models/assignment.rb
```diff
@@ -775,7 +775,7 @@ def get_num_assigned(ta_id = nil)
 
   def get_num_marked(ta_id = nil)
     if ta_id.nil?
-      groupings.count(marking_completed: true)
+      groupings.select(&:marking_completed?).count()
```
david-yz-liu: Omit the `()` here (I made a similar comment below)

app/views/layouts/result_content.erb
```diff
@@ -58,11 +58,40 @@
...
+            }else if(width > 35){
+                $bar.css('background-color', '#FBC02D')
+            }
+        }
+        update_bar(marked, total);
```
david-yz-liu: Please wrap this in a `$.ready(document)`

app/views/results/edit.html.erb
```diff
@@ -164,6 +168,27 @@
...
+        num_assigned = @assignment.get_num_assigned
+      else
+        num_marked = @assignment.get_num_marked(@current_user.id)
+        num_assigned = @assignment.get_num_assigned(@current_user.id)
+      end
+    %>
+    $( document ).ready(function() {
+        update_bar(<%= num_marked %>, <%= num_assigned %>);
+    });
+</script>
```
david-yz-liu: Better, but please merge this with the other `$(document).ready` callback earlier in this file.

app/views/results/marker/_update_progress_bar.js.erb
```diff
@@ -0,0 +1,13 @@
+<script>
+  function update_bar(score, outof) {
+      var width = (score/outof) * 100;
```
david-yz-liu: I realized that there should probably be a guard for when `outof` is 0 (in this case, maybe there should be a full-width bar).

app/views/results/marker/_update_progress_bar.js.erb
```diff
@@ -0,0 +1,13 @@
...
+      var $bar = $('#progress_span');
+      $bar.text(score + '/' + outof + ' Completed');
+      $bar.css('width', width + '%');
+      if(width > 75) {
```
david-yz-liu: Please put a space after the `if` here, and around the `else if` below.

app/views/layouts/result_content.erb
david-yz-liu: @DecathectZero I think this one still needs to be addressed.

david-yz-liu:
@DecathectZero I've merged this in, thanks! However, after looking at this some more, I'm worried about the vertical space the bar takes up. When you get a chance, please move it as follows:

1. In the "Normal view", it goes to the left of the heading links ("compact view")
2. In the "Compact view", it goes to the left of the top bar links ("normal view")

PR #3002, opened 2017-05-18
TA Criteria

DecathectZero:
I guess this should work and covers the cases.

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -784,7 +784,9 @@ class Assignment < ActiveRecord::Base
       if is_criteria_mark?(ta_id)
         n = 0
         ta_memberships.includes(grouping: :current_submission_used).where(user_id: ta_id).find_each do |x|
-          x.grouping.current_submission_used.get_latest_result.marks.joins('INNER JOIN criterion_ta_associations c ON c.criterion_id = markable_id AND c.criterion_type = markable_type').where('c.ta_id = ? AND mark IS NULL', ta_id).empty? && n += 1
+          x.grouping.current_submission_used.get_latest_result.marks
+            .joins('INNER JOIN criterion_ta_associations c ON c.criterion_id = markable_id AND c.criterion_type = markable_type')
+            .where('c.ta_id': ta_id, mark: nil).empty? && n += 1
         end
         n
       else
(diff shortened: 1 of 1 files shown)


app/models/assignment.rb
```diff
@@ -775,13 +779,31 @@ def get_num_assigned(ta_id = nil)
...
+        criteria = []
+        criteriatype = []
+        self.criterion_ta_associations.where(ta_id: ta_id).find_each do |x|
+          criteria << x.criterion_id
+          criteriatype << x.criterion_type
+        end
+        n = 0
+        ta_memberships.includes(grouping: :current_submission_used).where(user_id: ta_id).find_each do |x|
+          hasnull = false
+          for i in 0..criteria.size do
```
david-yz-liu: This is a good start. Let's try to remove the need for an inner loop by using a join. (http://guides.rubyonrails.org/active_record_querying.html#joining-tables) I think you need a custom string for the join, something like `"INNER JOIN on results.markable_id = criterion_ta_associations.criterion_id AND results.markable_type = criterion_ta_associations.criterion_type"`.

app/models/assignment.rb
```diff
@@ -782,19 +782,9 @@ def get_num_marked(ta_id = nil)
...
-          criteriatype << x.criterion_type
-        end
         n = 0
         ta_memberships.includes(grouping: :current_submission_used).where(user_id: ta_id).find_each do |x|
-          hasnull = false
-          for i in 0..criteria.size do
-            !(x.grouping.current_submission_used.get_latest_result.marks.where(markable_id: criteria.at(i), markable_type: criteriatype.at(i), mark: nil).empty?) && hasnull = true
-          end
-          !hasnull && n += 1
+          x.grouping.current_submission_used.get_latest_result.marks.joins('INNER JOIN criterion_ta_associations c ON c.criterion_id = markable_id AND c.criterion_type = markable_type').where('c.ta_id = ? AND mark IS NULL', ta_id).empty? && n += 1
```
david-yz-liu: Great! Please separate the method calls on different lines (put each period on a newline).

Also for the `where`, don't use a raw string - instead, you can do something like `.where('c.ta_id': ta_id, mark: nil)`. (In general, we prefer not using raw strings.)

PR #3003, opened 2017-05-18
Store tas for filter

rathukugan:
I forgot to do this pull request before working on the ta filter, but is this the right place I should store the ta list?


david-yz-liu:
@rathukugan this is the correct place - but you should really just store the TA user names, not the TA objects themselves. (You can use a `pluck` method here I think)

PR #3004, opened 2017-05-18
Refactor repository code to accept git shas

adisandro:
This is a work in progress, it's good to get some early reviews.
(the repo browser works reliably, everything else is still untested)


adisandro:
@david-yz-liu Is there any other page to be fixed after your collection fixes? If not, we could merge this in.
New things:
1) Fixed 3 tests that were failing because of MemoryRepository not being up to speed
2) Added a first hook to prevent forced push on master, it gets installed at bare creation time, you can try it with a simple db:reset

PR #3005, opened 2017-05-19
Section filter to ta

rathukugan:
Replaced the section filter with the TA filter, will separate in the next pull request.

Also do you have a suggestion for a good place to put the 'All TAs' phrase in the yml file(s)?

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -277,6 +277,12 @@ class SubmissionsController < ApplicationController
     @groupings = Grouping.get_groupings_for_assignment(@assignment,
                                                        current_user)
     @sections = Section.order(:name)
+
+    @tas = Set.new
+    @groupings.map do |grouping|
+      grouping.tas.pluck(:user_name).each{|ta| @tas.add(ta)}
+    end
+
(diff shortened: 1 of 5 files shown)


app/controllers/submissions_controller.rb
```diff
@@ -277,6 +277,12 @@ def browse
     @groupings = Grouping.get_groupings_for_assignment(@assignment,
                                                        current_user)
     @sections = Section.order(:name)
+
+    @tas = Set.new
+    @groupings.map do |grouping|
+      grouping.tas.pluck(:user_name).each{|ta| @tas.add(ta)}
```
david-yz-liu: I think there's a more elegant way of doing this - `@assignment.ta_memberships` should work, and from there you can join with `:tas` and then pluck `ta.user_name`

rathukugan:
Hey David,
The section and ta filters should be working but i'm getting an error before the submission table data is loaded
`ActiveRecord::AssociationNotFoundError (Association named 'current_result' was not found on Grouping; perhaps you misspelled it?):` from line 350 on submissions_controller.rb

Any idea what's going on or is it just something on my end?

PR #3009, opened 2017-05-23
created page for Exam Template under midterm1>Settings

jenachoi:
Currently it only shows number of pages and file name for the exam template. And it also has link for downloading exam template. But more things will be added!


app/views/exam_templates/_form.html.erb
```diff
@@ -0,0 +1,10 @@
+<%= form_for @exam_templates,
```
david-yz-liu: I think you should use the `exam_template` local variable (you pass this to the partial)

app/views/exam_templates/index.html.erb
```diff
@@ -0,0 +1,8 @@
+<% content_for :title, ExamTemplate.model_name.human.pluralize %>
+
+<% content_for :content do %>
+    <%= render partial: 'form',
+               formats: [:html],
```
david-yz-liu: You shouldn't need to specify the `formats` or `handlers`

app/views/exam_templates/index.html.erb
```diff
@@ -0,0 +1,8 @@
+<% content_for :title, ExamTemplate.model_name.human.pluralize %>
+
+<% content_for :content do %>
```
david-yz-liu: Good - so you don't actually need the outer `content_for`; the contents of the view is actually automatically put in `content` in the layout.

app/views/exam_templates/_form.html.erb
```diff
@@ -0,0 +1,10 @@
...
+             as: :exam_template,
+             url: assignment_exam_templates_path,
+             html: {multipart: true} do |t| %>
+    <div class='template-pane'>
+      <h4><%= link_to "Download Exam Template", download_assignment_exam_templates_path %></h4><br>
```
david-yz-liu: Let's make this a bit better. Pass in `id: exam_template.id` to the path helper here, and make the `download` route a `member` instead of `collection` (in the routes.rb file), and then update the `download` method in the controller to look up a specific exam template based on the `params[:id]`.

app/controllers/exam_templates_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+    @assignment = Assignment.find(params[:assignment_id])
+    @exam_templates = ExamTemplate.all
+  end
+
+  def download
+    assignment = Assignment.find(params[:assignment_id])
+    exam_template = ExamTemplate.find_by(assignment: assignment)
+    filename = exam_template.filename
+    basename = File.basename(filename, ".pdf")
+    send_file("#{Rails.root}/data/dev/exam_templates/#{basename}/#{filename}",
```
david-yz-liu: I believe there's an option in the config files (check `development.rb`) that specifies the path. You should look into `markus_config_*` methods to see how we use them to access configuration options.

app/controllers/exam_templates_controller.rb
```diff
@@ -0,0 +1,21 @@
+class ExamTemplatesController < ApplicationController
+
+  before_filter      :authorize_only_for_admin
+
+  layout 'assignment_content'
+
+  def index
+    @assignment = Assignment.find(params[:assignment_id])
+    @exam_templates = ExamTemplate.all
```
david-yz-liu: You should only grab the exam templates associated with the particular assignment.

config/routes.rb
```diff
@@ -94,6 +94,13 @@
         end
       end
 
+      resources :exam_templates do
+        member do
+          get 'download'
+          post 'update'
```
david-yz-liu: This is not necessary. `resources` will automatically create an `upload` route (see the Rails "Routing" guide), which you can use when you set up the form properly

app/views/exam_templates/index.html.erb
```diff
@@ -0,0 +1,4 @@
+<% content_for :title, ExamTemplate.model_name.human.pluralize %>
+
+<%= render partial: 'form',
+           locals: { exam_template: @exam_templates }%>
```
david-yz-liu: Do `@exam_templates[0]` here for now. Eventually you'll want a separate form for each exam template.

app/views/exam_templates/_form.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= render partial: 'exam_template_download',
+           locals: { exam_template: @exam_templates }%>
+<br><br>
```
david-yz-liu: Don't use `<br>` for styling. You should remove these for now.

app/views/exam_templates/_exam_template_upload.html.erb
```diff
@@ -0,0 +1,7 @@
+<label>Upload Exam Template</label>
+<%= form_tag({:action=>"update", :controller=>"exam_templates", :id=>exam_template.id}, :multipart => true) do %>
+    <p>
+      <%= file_field_tag 'new_template' %>
```
david-yz-liu: This actually shouldn't be a separate partial. Instead, this tag belongs in the form currently contained in `_exam_template_download` (whose contents should actually just go into `_form`)

app/controllers/exam_templates_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    # error checking when new_uploaded_io is not pdf
+    if new_uploaded_io.content_type != 'application/pdf'
+      flash_message(:failure, 'Exam Template should be in pdf format')
+      redirect_to action: 'index'
+      return
+    end
+    assignment = Assignment.find(params[:assignment_id])
+    old_exam_template = ExamTemplate.find_by(assignment: assignment, id: params[:id])
+    old_template_filename = old_exam_template.filename
+    ExamTemplate.create_with_file(new_uploaded_io.read, assignment_id: assignment.id, filename: old_template_filename)
```
david-yz-liu: You shouldn't be creating a new `ExamTemplate` object here. The point of an *update* is to replace an existing PDF with a new one.

config/routes.rb
jenachoi: oh I see! Do you mean update route by any chance?

david-yz-liu: yes, that's right :)

app/views/exam_templates/index.html.erb
jenachoi: local variable exam_template becomes nil when I try @exam_templates[0]. Should I leave it?

app/controllers/exam_templates_controller.rb
```diff
@@ -0,0 +1,37 @@
+class ExamTemplatesController < ApplicationController
+
+  before_filter      :authorize_only_for_admin
+
+  layout 'assignment_content'
+
+  def index
+    @assignment = Assignment.find(params[:assignment_id])
+    @exam_templates = ExamTemplate.find_by(assignment: @assignment)
```
david-yz-liu: Use `.where` here, which will grab all exam templates for an assignment.

config/locales/en.yml
```diff
@@ -1480,6 +1480,16 @@ en:
     date_validator:
       invalid_date: "is not a valid date"
 
+    #Exam Template
```
david-yz-liu: These should be put into separate locale files.

config/routes.rb
```diff
@@ -94,6 +94,17 @@
         end
       end
 
+      resources :exam_templates do
+        member do
+          get 'download'
+        end
+
+        collection do
+          get 'index'
```
david-yz-liu: You need to review http://guides.rubyonrails.org/routing.html#resource-routing-the-rails-default. Don't create routes that are already defined.

app/views/exam_templates/_form.html.erb
```diff
@@ -0,0 +1,21 @@
+<%= form_for exam_template,
+             as: :exam_template,
+             url: assignment_exam_templates_path,
+             html: {multipart: true} do |f| %>
+    <%= f.error_messages%>
+    <div class='template-pane'>
+      <h4>
+        <%= link_to t('exam_templates.download'), {:action=>'download', :controller=>'exam_templates', :id=>exam_template.id} %>
```
david-yz-liu: Use a path helper. See http://guides.rubyonrails.org/routing.html#adding-more-restful-actions for what is generated when you create a new route (like `download`). You are looking for a method that ends with `_path`.

app/views/exam_templates/_form.html.erb
```diff
@@ -0,0 +1,21 @@
+<%= form_for exam_template,
+             as: :exam_template,
```
david-yz-liu: You shouldn't need the `as` or `url` keys (rails is pretty good about defaults).

app/views/exam_templates/_form.html.erb
```diff
@@ -0,0 +1,20 @@
+<%= form_for exam_template,
+             url: assignment_exam_templates_path,
+             html: {multipart: true} do |f| %>
+    <%= f.error_messages%>
+    <div class='template-pane'>
+      <h4>
+        <%= link_to t('exam_templates.download'), download_assignment_exam_template_path(:id=>exam_template.id) %>
```
david-yz-liu: Good, but in general prefer using the new key syntax `id: exam_template.id`.

app/views/exam_templates/_form.html.erb
```diff
@@ -0,0 +1,20 @@
+<%= form_for exam_template,
+             url: assignment_exam_templates_path,
+             html: {multipart: true} do |f| %>
```
david-yz-liu: For hashes the conventional style is the include a blank space after the `{` and before the `}

app/controllers/exam_templates_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  layout 'assignment_content'
+
+  def index
+    @assignment = Assignment.find(params[:assignment_id])
+    @exam_templates = ExamTemplate.where(assignment: @assignment)
+  end
+
+  def download
+    assignment = Assignment.find(params[:assignment_id])
+    exam_template = ExamTemplate.find_by(assignment: assignment, id: params[:id]) # look up a specific exam template based on the params[:id]
```
david-yz-liu: Better to use associations: `assignment.exam_templates.find_by(...)`

app/views/exam_templates/_form.html.erb
```diff
@@ -0,0 +1,20 @@
+<%= form_for exam_template,
```
david-yz-liu: Use `form_for [@assignment, exam_template]` here - this will allow you to remove the `url` argument.

jenachoi: okay! I will make a change 👍

jenachoi: should i make this change in separate branch?

PR #3017, opened 2017-05-30
added condition to prevent error when we don't upload any file

jenachoi:
This is necessary since we will add nested form for updating template division. The only change in the code here is adding 'unless' statement to prevent the error that will occur when we update 'only' template division.

### DIFF
--- a/app/controllers/exam_templates_controller.rb
+++ b/app/controllers/exam_templates_controller.rb
@@ -22,7 +22,6 @@ class ExamTemplatesController < ApplicationController
   def update
     assignment = Assignment.find(params[:assignment_id])
     old_exam_template = assignment.exam_templates.find_by(id: params[:id])
-
     # updating exam template file
     new_uploaded_io = params[:exam_template][:new_template]
     unless new_uploaded_io.nil?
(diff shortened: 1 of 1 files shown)


app/controllers/exam_templates_controller.rb
```diff
@@ -20,18 +20,21 @@ def download
...
-    old_template_filename = old_exam_template.filename
-    old_exam_template.replace_with_file(new_uploaded_io.read, assignment_id: assignment.id, filename: old_template_filename)
-    flash_message(:success, t('exam_templates.update.success'))
+    # updating exam template file
+    new_uploaded_io = params[:exam_template][:new_template]
+    unless new_uploaded_io.nil?
+      # error checking when new_uploaded_io is not pdf
+      if new_uploaded_io.content_type != 'application/pdf'
+        flash_message(:error, t('exam_templates.update.failure'))
+        redirect_to action: 'index'
```
david-yz-liu: Don't put the `redirect_to` (or `return`) here. Instead, put the following code in an `else` block after this, and then keep the `redirect_to` on line 38.

david-yz-liu:
@jenachoi I'm not sure the problem is fixed (in particular, `params[:exam_template]` might be nil. Also, there's a conflict now that I merged in the responders branch, so please do a merge and then fix this as needed.

jenachoi:
I haven't finished this yet!

jenachoi:
finished- Please let me know if there is any changes needed

PR #3023, opened 2017-05-31
Enable deleting template division

jenachoi:


jenachoi:
destroy_all method doesn't seem to delete all exam templates yet. I'll be working on it soon

app/views/exam_templates/_form.html.erb
```diff
@@ -3,6 +3,12 @@
     <%= f.error_messages%>
     <div class='template-pane'>
       <h4>
+        <%= link_to t('exam_templates.delete.all_general'), action: 'destroy_all' %>
```
david-yz-liu: This is not necessary; you can remove this button (and route).

app/views/exam_templates/_form.html.erb
```diff
@@ -3,6 +3,12 @@
...
+      </h4>
+      <h4>
+        <%= link_to t('exam_templates.delete.general'), action: 'destroy', id: exam_template.id %>
```
david-yz-liu: This is not the normal Rails approach. See http://guides.rubyonrails.org/form_helpers.html#removing-objects.

jenachoi: I think the link is about deleting associated objects, which might be useful to delete template divisions. But for deleting exam template objects, aren't we supposed to write our own method in the controller?

app/views/exam_templates/_form.html.erb
```diff
@@ -22,6 +22,7 @@
       <p><%= raw t('assignment.required_fields') %> <span class='required_field'>*</span></p>
       <%= f.fields_for :template_divisions do |form| %>
         <p>
+          <%= form.check_box :_destroy%>
```
david-yz-liu: There should be a space before the `%>`. Also, this checkbox should have a label.

PR #3018, opened 2017-05-30
New template divisions nested-form

jenachoi:
User can update label, starting page, and ending page of template divisions from the exam template.

PR #3025, opened 2017-06-01
exam template db migration: add name field

jenachoi:
by default, name is set to the filename without file extension.

### DIFF
--- a/app/models/exam_template.rb
+++ b/app/models/exam_template.rb
@@ -7,7 +7,8 @@ require 'rmagick'
 
 class ExamTemplate < ActiveRecord::Base
   belongs_to :assignment
-  validates :assignment, :filename, :num_pages, presence: true
+  validates :assignment, :filename, :num_pages, :name, presence: true
+  validates :name, uniqueness: true
   validates :num_pages, numericality: { greater_than_or_equal_to: 0,
                                         only_integer: true }
 
--- a/app/models/exam_template.rb
+++ b/app/models/exam_template.rb
@@ -6,6 +6,8 @@ require 'zxing'
 require 'rmagick'
 
 class ExamTemplate < ActiveRecord::Base
+  after_initialize :set_defaults, unless: :persisted? # The set_defaults will only work if the object is new
+
   belongs_to :assignment
   validates :assignment, :filename, :num_pages, :name, presence: true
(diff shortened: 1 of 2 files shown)


david-yz-liu:
You need to grab the new migration file and the new `schema.rb` file from the virtual machine and include them in this pull request.

jenachoi:
okies! I will work on that after fixing merge conflicts from #3018

app/controllers/exam_templates_controller.rb
```diff
@@ -54,10 +54,30 @@ def update
...
+        flash_message(:error, t('exam_templates.update.failure'))
+      end
     end
     redirect_to action: 'index'
   end
+
+  def exam_template_params
+    params.require(:exam_template)
+      .permit(
+        :assignment,
```
david-yz-liu: The only attribute you should include in these params are `name` and the nested attributes for the template divisions `:start`, `:end`, and `:label`.

app/views/exam_templates/_form.html.erb
```diff
@@ -5,13 +5,18 @@
...
+      <h4><%= ExamTemplate.human_attribute_name('name') %>: <%= exam_template.name %></h4>
       <h4><%= ExamTemplate.human_attribute_name('num_pages') %>: <%= exam_template.num_pages %></h4>
       <h4><%= ExamTemplate.human_attribute_name('filename') %>: <%= exam_template.filename %></h4>
     </div>
 
     <fieldset>
       <legend><%= t('exam_templates.update.general') %></legend>
       <p><%= raw t('assignment.required_fields') %> <span class='required_field'>*</span></p>
+      <p>
+        <%= f.label ExamTemplate.model_name.human + ' ' + ExamTemplate.human_attribute_name('name') %>
```
david-yz-liu: You don't need to include the model name here.

app/controllers/exam_templates_controller.rb
jenachoi: When we don't include :id Inside template_divisions_attributes, it doesn't seem to update itself. So I left :id in template_divisions_attributes along with :start, :end, and :label

jenachoi:
Oops I forgot to undo the changes to the other two migration files. I will work on that right now.

PR #3026, opened 2017-06-01
Template Division Validation

jenachoi:
It should always be 1 <= start <= end


jenachoi:
I noticed that this causes error when we try rake db:reset. I will find another way

jenachoi:
This still seems to be causing error when I try running rake:db reset. I will fix that-

PR #3029, opened 2017-06-02
Added student ids and email columns to users

DecathectZero:
Not sure if I did this migration correctly. Feedback appreciated

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -23,6 +23,8 @@ class User < ActiveRecord::Base
 
   validates_presence_of     :user_name, :last_name, :first_name
   validates_uniqueness_of   :user_name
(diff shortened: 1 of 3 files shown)


david-yz-liu:
The migration file is good, but you need to run the migration and the download the `schema.rb` file from the virtual machine onto your local machine to commit the changes. You should see in the diff that two new columns are actually added to `schema.rb`.

app/models/user.rb
```diff
@@ -282,6 +284,12 @@ def strip_name
     if self.first_name
       self.first_name = self.first_name.strip
     end
+    if self.email
+      self.email = self.email.strip
+    end
+    if self.id_number
```
david-yz-liu: I don't think this is necessary - the `id_number` is a number, not a string.

PR #3040, opened 2017-06-06
Multiple n+plus query fixes

DecathectZero:


david-yz-liu:
@DecathectZero unfortunately this isn't the right spot for the includes (you can see the error in the failing test). You should put it inside the "get groupings" method, not after it.

PR #3046, opened 2017-06-06
Enabling adding new template division through ajax

jenachoi:


david-yz-liu:
The form needs to be tied to a concrete exam template (i.e., `member` not `collection`)

jenachoi:
I temporarily made it for only exam_template[0] since we only had rendered exam_template[0], but I am currently changing it for each template. (member)

david-yz-liu:
Right, so the point is the button shouldn't be a "heading link" but instead a separate link in the form itself.

jenachoi:
okay!

app/controllers/exam_templates_controller.rb
```diff
@@ -73,6 +73,38 @@ def update
...
+    redirect_to action: 'index'
+  end
+
+  # Dialog to create template division.
+  def create_template_division_dialog
+    @assignment = Assignment.find(params[:assignment_id])
+    @exam_template = @assignment.exam_templates.find_by(id: params[:id])
+
+    render partial: 'exam_templates/create_division_dialog',
+           formats: [:js], handlers: [:erb]
```
jenachoi: I needed to specify formats and handlers this time!

app/controllers/exam_templates_controller.rb
```diff
@@ -73,6 +73,37 @@ def update
     redirect_to action: 'index'
   end
 
+  def create_template_division
+    assignment = Assignment.find(params[:assignment_id])
+    template = assignment.exam_templates.find_by(id: params[:id])
```
david-yz-liu: Use `find`, not `find_by`, when searching by id.

app/controllers/exam_templates_controller.rb
```diff
@@ -73,6 +73,37 @@ def update
...
+    division_start = params[:create_division][:start]
+    division_end = params[:create_division][:end]
+    division_label = params[:create_division][:label]
+
+    new_template_division = template.template_divisions.create_with_associations(
```
david-yz-liu: I believe `create_*` saves already. You might want `new_with...` or `initialize_with...` But also I'm not sure why the association is necessary?

app/controllers/exam_templates_controller.rb
```diff
@@ -73,6 +73,37 @@ def update
...
+    if new_template_division.save
+      flash_message(:success, t('template_divisions.create.success'))
+    else
+      flash_message(:error, t('template_divisions.create.failure'))
+    end
+    redirect_to action: 'index'
+  end
+
+  # Dialog to create template division.
+  def create_template_division_dialog
```
david-yz-liu: If you rename this method and route to `create_division_dialog`, you won't have to include the explicit render at the bottom.

app/views/exam_templates/_create_new_division.html.erb
```diff
@@ -0,0 +1,19 @@
+<%= form_for :create_division,
+             url: create_template_division_assignment_exam_template_path(@assignment, id: @exam_template.id),
+             html: { multipart: true, size: 1 } do |f| %>
+  <h2><%= I18n.t('template_divisions.create.add_new') %></h2>
```
david-yz-liu: Use just `t`, not `I18n.t`

jenachoi:
I'll continue working on it tomorrow- For now, I have made changes according to the requests you've made.

jenachoi:
not finished yet!

jenachoi:
it finally works. Feedback would be appreciated!

jenachoi:
I have checked thoroughly on this PR. It correctly adds template division to corresponding exam template now. The only problem I have noticed while testing this is that it doesn't allow for more than one template division with same 'label' attribute. (even if those template divisions belong to different exam template / assignment). If it is okay, I will work on this on separate PR.

app/views/exam_templates/_boot.js.erb
```diff
@@ -1,3 +1,57 @@
...
+    });
 });
+
+function add_template_division(id) {
+    var new_id = new Date().getTime();
+    var input_id = 'exam_template_template_divisions_attributes' + new_id;
+    var new_division_row = `
+        <li id="${input_id}_holder" class="new">
+          <span class="label">
+            <input type="text" required="required" id="new_label">
```
david-yz-liu: Don't use `id` here; use the `name` attribute with values `label`, `start`, and `end`, and the form should just work (I think).

app/views/exam_templates/_form.html.erb
```diff
@@ -1,5 +1,14 @@
+<% content_for :head do %>
+    <script>
+        <%= render partial: 'boot',
+                   formats: [:js],
```
david-yz-liu: You can remove the `formats` (?)

app/views/exam_templates/_form.html.erb
```diff
@@ -1,5 +1,14 @@
...
+                   locals: { exam_template: exam_template } %>
+    </script>
+<% end %>
+
 <%= form_for [@assignment, exam_template],
-             html: { multipart: true } do |f| %>
+             html: { multipart: true,
+                     class: 'form_' + exam_template.id.to_s} do |f| %>
```
david-yz-liu: Use ids here, since these are unique.

jenachoi:
I made requested changes.

jenachoi:
The issue I previously mentioned is now fixed.

jenachoi:
It should work well with nested form if we test it with #3070

app/views/exam_templates/_boot.js.erb
```diff
@@ -1,3 +1,32 @@
...
   window.modal_create_new   = new ModalMarkus('#create_new_template');
 });
+
+function add_template_division(id) {
+    var new_id = new Date().getTime();
+    var input_id = 'exam_template_template_divisions_attributes' + new_id;
+    var new_division_row = `
+        <li id="${input_id}_holder" class="new">
+          <span class="label">
+            <input type="text" required="required" name="exam_template[template_divisions_attributes][${new_id}][label]">
```
david-yz-liu: Create (and use) a variable storing `exam_template[template_divisions_attributes][${new_id}]` to avoid the repetition.

app/views/exam_templates/_boot.js.erb
```diff
@@ -1,3 +1,32 @@
 $(document).ready(function() {
   window.modal_create_new   = new ModalMarkus('#create_new_template');
 });
+
+function add_template_division(id) {
+    var new_id = new Date().getTime();
```
david-yz-liu: In JavaScript and HTML files, use 2 spaces for indentation.

app/views/exam_templates/_boot.js.erb
```diff
@@ -1,3 +1,32 @@
...
+            <input type="text" required="required" name="exam_template[template_divisions_attributes][${new_id}][label]">
+          </span>
+          <span class="start">
+            <input type="number" required="required" name="exam_template[template_divisions_attributes][${new_id}][start]">
+          </span>
+          <span class="end">
+            <input type="number" required="required" name="exam_template[template_divisions_attributes][${new_id}][end]">
+          </span>
+          <span class="delete">
+            <a onclick="remove_template_division(\'${input_id}\'); return false;">
```
david-yz-liu: Eh, I think you can inline the function here. Something like `$(this).closest(li).remove()` should work.

app/views/exam_templates/_form.html.erb
```diff
@@ -33,11 +41,18 @@
 
   <div class="exam-template-section">
     <h3><%= ExamTemplate.human_attribute_name('template_divisions') %></h3>
+    <p class="alignright">
+      <%= link_to 'Add Template Division',
+                  '#',
+                  onClick: 'add_template_division(' + exam_template.id.to_s + '); return false;' %>
```
david-yz-liu: When in Ruby code, always prefer string interpolation (with double-quotes, and use `#{...}`).

PR #3047, opened 2017-06-06
fix message for filename

jenachoi:


config/locales/models/exam_templates/en.yml
```diff
@@ -5,7 +5,7 @@ en:
         one: "Exam Template"
     attributes:
       exam_template:
-        filename: "Name of file"
+        filename: "filename"
```
david-yz-liu: Should this be capitalized?

PR #3050, opened 2017-06-07
generate copies message fix

jenachoi:
removed semi-colon


david-yz-liu:
Note: you removed colons, not semi-colons.

PR #3051, opened 2017-06-07
Autocollects scanned exams after saving pages

DecathectZero:


david-yz-liu:
This is the right approach - however, I'm a little worried about scalability. It would be great to collect all the groupings in an array, and call `SubmissionsJob.perform` on all of them at once.

PR #3053, opened 2017-06-08
Splits the PDF to Cover/Extra

DecathectZero:


app/models/exam_template.rb
```diff
@@ -221,13 +221,21 @@ def save_pages(partial_exams)
             division.start <= page_num && page_num <= division.end
           end
         end
+        extra_pages.sort_by! { |page_num, _| page_num }
         extra_pdf = CombinePDF.new
-        extra_pdf << extra_pages.collect { |_, page| page }
+        cover_pdf = CombinePDF.new
+        cover_pdf << extra_pages[0][1]
```
david-yz-liu: You need to check whether the first page here is actually 1 or not. (Could be that the first page was missing.)

app/models/exam_template.rb
```diff
@@ -221,8 +221,20 @@ def save_pages(partial_exams)
...
           end
         end
+        extra_pages.sort_by! { |page_num, _| page_num }
         extra_pdf = CombinePDF.new
-        extra_pdf << extra_pages.collect { |_, page| page }
+        cover_pdf = CombinePDF.new
+        start_page = 0
+        if extra_pages[0][0] == 1
+          cover_pdf << extra_pages[0][1]
+          txn.add(File.join(assignment_folder,
```
david-yz-liu: Error-handling is a little tricky. If we only save a `COVER.pdf` file when it's there, there will be an error raised in the Assign Scans page. So maybe it's better to save an "empty PDF" (if that's even possible) so it always exists?

PR #3055, opened 2017-06-08
Exam templates split pdf

rathukugan:


app/controllers/exam_templates_controller.rb
```diff
@@ -87,6 +87,23 @@ def generate
...
+    assignment = Assignment.find(params[:assignment_id])
+    exam_template = assignment.exam_templates.find(params[:id])
+    split_exam = params[:exam_template][:pdf_to_split]
+    unless split_exam.nil?
+      if split_exam.content_type != 'application/pdf'
+        flash_message(:error, t('exam_templates.split.invalid'))
+      else
+        exam_template.split_pdf(split_exam.path)
+        flash_message(:success,t('exam_templates.split.success'))
+        redirect_to action: 'index'
```
david-yz-liu: In general, if you *don't* have a view whose name matches a controller method (which is the case here), then every control flow path in a controller method must have a `redirect_to` or `render`. So you should put the `redirect_to` here at the bottom of this method, to ensure that it is always called (even in the case of errors).

PR #3070, opened 2017-06-14
setting default associated criteria_assignment_files_join when emplate division is created or updated

jenachoi:
Updating nested attribute 'criteria_assginement_files_join' is necessary whenever template division is created or updated. This fixes the problem I had in #3046 where removing template division caused error.

### DIFF
--- a/app/models/template_division.rb
+++ b/app/models/template_division.rb
@@ -31,7 +31,6 @@ class TemplateDivision < ActiveRecord::Base
     criterion.destroy
   end
 
-
   def set_defaults_for_associated_criteria_assignment_files_join
     if criteria_assignment_files_join.nil?
       assignment_file = AssignmentFile.find_or_initialize_by(
@@ -45,7 +44,7 @@ class TemplateDivision < ActiveRecord::Base
       if criterion.new_record?
         criterion.update(max_mark: 1.0)
       end
-      criteria_assignment_files_join_object = CriteriaAssignmentFilesJoin.create(
+      criteria_assignment_files_join_object = CriteriaAssignmentFilesJoin.new(
         assignment_file: assignment_file,
         criterion: criterion
       )
--- a/app/models/template_division.rb
+++ b/app/models/template_division.rb
@@ -11,7 +11,7 @@ class TemplateDivision < ActiveRecord::Base
   validate :end_should_be_less_than_or_equal_to_num_pages
   validates :label, uniqueness: true, allow_blank: false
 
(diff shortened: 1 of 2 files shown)


app/models/template_division.rb
```diff
@@ -46,4 +31,30 @@ def destroy_associated_objects
...
+        assignment_id: exam_template.assignment.id
+      )
+      criteria_assignment_files_join_object = CriteriaAssignmentFilesJoin.create(
+        assignment_file: assignment_file,
+        criterion: criterion
+      )
+      self.update(criteria_assignment_files_join: criteria_assignment_files_join_object)
+    else
+      criteria_assignment_files_join.assignment_file.filename = "#{exam_template.name}-#{label}.pdf"
+      criteria_assignment_files_join.assignment_file.assignment_id = exam_template.assignment.id
```
david-yz-liu: The `assignment_file` should not need to have its `assignment_id` changed.

david-yz-liu: And same for the `criterion` as well.

app/models/template_division.rb
```diff
@@ -46,4 +31,30 @@ def destroy_associated_objects
...
+
+  def set_defaults_for_associated_criteria_assignment_files_join
+    if criteria_assignment_files_join.nil?
+      assignment_file = AssignmentFile.create(
+        filename: "#{exam_template.name}-#{label}.pdf",
+        assignment_id: exam_template.assignment.id
+      )
+      criterion = FlexibleCriterion.create(
+        max_mark: 1.0,
+        name: "#{exam_template.name}-#{label}",
```
david-yz-liu: This isn't what we want. The criterion should only have the `label` as its name; this will allow different exam templates to use the same criteria. Here you can use `find_or_initialize_by` with an assignment id; if it's new, set the max_mark to 1.

app/models/template_division.rb
```diff
@@ -46,4 +31,30 @@ def destroy_associated_objects
...
+      self.update(criteria_assignment_files_join: criteria_assignment_files_join_object)
+    else
+      criteria_assignment_files_join.assignment_file.filename = "#{exam_template.name}-#{label}.pdf"
+      criteria_assignment_files_join.assignment_file.assignment_id = exam_template.assignment.id
+      criteria_assignment_files_join.criterion.name = "#{exam_template.name}-#{label}"
+      criteria_assignment_files_join.criterion.assignment_id = exam_template.assignment.id
+      criteria_assignment_files_join.save!
+    end
+  end
+
```
david-yz-liu: Remove this blank line.

PR #3074, opened 2017-06-15
create an instance of SplitPdfLog when split_pdf is run

jenachoi:


app/jobs/split_pdf_job.rb
```diff
@@ -137,4 +153,16 @@ def save_pages(exam_template, partial_exams)
...
+  def get_num_groups_in_dir(dir)
+    num_groups_in_dir = 0
+    if Dir.exists?(dir)
+      Dir.foreach(dir) do |filename|
+        if File.directory?(File.join(dir, filename)) && !filename.start_with?('.')
+          num_groups_in_dir += 1
+        end
+      end
+    end
+    return num_groups_in_dir
```
david-yz-liu: You don't need the `return` here.

lib/tasks/scanned_exam.rake
```diff
@@ -40,6 +40,6 @@ namespace :db do
 
     template.generate_copies 5
     template_path  = File.join(File.dirname(__FILE__), '/../../db/data/scanned_exams/midterm1_scanned_sample.pdf')
-    template.split_pdf template_path
+    template.split_pdf(template_path, current_user)
```
david-yz-liu: It doesn't seem like you changed the implementation of `split_pdf` in this branch. You should revert this change.

app/jobs/split_pdf_job.rb
```diff
@@ -85,50 +102,51 @@ def save_pages(exam_template, partial_exams)
 
       group.access_repo do |repo|
         assignment_folder = exam_template.assignment.repository_folder
-        txn = repo.get_transaction(Admin.first.user_name)
-
+        unless Admin.first.nil?
```
jenachoi: The only change I made from line 105 - 148 is putting existing codes inside unless Admin.first.nil? in order to handle error that occurs when Admin.first is nil.

david-yz-liu: Don't use a block here. Do a one-liner, something like `return if Admin.empty?` (I believe `.empty?` or something like it can be used here too).

app/jobs/split_pdf_job.rb
```diff
@@ -35,14 +36,46 @@ def perform(exam_template, path)
...
+
+      # creating an instance of split_pdf_log
+      filename = original_filename.nil? ? File.basename(path) : original_filename
+      num_pages = pdf.pages.length
+      complete_dir = File.join(exam_template.base_path, 'complete')
+      incomplete_dir = File.join(exam_template.base_path, 'incomplete')
+      num_groups_in_complete = get_num_groups_in_dir(complete_dir)
+      num_groups_in_incomplete = get_num_groups_in_dir(incomplete_dir)
+
+      success = !partial_exams.empty?
```
david-yz-liu: This variable is redundant. You should remove it from here and from the schema. Please do not make any schema changes without consulting me first, as you did for the exam template id.

app/jobs/split_pdf_job.rb
```diff
@@ -85,9 +118,9 @@ def save_pages(exam_template, partial_exams)
 
       group.access_repo do |repo|
         assignment_folder = exam_template.assignment.repository_folder
+        return if Admin.first.nil?
```
david-yz-liu: Use https://apidock.com/rails/ActiveRecord/Relation/empty%3F.

.travis.yml
```diff
@@ -4,6 +4,8 @@ rvm:
 
 before_install:
   - sudo apt-get install -y cmake libssh2-1-dev
+  - sudo apt-get install -y ghostscript
```
david-yz-liu: Put all dependencies on the same line (i.e., merge with line 6).

app/jobs/split_pdf_job.rb
jenachoi: Admin.empty? causes *** NoMethodError Exception: undefined method `empty?' for #<Class:0x00000007eb5dc8> so I used Admin.blank? instead.

PR #3089, opened 2017-06-22
Create AssignmentFile after template division instead of CriteriaAssignmentFilesJoin & schema change

jenachoi:


david-yz-liu:
@jenachoi this is okay, although it seems to contain at least one unnecessary change (the whitespace in filenames) and there should be only one migration file, not two. Please create a new branch off of master.

PR #3091, opened 2017-06-23
Generate progress bar

rathukugan:


app/views/exam_templates/_poll_generate_job.js.erb
```diff
@@ -0,0 +1,4 @@
+function download_generate_pdf() {
+    window.location="<%= download_generate_assignment_exam_template_path(file_path: file_path) %>";
+}
+poll_job("<%= job_id %>", undefined, download_generate_pdf);
```
david-yz-liu: Don't define a separate function; instead, use an *anonymous* function here.

app/controllers/exam_templates_controller.rb
```diff
@@ -82,12 +82,21 @@ def generate
...
+    respond_to do |format|
+      format.js { render 'exam_templates/_poll_generate_job.js.erb',
+                         locals: { file_path: "#{exam_template.base_path}/#{generated_filename}",
+                                   job_id: current_job.job_id} }
+    end
+  end
+
+  def download_generate
+    send_file(params[:file_path],
+              filename: File.basename(params[:file_path]),
```
david-yz-liu: This isn't great because it gives arbitrary access to the server's files. Rather than specifying the full path, just send back the key elements -- exam template id and generated file name -- and turn those into query parameters for the URL, and do a lookup here to compute the correct URL.

PR #3092, opened 2017-06-23
Template division change & Create AssignmentFile after template division instead of join table

jenachoi:


app/models/assignment_file.rb
```diff
@@ -2,6 +2,7 @@ class AssignmentFile < ActiveRecord::Base
 
   belongs_to :assignment
   has_many :criteria_assignment_files_joins, dependent: :destroy
+  has_one :template_division
```
david-yz-liu: This isn't correct. It could have many template divisions.

PR #3094, opened 2017-06-26
Criterion chooses assignment file

jenachoi:


app/controllers/criteria_controller.rb
```diff
@@ -49,15 +50,29 @@ def destroy
...
+      properly_updated = @criterion.update(flexible_criterion_params.except(:assignment_files))
+      assignment_file = AssignmentFile.find(flexible_criterion_params[:assignment_files].to_i)
     else
-      properly_updated = @criterion.update(checkbox_criterion_params)
+      properly_updated = @criterion.update(checkbox_criterion_params.except(:assignment_files))
+      assignment_file = AssignmentFile.find(checkbox_criterion_params[:assignment_files].to_i)
+    end
+    # delete old associated criteria_assignment_files_join
+    old_criteria_assignment_files_join = @criterion.criteria_assignment_files_joins
+    if old_criteria_assignment_files_join.exists?
```
david-yz-liu: It isn't necessary to make this check.

app/controllers/criteria_controller.rb
```diff
@@ -36,6 +36,7 @@ def create
   end
 
   def edit
+    @assignment = Assignment.find(params[:assignment_id])
```
david-yz-liu: This is not necessary. You can access `@criterion.assignment` through the assignment. In general you should aim to have few instance variables.

app/controllers/criteria_controller.rb
```diff
@@ -49,14 +50,36 @@ def destroy
...
 
   def update
+    @assignment = Assignment.find(params[:assignment_id])
     criterion_type = params[:criterion_type]
     @criterion = criterion_type.constantize.find(params[:id])
     if criterion_type == 'RubricCriterion'
-      properly_updated = @criterion.update(rubric_criterion_params)
+      properly_updated = @criterion.update(rubric_criterion_params.except(:assignment_files))
+      unless rubric_criterion_params[:assignment_files].nil?
+        assignment_files = rubric_criterion_params[:assignment_files].map { |id| AssignmentFile.find_by(id: id.to_i) }
```
david-yz-liu: 1. Use `find` when searching by id.
2. You can pass `find` a list of ids.

app/controllers/criteria_controller.rb
```diff
@@ -49,14 +50,36 @@ def destroy
...
+    end
+    # delete old associated criteria_assignment_files_join
+    old_criteria_assignment_files_join = @criterion.criteria_assignment_files_joins
+    old_criteria_assignment_files_join.destroy_all
+    # create new corresponding criteria_assignment_files_join
+    assignment_files.to_a.each do |assignment_file|
+      unless assignment_file.nil?
+        @criterion.criteria_assignment_files_joins.create(
+          assignment_file: assignment_file,
+          criterion: @criterion
```
david-yz-liu: You don't need to pass in the criterion because you're accessing the association from `@criterion` already.

app/controllers/criteria_controller.rb
```diff
@@ -49,14 +50,36 @@ def destroy
...
+      unless checkbox_criterion_params[:assignment_files].nil?
+        assignment_files = checkbox_criterion_params[:assignment_files].map { |id| AssignmentFile.find_by(id: id.to_i) }
+      end
+    end
+    # delete old associated criteria_assignment_files_join
+    old_criteria_assignment_files_join = @criterion.criteria_assignment_files_joins
+    old_criteria_assignment_files_join.destroy_all
+    # create new corresponding criteria_assignment_files_join
+    assignment_files.to_a.each do |assignment_file|
+      unless assignment_file.nil?
```
david-yz-liu: You can remove this check if you do what I said above.

app/views/checkbox_criteria/_criterion_editor.html.erb
```diff
@@ -39,6 +39,14 @@
       <%= f.text_field :max_mark %>
     </span>
     <br>
+    <span id="<%= "criterion_inputs_#{criterion.id}" %>">
+      <% selected_files = criterion.criteria_assignment_files_joins.exists? ? criterion.criteria_assignment_files_joins.map { |join| join.assignment_file }: [] %>
```
david-yz-liu: The check is unnecessary. Calling `map` on `criterion.criteria_assignment_files_joins` even when it's empty should be fine.

jenachoi:
It causes RecordNotFound error when I use `find` in criteria_controller.rb:57 because for some reason params[:assignment_files] contains empty string as an element. So I used `where` for finding grabbing corresponding assignment files.

db/migrate/20170623163309_replace_join_table_with_assignment_file_in_template_divisions.rb
```diff
@@ -1,8 +1,8 @@
 class ReplaceJoinTableWithAssignmentFileInTemplateDivisions < ActiveRecord::Migration
   def change
     # Remove criteria_assignment_files_join column from TemplateDivision
-    remove_reference :template_divisions, :criteria_assignment_files_join, foreign_key: true
+    remove_reference :template_divisions, :criteria_assignment_files_join
```
david-yz-liu: Don't change migration files after they've happened. Also the foreign key is appropriate here. If there's an error elsewhere, you need to fix the problem there, not by changing the schema.

david-yz-liu:
@jenachoi would be good to update the seed file to include an assignment with required files, and associated criteria.

PR #3095, opened 2017-06-26
Split progress bar

rathukugan:


app/views/exam_templates/view_logs.html.erb
```diff
@@ -1,5 +1,11 @@
 <% content_for :title, "#{@assignment.short_identifier} #{SplitPdfLog.model_name.human.pluralize}" %>
 
+<% if flash[:split] %>
```
david-yz-liu: This is what I meant by "good basic idea, but use `session` instead of `flash`."

app/views/exam_templates/view_logs.html.erb
```diff
@@ -1,5 +1,11 @@
...
+    <script type="text/javascript">
```
david-yz-liu: Use 2 spaces for indentation.

app/views/exam_templates/view_logs.html.erb
```diff
@@ -1,5 +1,11 @@
...
+        poll_job("<%= flash[:split] %>", undefined, function() {window.location.reload()});
```
david-yz-liu: You can just pass in `window.location.reload` (without calling it); you don't need to wrap it in an anonymous function.

app/controllers/exam_templates_controller.rb
```diff
@@ -109,14 +109,18 @@ def split
     unless split_exam.nil?
       if split_exam.content_type != 'application/pdf'
         flash_message(:error, t('exam_templates.split.invalid'))
+        redirect_to action: 'index'
       else
-        exam_template.split_pdf(split_exam.path, split_exam.original_filename, @current_user)
-        flash_message(:success,t('exam_templates.split.success'))
+        current_job = exam_template.split_pdf(split_exam.path, split_exam.original_filename, @current_user)
+        session[:job_id] = current_job.job_id
+        respond_to do |format|
```
david-yz-liu: You don't need the `respond_to` here.

PR #3097, opened 2017-06-27
check whether instructor uploaded right exam template in split pdf (with QR code)

jenachoi:


app/jobs/split_pdf_job.rb
```diff
@@ -38,8 +38,13 @@ def perform(exam_template, path, original_filename=nil, current_user=nil)
           new_page.save File.join(error_dir, "#{basename}-#{i}.pdf")
           num_pages_qr_scan_error += 1
         else
-          partial_exams[m[:exam_num]] << [m[:page_num].to_i, page]
-          m_logger.log("#{m[:short_id]}: exam number #{m[:exam_num]}, page #{m[:page_num]}")
+          if m[:short_id] == exam_template.assignment.short_identifier # if QR code contains corresponding exam template
```
david-yz-liu: This is also not good. It should be compared to the exam template's name.

david-yz-liu:
@jenachoi I merged this in, but the `SplitPdfLog` model should be updated to keep track of these pages as well. (Otherwise the log displays that pages were scanned, but doesn't explain why corresponding groups were not created.)

PR #3099, opened 2017-06-27
Improve on summaries table view

jenachoi:
Haven't thoroughly tested it out yet, but it seems to work based on what I've tested so far. Will do more work on it

### DIFF
--- a/app/controllers/summaries_controller.rb
+++ b/app/controllers/summaries_controller.rb
@@ -1,6 +1,5 @@
 class SummariesController < ApplicationController
   include SummariesHelper
-  include SubmissionsHelper
 
   before_filter  :authorize_for_ta_and_admin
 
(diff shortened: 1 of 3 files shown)


app/views/summaries/_summaries_table.js.jsx.erb
```diff
@@ -206,6 +185,10 @@
         return s;
       }.bind(this));
 
+      if (render_bonus_deductions == false) {
```
david-yz-liu: In general you shouldn't mutate `props` of a React component. Don't do this check here; instead, do it in the server code, similar to how the `sections` column works.

app/controllers/summaries_controller.rb
```diff
@@ -1,5 +1,6 @@
 class SummariesController < ApplicationController
   include SummariesHelper
+  include SubmissionsHelper
```
david-yz-liu: I'm not sure why this was added, or in general why there are so many changes in this file. You can keep just sending `json`, and then on the Javascript side check the values in the JSON to decide whether to splice the column. I.e., don't use `@render_bonus_deductions_column`, check whether the `total_extra_points` key exists in at least one row.

PR #3101, opened 2017-06-29
check for whether it successfully splits PDF with multiple page_num shuffled in random order

jenachoi:


app/jobs/split_pdf_job.rb
```diff
@@ -104,7 +104,7 @@ def save_pages(exam_template, partial_exams)
 
       group.access_repo do |repo|
         assignment_folder = exam_template.assignment.repository_folder
-        return unless Admin.exists?
+        next unless Admin.exists?
```
david-yz-liu: This is a condition that doesn't depend on the loop, and so shouldn't be inside the loop at all. You can move this to the top of the function, and return immediately if there are no admins.

You can also modify your test so that an admin is created before the tests are run.

PR #3103, opened 2017-06-29
Added Fix Scan Errors

DecathectZero:


app/controllers/exam_templates_controller.rb
```diff
@@ -138,6 +138,36 @@ def view_logs
                                  .includes(:user)
   end
 
+  def assign_errors
+    @assignment = Assignment.find(params[:assignment_id])
+    @error_files = []
+    Dir.foreach(File.join(
+      MarkusConfigurator.markus_exam_template_dir,
+      @assignment.short_identifier, 'error'
+    )) { |file|
```
david-yz-liu: Use `do` syntax for multi-line blocks.

app/controllers/exam_templates_controller.rb
```diff
@@ -138,6 +138,36 @@ def view_logs
...
+      @error_files << file unless file =~ /^\.\.?$/
+    }
+    @error_files = @error_files.sort do |a, b|
```
david-yz-liu: Hmm. The comparison function you're using looks like the default, so I would think that just `@error_files.sort` is fine.

app/controllers/exam_templates_controller.rb
```diff
@@ -138,6 +138,36 @@ def view_logs
...
+      a <=> b
+    end
+  end
+
+  def download_error
+    @assignment = Assignment.find(params[:assignment_id])
+    send_file(File.join(MarkusConfigurator.markus_exam_template_dir,
+                        @assignment.short_identifier, 'error') +
+              '/' + params[:file_name],
```
david-yz-liu: Pass `params[:file_name]` into the `join` (that's the point of `File.join`.

app/views/exam_templates/assign_errors.html.erb
```diff
@@ -0,0 +1,279 @@
...
+                               'pdfjs',
+                               'testresults_config',
+                               'Results/context_menu' %>
+
+    <%= javascript_tag do -%>
+        PDFJS.workerSrc = "<%= @host %>/assets/pdfjs_lib/pdf.worker.js";
+    <% end -%>
+<% end %>
+
+<% # SourceCodeGlower, ImageAnnotations, and SyntaxHighlighter brushes %>
```
david-yz-liu: Yikes, did I miss in on Assign scans? You shouldn't need the syntax highlighter stuff in either view. (Remove it from this view in this branch, make a separate branch for the other view.)

app/views/exam_templates/assign_errors.html.erb
```diff
@@ -0,0 +1,279 @@
...
+                <div class="buttonRow">
+                  <button id="documentPropertiesClose" class="overlayButton"><span data-l10n-id="document_properties_close">Close</span></button>
+                </div>
+              </div>
+            </div>
+          </div>  <!-- overlayContainer -->
+        </div>
+      </div>
+    </div>
+    <script>
```
david-yz-liu: This script should go in a separate file and be rendered as a partial.

app/views/exam_templates/assign_errors.html.erb
```diff
@@ -0,0 +1,279 @@
...
+    <div id='drag'></div>
+
+    <div id='right-pane' class='input_pane'>
+      <div class='ui-tabs-panel flex-col'>
+        <div class="side_header">
+          <h2>Document Information</h2>
+        </div>
+        <form id="assign_error" class="side_header">
+          <div class="side_header">
+            <label class="form_label" for="exam_template">Exam Template: </label>
```
david-yz-liu: Use the internationalized name here. (Jenna can tell you if you don't remember)

app/views/exam_templates/assign_errors.html.erb
```diff
@@ -0,0 +1,279 @@
...
+            <select id="exam_template">
+              <!--<option disabled selected>Select Template</option>-->
+              <% ExamTemplate.where(assignment_id: @assignment.id).each do |e| %>
+                  <option><%= e.name %></option>
+              <% end %>
+            </select>
+          </div>
+          <br>
+          <input id="copy" name="names" type="text" placeholder="Copy Number">
```
david-yz-liu: The `name` attribute should be more meaningful. Also, it would be good to have a label tag for each of these. The placeholder text is long compared to the space that is required for the actual numbers.

app/views/exam_templates/assign_errors.html.erb
```diff
@@ -0,0 +1,279 @@
...
+
+    <div id='right-pane' class='input_pane'>
+      <div class='ui-tabs-panel flex-col'>
+        <div class="side_header">
+          <h2>Document Information</h2>
+        </div>
+        <form id="assign_error" class="side_header">
+          <div class="side_header">
+            <label class="form_label" for="exam_template">Exam Template: </label>
+            <select id="exam_template">
```
david-yz-liu: I'd like you to look into the Rails standard way of creating forms here (http://guides.rubyonrails.org/form_helpers.html). In particular, there's a nice `options_for_select` method here.

Also, would be good to automatically select an exam template if there's only one.

config/routes.rb
```diff
@@ -103,6 +103,9 @@
         end
 
         collection do
+          get 'download_error'
+          get 'download_error_path'
```
david-yz-liu: I imagine you were trying to use a path helper to generate the path. You should stick with `download_error_file` (just "error" by itself has different connotations).

For the path helper, make sure you're in Ruby mode (`<%= ... %>`) and it'll be long, something like `download_error_file_assignment_exam_templates_path(you need to pass in the assignment id)`.

app/views/exam_templates/assign_errors.html.erb
DecathectZero: No it wasn't on Assign scans, but unfortunately for me to get the PDF viewer toolbar/rotation working it required a AnnotationManager object to be defined. This was a temporary workaround.

app/views/exam_templates/assign_errors.html.erb
DecathectZero: Still working on the branch, I'll definitely internationalized everything before I finish up.

app/views/exam_templates/assign_errors.html.erb
DecathectZero: By default it just selects the first available one as the option, so it will definitely auto select if there is only one.

DecathectZero:
I still need to implement a checkbox button to reorientate the file in the case it is upside down.

app/views/exam_templates/view_logs.html.erb
```diff
@@ -2,8 +2,13 @@
 
 <% @heading_buttons = [
   { link_text: t('exam_templates.back_to_exam_template_page'),
-    link_path: assignment_exam_templates_path }
-] %>
+    link_path: assignment_exam_templates_path }]
+  if !Dir[File.join(MarkusConfigurator.markus_exam_template_dir,
```
david-yz-liu: Use `unless` instead of `if !`

app/views/exam_templates/_assign_errors.js.erb
```diff
@@ -0,0 +1,69 @@
+<script>
+  function rotate() {
+      PDFView.rotatePages(180);
+  }
+  PDFView.onLoadComplete = function() {
```
david-yz-liu: Can you try to remove this part and see if the pages still load properly? `PdfAnnotationManager` is a class that's used for grading, but not here.

app/views/exam_templates/_assign_errors.js.erb
```diff
@@ -0,0 +1,69 @@
...
+      annotation_manager = new PdfAnnotationManager(PDFView, 'viewer');
+      annotation_manager.resetAngle();
+  };
+  var index = 0;
+  var files = <%= raw @error_files %>;
+
+  function getFile(filename) {
+      checkLastFile();
```
david-yz-liu: A couple of style things for this file.

1. Use 2 spaces for indentation.
2. Use single-quotes, not double-quotes.
3. Use whitespace after `if`, after `)`, etc. (same comments as usual).

app/views/exam_templates/assign_errors.html.erb
```diff
@@ -0,0 +1,231 @@
...
+<% content_for :title do %>
+    <%= t('exam_templates.assign_errors.fix_errors') %>
+    <span class='title-help assignment_groups_help'>
+      <p class='help-message-title assignment_groups_help'>
+        <%= t('assignment.help.assign_scans') %>
+      </p>
+    </span>
+<% end %>
+
+<div id='criterion_incomplete_error' class='error' style='display:none'></div>
```
david-yz-liu: You can delete this. (And for the `assign_scans` page too, in a separate PR)

app/views/exam_templates/assign_errors.html.erb
```diff
@@ -0,0 +1,231 @@
...
+
+<div id='student_selector'>
+  <span class='left'>
+      <a id='previous_file' onclick='previousFile(); return false'><%= t('marker.previous_submission') %></a>
```
david-yz-liu: Please keep indentation consistent at 2 spaces.

app/views/exam_templates/assign_errors.html.erb
```diff
@@ -0,0 +1,231 @@
...
+    <div id='drag'></div>
+
+    <div id='right-pane' class='input_pane'>
+      <div class='ui-tabs-panel flex-col'>
+        <div class="side_header">
+          <h2><%= t('exam_templates.assign_errors.document_info') %></h2>
+        </div>
+        <%= form_tag(nil, id: 'info_form', class:'side_header') do %>
+          <div class="side_header">
+            <%= label_tag(:exam_template, ExamTemplate.model_name.human + ':') %>
```
david-yz-liu: Please remove the colons; I'm trying to not use them for input labels any more.

app/views/exam_templates/assign_errors.html.erb
```diff
@@ -0,0 +1,231 @@
...
+            <%= select_tag(:exam_template,
+                           options_from_collection_for_select(ExamTemplate.where(assignment_id: @assignment.id), :id, :name))
```
david-yz-liu: Use an association: `@assignment.exam_templates`

PR #3107, opened 2017-06-30
added fix_error function in exam template model

jenachoi:


david-yz-liu:
@jenachoi you're missing the point of what I asked you to do, which is not merge your work with @DecathectZero's, but instead create a separate model method that correctly handles the addition of a new page. In particular, your `fix_error` method currently only adds a new page to the directory; it doesn't do any of the creation of submission files that we discussed. Not only should you do this, but you should also revert the merge that you did.

jenachoi:
Oh I see! Thank you for letting me know, I will make changes on them tomorrow!

PR #3109, opened 2017-07-05
Arrow keys to switch submissions

DecathectZero:
I had to switch the unicode for the arrow keys because apparently for the current style, right arrow unicode symbols do not exist? It was a bit odd.


david-yz-liu:
@DecathectZero don't forget to change the keyboard shortcut help window too.

PR #3111, opened 2017-07-05
add fix_error function

jenachoi:
Sorry it took some time. I tested the function with the view and controller I have written by mistake and it seems to work well. Feedback would be appreciated!

### DIFF
--- a/app/models/exam_template.rb
+++ b/app/models/exam_template.rb
@@ -104,50 +104,70 @@ class ExamTemplate < ActiveRecord::Base
     error_file = File.join(
       base_path, 'error', filename
     )
-    if File.exists? error_file
-      complete_dir = File.join(
-        base_path, 'complete', exam_num
-      )
-      incomplete_dir = File.join(
-        base_path, 'incomplete', exam_num
-      )
-      # if file is missing from complete group
-      unless File.exists? File.join(complete_dir, page_num)
-        # if there isn't corresponding file in incomplete group
-        unless File.exists? File.join(incomplete_dir, page_num)
-          # if incomplete directory doesn't exist yet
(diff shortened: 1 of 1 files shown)


app/models/exam_template.rb
```diff
@@ -100,6 +100,59 @@ def split_pdf(path, original_filename=nil, current_user=nil)
...
+              (template_division.start..template_division.end).each do |i|
+                path = File.join(incomplete_dir, "#{i}.pdf")
+                pdf = CombinePDF.load path
+                submission_file << pdf.pages[0]
+              end
+              txn.replace(
+                File.join(assignment_folder, "#{template_division.label}.pdf"),
+                submission_file.to_pdf,
+                'application/pdf',
+                repo.get_latest_revision().revision_identifier
```
david-yz-liu: I'm not sure whether this argument is necessary. You should look into the `SubmissionsController` and see how student files are replaced when they upload them.

app/models/exam_template.rb
```diff
@@ -100,6 +100,59 @@ def split_pdf(path, original_filename=nil, current_user=nil)
     SplitPDFJob.perform_later(self, path, original_filename, current_user)
   end
 
+  def fix_error(filename, exam_num, page_num)
+    error_file = File.join(
+      base_path, 'error', filename
+    )
+    if File.exists? error_file
```
david-yz-liu: This type of error checking is best done with early returns. E.g., `return unless File.exists? error_file`. This prevents the extreme nesting that we currently have here.

david-yz-liu:
@jenachoi okay you can attach this method to the form now, as both branches have been merged in.

PR #3114, opened 2017-07-06
scale marks when max_mark is updated

jenachoi:


david-yz-liu:
I'm not sure why you didn't just put this into the `Criterion` class to avoid the duplication.

Also the query doesn't need to be this complicated. First, access `self.marks` (or better, just `marks`) and scale them all. Then, call `update_total_mark` on *all* of the results for the current assignment.

jenachoi:
I've been working on fixing rspec errors that I had. I had to revert back to the previous self.marks query inside results query in order to fix the errors.

app/models/mark.rb
```diff
@@ -37,6 +37,11 @@ def valid_mark
     end
   end
 
+  def scale_mark(curr_max_mark, prev_max_mark)
+    new_mark = self.mark * (curr_max_mark.to_f / prev_max_mark)
```
david-yz-liu: This actually needs to be implemented separately for each criterion type (this helps avoid potential floating point errors).
Note that for a Rubric criterion, the value of `(mark / max_mark) * 4` should always be an *integer* between 0 and 4, and for a Checkbox criterion, the value of `mark` should always be either 0 or `max_mark`.

The logic you've written here is only appropriate for Flexible Criterion (where an arbitrary scaling is necessary). However, in this case you should round the mark to two decimal places.

Finally, for all three criterion types, you should do a check for `prev_max_mark` being 0; in this case, no scaling should occur.

app/models/mark.rb
```diff
@@ -37,6 +37,19 @@ def valid_mark
     end
   end
 
+  def scale_mark(curr_max_mark, prev_max_mark)
+    return if prev_max_mark == 0 || mark == 0 # no scaling occurs if prev_max_mark is 0 or mark is 0
+    if markable.is_a? RubricCriterion
+      new_mark = mark * (curr_max_mark / prev_max_mark)
```
david-yz-liu: Round the new mark to two decimal places.

jenachoi: Changed it-

david-yz-liu:
@jenachoi please write some tests for this change as well.

PR #3117, opened 2017-07-07
Job class status

rathukugan:
- Job class was added to split pdf job's status, and added internationalized strings for job classes.

Note:
Instead of adding the job class to the session I figured it's easier if I updated the status before the job is queued (so I also avoid the nil class error when calling it on the JS poll_job side).

Also I need to make similar changes to the generate job / other jobs, which I can do in a separate pull request.


david-yz-liu:
@rathukugan now that the infrastructure is in place, choose a better progress message for the job than "x out of y" -- the latter can remain as a default, should in general shouldn't be used.

Also make use of the other flash keys (besides just `:success`) when sending back the status.

PR #3118, opened 2017-07-07
keeping track of which pages have QR code with wrong exam template in SplitPDFLog

jenachoi:
screen shot 2017-07-07 at 3 56 20 pm


david-yz-liu:
I'll have you support page-level granularity of types of errors later.

PR #3124, opened 2017-07-11
Better Progress messages

rathukugan:
Added better progress messages for split/generate jobs and changed flash keys for status updates
Ex:
"Processing page 1 out of 40 for midterm1 (1-5.pdf)"
"Generating copy 1 of 3 for midterm1"

### DIFF
--- a/app/controllers/job_messages_controller.rb
+++ b/app/controllers/job_messages_controller.rb
@@ -11,16 +11,15 @@ class JobMessagesController < ApplicationController
   def get
     status = ActiveJob::Status.get(params[:job_id])
     if status.working?
-      flash_now(:success, t('poll_job.working_message', progress: status[:progress], total: status[:total]))
+      flash_now(:notice, t('poll_job.working_message', progress: status[:progress], total: status[:total]))
     else
       if status.queued?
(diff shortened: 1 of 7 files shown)


david-yz-liu:
@rathukugan this is a good start, but the `job_suffix` handling is ugly. Instead, you should create a separate *method* for each job class called `show_status(status)` or something, when then calls the corresponding `t(...)` method. This way you'll be able to customize different status keys being passed to `t` for different classes.

app/controllers/job_messages_controller.rb
```diff
@@ -11,16 +11,19 @@ def show
...
-      flash_now(:success, t('poll_job.working_message', progress: status[:progress], total: status[:total]))
+      if status[:job_class]
+        flash_now(:notice, status[:job_class].show_status(status))
+      else #default x out of y message
+        flash_now(:notice, t('poll_job.working_message', progress: status[:progress], total: status[:total]))
+      end
     else
       if status.queued?
-        status_msg = t('poll_job.queued')
+        flash_now(:notice,t('poll_job.queued'))
```
david-yz-liu: @rathukugan please put in a space after `:notice`.

PR #3128, opened 2017-07-12
Instantiate SplitPage in split_pdf

jenachoi:


jenachoi:
working on it

app/jobs/split_pdf_job.rb
```diff
@@ -46,13 +46,18 @@ def perform(exam_template, path, original_filename=nil, current_user=nil)
         if m.nil?
           new_page.save File.join(error_dir, "#{filename}-#{i}.pdf")
           num_pages_qr_scan_error += 1
+          status = 'ERROR: QR code not found'
+          m_logger.log(status)
+          SplitPage.create(filename: filename, raw_page_number: i + 1, status: status)
```
david-yz-liu: You're jumping through hoops here because `SplitPdfLog` is currently created after the `SplitPage`. You should do this the other way around - create the `SplitPdfLog` at the very beginning, and then create the `SplitPage` objects as an association.

Create the groups earlier too. This way you can populate all the `SplitPage` fields and save them immediately, with no subsequent update.

app/models/split_page.rb
```diff
@@ -1,8 +1,9 @@
 class SplitPage < ActiveRecord::Base
   belongs_to :split_pdf_log
-  belongs_to :group, optional: true
-  validates :split_pdf_log, :filename, :raw_page_number, presence: true
+  belongs_to :group
```
david-yz-liu: No changes to this file should be made.

PR #3129, opened 2017-07-13
Git hooks

adisandro:
This pull request adds:
- conf option for git hooks
- support for multiple hooks
- hook to prevent top level changes to files/dirs
- handling of push time through the git reflog (which already stores the server push time if properly configured, no need to have a hook + database table)


adisandro:
Other than the hook for required files, this is ready

david-yz-liu:
@adisandro thanks, looks great.

PR #3130, opened 2017-07-13
Saved filename with id of SplitPage (in error/raw directory)

jenachoi:
Modified it so that filenames saved into `error` and `raw` directory to have raw page number of the uploaded file instead, (e.g. **40** from `midterm1_scanned_sample-40.pdf`) because we need ways to handle the case when uploaded file consists of **multiple exam versions** and situation as in line **108-109** happens. 

When I uploaded `midterm1_scanned_sample.pdf` from `db/data/scanned_exams` twice to split them, only `midterm1_scanned_sample-1.pdf` ~ `midterm1_scanned_sample-8.pdf` were saved in error directory because of the overlaps in page number. Now, it saves `midterm1_scanned_sample-1.pdf` ~ `midterm1_scanned_sample-40.pdf` in the directory.

**UPDATE**: we changed filename to contain id of SplitPage instead. 

The problem that we mentioned above is now fixed.


david-yz-liu:
@jenachoi it's a good problem to raise, but I'm not convinced of the solution. A better solution would be to use the id of the corresponding `SplitPage` instance, which is guaranteed to be unique. (This might simplify other things, too.)

jenachoi:
still working on it

app/jobs/split_pdf_job.rb
```diff
@@ -146,23 +143,24 @@ def save_pages(exam_template, partial_exams, filename=nil, split_pdf_log=nil)
...
+          filename: filename,
+          exam_page_number: exam_num.to_i,
+          raw_page_number: raw_page_num,
+          group: group,
+          split_pdf_log: split_pdf_log
+        )
         # if a page already exists, move the page to error directory instead of overwriting it
         if File.exists?(File.join(destination, "#{page_num}.pdf"))
-          new_pdf.save File.join(error_dir, "#{filename}-#{page_num}.pdf")
+          new_pdf.save File.join(error_dir, "#{filename}-#{split_page.id}.pdf")
```
david-yz-liu: You no longer need the `filename` here.

test/jobs/split_pdf_job_test.rb
```diff
@@ -56,12 +56,6 @@ class SplitPDFJobTest < ActiveJob::TestCase
       should 'have 1 pdf in incomplete directory' do
         assert_equal Dir.entries(@exam_template.base_path + '/incomplete/21/').sort, %w[. .. 1.pdf].sort
       end
-
-      should 'have 1 pdf in error directory' do
```
david-yz-liu: Deleting tests is often a bad sign. If there are changes that invalidate existing tests, you should work to modify them properly.

jenachoi:
going to work on adding test cases!

test/jobs/split_pdf_job_test.rb
```diff
@@ -6,9 +6,10 @@ class SplitPDFJobTest < ActiveJob::TestCase
   context 'split PDF job' do
     setup do
       Admin.make
-      @assignment = Assignment.make(short_identifier: 'midterm1')
-      f = File.open('db/data/scanned_exams/midterm1.pdf')
-      @exam_template = ExamTemplate.create_with_file(f.read, assignment_id: @assignment.id, filename: 'midterm1.pdf')
+      FileUtils.rm_rf(Dir.glob('data/dev/exam_templates/mdt'))
```
david-yz-liu: Don't do this here. Create the Assignment first, then do a `File.join` with the base path (which should be grabbed from the config file, actually) with the assignment short_identifier.

test/jobs/split_pdf_job_test.rb
```diff
@@ -57,10 +58,9 @@ class SplitPDFJobTest < ActiveJob::TestCase
         assert_equal Dir.entries(@exam_template.base_path + '/incomplete/21/').sort, %w[. .. 1.pdf].sort
       end
 
-      should 'have 1 pdf in error directory' do
-        error_dir = Dir.entries(@exam_template.base_path + '/error')
-        error_generated_files = ['midterm21-1.pdf']
-        assert_empty error_generated_files-error_dir
+      should 'have 2 pdf in error directory' do
```
david-yz-liu: Pluralize this.

PR #3132, opened 2017-07-13
Select groups to assign members

DecathectZero:
Also updated the get group code to allow cases where `COVER.pdf `does not exist.

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -119,9 +119,34 @@ class GroupsController < ApplicationController
 
   def assign_scans
     @assignment = Assignment.find(params[:assignment_id])
-    @current_grouping = Grouping.get_assign_scans_grouping(@assignment)
-    if @current_grouping.nil?
+    if params.has_key?(:group)
+      next_grouping = Grouping.find(params[:group])
+    else
+      next_grouping = Grouping.get_assign_scans_grouping(@assignment)
+    end
+    if next_grouping.nil?
       redirect_to(:back)
+    end
+    names = next_grouping.memberships.map do |u|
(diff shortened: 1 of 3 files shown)


app/views/groups/_groups_manager.js.jsx.erb
```diff
@@ -280,7 +280,11 @@
       var groups_data = this.props.groups.map(function(group) {
         var g = {};
         g['id'] = group.id;
+        <% if @assignment.scanned_exam %>
+        g['name'] = <a href={'groups/assign_scans?group=' + group.id}>{group.name}</a>;
```
david-yz-liu: Use a ruby path helper here.

app/controllers/groups_controller.rb
```diff
@@ -119,10 +119,35 @@ def index
 
   def assign_scans
     @assignment = Assignment.find(params[:assignment_id])
-    @current_grouping = Grouping.get_assign_scans_grouping(@assignment)
-    if @current_grouping.nil?
+    if params.has_key?(:group)
+      next_grouping = Grouping.find(params[:group])
```
david-yz-liu: Name this key `:grouping` rather than `:group`.

app/views/groups/_groups_manager.js.jsx.erb
DecathectZero: does a path helper still work if I need to input javascript variables like `group.id` and `group.name`?

david-yz-liu: @DecathectZero sorry I missed this one. You can use the `Routes` Javascript object to generate routes. Can't remember if you've done that before - https://github.com/railsware/js-routes#usage for examples.

PR #3133, opened 2017-07-13
fix error that occurs when exam template's name doesn't match its assignment's short_identifier

jenachoi:
I was testing by adding new exam template called `midterm1` under an assignment called `csc236-midterm1` and I noticed that even though correctly generated file was uploaded in `split_pdf`, pages were created in `error` directory because of the mistake made in my previous PR. Since QR code content is made with `#{exam_template.assignment.short_identifier}-#{exam_num}-#{page_num + 1}`, it should be compared to `exam_template.assignment.short_identifier`, not `exam_template.name`


david-yz-liu:
@jenachoi the QR code is the thing that should change - it should encode the name of the exam template.

PR #3136, opened 2017-07-14
update COVER.pdf and EXTRA.pdf when we fix error page

jenachoi:


app/models/exam_template.rb
```diff
@@ -155,6 +155,39 @@ def fix_error(filename, exam_num, page_num, upside_down)
...
+            pdf = CombinePDF.load path
+            cover_pdf << pdf.pages[0]
+          end
+          txn.replace(File.join(assignment_folder, "COVER.pdf"), cover_pdf.to_pdf,
+                      'application/pdf', revision.revision_identifier)
+        end
+
+        # update EXTRA.pdf if error page given doesn't belong to any template division
+        update_extra_pdf = self.template_divisions.all? { |division| division.start > page_num.to_i || division.end < page_num.to_i }
+        if update_extra_pdf
```
david-yz-liu: This check is unnecessary. As far as I know we do not do an analogous check when first creating `EXTRA.pdf`.

app/models/exam_template.rb
```diff
@@ -155,6 +155,37 @@ def fix_error(filename, exam_num, page_num, upside_down)
...
+
+        # update COVER.pdf if error page given is first page of exam template
+        if page_num.to_i == 1
+          path = File.join(incomplete_dir, "#{page_num}.pdf")
+          if File.exists? path
+            cover_pdf = CombinePDF.new
+            pdf = CombinePDF.load path
+            cover_pdf << pdf.pages[0]
+          end
+          txn.replace(File.join(assignment_folder, "COVER.pdf"), cover_pdf.to_pdf,
```
david-yz-liu: This needs to go inside the `if`. It uses a variable initialized in that block.

PR #3137, opened 2017-07-14
only updating submission file for corresponding error page

jenachoi:
Everytime we fix an error page, PDFs for every template division were replaced. Now it only updates one with corresponding page number that should be updated.


jenachoi:
oops it seems like I made a mistake in this PR, I will fix it after the meeting

PR #3138, opened 2017-07-17
add submission file in case new template division has been created

jenachoi:
If instructors add new template division after having scanned exams, it updates submission files based on template divisions in `fix_error` function next time they want to scan same exams again.


david-yz-liu:
You are not doing this part properly. But it also is not a good change. Instead, the server should reject any changes to template divisions once exams have been uploaded. That's a change you can make.

PR #3141, opened 2017-07-18
generate correct file after exam template file is changed

jenachoi:
When we upload new file for `exam template` and we generate QR code, it used to still generate on old file. Now, it generates QR code on updated one instead.


david-yz-liu:
I don't know why you wouldn't just change line 85.

PR #3142, opened 2017-07-18
seed multiple templates in one assignment

jenachoi:


david-yz-liu:
@jenachoi this is the right idea, but it's not good to put these files on a public repository without permission. Generate your own template files. Also delete this branch.

PR #3143, opened 2017-07-18
create extra directory for each exam template under assignment (for split_pdf)

jenachoi:
So far we have a directory for each assignment where we store pages in `complete`, `incomplete`, `error`, `raw` directory. When we have multiple exam templates in an assignment and we split generated PDF for each exam template, the pages go into `error` directory because `incomplete/exam_num.pdf` or `complete/exam_num.pdf` already exists in the corresponding directory. We should have a directory layer for each exam template underneath assignment directory.
screen shot 2017-07-19 at 2 28 02 pm

### DIFF
--- a/app/controllers/exam_templates_controller.rb
+++ b/app/controllers/exam_templates_controller.rb
@@ -96,7 +96,7 @@ class ExamTemplatesController < ApplicationController
   def download_generate
     assignment = Assignment.find(params[:assignment_id])
     exam_template = assignment.exam_templates.find(params[:id])
-    send_file(exam_template.base_path + '/' + params[:file_name],
+    send_file(File.join(exam_template.base_path, params[:file_name]),
               filename: params[:file_name],
               type: "application/pdf")
   end
--- a/app/controllers/exam_templates_controller.rb
+++ b/app/controllers/exam_templates_controller.rb
@@ -40,8 +40,7 @@ class ExamTemplatesController < ApplicationController
     assignment = Assignment.find(params[:assignment_id])
     exam_template = assignment.exam_templates.find_by(id: params[:id]) # look up a specific exam template based on the params[:id]
     filename = exam_template.filename
-    assignment_name = assignment.short_identifier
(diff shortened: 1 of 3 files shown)


jenachoi:
I haven't tested this yet.

app/controllers/exam_templates_controller.rb
```diff
@@ -40,8 +40,7 @@ def download
     assignment = Assignment.find(params[:assignment_id])
     exam_template = assignment.exam_templates.find_by(id: params[:id]) # look up a specific exam template based on the params[:id]
     filename = exam_template.filename
-    assignment_name = assignment.short_identifier
-    send_file("#{EXAM_TEMPLATE_DIR}/#{assignment_name}/#{filename}",
+    send_file(exam_template.base_path + '/' + filename,
```
david-yz-liu: Use `File.join`.

app/jobs/split_pdf_job.rb
```diff
@@ -22,12 +22,14 @@ def perform(exam_template, path, original_filename=nil, current_user=nil)
     m_logger = MarkusLogger.instance
     begin
       # Create directory for files whose QR code couldn't be parsed
-      error_dir = File.join(exam_template.base_path, 'error')
-      raw_dir = File.join(exam_template.base_path, 'raw')
+      error_dir = File.join(MarkusConfigurator.markus_exam_template_dir,
+                            exam_template.assignment.short_identifier, 'error')
```
david-yz-liu: This change seems like it goes in the opposite direction. You seem to have made the proper change to `base_path` already.

jenachoi: Unlike `complete` or `incomplete`, aren't `error` and `raw` directories supposed to be shared among every exam template under same assignment? (especially `error` for `assign_error` page)

jenachoi:
I tested this along with #3146 and it works successfully. I will work on changing the `assign_error` page to be a `member` route of the `exam templates`.

david-yz-liu:
@jenachoi merge the two branches into one.

PR #3148, opened 2017-07-19
create extra directory for each exam template under assignment (for split_pdf) & assign_error to be a member route of exam templates

jenachoi:
screen shot 2017-07-20 at 4 16 32 pm

When we click the number of QR scan errors, we can fix the pages for corresponding exam template.


jenachoi:
working on it

david-yz-liu:
@jenachoi I merged this in, but the `title` of the warning icon should be internationalized.

PR #3151, opened 2017-07-21
FIX: Exact same file after switching submissions

DecathectZero:


app/views/results/common/_file_selector_dropdown.html.erb
```diff
@@ -14,23 +14,33 @@
...
-            } else {
-              first_file_id = <%= files[0].id %>;
-              first_file_path = '<%= files[0].path %>/<%= files[0].filename %>';
+        <% if @current_user.ta? && (assignmentfile =
+              AssignmentFile.select(:filename)
+              .joins('JOIN criteria_assignment_files_joins c ON c.assignment_file_id = assignment_files.id')
+              .joins('JOIN criterion_ta_associations t ON c.criterion_id = t.criterion_id')
+              .where('t.ta_id = ? AND assignment_files.assignment_id = ?', @current_user.id, @assignment.id)
+              .first) %>
+        <% index = files.find_index{|x| x.filename == assignmentfile.filename} %>
```
david-yz-liu: Please indent the lines of code under the `if` and `else`.

app/views/results/common/_file_selector_dropdown.html.erb
```diff
@@ -14,23 +14,33 @@
...
+        first_file_id = <%= files[index].id %>;
+        first_file_path = '<%= files[index].path %>/<%= files[index].filename %>';
+        <% else %>
+        $(document).ready(function(){
```
david-yz-liu: Missing some whitespace here

app/views/results/common/_file_selector_dropdown.html.erb
```diff
@@ -14,23 +14,33 @@
...
+          first_file_id = <%= files[0].id %>;
+          first_file_path = '<%= files[0].path %>/<%= files[0].filename %>';
+          if (localStorage.getItem('assignment_id') == <%= @assignment.id %> && localStorage.getItem('file_path')) {
```
david-yz-liu: In general use `===` not `==` (this occurs below as well).

DecathectZero: Some of these variables have different types, is it still better to use === and convert the types before comparison?

david-yz-liu: Great question. Yes, this is still strongly preferred. Note that you can kind of cheat in the Ruby code and do something like `'<%= @asignment.id %>'` to get a string.

PR #3156, opened 2017-07-24
Delete exam template directory in server when exam template is deleted

jenachoi:
Instead of deleting only exam template, we also delete exam template directory in server.


app/models/exam_template.rb
```diff
@@ -96,6 +96,16 @@ def replace_with_file(blob, attributes={})
     self.update(num_pages: pdf.pages.length, filename: attributes[:new_filename])
   end
 
+  def delete_with_file
+    template_path = File.join(
```
david-yz-liu: Use `base_path` here

app/models/exam_template.rb
```diff
@@ -96,6 +96,16 @@ def replace_with_file(blob, attributes={})
...
+      MarkusConfigurator.markus_exam_template_dir,
+      self.assignment.short_identifier,
+      self.name
+    )
+    FileUtils.rm_rf(template_path)
+    return self.destroy
```
david-yz-liu: No need for the `return`

PR #3158, opened 2017-07-25
Fixed annotations not being added to categories

DecathectZero:
If an annotation has no user attached to it (i.e. autogenerated by rake), it will still appear to all users as it is assumed it is default.

### DIFF
--- a/app/views/annotations/_annotation_list.html.erb
+++ b/app/views/annotations/_annotation_list.html.erb
@@ -1,17 +1,14 @@
 <ul class='annotation_text_list'>
   <%
-    text = []
-    annotation_category.annotation_texts.each do |annotation_text|
-      if (annotation_text.user.nil? || (annotation_text.user.admin? || annotation_text.user == @current_user))
-        text << annotation_text
-      end
+    texts = annotation_category.annotation_texts.select do |annotation_text|
+      !annotation_text.user.nil? && (annotation_text.user.admin? || annotation_text.user == @current_user)
     end
   %>
-  <% if text.empty? %>
+  <% if texts.empty? %>
     <li><%= t('annotation_categories.no_annotations') %></li>
   <% end %>
 
-  <% text.each do |annotation_text| %>
+  <% texts.each do |annotation_text| %>
     <li id='annotation_text_<%= annotation_text.id %>'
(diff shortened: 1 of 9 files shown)


app/views/annotations/_annotation_list.html.erb
```diff
@@ -1,9 +1,17 @@
 <ul class='annotation_text_list'>
-  <% if annotation_category.annotation_texts.empty? %>
+  <%
+    text = []
+    annotation_category.annotation_texts.each do |annotation_text|
+      if (annotation_text.user.nil? || (annotation_text.user.admin? || annotation_text.user == @current_user))
```
david-yz-liu: This is interesting for a few reasons. 

1. In general, when building a new list like this you should be using the Enumerable `select` method (I think that's what Ruby calls it)
2. I'm not sure under which conditions `user` is nil (that's strange). No need for parentheses around the whole condition.

app/views/annotations/_annotation_list.html.erb
```diff
@@ -1,9 +1,17 @@
 <ul class='annotation_text_list'>
-  <% if annotation_category.annotation_texts.empty? %>
+  <%
+    text = []
```
david-yz-liu: Pick a plural noun.

app/views/annotations/_annotation_list.html.erb
DecathectZero: The user is nil when the category/annotation is autogenerated when building the database, so I assume all those should be accessible to everyone.

david-yz-liu: Ah, good catch. This is an error in the development environment seeding (please fix); we never generate global annotations like that in production. In fact, it would be great if you added a validation for `user` for both category and annotation.

DecathectZero: ok sounds good, I'll get on that thanks.

david-yz-liu:
@DecathectZero This is definitely the right change to make - but it looks like the tests need to be updated as well. (Look in the factories/blueprints; should be a simple addition to specify an associated user)

david-yz-liu:
@DecathectZero I think you fixed these for the `spec/` tests, but not the `test/` tests.

PR #3163, opened 2017-07-26
reject any changes to template divisions once exams have been uploaded

jenachoi:
I'll focus on fixing QR code errors for now. Please let me know if there are any changes to be made here!

### DIFF
--- a/app/models/exam_template.rb
+++ b/app/models/exam_template.rb
@@ -264,6 +264,6 @@ class ExamTemplate < ActiveRecord::Base
 
   # any changes to template divisions should be rejected once exams have been uploaded
   def exam_been_uploaded?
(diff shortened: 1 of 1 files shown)


app/models/exam_template.rb
```diff
@@ -261,4 +261,9 @@ def rename_exam_template_directory
       File.rename old_directory_name, new_directory_name
     end
   end
+
+  # any changes to template divisions should be rejected once exams have been uploaded
+  def exam_been_uploaded?
+    SplitPdfLog.where(exam_template: self).length > 0
```
david-yz-liu: 1. Use an association (`self.split_pdf_logs`)
2. Use `.any?`

PR #3166, opened 2017-07-30
Large copies -  split pdf fix

rathukugan:
- Changed loading entire pdf from combinepdf to using rmagick's imagelist to convert pdf to images before looping through them.
- Improved performance
- Split pdf is working on pages over 100,200+ (10, 20 copies etc) where as it was failing at around 100 pages before.

Also CombinePDF load (line 59) is causing some warnings while parsing through pages like so:
`Shouldn't add objects to the file unless they are PDF objects or PDF pages (an Array or a single PDF page).`
`Couldn't connect reference for {:is_reference_only=>true, :indirect_generation_number=>0, :indirect_reference_id=>12, :referenced_object=>nil}
`
I'll have to look more into this but split pdf is still working regardless.

### DIFF
--- a/app/jobs/split_pdf_job.rb
+++ b/app/jobs/split_pdf_job.rb
@@ -55,15 +55,13 @@ class SplitPDFJob < ActiveJob::Base
         split_page = SplitPage.create(filename: filename,
                                       raw_page_number: i + 1,
                                       split_pdf_log: split_pdf_log)
-
         page_img.write File.join(raw_dir, "#{split_page.id}.pdf")
         page = CombinePDF.load(File.join(raw_dir, "#{split_page.id}.pdf"))
 
-        #Crop and save QR Image from pdf
+        # Snip out the part of the PDF that contains the QR code.
         qr_img = page_img.crop 0, 10, page_img.columns, page_img.rows / 5
         qr_img.write File.join(raw_dir, "#{split_page.id}.png")
 
-        #Decode QR code
         qrcode_string = ZXing.decode (qr_img.to_blob{ self.format = "png" })
         qrcode_regex = /(?<short_id>\w+)-(?<exam_num>\d+)-(?<page_num>\d+)/
         m = qrcode_regex.match qrcode_string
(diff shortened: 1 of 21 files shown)


app/jobs/split_pdf_job.rb
```diff
@@ -45,32 +45,30 @@ def perform(exam_template, path, original_filename=nil, current_user=nil)
...
+      pdf.each_with_index do |page_img, i|
         split_page = SplitPage.create(filename: filename,
                                       raw_page_number: i + 1,
                                       split_pdf_log: split_pdf_log)
-        page = pdf.pages[i]
-        new_page = CombinePDF.new
-        new_page << page
-        new_page.save File.join(raw_dir, "#{split_page.id}.pdf")
+        page_img.write File.join(raw_dir, "#{split_page.id}.pdf")
+        page = CombinePDF.load(File.join(raw_dir, "#{split_page.id}.pdf"))
```
david-yz-liu: This line probably isn't necessary any more - we shouldn't need to load the PDF here, given that we don't need it for QR code parsing below any more. So you can remove it.

However, what you *should* check (manually) is that the PDFs that are saved here are actually valid PDFs. You should download a couple from the virtual machine and check that you can open them normally in a PDF viewer.

rathukugan:
Hey David,

So I removed the combine pdf loading and was able to replace it using rmagick functionality. The PDF's I downloaded from the vm are valid, the only problem is I feel like the quality/density is too low. ImageList defaults at a density of 72 when reading the files, I've experimented with a density of 100-200 but it takes too long for larger numbers of pages when running the job and the server gets killed.

Do you have any suggestions on how to get around this? 
Here are some examples of the quality differences:
[default-density.pdf](https://github.com/MarkUsProject/Markus/files/1188279/default-density.pdf)
[density-100.pdf](https://github.com/MarkUsProject/Markus/files/1188274/density-100.pdf)
[density-300.pdf](https://github.com/MarkUsProject/Markus/files/1188275/density-300.pdf)

rathukugan:
Closing this PR:
Found simpler solution, see https://github.com/MarkUsProject/Markus/pull/3170

PR #3169, opened 2017-08-02
Marks spreadsheets performance

brendahuang93:
## Ready for Review

This pull request increases the performance for the grade entry forms (marks spreadsheets) graph as well as the calculation of the info summary.

- Utilized eager_load to load in data, reduced repetitive code, and eliminated grade arrays being repeatedly calculated in grade_entry_form.rb

### DIFF
--- a/app/models/grade_entry_form.rb
+++ b/app/models/grade_entry_form.rb
@@ -9,6 +9,7 @@ require 'histogram/array'
 class GradeEntryForm < ActiveRecord::Base
   attr_accessor             :released_grades_array
   attr_accessor             :grades_array
+  attr_accessor             :ges_total_grade
   has_many                  :grade_entry_items,
                             -> { order(:position) },
                             dependent: :destroy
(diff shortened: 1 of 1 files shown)


app/models/grade_entry_form.rb
```diff
@@ -44,44 +44,56 @@ def out_of_total
...
+
+    ges.each do |grade_entry_student|
+      @ges_total_grade = grade_entry_student.total_grade
+      unless @ges_total_grade.nil?
         grades.push(calculate_total_percent(grade_entry_student))
       end
     end
 
-    return grades
+    @released_grades_array = grades
```
david-yz-liu: Using an instance variable to store this is a good idea. But you should put a conditional `if defined? @released_grades_array ... return @released_grades_array` at the very top of this method. This is the proper way of caching the result of this method. That way, you can actually keep the exact same code in your `calculate_*` methods that you had before, and the work will only be done the first time `released_percentage_grades_array` is called.

app/models/grade_entry_form.rb
```diff
@@ -44,44 +44,56 @@ def out_of_total
...
   def percentage_grades_array
-    grades = Array.new
-    grade_entry_students.each do |grade_entry_student|
-      if !grade_entry_student.total_grade.nil? && out_of_total > 0
+    grades = Array.new()
+    ges = GradeEntryStudent.includes(:grades)
+                           .where(grade_entry_form: self)
+
+    ges.each do |grade_entry_student|
+      @ges_total_grade = grade_entry_student.total_grade
```
david-yz-liu: On the other hand, it is *not* appropriate to make `ges_total_grade` an instance variable. It should be local to this method only. (Similar problem occurs above.)

app/models/grade_entry_form.rb
```diff
@@ -44,44 +44,56 @@ def out_of_total
...
+      unless @ges_total_grade.nil?
         grades.push(calculate_total_percent(grade_entry_student))
       end
     end
 
-    return grades
+    @grades_array = grades
```
david-yz-liu: Similar comment as with the percentage array version.

PR #3170, opened 2017-08-02
split pdf fix - Replace rmagick read with from blob

rathukugan:
Simpler solution and is working on large numbers of pages (300+)


david-yz-liu:
@rathukugan there's a failing test. Not sure why, but please look into this. (Could it be that this change makes a page that previously had an error actually parse successfully?)

test/jobs/split_pdf_job_test.rb
```diff
@@ -182,7 +182,7 @@ class SplitPDFJobTest < ActiveJob::TestCase
 
         should 'have 3 pdfs in error directory' do
```
david-yz-liu: @rathukugan please change the text here as well.

david-yz-liu:
@rathukugan and @jenachoi I merged this in now so that you are using it in your work improving the error rate.

PR #3173, opened 2017-08-03
Autocomplete Scan

DecathectZero:
Sorry this took awhile.

### DIFF
--- a/app/controllers/exam_templates_controller.rb
+++ b/app/controllers/exam_templates_controller.rb
@@ -147,7 +147,7 @@ class ExamTemplatesController < ApplicationController
     @exam_template = @assignment.exam_templates.find(params[:id])
     @error_files = []
     error_path = File.join(@exam_template.base_path, 'error')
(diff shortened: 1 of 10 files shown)


app/controllers/exam_templates_controller.rb
```diff
@@ -146,10 +146,38 @@ def assign_errors
     @assignment = Assignment.find(params[:assignment_id])
     @exam_template = @assignment.exam_templates.find(params[:id])
     @error_files = []
-    Dir.foreach(File.join(@exam_template.base_path, 'error')) do |file|
-      @error_files << file unless file =~ /^\.\.?$/
+    error_path = File.join(@exam_template.base_path, 'error')
+    if(File.directory?(error_path))
```
david-yz-liu: There are a couple of style issues throughout these changes - needs a once-over.

Perhaps a new one I haven't commented on before is that in Ruby it's preferred to omit parentheses around a single-expression condition for an `if`.

PR #3178, opened 2017-08-04
more indentation on testing PDFs

jenachoi:
We've been testing with these PDFs to make sure that there aren't any overlaps between QR code & QR text label and questions.


david-yz-liu:
@jenachoi I'm not sure that you actually want me to merge this in? If this is just for testing purposes, then I'm okay with you just sharing the test files electronically, rather than changing the repo. We should only change the test files after testing has been completed.

(As a side note, the margins for these PDFs are so large that this would be unusable for an actual test.)

PR #3180, opened 2017-08-05
Submissions progress bar

rathukugan:
Converting submissions job to use the new infrastructure for progress bar


david-yz-liu:
@rathukugan the change you made is good, but broke some old tests. You can fix the tests just by asserting that a certain partial is rendered.

PR #3183, opened 2017-08-07
Progress bar change for create indiv groups job

rathukugan:


app/views/groups/_groups_manager.js.jsx.erb
```diff
@@ -313,7 +313,6 @@
       return (
         <div className={'col-right'}>
           <div className={'tab-pane ui-tabs ui-widget ui-widget-content ui-corner-all'} id='groups-tabs'>
-            <AddGroupButton refresh={this.props.refresh} />
```
david-yz-liu: I'm not sure why you deleted this button?

rathukugan: Oops, that was a mistake. I will push a commit with that re-added, thanks for catching that!

PR #3189, opened 2017-08-10
Submission Rules into tables

DecathectZero:


app/views/assignments/_form.html.erb
```diff
@@ -304,27 +304,42 @@
                         '.form_description') %>
       </p>
       <div id='grace_periods'>
+        <h2><%= t('submission_rules.grace_period_submission_rule.short_name') %></h2>
+        <table id='grace_periods_table'>
+          <thead>
+          <tr>
+            <th><%= label :hours,
```
david-yz-liu: The HTML semantics might be a little misleading. The column header is not a label for any one particular input tag, so I don't think "label" is actually appropriate here.

app/views/assignments/_form.html.erb
```diff
@@ -304,27 +304,42 @@
...
+          </tr>
+          </thead>
+          <tbody>
+            <% if rule.object.type.to_s == 'GracePeriodSubmissionRule' %>
+                <%= rule.fields_for :periods do |period_form| %>
+                    <%
+                      # The below conditional is meant to filter out periods that do not match
+                      # the currently selected submission rule. This only happens when we fail to
+                      # save a new submission rule.
+                      if rule.object.type.to_s == rule.object.class.to_s or period_form.object.id.nil? %>
```
david-yz-liu: Use `||` instead of `or`.

PR #3195, opened 2017-08-12
convert student_test to spec

jenachoi:


david-yz-liu:
@jenachoi looks like there's an `include` error on this one. Let me know if you want to fix it up!

Issue #2775, opened 2016-10-04
_ta_grade_distribution_graph.html.erb has error when assignment has no criteria


david-yz-liu:
An instructor ran into this problem where graders had been assigned to mark an assignment, but no criteria were created. This causes the dashboard to fail to load (some kind of division by zero error).

PR #3201, opened 2017-09-25
Assignment no criteria - Issue #2775

brendahuang93:
## Ready for Review
This pull request fixes issue #2775 where the assignment dashboard does not load when graders are assigned to mark the assignment but no criteria were created.

- Added an extra case in assignment.rb to handle when the current result of a grouping is nil


david-yz-liu:
@brendahuang93 thanks for doing this -- I'd actually fixed it in the 1.6.0 branch (just merged in), but forgot to close the issue.

FYI your fix actually has a logical error of resetting `count` to 0, rather than not adding to it (i.e., `count += 0` vs. `count = 0`).

PR #3204, opened 2017-09-26
Harmonize test results

adisandro:
This unifies the 3 test result interfaces (student tests, student results, instructor results).
Subtle differences were present in each one because changes were rarely done to all at the same time, leading to things breaking at each major autotest change.

### DIFF
--- a/app/views/automated_tests/_test_script_results.html.erb
+++ b/app/views/automated_tests/_test_script_results.html.erb
@@ -3,23 +3,11 @@
 <%#                |   'student'   |         'instructor'         |%>
 <%#    .student?   | student tests | instr tests, results tab     |%>
 <%#   !.student?   | XXXXXXXXXXXXX | instr tests, submissions tab |%>
-<% if ran_by == 'student' %>
-  <% marks = grouping.last_student_test_marks %>
-  <% total_marks = assignment.total_student_test_script_marks %>
-<% elsif ran_by == 'instructor' && local_assigns[:submission] %>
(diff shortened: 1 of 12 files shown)


app/views/automated_tests/_test_script_results.html.erb
```diff
@@ -0,0 +1,38 @@
...
+  <% marks = grouping.last_instructor_test_marks %>
+  <% total_marks = assignment.total_instructor_test_script_marks %>
+<% end %>
+<h2><%= t('automated_tests.test_results') %></h2>
+<div class="block_content">
+  <div class="sub_block test_points">
+    <% if results.empty? %>
+      <%= t('automated_tests.no_results') %>
+    <% elsif !current_user.student? || ran_by == 'student' %>
+      <%= raw t('automated_tests.marks_obtained', marking: "#{marks}/#{total_marks}") %>
```
david-yz-liu: Everyone finds this confusing, not just students. You can remove it altogether. Even for the student-run tests.

adisandro: I agree but I don't know how to make it better. That one time when you get some test numbers wrong, it's actually useful to debug.
I will remove it, and in general we should make the mapping between tests and marks more intuitive.

david-yz-liu: Yes, I agree. I could see, for example, having a new column in the test result table that shows "points awarded" or something.

PR #3207, opened 2017-09-29
Autotest fixes

adisandro:


david-yz-liu:
I don't think students can use the API (see `controllers/api/main_api_controller.rb:60`), so I'm not really sure what the first commit is for. Even when students run self-tests, shouldn't it be the TestServer user that actually runs the tests and adds the results?

adisandro:
The api key of the requesting user is passed to the server and to the test scripts:
- The server sends it back for Markus to know who requested the test (the requested_by field).
- The test script can use it to invoke apis like upload_feedback_file or add_annotation, so that in fact api calls from students are rejected (if it were the TestServer's api key, a bad script would allow privilege escalation).

The test results only are sent back by api call using the TestServer's api key, yes, but we do need both the api keys.

david-yz-liu:
Okay thanks @adisandro.

PR #3209, opened 2017-10-02
Fixed issue of not being able to sort Graders in an assignment

brendahuang93:
## Ready for Review
This pull request fixed the issue of not being able to sort Graders for an assignment. The original issue was that _graders_manager.js.jsx.erb was calling table.js, which was trying to sort the checkboxes of each grader instead of the graders' names. 

- Added sort_data in _graders_manager.js.jsx.erb: sort_data contains the names of each grader for each row
- sort_data is passed in as a props named "sort" to table.js
- method sort_by_column in table.js sorts the graders' names if the data exists

### Screenshots
screen shot 2017-10-02 at 12 28 46 am
screen shot 2017-10-02 at 12 28 40 am

### DIFF
--- a/app/assets/javascripts/ReactComponents/table.js
+++ b/app/assets/javascripts/ReactComponents/table.js
@@ -53,7 +53,6 @@ if (typeof React !== 'undefined') {
 var Table = React.createClass({displayName: 'Table',
   propTypes: {
     data: React.PropTypes.array,
-    sort: React.PropTypes.array,
     search_placeholder: React.PropTypes.string,
     columns: React.PropTypes.array,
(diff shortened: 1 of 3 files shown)


david-yz-liu:
Adding a new prop to the base `Table` class is probably overkill, and may have unintended consequences.

Instead, what you should be using to customize sorting behaviour is to:

1. Add a new function to `table_sorts.js`.
2. Specify your function when defining the graders column in the actual React table. Note that each column is specified as a hash---they key you'll want to set is `compare`.

You can search for the other sorting functions in `table_sorts.js` in the repo to see examples of how they're used.

PR #3210, opened 2017-10-05
Repo fixes

adisandro:
1) Fixes the seeded database to work with git.
2) Fixes the 'there is a commit after deadline' message for collected submissions to look for the assignment directory only.
3) Reorders svn revisions from last to first instead of the opposite
4) Fixes missing marks with autotesting when criteria are created after collection

### DIFF
--- a/app/helpers/submissions_helper.rb
+++ b/app/helpers/submissions_helper.rb
@@ -106,13 +106,7 @@ module SubmissionsHelper
           g[:tags] = grouping.tags
           g[:commit_date] = grouping.last_commit_date
           g[:has_files] = grouping.has_files_in_submission?
-          g[:late_commit] =
-            # TODO: Enable this check for Git backend. See issue #1866.
-            if MarkusConfigurator.markus_config_repository_type == 'git'
-              false
-            else
-              grouping.past_due_date?
-            end
+          g[:late_commit] = grouping.past_due_date?
           g[:grace_credits_used] =
             if assignment.submission_rule.is_a? GracePeriodSubmissionRule
               grouping.grace_period_deduction_single
--- a/app/models/grouping.rb
+++ b/app/models/grouping.rb
@@ -513,12 +513,16 @@ class Grouping < ActiveRecord::Base
 
   # Returns last modified date of the assignment_folder in this grouping's repository
   def assignment_folder_last_modified_date
-    repo = self.group.repo
-    rev = repo.get_latest_revision
-    last_date = rev.timestamp
-
-    repo.close()
-    last_date
+    revision = nil
+    group.access_repo do |repo|
+      # we use this function because it allows to specify a path
+      revision = repo.get_revision_by_timestamp(Time.zone.now, assignment.repository_folder)
+      # an alternative could have been:
+      # repo.get_latest_revision.directories_at_path('')[assignment.repository_folder].last_modified_date
+      # but in git, the latter returns commit times instead of push times, and would be less efficient because it has to
+      # walk through the entire history
+    end
+    revision.timestamp
   end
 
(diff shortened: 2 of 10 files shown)


app/helpers/submissions_helper.rb
```diff
@@ -106,13 +106,7 @@ def get_submissions_table_info(assignment, groupings)
...
           g[:commit_date] = grouping.last_commit_date
           g[:has_files] = grouping.has_files_in_submission?
-          g[:late_commit] =
-            # TODO: Enable this check for Git backend. See issue #1866.
-            if MarkusConfigurator.markus_config_repository_type == 'git'
-              false
-            else
-              grouping.past_due_date?
-            end
+          g[:late_commit] = grouping.past_due_date?
```
david-yz-liu: @adisandro The previous issue here was about performance, not correctness. Just want to make sure you're aware of that and tried this out on a fairly large set of submissions.

app/models/grouping.rb
```diff
@@ -513,12 +513,16 @@ def number_of_submitted_files
...
+    revision = nil
+    group.access_repo do |repo|
+      # we use this function because it allows to specify a path
+      revision = repo.get_revision_by_timestamp(Time.zone.now, assignment.repository_folder)
+      # an alternative could have been:
+      # repo.get_latest_revision.directories_at_path('')[assignment.repository_folder].last_modified_date
+      # but in git, the latter returns commit times instead of push times, and would be less efficient because it has to
+      # walk through the entire history
+    end
+    revision.timestamp
```
david-yz-liu: I'm wondering about a possible `nil` error here. I guess in the old version, `rev` was guaranteed not to be nil, but I'm not sure this is true here.

app/models/grouping.rb
```diff
@@ -664,18 +668,11 @@ def section
...
-                         due_dates.where(section_id: section).first.due_date
-                       end
-
-    if !section_due_date.blank?
-      timestamp > section_due_date
-    else
+    if inviter.blank? || inviter.section.blank? || assignment.section_due_dates.blank?
       timestamp > assignment.due_date
+    else
+      section_due_date = assignment.section_due_dates.where(section_id: inviter.section).first.due_date
```
david-yz-liu: Using `find` instead of `.where().first`

lib/tasks/groups.rake
```diff
@@ -38,12 +38,11 @@ namespace :db do
         end
 
         group.access_repo do |repo|
+            # add files to the root folder of the repo (e.g. "A1")
```
david-yz-liu: The indentation is off here.

app/helpers/submissions_helper.rb
adisandro: Yes, I was very careful about performance when fixing git this summer, following that bug and the related discussion on the rugged bug tracker. The bottleneck was the function to detect if a file/directory changed with a specific commit, and I implemented all the optimizations suggested by the rugged people. Still, some repo operations are always going to be quite expensive in the worst case scenario, like having to walk all histories from top to the first commit.
I haven't tried this out with a lot of submissions though. My plan would be to enable it, try it in the real 207 scenario, and revert if there are problems.

app/models/grouping.rb
adisandro: That function always returns a revision for git and svn, with the first revision as a fallback.
(I'm sure about git, and the svn comments say that at least)

app/models/grouping.rb
adisandro: Right, I copied and pasted the previous code. Will change it.

lib/tasks/groups.rake
adisandro: Will fix.

Issue #2776, opened 2016-10-04
Investigate spreadsheet column order


david-yz-liu:
Instructor reports: had two spreadsheet columns and created a third; the new one appeared between the old ones.

PR #3214, opened 2017-10-21
Resolved issue #2776

gilltan97:
# Resolved Issue #2776
Columns created in the spreadsheet were being viewed in the order they were created in.

## Before 
The maximum position any of the existing columns was hardcoded to 1 and because of this, other columns were being assigned the same position which led to this issue. 

## After 
Now the max position is assigned to the count of existing columns in the particular spreadsheet. Due to this change, the position of columns gets assigned the position of the last column plus 1 which resolves this issue.

### DIFF
--- a/app/helpers/grade_entry_forms_helper.rb
+++ b/app/helpers/grade_entry_forms_helper.rb
@@ -29,7 +29,7 @@ module GradeEntryFormsHelper
 
     unless grade_entry_items.nil?
       # Update the attributes hash
-      max_position = 1
+      max_position = GradeEntryItem.where(:grade_entry_form_id => 1).count
       grade_entry_items.each do |_, item|
         # Some items are being deleted so don't update those
         unless item[:_destroy]


app/helpers/grade_entry_forms_helper.rb
```diff
@@ -29,7 +29,7 @@ def update_grade_entry_form_params(attributes)
 
     unless grade_entry_items.nil?
       # Update the attributes hash
-      max_position = 1
+      max_position = GradeEntryItem.where(:grade_entry_form_id => 1).count
```
david-yz-liu: This is the right idea, but this hard-codes a grade entry form id of `1`, which isn't good. You should be able to get the correct id from `attributes`.

david-yz-liu: Actually I'm looking at the code immediately beneath this now, which does seem to correctly increment `max_position` for each grade entry item. The way the form is set up, both existing and new grade entry items should appear in `params[:grade_entry_form][:grade_entry_items_attributes]`, so there's a mixture of both items with existing positions, and ones that are new.

I wonder if what's actually causing the problem is that the items aren't sorted in `grade_entry_items`, so they're assigned random new positions. You could try fixing this by explicitly sorting `grade_entry_items` by `position`, putting the ones with `nil` last.

Issue #2833, opened 2016-12-01
Bug: Error thrown in Tags -> Download yml


b2simms:
**Problem:** An error is thrown in the Tags controller when attempting to download the **yml** file

To reproduce:
1. Go to: _local-url_/en/assignments/2/tags
2. Click on `download` in top right corner
3. Click the `yml` option

Result:
Should see an error thrown in the Tags controller class
![tags_controller_download_yml_error](https://cloud.githubusercontent.com/assets/12561579/20799977/3f6525e0-b7ba-11e6-87e4-d0a4d16cd44a.PNG)

PR #3215, opened 2017-10-21
Fixed a syntax error to resolve the issue #2833

gilltan97:
# Resolved issue #2833 
An error message was being thrown when attempted to download assignment tags in a file with .yaml format. 

This was happening due to the bad syntax for getting all the tags ordered by name in file ``` tags_controller.rb```.


david-yz-liu:
Thanks, @gilltan97!

PR #3222, opened 2017-10-23
Ready for Review: Assignment summaries grade sort

brendahuang93:
## Ready for Review
This pull request converts the types of grades per criterion under Assignment --> Summaries from string to float. This allows proper sorting of the grades; previously they were being sorted by the first digit of each grade.

- Changed the mark for each criterion from string to float

### Screenshots
screen shot 2017-10-23 at 11 55 03 am
screen shot 2017-10-23 at 11 54 57 am


david-yz-liu:
Thanks, @brendahuang93!

Issue #2936, opened 2017-05-02
Bug: inconsistent decimal when comma used as period


mishaschwartz:
In the Manage Marks Spreadsheet view, when adding a mark to a cell, if a comma is used instead of period the cell total is counted as the digits up to the comma and no format error is raised (cell does not turn red)

![screen shot 2017-05-02 at 11 43 50](https://cloud.githubusercontent.com/assets/4380924/25626188/b34053d8-2f2c-11e7-9e65-9fdfe431f552.png)
![screen shot 2017-05-02 at 11 44 05](https://cloud.githubusercontent.com/assets/4380924/25626191/b6413822-2f2c-11e7-9f2d-8a8646a3f03e.png)

adisandro:
Duplicate of #2925

PR #3223, opened 2017-10-23
Ready for Review: Marks Spreadsheet Grade with Non-Numerical Characters - Issue #2936

brendahuang93:
## Ready for Review
This pull request fixes issue #2936 and updates code so that all marks spreadsheet grades that contain non-numerical characters (other than period) are disregarded. 

- Updated parseFloat to Number, which returns NaN if the string contains non-numerical characters (other than period)
- Added timeout of 1.5 seconds so that ajax request is not sent until user is done typing 

### Screenshots
screen shot 2017-10-24 at 9 14 09 am

### DIFF
--- a/app/views/grade_entry_forms/_react_grades_table.js.jsx.erb
+++ b/app/views/grade_entry_forms/_react_grades_table.js.jsx.erb
@@ -56,7 +56,7 @@
         this.props.value = this.props.default_value;
     },
     handleChange: function(event) {
-      var updated_grade = event.target.value === '' ? null : parseFloat(event.target.value);
+      var updated_grade = event.target.value === '' ? null : Number(event.target.value);
       var params = {
         'updated_grade':       updated_grade,
         'student_id':          this.props.student_id,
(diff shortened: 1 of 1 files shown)


app/views/grade_entry_forms/_react_grades_table.js.jsx.erb
```diff
@@ -53,23 +53,30 @@
 
   var GradeEntryBox = React.createClass({
     componentDidMount: function(){
-        this.props.value = this.props.default_value;
```
david-yz-liu: Revert the change on this line.

app/views/grade_entry_forms/_react_grades_table.js.jsx.erb
```diff
@@ -53,23 +53,30 @@
...
-      $.ajax({
-        url:  this.props.data_action,
-        data: params,
-        type: 'POST',
-        dataType: 'script',
-      });
-      this.props.value = parseFloat(event.target.value);
+
+      var data_action = this.props.data_action;
+      this.typing_timer = setTimeout(function() {
```
david-yz-liu: It seems like this should only happen if `updated_grade` is not NaN? Ideally the cell would be set to red immediately, rather than making a request to the server that's going to fail.

PR #3226, opened 2017-10-30
Fix TypeError in Results#edit

MeowsterEric:
add .to_i to explicitly convert nil to 0.

markus crash


david-yz-liu:
Thanks again, @MeowsterEric!

PR #3228, opened 2017-10-30
Restore proper text for student tests that regenerate

adisandro:
After the introduction of non_regenerating_tokens, the ui for student tests was not showing the same info it used to show.


adisandro:
I really don't understand the test failure, I didn't change anything relevant in the code.
I can see that it appeared before this, intermittently.

Issue #3237, opened 2017-11-24
CSRF Vulnerability


peterzhu2118:
I believe MarkUs is vulnerable to CSRF. I am able to send POST requests through Postman without sending the CSRF token. This is caused by the fact that in the [ApplicationController](https://github.com/MarkUsProject/Markus/blob/master/app/controllers/application_controller.rb#L15) it uses `protect_from_forgery` rather than `protect_from_forgery with: :exception`. I believe this choice was to enable support API, however `protect_from_forgery with: :exception` can be enabled in the controllers that do not serve API requests.

Also odd on how the logout is a GET request instead of a POST request, which would not be protected against CSRF (if it was enabled).

david-yz-liu:
Thanks @peterzhu2118!

PR #3238, opened 2017-11-25
Change logout to POST request

peterzhu2118:
Change logout to a POST request (from a GET request), following regular conventions and allows for CSRF protection. Fixes second half of issue #3237.

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -23,10 +23,9 @@
                     remote: true %>
       <% end %>
 
-      <%= button_to(t(:log_out),
+      <%= link_to(t(:log_out),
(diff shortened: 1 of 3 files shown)


adisandro:
Thanks, @peterzhu2118!
(I made a minor change to restore the link rather than using the button, this is good to go @david-yz-liu)

peterzhu2118:
@adisandro The link wouldn't work if there were proper CSRF protection because the link just sends the request. The button on the other hand, generates a form instead to send the CSRF token along with the request, allowing CSRF protection.

adisandro:
Yeah, but link_to with a method parameter is processed by jquery_ujs into a proper post request.

PR #3245, opened 2017-12-08
New autotesting features

adisandro:
This pull request is dependent on merging 1.6.0 into master first (well, it contains the merge itself from the merge-1.6 pull request).

I still need to write a few tests to cover all the new features (markus-autotesting's kind of tests, not Markus'), but everything else is ready for review.

- Remove user input for max points, and have the tests specify how many points they are worth in the xml protocol.
- Support floats as test points.
- Replace ruby net-scp gem with native scp (~10x faster)
- Allow replacing test scripts and support files with same name.
- Send back errors.txt to markus.
- Get rid of a few n+1 queries.
- Show the user that ran the test.
- Remove experimental label for student test option.


adisandro:
@david-yz-liu: I don't know what's happening with the tests, they were fine before the rebase and now every single attempt to open a repository fails. I'll need more time to check this tomorrow.

app/assets/javascripts/testframework_config.js
```diff
@@ -13,27 +13,27 @@ function toggleSettings ( collapse_lnk ) {
...
-  max_marks = box.find('.maxmarks');
+  //max_marks = box.find('.maxmarks');
   desc = left_side.find('.desc');
   desc_box = desc.find('textarea');
 
   if( collapse_lnk.data('collapsed') ) {
     // Was collapsed. Need to expand.
     desc.nextAll('*').show();
-    max_marks.nextAll('*').show();
+    //max_marks.nextAll('*').show();
```
david-yz-liu: I suppose you can delete these.

app/controllers/automated_tests_controller.rb
```diff
@@ -14,48 +14,27 @@ def update
...
-          File.open(
-              assignment_tests_path, 'wb') { |f| f.write contents }
+    begin
+      @assignment.transaction do
+        files = process_test_form(@assignment, params, assignment_params)
+        if @assignment.save
+          # write the uploaded files
+          files.each do |file|
+            File.open(file[:path], 'wb') { |f| f.write file[:upload].read }
+            if file.has_key?(:delete)
```
david-yz-liu: Please put the `File.exist?` here, in an `&&`

app/models/grouping.rb
```diff
@@ -22,10 +22,6 @@ class Grouping < ActiveRecord::Base
   has_many :accepted_student_memberships,
            -> { where 'memberships.membership_status' => [StudentMembership::STATUSES[:accepted], StudentMembership::STATUSES[:inviter]] },
            class_name: 'StudentMembership'
-  has_many :accepted_students,
```
david-yz-liu: I'm not sure why you deleted this.

app/controllers/automated_tests_controller.rb
```diff
@@ -14,48 +14,27 @@ def update
...
+              File.delete(file[:delete]) if File.exist?(file[:delete])
+            end
+          end
+          flash_message(:success, I18n.t('assignment.update_success'))
```
david-yz-liu: You should be able to do this without the `I18n` now.

app/helpers/automated_tests_client_helper.rb
```diff
@@ -14,128 +14,97 @@ def create_test_repo(assignment)
...
-            File.delete(old_file_path)
-          end
-        end
+      updated_form_file[name_field] = new_file_name
+      new_file_path = File.join(MarkusConfigurator.markus_ate_client_dir, assignment.repository_folder, new_file_name)
+      files.push({path: new_file_path, upload: new_file})
+    # 5) Possibly replace existing test file
+    else
+      return updated_form_file unless form_file[name_field].nil? # replacing a test file resets the old name
+      old_file_name = model_class.where(id: form_file[:id]).pluck(name_field).first
```
david-yz-liu: You should be able to use `find` here.

app/helpers/automated_tests_client_helper.rb
```diff
@@ -14,128 +14,97 @@ def create_test_repo(assignment)
...
+      upd_file = params[("#{upd_name}_#{old_file_name}").to_sym]
+      upd_file_name = upd_file.original_filename
+      updated_form_file[name_field] = upd_file_name
+      mod_file_path = File.join(MarkusConfigurator.markus_ate_client_dir, assignment.repository_folder, upd_file_name)
+      f = {path: mod_file_path, upload: upd_file}
+      files.push(f)
```
david-yz-liu: This is very small, but I'd prefer seeing this line after the `unless`.

app/helpers/automated_tests_client_helper.rb
```diff
@@ -428,24 +392,121 @@ def self.perform(host_with_port, test_scripts, user_api_key, server_api_key, gro
     end
   end
 
-  def self.process_test_result(raw_result, test_scripts_ran, assignment, grouping, submission, requested_by)
+  def self.process_test_result(xml, test_script_result)
```
david-yz-liu: This should also be an instance method, something like `TestScriptResult#build_test_result_from_xml`.

app/helpers/automated_tests_client_helper.rb
```diff
@@ -428,24 +392,121 @@ def self.perform(host_with_port, test_scripts, user_api_key, server_api_key, gro
...
+      raise 'Malformed xml'
+    end
+
+    marks_earned = xml['marks_earned'].nil? ? 0.0 : xml['marks_earned'].to_f
+    marks_total = xml['marks_total'].nil? ? nil : xml['marks_total'].to_f
+    input = xml['input'].nil? ? '' : xml['input']
+    expected = xml['expected'].nil? ? '' : xml['expected']
+    actual = xml['actual'].nil? ? '' : xml['actual']
+    status = xml['status']
+    if status.nil? or not status.in?(%w(pass partial fail error error_all))
```
david-yz-liu: Use `||` instead of `or`

app/helpers/automated_tests_client_helper.rb
```diff
@@ -428,24 +392,121 @@ def self.perform(host_with_port, test_scripts, user_api_key, server_api_key, gro
...
+
+    add_test_result(test_script_result, test_name, input, actual, expected, marks_earned, marks_total, status)
+    if stop_processing
+      raise 'Test script reported a serious failure'
+    end
+
+    return marks_earned, marks_total
+  end
 
+  def self.process_test_script_result(xml, assignment, grouping, submission, requested_by)
```
david-yz-liu: As above, I could see this moving this method to something like `TestScriptResult.create_from_xml`.

app/helpers/automated_tests_client_helper.rb
```diff
@@ -428,24 +392,121 @@ def self.perform(host_with_port, test_scripts, user_api_key, server_api_key, gro
...
+
+    # create test result
```
david-yz-liu: This should be `create test script result` I think

app/assets/javascripts/testframework_config.js
adisandro: That js file is not currently used, but I wanted to keep track of the original behavior in case we want to use it again.
Commenting means that if somebody puts it back as-is, at least it does not break. I'm open to leave it without the comments or even dump it altogether.

app/controllers/automated_tests_controller.rb
adisandro: Do you mean there is a default message for a :success?

app/assets/javascripts/testframework_config.js
david-yz-liu: Ah, got it. Please just dump this file. If we later want to re-introduce some of this functionality, we'll use libraries to do so.

app/controllers/automated_tests_controller.rb
david-yz-liu: No I mean you should be able to use `t(...)` rather than `I18n.t(...)` in all controllers now.

app/assets/javascripts/testframework_config.js
adisandro: Well it turns out it's being used instead for minor things (setting the file name when clicking browse and greying the form when clicking delete). I'll dump the unused stuff.

app/helpers/automated_tests_client_helper.rb
adisandro: Yes but I can't pluck that way, I have to get the whole row. It's efficiency vs readability, I was choosing the first.

app/helpers/automated_tests_client_helper.rb
adisandro: +not: too much python here

app/helpers/automated_tests_client_helper.rb
david-yz-liu: I prefer readability on this line. Please change it.

app/helpers/automated_tests_client_helper.rb
```diff
@@ -428,24 +392,121 @@ def self.perform(host_with_port, test_scripts, user_api_key, server_api_key, gro
...
+    unless tests.is_a?(Array) # same workaround as above, Hash.from_xml returns a hash if it's a single test
+      tests = [tests]
+    end
+
+    # process tests
+    all_marks_earned = 0.0
+    all_marks_total = 0.0
+    tests.each do |test|
+      begin
+        marks_earned, marks_total = AutomatedTestsClientHelper.process_test_result(test, new_test_script_result)
```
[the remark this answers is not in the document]
adisandro: It won't work in ruby if it's a class method, you either need the class or self.class.

app/models/grouping.rb
adisandro: Because it has two of those, it was duplicate.

app/helpers/automated_tests_client_helper.rb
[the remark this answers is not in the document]
david-yz-liu: Good call, I lost track of which ones are class methods and which ones are instance methods.

app/helpers/automated_tests_client_helper.rb
```diff
@@ -14,128 +14,97 @@ def create_test_repo(assignment)
...
-      # neither element is nil
-      if testscripts[file_num][:criterion_id].nil? || testscripts[file_num][:criterion_id][0].nil? || testscripts[file_num][:criterion_id][1].nil?
-        updated_script_files[file_num][:criterion_type]
+  def process_test_file(assignment, params, files, form_file, is_script)
+
+    # custom variables
+    if is_script
+      new_file_param = :new_script
+      model_class = TestScript
+      name_field = :script_name
```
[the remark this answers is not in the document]
adisandro: Sure, can do that. You don't want to go all the way to table inheritance, right?

app/helpers/automated_tests_client_helper.rb
```diff
@@ -428,24 +392,121 @@ def self.perform(host_with_port, test_scripts, user_api_key, server_api_key, gro
...
+      marks_earned = 0.0
+    end
+    if status == 'error_all'
+      status = 'error'
+      stop_processing = true
+    else
+      stop_processing = false
+    end
+    if !marks_total.nil? && marks_earned > marks_total
+      marks_earned = marks_total
```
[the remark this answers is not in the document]
adisandro: Can do that, needs a bit of logic refactoring.

PR #3255, opened 2017-12-15
Break automated_test_client_helper into pieces

adisandro:


adisandro:
1 Sounds good, I don't like ATE either. What about using Autotest as a prefix for this kind of things, or you just want to be done with prefixes and just write something meaningful for the rest? (I like having a prefix to categorize things)
2 I was using the same name as the queue, more or less like the other jobs. I would like to keep them consistent so maybe I should change the queue name too.
3 Regarding Rails5, I made this pull request for 1.6.1. Should this go into master instead?

david-yz-liu:
1. I'm okay with a prefix of "Autotest".
2. Yeah, please change the queue name too then. (The configurator option)
3. You're right, sorry. Targeting this branch is fine, I'll handle the update when I merge this branch into master.

adisandro:
My ocd took over and I renamed all settings from ATE to Autotest.

PR #3257, opened 2017-12-19
Rails 5: introduce Application superclasses

david-yz-liu:


david-yz-liu:
@adisandro please take this into account to update #3255.

PR #3265, opened 2018-01-12
Make all assets local

adisandro:
* js: React, JSXTransformer, MathJax (upgraded to 2.7.2)
 * css: OpenSans

I kept the locality of the various js imports, but I think we could reorganize our assets pipeline to be faster (e.g. test if a big application[.js|.css] speeds up things, precompiling assets in production into static files rather than do it on the fly).
Also, the fix seems trivial but I had a lot of issues to make the huge MathJax load. I'm not completely sure I've done it correctly, do you have files to test it?


adisandro:
I don't have time to do it now, but I think we can simplify the fonts directory to have just the .woff files.
I can do a forced push on Monday to override this last commit.

adisandro:
@david-yz-liu Done now, the fonts dir is as small as possible.

PR #3279, opened 2018-02-12
Create rake task to seed an assignment with autotesting scripts

mishaschwartz:


db/data/autotest_files/autotest_01.sh
```diff
@@ -0,0 +1,41 @@
...
+    <status>pass</status>
+</test>
+<test>
+    <name>test_2</name>
+    <input>input string</input>
+    <expected>expected result value</expected>
+    <actual>actual result value</actual>
+    <marks_earned>1</marks_earned>
+    <marks_total>5</marks_total>
+    <status>fail</status>
```
david-yz-liu: Seems like we should be enforcing an invariant that status `fail` implies marks earned = 0. @adisandro is there a reason not to do this?

But even if we permit this to be a convention rather than requirement, I'd rather not our sample test illustrate this behaviour.

Same comment for the `error` output below.

lib/tasks/autotest.rake
```diff
@@ -0,0 +1,65 @@
+namespace :db do
+  desc 'Create autotest for assignment 5'
+  task :autotest => :environment do
+    puts 'Add autotest scripts'
+
+    test_file_location = File.join('db', 'data', 'autotest_files')
+    test_file_destination = File.join('data', 'dev', 'autotest')
```
david-yz-liu: You should use the `MarkusConfigurator.autotest_client_dir` value here.

lib/tasks/autotest.rake
```diff
@@ -0,0 +1,65 @@
...
+    # create A5 directory to put new autotest files into
+    test_file_destination = File.join(test_file_destination, 'A5')
+    FileUtils.makedirs test_file_destination
+
+    # copy test files into the destination directory
+    FileUtils.cp original_test_files, test_file_destination
+
+    assignment = Assignment.find_by(short_identifier: 'A5')
+
+    unless assignment.nil?
```
david-yz-liu: Eh, you can remove this check.

lib/tasks/autotest.rake
```diff
@@ -0,0 +1,65 @@
...
+      # get the criteria from assignment 5
+      criteria = assignment.flexible_criteria + assignment.rubric_criteria + assignment.checkbox_criteria
```
david-yz-liu: You want `assignment.get_criteria` here.

lib/tasks/autotest.rake
```diff
@@ -0,0 +1,65 @@
...
+
+      # get the test files stored in db/data/autotest_files
+      test_files = Dir.glob(File.join(test_file_destination, '*')).select {|f| File.file?(f)}
+
+      test_files.each do |test_file|
+        # get an arbitrary criterion from assignment 5
+        criterion = criteria.pop
```
david-yz-liu: Prefer using [zip](https://apidock.com/ruby/Enumerable/zip) here.

lib/tasks/autotest.rake
```diff
@@ -0,0 +1,65 @@
...
+          display_input: "do_not_display",
+          display_expected_output: "do_not_display",
+          display_actual_output: "display_after_submission",
+          timeout: 30,
+          criterion: criterion
+        )
+      end
+
+      # collect the submissions from all groupings for assignment 5 so they can be autotested
+      assignment.groupings.all.each do |grouping|
```
david-yz-liu: Use `assignment.groupings.find_each do ...`

lib/tasks/autotest.rake
```diff
@@ -0,0 +1,65 @@
...
+        time = assignment.submission_rule.calculate_collection_time.localtime
+        Submission.create_by_timestamp(grouping, time)
+        # collect submission
+        grouping.is_collected = true
+        grouping.save
+      end
+    end
+  end
+end
+
```
david-yz-liu: End the file with just one blank line.

db/data/autotest_files/autotest_01.sh
adisandro: We are enforcing it in the testers, i.e. the MarkusTester base class they use.
It's true that it is not enforced on the db side, so a test script created by instructors that doesn't use it (or this one in fact) is not constrained at all.
I think we should enforce it, but also care about the messages to display in those cases.

PR #3296, opened 2018-02-27
Prevent concurrent creation of groups for the same student

adisandro:
(the disable_with does not fully work on Safari, it will prevent the double click but won't display the waiting message)


adisandro:
1. Sure.
2. For blocks, yes. You want it for hashes too? I've never done that so far..

david-yz-liu:
2. Yes, I want this for hashes as well. I should have commented on this earlier, but oh well.

adisandro:
Now that I think about it, I should review the Work Alone function too, I bet there are problems there as well.

david-yz-liu:
@adisandro thanks!

PR #3309, opened 2018-03-07
tests: use a git repository instead of svn

mishaschwartz:


mishaschwartz:
This addresses Issue #3305 but ideally we would want to use a memory repository instead (of git or svn)

PR #3312, opened 2018-03-13
autotest: converts newlines in all uploaded test scripts to LF

mishaschwartz:
When test scripts and test script files and test support files are uploaded, their newline characters are converted to LF only (ie. no CRLF)

### DIFF
--- a/app/controllers/automated_tests_controller.rb
+++ b/app/controllers/automated_tests_controller.rb
@@ -22,13 +22,13 @@ class AutomatedTestsController < ApplicationController
           files.each do |file|
             File.open(file[:path], 'wb') do |f|
               content = file[:upload].read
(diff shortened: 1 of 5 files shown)


app/controllers/automated_tests_controller.rb
```diff
@@ -20,7 +20,10 @@ def update
         if @assignment.save
           # write the uploaded files
           files.each do |file|
-            File.open(file[:path], 'wb') { |f| f.write(file[:upload].read) }
+            File.open(file[:path], 'wb') do |f|
+              content = file[:upload].read
+              f.write(content.encode(content.encoding, universal_newline: true))
```
adisandro: This should support binary files too (e.g. images, zips).

PR #3315, opened 2018-03-22
test files: remove callbacks to methods that delete or write test files

mishaschwartz:
Adding, removing and replacing test files and test support files should be handled exclusively in automated_tests_controller.rb now.
These callbacks may have been causing test files and test support files to go missing after a save is called unsuccessfully.


adisandro:
Looks good!

PR #3317, opened 2018-03-22
Add support for required files in subdirs

### LINKED ISSUE #3290: assignment: Allow required files to be in subdirs

adisandro:
This pull request adds support for specifying required files in subdirs, using the already existing interface (closes #3290):
1. In the assignment properties, you can specify a usual 'file.xxx' or a new 'subdir/file.xxx'.
2. Required files in subdirs are then enforced by:
  a. Git hooks
  b. File browser
  c. Autotest repo exporter
3. As a side effect, you can now add and delete files in subdirs through the file browser.

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -338,7 +338,7 @@ class SubmissionsController < ApplicationController
     @file_manager_errors = Hash.new
     required_files = AssignmentFile.where(
                            assignment_id: @assignment).pluck(:filename)
-    students_filename = []
+    students_filenames = []
     @path = params[:path] || '/'
     @grouping = current_user.accepted_grouping_for(assignment_id)
     unless @grouping.is_valid?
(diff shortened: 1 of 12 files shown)


app/controllers/submissions_controller.rb
```diff
@@ -338,7 +338,7 @@ def update_files
     @file_manager_errors = Hash.new
     required_files = AssignmentFile.where(
                            assignment_id: @assignment).pluck(:filename)
-    students_filename = []
+    students_filenames = []
```
david-yz-liu: Since you want to change this anyways, please rename to just `filenames`.

app/controllers/submissions_controller.rb
```diff
@@ -383,67 +383,57 @@ def update_files
 
       log_messages = []
       begin
+
```
david-yz-liu: Please remove this blank line.

app/controllers/submissions_controller.rb
```diff
@@ -383,67 +383,57 @@ def update_files
...
         if new_files.empty?
           # delete files marked for deletion
           delete_files.each do |filename|
-            txn.remove(File.join(assignment_folder, filename),
-                       file_revisions[filename])
-            log_messages.push("Student '#{current_user.user_name}'" +
-                              " deleted file '#{filename}' for assignment" +
-                              " '#{@assignment.short_identifier}'.")
+            file_path_full = File.join(assignment_folder, filename)
+            file_path_relative = File.join(@path, filename)[1..-1] # remove trailing '/' to match required files
```
david-yz-liu: This seems to happen in a couple of places.

I think a better approach is to call [`Pathname.cleanpath`](https://ruby-doc.org/stdlib-2.5.0/libdoc/pathname/rdoc/Pathname.html#method-i-cleanpath) ~~both here~~ and on the actual user input for the assignment's required files.

(**Updated**) Okay I stand by what I said about the user input, but here I actually prefer using [`Pathname.relative_path_from`](https://ruby-doc.org/stdlib-2.5.0/libdoc/pathname/rdoc/Pathname.html#method-i-relative_path_from). It makes the connection between `file_path_full` and `file_path_relative` more obvious.

david-yz-liu: Related: I was confused at first because I didn't realize `assignment_folder` was a path into a subdirectory, not just `@assignment.repository_folder`. If you could rename that variable that would be great. :)

adisandro: 1. With user input you mean the assignment config page, right? I can certainly do that
2. Ok, I'll just have to make things work with Pathnames instead of strings
3. Yes

adisandro: I gave up on the .. matching with a regex.

PR #3318, opened 2018-03-23
Refactor rake task to generate repo permissions

adisandro:
This is a new task to statically create the repo auth file, by invoking the same code that Markus use.
(changes in repository.rb are only cosmetic)


adisandro:
@david-yz-liu This task fails in master because of that conditional import of subversion_repository.rb in repository.rb; I remember it was happening for some test too before I had surgery, what did we decide to do about it?

david-yz-liu:
Hi @adisandro! Two quick things:

1. Can you please target `master` rather than `1.6.1` for this. Even if this is needed this term we can just send over the rake task separately.
2. About the conditional import: we fixed the test so it no longer fails, just by changing the config setting for repository type. I'm not sure why this task should fail on a call to `Repository.get_class` (I'm assuming that's the issue), given `SubversionRepository` shouldn't even be accessed unless the config is set to `'svn'`? We can chat more offline to give me a better sense of what the problem is.

adisandro:
1. Sounds good.
2. It fails when trying to access MarkusConfigurator, because it thinks it's Repository::MarkusConfigurator

PR #3321, opened 2018-03-27
Refactor rake task to generate repo permissions

adisandro:
I had to move the Repository dependency loading into the get_class function, it's the best solution I came up with.
When running from a rake task, MarkusConfigurator would not be defined at the time the dependencies were loaded, not even if required explicitly.

(I'm creating the pull request to get your comments @david-yz-liu before doing more work, tests are broken with this change)


david-yz-liu:
@adisandro man this is annoying. Your change looks fine, but I can't help but wonder if there's a better way. Oh well, I can't articulate anything very well right now. We can talk more about this on Monday too if you want.

adisandro:
@david-yz-liu Yeah let's talk about this at the meeting.

PR #3323, opened 2018-03-28
Refactor work alone group to update repo permission once and outside the transaction

adisandro:
This applies the same principles of #3296 for the regeneration of the repo auth file.
In addition, refactors the logic around a weird corner case.


app/models/student.rb
```diff
@@ -141,45 +138,33 @@ def create_group_for_working_alone_student(aid)
         end
       end
 
-      @grouping.group = @group
-      begin
-        unless @grouping.save
-          m_logger = MarkusLogger.instance
-          m_logger.log("Could not create a grouping for Student '#{user_name}'"\
+      empty_grouping = Grouping.find_by(assignment_id: aid, group_id: @group.id)
```
david-yz-liu: I looked at the logic and the old issue. This seems unnecessary (at least now, maybe useful before). If a grouping already exists, it should still be the case that we can add the student to the grouping without destroying and recreating it.

app/controllers/assignments_controller.rb
```diff
@@ -104,15 +104,7 @@ def student_interface
...
-          # when saving a grouping throws an exception
-          unless @student.create_group_for_working_alone_student(@assignment.id)
-            # if create_group_for_working_alone_student returned false then the student
-            # must have an ( empty ) existing grouping that he is not a member of.
-            # we must delete this grouping for the transaction to succeed.
-            Grouping.find_by_group_id_and_assignment_id( Group.find_by_group_name(@student.user_name), @assignment.id).destroy
-          end
+          @student.create_group_for_working_alone_student(@assignment.id)
         rescue RuntimeError => @error
           flash_message(:error, 'Error')
```
david-yz-liu: Do you mind changing this to use the error's message? (And I think you can change `@error` to `error`). I know it's not exactly relevant, but I think it's a related-enough improvement.

app/models/student.rb
```diff
@@ -141,45 +138,33 @@ def create_group_for_working_alone_student(aid)
...
           ' come back to this page, you should inform the course instructor.'
-        end
-      # This exception will only be thrown when we try to save to a grouping that already exists
-      rescue ActiveRecord::RecordNotUnique
-        # transaction has failed, so quit it
-        return false
-      end
-
-      # We give students the tokens for the test framework
-      if @assignment.enable_test
```
david-yz-liu: Here I'm just checking that this happens somewhere else? Or is unnecessary?

app/models/student.rb
```diff
@@ -141,45 +138,33 @@ def create_group_for_working_alone_student(aid)
...
-      @grouping.update_repository_permissions
-      # Add permissions for TAs and admins.  This completely rerwrites the auth file
-      # but that shouldn't be a big deal in this case.
-      @group.set_repo_permissions
+      self.destroy_all_pending_memberships(aid)
     end
-    return true
+    Repository.get_class.__set_all_permissions
+
+    true
```
david-yz-liu: I'd prefer if you remove the return value altogether now. There are two tests in `student_test` that you'll need to modify, but that shouldn't be too long.

app/models/student.rb
adisandro: There is a unique index in the database on (assignement_id + group_id) in the groupings table.
I was able to reproduce the issue: you create your group as a student, then log in as the instructor and remove the student membership from the group (without deleting the group), then try to create it again as a student.

app/models/student.rb
adisandro: It's unnecesary, tokens are given at test request time if needed.
(but thinking about it, let me double check the interface instead, whether it hides the button without tokens)

adisandro: All good!

david-yz-liu:
@adisandro since you're still working on this: I went through and deleted the Hound comments I don't care about and will disable, but there were two that I liked and I ask that you change. Thanks!

adisandro:
@david-yz-liu I was done actually, but I'll look into these Hound comments.

app/models/student.rb
david-yz-liu: @adisandro weren't you going to replace this with a `find_or_create_by`?

adisandro: Shoot I confused this pull request with the other about git.

app/models/student.rb
```diff
@@ -141,45 +138,31 @@ def create_group_for_working_alone_student(aid)
         end
       end
 
-      @grouping.group = @group
-      begin
-        unless @grouping.save
-          m_logger = MarkusLogger.instance
-          m_logger.log("Could not create a grouping for Student '#{user_name}'"\
+      empty_grouping = Grouping.find_by(assignment_id: aid, group_id: @group.id)
+      unless empty_grouping.nil?
```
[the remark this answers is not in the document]
adisandro: I'd like to, if we weren't running ruby 2.2 in production

PR #3325, opened 2018-04-03
Use push date instead of commit date with get_all_revisions

adisandro:
With git, the repo browser was displaying revision commit times to avoid the overhead of getting push times. Push times were used for collection purposes only, and displayed in the submission list page.

That is confusing for instructors because they may think Markus has some clock problem, so this pull request now displays push times too for the current and collected revisions in git.

The long dates now include seconds, because it can be a needed level of detail to waive submissions exceeding the deadline, and gives more confidence to the instructor on the precision of results.

image

Technically, now revisions support two dates, timestamp and server_timestamp, indicating the client and server clocks. SVN just sets them to the same value, while GIT uses them to store the commit and push dates.

### DIFF
--- a/lib/repo/git_repository.rb
+++ b/lib/repo/git_repository.rb
@@ -197,12 +197,32 @@ module Repository
     end
 
     def get_all_revisions
+      # use the git reflog to get a list of pushes
+      repo_path, _sep, repo_name = @repos_path.rpartition(File::SEPARATOR)
+      bare_path = File.join(repo_path, 'bare', "#{repo_name}.git")
+      bare_repo = Rugged::Repository.new(bare_path)
+      reflog = bare_repo.ref('refs/heads/master').log.reverse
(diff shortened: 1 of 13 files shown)


lib/repo/git_repository.rb
```diff
@@ -197,12 +197,32 @@ def get_revision_by_timestamp(target_timestamp, path = nil)
...
+      push_info = OpenStruct.new
+      push_info.index = -1
+      # walk through the commits and get revisions
+      revisions = []
       walker = Rugged::Walker.new(@repos)
       walker.sorting(Rugged::SORT_TOPO | Rugged::SORT_DATE)
       walker.push(@repos.last_commit)
-      walker.map do |commit|
-        get_revision(commit.oid)
+      walker.each do |commit|
```
david-yz-liu: I'd prefer if you stick with the `map` here.

PR #3330, opened 2018-04-10
svn: try installing svn from two different mirrors

mishaschwartz:
The mirror that was previously being used to get SVN for travis-ci is either down permanently or is unreliable.  
This will try to get SVN from two (hopefully more reliable) mirrors before giving up.

### DIFF
--- a/script/install-svn.sh
+++ b/script/install-svn.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
 SVN_BASE=subversion-1.9.7
-MIRROR_PRIMARY=http://www-us.apache.org/dist
-MIRROR_SECONDARY=http://www-eu.apache.org/dist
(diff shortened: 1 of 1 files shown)


script/install-svn.sh
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 
 SVN_BASE=subversion-1.9.7
+MIRROR_PRIMARY=http://www-us.apache.org/dist
+MIRROR_SECONDARY=http://www-eu.apache.org/dist
```
adisandro: @mishaschwartz You're missing the final /subversion dir

mishaschwartz: @adisandro I think I had it lower down initially but I must have removed it. I'll fix it now.

adisandro:
We should also backport a more trivial version of this in the 1.6.1 branch (i.e. a new address only)

Issue #3298, opened 2018-02-28
feature request: show assignment median to students


kshvmdn:
For course assignments, students are shown the class average (i.e., [the mean](https://github.com/MarkUsProject/Markus/blob/35d00123cba9bea5ca750f7d01c35f89aa4015ce/app/models/assignment.rb#L286:L288)), but not the median. Was this just an oversight or done intentionally? I personally believe that median says more than mean when it comes to grades and think it'd be beneficial if students had access to both numbers.

I haven't seen the instructor/TA interface, but based on #3007, the data seems to be readily available. I doubt this would take a lot of work to implement, but I'd love to take a look if there's interest.

david-yz-liu:
Hi there @kshvmdn! Sorry for the delay in replying to this. Both assignment median and grade distribution data is currently available to instructors, so you're perfectly correct in saying that it's readily available.

You're more than welcome to implement showing this for TAs as well. However, I would not merge in such a change *unless* it was accompanied by admin-configurable settings to choose whether to display this information to students or not. This could be implemented by adding new columns to the `Assignments` table in the database.

PR #3335, opened 2018-04-11
Add "show assignment median" option to assignments

kshvmdn:
Gives admins the option to display assignment medians (set to true by default).

Here's what the student sees (hidden for A0, but shown for A2):

screen shot 2018-04-11 at 7 12 52 pm

Adds a new column to the `Assignments` table, as suggested by @david-yz-liu in https://github.com/MarkUsProject/Markus/issues/3298#issuecomment-379504913. Not very familiar with Ruby/Rails, so let me know if I missed anything.


david-yz-liu:
Thanks, @kshvmdn! Sorry t=for the delay in getting to this. Your work was great! The only change I made was for the default of the column to change from true to false (this is the conservative default).

PR #3337, opened 2018-04-13
Empty file upload warning and flash_message rewrite

mishaschwartz:
- Warn students if they upload an empty file
- Update to the flash_message function so that it can render partial html pages as well as just strings. 
- Update to the submission_controller so that it now uses the assignment_content layout


app/controllers/submissions_controller.rb
```diff
@@ -33,6 +33,8 @@ class SubmissionsController < ApplicationController
                        :populate_peer_submissions_table]
   before_filter :authorize_for_user, only: [:download, :downloads]
 
+  layout 'assignment_content'
```
david-yz-liu: Let's not change it globally for all routes here. Instead, please use the layout specifically for the actions you changed (option to `render`).

app/helpers/application_helper.rb
```diff
@@ -1,29 +1,31 @@
 # Methods added to this helper will be available to all templates in the application.
 module ApplicationHelper
 
-  # A more robust flash method. Easier to add multiple messages of each type:
-  # :error, :success, :warning and :notice
-  def flash_message(type, text)
+  # # A more robust flash method. Easier to add multiple messages of each type:
```
david-yz-liu: Revert this change (and the next line)

app/views/shared/_flash_message.html.erb
```diff
@@ -1,19 +1,10 @@
...
-    <% end %>
-  <% else %>
-    <% unless flash[key] %>
-      <%= content_tag :div, nil, class: "#{key}", style: "display: none" %>
-    <% else %>
-      <%= content_tag :div, flash[key], class: "#{key}" %>
-    <% end %>
+<% [:notice, :error, :success, :warning].each do |key| %>
+  <% unless flash[key].nil? %>
+    <% flash[key] = [flash[key]] unless flash[key].is_a?(Array)%>
```
david-yz-liu: In general please use a space inside `<% ... %>` (so add one before the closing `%>` on this line. Also see below.

app/controllers/submissions_controller.rb
```diff
@@ -440,8 +453,7 @@ def update_files
         end
         # finish transaction
         unless txn.has_jobs?
-          flash[:transaction_warning] =
-              I18n.t('student.submission.no_action_detected')
+          flash_message(:warning, I18n.t('student.submission.no_action_detected'))
           # can't use redirect_to here. See comment of this action for details.
```
david-yz-liu: I think you can delete this comment now.

app/controllers/submissions_controller.rb
```diff
@@ -455,13 +467,12 @@ def update_files
             m_logger.log(msg)
           end
         else
-          @file_manager_errors[:update_conflicts] = txn.conflicts
+          flash_message(:error, partial: 'submissions/file_conflicts_list')
```
david-yz-liu: This partial seems to use the removed `@file_manager_errors`. You could fix this by passing `txn.conflicts` as a [local variable](http://guides.rubyonrails.org/layouts_and_rendering.html#local-variables) to the partial.

app/views/submissions/_missing_assignment_file_toggle_list.html.erb
```diff
@@ -0,0 +1,20 @@
+<%= javascript_include_tag "jquery-ui" %>
```
david-yz-liu: This shouldn't be necessary. Generally you shouldn't include JS files in a partial. But also the only use of jQuery in this file is on line 4 (and this uses `jquery` not `jquery-ui`) but jQuery should already be included in this template.

app/controllers/submissions_controller.rb
```diff
@@ -350,16 +359,20 @@ def update_files
...
              f.original_filename +
              '": The size of the uploaded file exceeds the maximum of ' +
              "#{(MarkusConfigurator.markus_config_max_file_size/ 1000000.00)
 	          .round(2)}" +
              'MB.'
+          flash_message(:error, error_message)
           render :file_manager
           return
         end
+        if f.size == 0
```
david-yz-liu: Let's merge this in with the previous conditional so this becomes an `elsif`.

david-yz-liu:
Also I noticed some tests fail due to an annoying reason. I recommend thinking about how you want to fix this and running it by me before you just change all the tests (which is a pain).

app/controllers/submissions_controller.rb
mishaschwartz: I'm happy to do this.  Does it make sense then to add the layout option to **all** calls to `render: :file_manager`? Specifically, would you recommend changing the two calls in assignment_controller.rb (lines 732 and 810) to also use `layout: 'assignment_content'` as a keyword argument to render?

mishaschwartz: @david-yz-liu . I agree that this shouldn't be changed for all routes but the problem is if I only pass the layout as an keyword argument when I call render, then flash messages are not rendered properly if this page is not displayed through a call to `render` (ie. by navigating to `.../assignments/X/submissions/file_manager`). 
What I would suggest instead is changing this line to 

    layout 'assignment_content', only: [:file_manager]

That way this page will be displayed with the assignment_content layout by default **unless** the `layout` keyword argument is used when calling render.

mishaschwartz:
@david-yz-liu .  I also noticed that there is an `update_files` method in the assignment_controller as well (though it looks like it isn't currently used). Should I update this method to mirror the changes that have been made to the `update_files` method in the submissions_controller? Should this method be removed?

mishaschwartz:
@david-yz-liu 
My suggestion for handling the failing tests on the flash message content is to strip off all html content. 
For example, a line in a test currently looks like:

    expect([t('criterion_deleted_success')]).to eql(flash[:success])

so we change it to:

    expect(t('criterion_deleted_success')).to eql(Nokogiri::HTML(flash[:success][0]).text )

or even:

    expect(Nokogiri::HTML(t('criterion_deleted_success')).text).to eql(Nokogiri::HTML(flash[:success][0]).text )

if we expect there to be tags in some of our internationalized strings. 

We could even add some other string normalization steps after stripping the tags just in case we change the spacing (for example) when adding tags to the flash messages.  In either case I would probably like to create a helper function that we can call from the tests just to keep the test code clean looking.

david-yz-liu:
@mishaschwartz your proposal for changing the tests looks good to me.

PR #3346, opened 2018-04-18
repo: fix badly formatted repo checkout file

### LINKED ISSUE #2773: Bug: Malformed repository checkout list for assignments containing spaces
When using the "Subversion Checkout File" command, a malformed sequence of commands is generated if the assignment contains a space. For example, if an assignment is named "Assignment 1", the checkout file contains

```
 svn checkout -r 1 https://url/username/Assignment 1 "username"
```

The expected behaviour is that the commands can be executed in a shell to check out the file. This line should therefore read

```
 svn checkout -r 1 https://url/username/Assignment%201 "username" # Using URI escaping
```

Or, alternately

```
 svn checkout -r 1 "https://url/username/Assignment 1" "username" # Using shell quotes to group arguments.
```

MarkUs version: 1.5.0.DEV
Ruby version: 2.2.5
Rails version: 4.2.4

mishaschwartz:
Add quotation marks around repo urls in repo checkout file to handle the case where a repo url contains a space.
If it does contain a space, not wrapping the url in quotation marks makes the suggested repo checkout command malformed. 

This change was made for both git and svn repo types. 

Fixes #2773

### DIFF
--- a/lib/repo/git_repository.rb
+++ b/lib/repo/git_repository.rb
@@ -133,7 +133,7 @@ module Repository
     end
 
     def self.get_checkout_command(external_repo_url, revision_hash, group_name, repo_folder=nil)
-      "git clone #{external_repo_url} \"#{group_name}\" && "\
+      "git clone \"#{external_repo_url}\" \"#{group_name}\" && "\
       "cd \"#{group_name}\" && "\
       "git reset --hard #{revision_hash} && "\
       "cd .."
--- a/lib/repo/subversion_repository.rb
+++ b/lib/repo/subversion_repository.rb
@@ -183,7 +183,7 @@ module Repository
       unless repo_folder.nil?
         external_repo_url += "/#{repo_folder}"
       end
-      "svn checkout -r #{revision_number} #{external_repo_url} \"#{group_name}\""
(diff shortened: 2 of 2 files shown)


adisandro:
How can we end up with the repo url containing spaces?
User and group names don't, although they could in theory, so this still makes sense.

(I'm looking at the bug and I'm wondering what does it mean to check out an assignment)

mishaschwartz:
@adisandro 
It looks like the SVN urls contain the assignment short identifier. The current format seems to be:

    <host name>/markus/svn/<group name>/<assignment short id>

So for example, I can replicate the problem described in the issue if I have an instance that uses SVN and has an assignment with the short id "Assignment 1" and a group named "group001" , the repo checkout file contains the following line :

    svn checkout -r 2 http://www.example.com/markus/svn/group001/Assignment 1 "group001"

Which is where the issue with the space comes in.

adisandro:
Ah crap, svn allows to checkout subdirs, totally forgot. I'm sorry.

Alessio

david-yz-liu:
It's also true that if the assignment setting "Autogenerated group names" is false, an instructor can specify whatever group name they want, presumably including spaces.

PR #3370, opened 2018-05-09
Simplify and do fewer writes to repo permissions file 

### LINKED ISSUE #3300: repos: Always rewrite the whole auth file in a background task when permissions change

mishaschwartz:
**CHANGES:**
- all writes to the permission file are done by a single method (previously multiple methods were used)
- users with permission to access all repos can be specified once (instead of for each repo)
- write only once to the permission file after a batch update
- ensures that the permission file reflects the most up to date version of the database when multiple concurrent requests to update the file are made
- only grant TAs permission to access repos that they are currently grading
- reduce the number of times the permissions file is updated 

**ONLY UPDATE PERMISSION FILE WHEN:**
- an admin is created or destroyed
- a TA is assigned or removed as a grader
- the `vcs_submit` value is changed for an assignment
- `vcs_submit` is true for an assignment and a student membership is created destroyed or updated

For all of the cases above, the option to write the files after a batch update was used whenever appropriate. 

Closes #3300


app/models/student.rb
```diff
@@ -98,14 +98,12 @@ def invite(gid)
...
       membership.save
-      # update repo permissions (for accepted memberships - includes inviter)
-      # if grouping is valid
-      grouping = Grouping.find(gid)
-      grouping.update_repository_permissions
     end
   end
 
   def destroy_all_pending_memberships(aid)
+    # NOTE: no repository permission updates needed since user's with
```
david-yz-liu: There shouldn't be an apostrophe here.

lib/repo/repository.rb
```diff
@@ -60,8 +60,21 @@ class ExportRepositoryAlreadyExists < Exception;  end
 
   class RepositoryCollision < Exception; end
 
+  class InvalidLocation < Exception;
```
david-yz-liu: Remove the semicolon (?)

lib/repo/repository.rb
```diff
@@ -149,73 +162,104 @@ def get_revision_by_timestamp(timestamp, path = nil)
...
+        raise NotAuthorityError.new('Unable to get permissions: Not in authoritative mode!')
+      end
+      repo_name = get_repo_name
+      permissions = self.get_all_permissions
+      users = permissions[repo_name] + self.get_full_access_users
 
-    # Gets permissions for a particular user
-    def get_permissions(user_id)
-      raise NotImplementedError, "Repository.get_permissions: Not yet implemented"
+      return if users.empty?
```
david-yz-liu: This isn't type-safe (if this returns `nil`, not compatible with usage below). Better to just return an empty list, i.e., just `users`.

lib/repo/repository.rb
mishaschwartz: Actually I can remove the whole Exception. It isn't used anymore

Issue #3364, opened 2018-05-03
Don't show parent link (../) when at the top level of an assignment


david-yz-liu:
Such a link appears in both the repo browser (admins and TAs) and for the student file submission interface. We need it for navigating into subdirectories, but it shouldn't appear when at the top-level directory for that assignment (clicking on it doesn't do anything).

imvijaybala:
Claim

PR #3377, opened 2018-05-15
Dont show parent link at the top level of an assignment

imvijaybala:
Top level and current directory links will not be available as they navigate to the same page.

Fixes https://github.com/MarkUsProject/Markus/issues/3364


david-yz-liu:
Hey @imvijaybala, sorry for the delay in looking at this. I made a similar comment on #... that you can check out, but basically I prefer you not to delete the "parent link" on the client side, but instead prevent the server from including the parent link in the first place. You can start by looking at how the `repo_browser` controller method generates the data the first place and go from there.

Issue #3361, opened 2018-05-03
Hide .gitkeep in assignment subdirectories


david-yz-liu:

amrrsharaff:
Hi, I will work on this issue.

reidka:
This isn't really an easy fix.

1. First question is where to hide the subdirectories: student view, instructor view, TA view
2. Next question is changing the code so that a directory that only has a .gitkeep file in it is empty for the purposes of auto-testing or whether a student has submitted it.

reidka:
Alessio recommends to create a new global variable in the git_repository code to avoid hard-coding the name.

david-yz-liu:
The file should be hidden for the student and grading views. It can be displayed in the repo browser.

PR #3383, opened 2018-05-18
Removed .gitkeep

amrrsharaff:
This PR addresses issue #3361 
Summary: I removed .gitkeep file from the drop_down_menu in the marker's view and from the file submissions in a repo in the student's view.
Before: (Drop down menu)

screen shot 2018-05-18 at 7 40 04 pm

After: (Drop down menu)
screen shot 2018-05-18 at 7 41 41 pm

Before: (File submissions)

screen shot 2018-05-18 at 7 42 57 pm

After: (File submissions)

screen shot 2018-05-18 at 7 43 37 pm

**NOTE**:
The first selected file in the drop-down menu is hardcoded to be .gitkeep so regardless of whether the file is in the menu, it will still appear. I am open to suggestions on whether I should leave it and it will be an easy fix or do I get the first file (and what if there is no file).

### DIFF
--- a/app/views/results/common/_file_selector_dropdown.html.erb
+++ b/app/views/results/common/_file_selector_dropdown.html.erb
@@ -20,12 +20,14 @@
           <% end %>
           <% if dir[:files] != nil %>
               <% dir[:files].each do |file| %>
+                <% if file.filename != ".gitkeep" %>
                   <li class="file_item" data-path="<%= file.path %>/<%= file.filename %>" data-id="<%= file.id %>">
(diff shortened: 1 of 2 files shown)


david-yz-liu:
@amrrsharaff your comment means you haven't filtered out the `.gitkeep` file early enough in the grading view. Instead of where you currently have your filter, look for where `show_folder_recursive` is being called, and what the input file data is. Use that to find where that data is originally generated, and do the filtering there.

PR #3386, opened 2018-05-22
Adding student number column to student summary csv file

### LINKED ISSUE #3355: Student Numbers from MarkUs
Downloading detailed reports and marks sheets from MarkUs only provides the students' UTORIDs. 
This is unfortunate because submitting the final grades via CSV eMarks requires studetns' student numbers.

Is there a way for MarkUs to provide student numbers in the downloadable spreadsheets?

partoo-vk:
Fixed #3355 
Adding student number to student summary csv file:
Added column names and their values in course_summaries_controller.rb 
Added id_number to data section in course_summaries_helper.rb to be accessed by course_summaries_controller
updated all language files in config to contain column names in all language

### DIFF
--- a/app/controllers/course_summaries_controller.rb
+++ b/app/controllers/course_summaries_controller.rb
@@ -45,7 +45,7 @@ class CourseSummariesController < ApplicationController
     grade_entry_forms = GradeEntryForm.all
     marking_schemes = MarkingScheme.all
 
-    header = [I18n.t('user.user_name'),I18n.t('user.id_number')]
+    header = [I18n.t('user.user_name'), I18n.t('user.id_number')]
     header.concat(assignments.map(&:short_identifier))
     header.concat(grade_entry_forms.map(&:short_identifier))
     header.concat(marking_schemes.map(&:name))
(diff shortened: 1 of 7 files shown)


app/controllers/course_summaries_controller.rb
```diff
@@ -45,7 +45,8 @@ def generate_csv_header(csv)
     grade_entry_forms = GradeEntryForm.all
     marking_schemes = MarkingScheme.all
 
-    header = ['Username']
+    header = [I18n.t('grade_entry_forms.username')]
+    header.concat([I18n.t('grade_entry_forms.id_number')])
```
david-yz-liu: It's unnecessary to use two separate lines; just initialize the `header` to be a list containing both strings. Similar to the test down below.

config/locales/en.yml
```diff
@@ -565,6 +565,7 @@ en:
         last_name:      "Last Name"
         ngr:            "No Grade Received"
 
+
```
david-yz-liu: Please delete this line.

config/locales/en.yml
```diff
@@ -582,6 +583,8 @@ en:
       properties: "Marks Spreadsheet Properties"
       required_fields: "Required Fields <span class='required_field'>*</span>"
       show_total: "Show Total"
+      id_number: "Student Number"
```
david-yz-liu: Good idea, but please don't add new keys. Instead, use the existing labels for the attributes under `user` (roughly line 963).

Issue #3359, opened 2018-05-03
Allow a GradeEntryForm to be created with no columns (GradeEntryItems) more easily


david-yz-liu:
Right now a blank nested form for a GradeEntryItem appears by default, and if you leave it blank when creating a GradeEntryForm, the validation fails.

A fix could be to ignore blank GradeEntryItem input fields in the controller when the form is submitted.

czack810150:
Should the red * beside "name" and "out_of" be removed? I think it makes sense either way because when both fields are not filled, they are not required(validations ignored), but if one of the fields are specified, then the other one becomes required.

Screenshot: 
<img width="538" alt="screen shot 2018-05-17 at 1 47 21 pm" src="https://user-images.githubusercontent.com/17039493/40194583-f1a8757e-59d8-11e8-83a9-bef387c3ccd1.png">

david-yz-liu:
Ah, sure that's fine to remove them.

czack810150:
As discussed on Slack with David, the UI will stay as is.

PR #3391, opened 2018-05-24
Fixed issue 3359

czack810150:
Fixed [issue #3359](https://github.com/MarkUsProject/Markus/issues/3359) so that empty grade entry item won't trigger validation error when creating or editing grade entry form.


david-yz-liu:
Hi @czack810150, are you sure this is exactly the check you want? My understanding is `params[:grade_entry_items]` is actually a list of grade entry items (since there are multiple "column subforms" on the main form, one per column); if this is the case, you should really *filter out* the elements that are blank in the list, rather than dropping the list itself. You could do this in the `update_grade_entry_form_params` helper.

Issue #2538, opened 2016-05-03
Spreadsheet csv download: "Total" column


david-yz-liu:
When downloading a CSV of spreadsheet grades, the "Total" column should only appear if "show total" attribute is true on the spreadsheet. It should also have the right header and "out of" value (right now both are blank).

PR #3396, opened 2018-05-29
Spreadsheet csv download-Grade entry form, "Total" column

partoo-vk:
fixed issue #2538

Added Total and out of columns that used to be blank in the spreadsheet

Also checking if show_total is set, then show total column in csv otherwise don't generate the column
screen shot 2018-05-29 at 12 57 12 am


david-yz-liu:
@partoo-vk Can you change the "Total" column so that it shows the numerical total, rather than the percentages. (This is in line with what's shown in the web interface.) Also, there's a failing test.

PR #3399, opened 2018-05-30
Issue-3359

czack810150:
Modified grade_entry_forms_helper so that grade_entry_items with empty name and out_of will be ignored and thus no validation will be triggered on empty grade entry items.

[Demo](https://www.dropbox.com/s/9rtbnrp3s197zzk/issue-3359-demo.mov?dl=0)


david-yz-liu:
@czack810150 the deletion inside iteration isn't a good approach. Please use `delete_if`: https://apidock.com/ruby/Hash/delete_if

PR #3413, opened 2018-06-11
Merge changes from #3408 into master

david-yz-liu:
@mishaschwartz please take a look!

### DIFF
--- a/app/controllers/criteria_controller.rb
+++ b/app/controllers/criteria_controller.rb
@@ -7,18 +7,27 @@ class CriteriaController < ApplicationController
 
   def index
     @assignment = Assignment.find(params[:assignment_id])
-    if @assignment.past_all_due_dates?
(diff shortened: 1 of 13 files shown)


app/controllers/criteria_controller.rb
```diff
@@ -7,18 +7,27 @@ class CriteriaController < ApplicationController
...
+      flash_now(:notice, I18n.t('marking_started_warning'))
     end
     @criteria = @assignment.get_criteria
   end
 
   def new
     @assignment = Assignment.find(params[:assignment_id])
+    if @assignment.released_marks.any?
+      flash_now(:error, t('criteria.errors.messages.released_marks'))
+      head :bad_request
```
david-yz-liu: Note: if you don't do a render/redirect, `new.js.erb` will still be rendered automatically, causing the pane to appear. I've changed this for the actions except `edit`, since I guess users might want to see the criterion details, even if they can't actually make changes.

PR #3427, opened 2018-06-28
Converting course_summary_table to react table

partoo-vk:
Convert course_summary table to use react tables.


app/controllers/course_summaries_controller.rb
```diff
@@ -15,7 +15,15 @@ def index
 
   def populate
     if current_user.admin?
-      render json: get_table_json_data
+      table = JSON.parse(get_table_json_data)
+      data = (table[0]['assignment_marks'].map do |marks|
+
+          {
+            accessor: "assignment_marks.#{marks[0]}",
+            Header: Assignment.select("short_identifier").find(marks[0])['short_identifier']
```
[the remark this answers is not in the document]
adisandro: What you wrote is correct, we just usually prefer the syntax with symbols and object accessors:
`Assignment.select(:short_assignment).find(marks[0]).short_assignment`
There's a shorter version, but you give up the select optimization:
`Assignment.find(marks[0]).short_assignment`

If what you need is only assignment ids and short identifiers though, you can avoid json-parsing the entire table (which is big) and loop directly on a simpler db query instead, something like:
`Assignment.pluck(:id, :short_identifier)` or `Assignment.pluck_to_hash(:id, :short_identifier)`
That'll give you an array of arrays (first version) or hashes (second version) with only the info you need. You may need to tweak the ordering to match that of the table.

Anyway, this is just to give you options, show you different ways of achieving the same thing and possibly measure their efficiency if you want to. I'm totally fine with your solution.

app/assets/javascripts/Components/course_summaries_table.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+    $.ajax({
+      url: Routes.populate_course_summaries_path(),
+      dataType: 'json',
+    }).then(res => {
+      let grade_column = res.marks.map(c =>
+        Object.assign([],
+          c)
+      );
+      this.setState({data: JSON.parse(res.data),grade_columns:res.marks})
```
david-yz-liu: We discussed in our meeting that you can remove the `JSON.parse`. Also the whitespace is wonky here.

app/assets/javascripts/Components/course_summaries_table.jsx
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  fetchData() {
+
+    $.ajax({
+      url: Routes.populate_course_summaries_path(),
+      dataType: 'json',
+    }).then(res => {
+      let grade_column = res.marks.map(c =>
+        Object.assign([],
```
david-yz-liu: This is not a good use of `Object.assign`, and possibly contributes to the problem of "having" to use `JSON.parse` below. If you need to reformat the data a little bit, either do it in the controller or just manually return the correct Object here (i.e., the right dictionaries with good keys and values).

app/controllers/course_summaries_controller.rb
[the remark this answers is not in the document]
david-yz-liu: @partoo-vk yes, you should change this query based on the above comment. And please use `pluck` (or `pluck_to_hash`) for the marks spreadsheets and marking schemes too.

app/controllers/course_summaries_controller.rb
```diff
@@ -15,7 +15,15 @@ def index
...
+          }
+        end)
+      render json: {data: get_table_json_data,marks: data}
```
david-yz-liu: Overall, you should pick better key names (both `data:` and `marks:`), and pick a better variable name than `data` too.

app/assets/javascripts/Components/course_summaries_table.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+import ReactTable from 'react-table';
+
+
+class CourseSummaryTable extends React.Component {
+  constructor() {
+    super();
+    this.state = {
+      data: [],
+      grade_columns: [],
```
david-yz-liu: This and the next two field names aren't good. This particular one is really about assignments only. The next two should be pluralized.

But also it's better to just store one attribute that has *all* the columns anyway. There's no reason to store the three columns separately.

app/controllers/course_summaries_controller.rb
```diff
@@ -15,9 +15,54 @@ def index
 
   def populate
     if current_user.admin?
-      render json: get_table_json_data
+      table = get_table_json_data
+      assignment_marks = []
+      gefms = []
+      markschemes = []
+      table[0][:assignment_marks].each do |key|
```
david-yz-liu: Similar to my comment on the other PR, it is much better to use your own query here. To start try plucking just the `:id` and `:short_identifier` for assignments, and then using a `map` to generate the column data.

app/views/assignments/student_assignment_list.html.erb
```diff
@@ -1,5 +1,13 @@
 <%= content_for :head do %>
-  <%= render partial: 'course_summaries/course_summaries_table', formats: [:'js.jsx'], handlers: [:erb] %>
+  <script>
+    document.addEventListener('DOMContentLoaded', () => {
+      makeCourseSummaryTable(document.getElementById('course_summaries_table'),
+        {
+          is_admin: <%= @current_user.admin? %>
```
david-yz-liu: It doesn't seem like this is being used anywhere.

PR #3432, opened 2018-07-05
query TestRun status more easily

mishaschwartz:
Added a `status` method to the TestRun model. This method can return one of the following:

- `'complete'`
- `'in_progress'`
- `'cancelled'`

Also added the following methods to query for a specific status (returns a boolean) if the current status is the same as the method name:

- `complete?`
- `in_progress?`
- `cancelled?`

When a TestRun job is cancelled, its time_to_completion attribute is set to `0` which indicates that it has been cancelled.  Note that if the cancellation is unsuccessful for any reason (maybe the cancellation request was sent too late and the job is no longer in the queue) then a result from that test _will_ be created meaning the status of the TestRun will become `'complete'` (not `'cancelled'`).

### DIFF
--- a/app/models/test_run.rb
+++ b/app/models/test_run.rb
@@ -7,7 +7,7 @@ class TestRun < ApplicationRecord
 
   validates_presence_of :revision_identifier
   validates_numericality_of :time_to_service_estimate, greater_than_or_equal_to: 0, only_integer: true, allow_nil: true
-  validates_numericality_of :time_to_service, greater_than_or_equal_to: 0, only_integer: true, allow_nil: true
+  validates_numericality_of :time_to_service, greater_than_or_equal_to: -1, only_integer: true, allow_nil: true
 
   STATUSES = {
     complete: 'complete',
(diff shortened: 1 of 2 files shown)


adisandro:
I like this!

app/models/test_run.rb
```diff
@@ -9,6 +9,24 @@ class TestRun < ApplicationRecord
...
 
+  STATUSES = {
+    complete: 'complete',
+    in_progress: 'in_progress',
+    cancelled: 'cancelled'
+  }.freeze
+
+  def status
+    return STATUSES[:complete] unless test_script_results.empty?
+    return STATUSES[:cancelled] if time_to_service&.zero?
```
david-yz-liu: I'm not sure about using zero instead of `nil` here. Is the server guaranteed never to send back 0 for the estimated time to completion?

adisandro: Mmm you are right, it should send 0 if it's going to be served immediately.

adisandro: +1 (or maybe -1) for the negative value, a new column would be overkill

PR #3436, opened 2018-07-07
Issue 3291: Added button to stop test in queue

amrrsharaff:
This PR addresses issue #3291 and is still in progress.

**Changes:**
- There is a button "Stop Tests" which shows up for every test that is executing or in the queue.
- Each test_run is identified by the time it was created at and its id.
- route stop_tests is defined and is handled by function stop_tests


app/controllers/results_controller.rb
```diff
@@ -179,6 +179,16 @@ def run_tests
     redirect_back(fallback_location: root_path)
   end
 
+  def stop_tests
+    test_id = params[:test_run_id].to_i
+    begin
```
david-yz-liu: Don't bother with the `rescue` here (just call `AutotestCancelJob.perform_later`)

app/views/automated_tests/_test_run.html.erb
```diff
@@ -13,7 +13,11 @@
   <% end %>
   <div id="<%=test_run.created_at.to_s.delete(' ')%>">
     <% if test_script_results.empty? %>
-      TODO Test run is enqueued for execution (or lost) + button to stop it
+      Test run with id <%= test_run.id.to_s %> is in progress .. (created at: <%= test_run.created_at.to_s %>)
```
david-yz-liu: This is also a string that needs to be internationalized.

PR #3440, opened 2018-07-12
[WIP] Converting marking_schemes to react table

partoo-vk:
Converting Marking Schema table to a react table.

### DIFF
--- a/app/assets/javascripts/Components/marking_schemes_table.jsx
+++ b/app/assets/javascripts/Components/marking_schemes_table.jsx
@@ -26,11 +26,10 @@ class MarkingSchemeTable extends React.Component {
       this.setState({
         data: res.data,
         marks: res.marks,
-
       });
     });
   }
-  //edit_link = this.data['edit_link'];
+
   nameColumns = [
     {
       Header: I18n.t('marking_schemes.name'),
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/Components/marking_schemes_table.jsx
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  nameColumns = [
+    {
+      Header: I18n.t('marking_schemes.name'),
+      accessor: 'name',
+      filterable: true,
+    }
+  ];
+  modifyColumn = [
```
david-yz-liu: Add a blank line above this.

app/controllers/marking_schemes_controller.rb
```diff
@@ -12,7 +12,24 @@ def index
   end
 
   def populate
-    render json: get_table_json_data
+    table = JSON.parse(get_table_json_data)
+    assignment_weights = []
+    spreadsheet_weights = []
+    table.map do |scheme|
```
david-yz-liu: This loop doesn't make sense. You can generating columns; the number of columns should depend only on the number of assignments and spreadsheets, and nothing else. If you can't do this easily from the given data, don't use it and instead write your own entirely new query.

Issue #2926, opened 2017-04-18
Interacting with Marks Spreadsheet doesn't stop session timeout


gfbee:
I'm reviewing a TA's mark entries right now, clicking the checkboxes to keep track. That interaction doesn't appear to be enough to stop the session timeout.
Is it documented somewhere what stops session timeout for each page?

arkon:
I believe that it requires an actual page navigation (i.e. something that will actually hit the server to update the session). I don't think the way that it's being handled right now is that elegant though. While having the popup to indicate that a session is about to expire is a decent stopgap, it'd probably be better to actually ping the server in the background to keep the session rolling as long as the user is doing something that isn't necessarily a page navigation (i.e. moving their mouse, clicking things, typing, etc.).

adisandro:
Duplicate of #3303.

adisandro:
This was mostly fixed by #3448

Issue #2384, opened 2015-12-08
Disabling alerts


samirp91:
When the markus tab is kept open for a long period of time, an alert pops up and interrupts whatever is going on. There should be an option where the alert could be turned off.

reidka:
Do you mean the alert that says that the session is about to end?  (We added that so TAs would not add annotations or marks when the session was ended and their work would be lost when the page was refreshed to tell them that they need to log in again.)
If the markus tab does not have focus, does the alert interrupt the currently open tab?
What browser are you using?
Thanks.

samirp91:
It interrupts whatever I am doing whether I'm using the browser or not. I'm using chrome. Also my suggestion was not to disable it completely but have the option to disable it.

qasim:
Could [this](https://developer.mozilla.org/en-US/docs/Web/API/notification) be an alternative method of notifying the user?

arkon:
I don't think notification support is good enough to be a valid alternative.

qasim:
For reference, the browser support table for notifications is located [here](http://caniuse.com/#feat=notifications).

david-yz-liu:
Fixed in #3448.

arkon:
Also worth mentioning the behaviour of browser dialogs changed in recent(ish) versions of Chrome, so it's a bit of a non-issue: https://developers.google.com/web/updates/2017/03/dialogs-policy

PR #3448, opened 2018-07-25
Timeout imminent warning modal

### LINKED ISSUE #3303: timeout: Ajax requests fail silently

mishaschwartz:
- timeout imminent now warns as a modal instead of an alert
- users can click the `"Continue"` button on the modal to refresh the session timeout
- if the session has already timed out, the user will be redirected to the login page
- if the session times out, all ajax requests will now redirect to the login page 
- removed the `redirectModal` in favour of simply redirecting to the login page
     - this modal was ugly and buggy and interfered with the behaviour of the timeout imminent modal

Closes #3303, #2926, #2384


david-yz-liu:
After thinking about this, I'm not satisfied with a "logout on 403" behaviour, as I worry it's too sudden for the user. The old redirect modal was a better idea, even though it wasn't implemented perfectly.

Why don't you try keeping that modal but fixing the styling and message text so that its "Continue" button *always* directs to the login page.

mishaschwartz:
@david-yz-liu I completely agree, requiring four separate files to create a single modal seems like overkill. I was keeping it the same as the other modals just for consistency but I'm happy to simplify things. 

On the other hand, does it make more sense to keep this as is for this pull request and open up a new branch where I can rewrite all the modals (not just this one)? That way we can keep the way we do modals consistent and I can even try to move them all over to using react-modal in one shot.

david-yz-liu:
While it is true that the existing modals you saw under "layouts" are structured in this way, there's already been a standardization of modals that has reached every other modal except for these ones. So in fact what I want you to do is use that approach anyway (in this PR).

See, e.g., https://github.com/MarkUsProject/Markus/blob/master/app/views/assignments/index.html.erb#L45.

PR #3449, opened 2018-07-26
Made all numeric values right-aligned in react tables

partoo-vk:
Added className: 'number' to the numeric columns and text-align: right; in the matching scss files.

How the tables used to be:
screen shot 2018-08-01 at 11 22 47 pm

How the numeric values are aligned now:
screen shot 2018-08-01 at 11 24 29 pm


david-yz-liu:
@partoo-vk this is a good start, but I'd like you to not change the individual stylesheets, but instead add this to either `_core.scss` or `_markus.scss` (look for where the other custom react-table styling is currently located).

PR #3455, opened 2018-08-02
Fix the filter method for members in assignment group table

partoo-vk:
The filter used to only filter the first group member and not all.
Added a custom filtering to search for all members

### How it worked before:
screen shot 2018-08-02 at 4 51 58 pm

### How it filters now:
screen shot 2018-08-02 at 4 45 49 pm

### DIFF
--- a/app/assets/javascripts/Components/groups_manager.jsx
+++ b/app/assets/javascripts/Components/groups_manager.jsx
@@ -240,6 +240,21 @@ class RawGroupsTable extends React.Component {
           );
         }
       },
+      filterMethod: (filter, row) => {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/Components/groups_manager.jsx
```diff
@@ -240,6 +240,15 @@ class RawGroupsTable extends React.Component {
           );
         }
       },
+      filterMethod: (filter, row) => {
+        if (filter.value) {
+
```
david-yz-liu: Please remove this blank line, and the one below.

PR #3457, opened 2018-08-03
Set empty email addresses to nil

mishaschwartz:
Previously, not filling in the 'email' input when editing a user's information would set the user's email to the empty string.  If there were multiple users edited in this way, this would cause a validation failure since there would be multiple identical email addresses (multiple empty strings). This converts all empty/whitespace strings to nil before validating email addresses.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -312,7 +312,6 @@ class User < ApplicationRecord
   end
 
   def nillify_empty_email
-    self.email.present? || self.email = nil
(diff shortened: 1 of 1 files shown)


app/models/user.rb
```diff
@@ -310,4 +311,8 @@ def strip_name
     end
   end
 
+  def nillify_empty_email
+    self.email.present? || self.email = nil
```
david-yz-liu: Prefer using an inline `unless` here.

PR #3469, opened 2018-08-20
Fix infinite dependency loop (with ruby 2.3) 

mishaschwartz:
Doing an explicit "include" of a gem module when using ruby 2.3 (current production) causes an infinite dependency loop. In this case the problem was caused by including the TranslationHelper module from the I18n gem. Note that this does not appear to cause problems if using ruby 2.5


david-yz-liu:
@mishaschwartz can you change `production.rb` as well?

mishaschwartz:
the eager load settings in production.rb are already how we want them

PR #3491, opened 2018-08-30
Update 1.7.0

david-yz-liu:


app/jobs/autotest_scripts_job.rb
```diff
@@ -34,7 +34,7 @@ def perform(host_with_port, assignment_id)
           server_path = ssh.exec!("mktemp -d --tmpdir='#{server_path}'").strip # create temp subfolder
           # copy all files using passwordless scp (natively, the net-scp gem has poor performance)
           Open3.capture3('scp', '-o', 'PasswordAuthentication=no', '-o', 'ChallengeResponseAuthentication=no', '-rq',
-                         "'#{assignment_tests_path}'/.", "#{server_username}@#{server_host}:'#{server_path}'")
+                         "#{assignment_tests_path}/.", "#{server_username}@#{server_host}:'#{server_path}'")
```
adisandro: I'm sorry I'm a bit late here.
What is the problem with that quoting? The same is used for server_path.

PR #3511, opened 2018-08-31
Revert "Bump webpack-dev-server from 2.11.2 to 3.1.7"

david-yz-liu:
Reverts MarkUsProject/Markus#3480

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,6 @@
     "react-tabs": "^2.3.0"
   },
   "devDependencies": {
-    "webpack-dev-server": "3.1.7"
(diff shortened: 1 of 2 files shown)


david-yz-liu:
@adisandro looks like this isn't compatible with versions of webpack < 4.x, and webpacker only has supports this on a [pre-release](https://github.com/rails/webpacker#installation). I'm good if you want to update to that pre-release, but also if you want to wait that's okay too.

adisandro:
@david-yz-liu I didn't see that, sorry.
The problem is that I don't run it at all in development, so I just checked that it would install. Do you use it in the vm?

PR #3523, opened 2018-09-04
Clean up repos at the beginning of seeding rather than before groups

adisandro:
Starter code repos were created for the first 4 assignments and then deleted.


david-yz-liu:
Thanks!

PR #3535, opened 2018-09-11
js: Manage i18n-js with yarn/webpack

david-yz-liu:
@adisandro I looked at this last night and I think this change works okay. Please test it out and let me know if you see any issues.

### DIFF
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -14,6 +14,11 @@ import 'javascripts/jquery.easyModal';
 import Rails from 'rails-ujs';
 Rails.start();
 
+// i18n-js
+import * as I18n from 'i18n-js';
+window.I18n = I18n;
+require('translations');
+
 // chart.js
 import { Chart } from 'chart.js';
 import 'javascripts/chart_config';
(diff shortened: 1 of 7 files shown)


app/javascript/packs/application.js
```diff
@@ -14,6 +14,11 @@ import 'javascripts/jquery.easyModal';
 import Rails from 'rails-ujs';
 Rails.start();
 
+// i18n-js
+import * as I18n from 'i18n-js';
+window.I18n = I18n;
+require('translations');
```
adisandro: I was missing the require translations part, nice.
(but it breaks the build and I'm working to understand why)

adisandro: @david-yz-liu found the problem, it's working locally for you because you kept the i18n-js gem, without it we need to find a way to generate that public/javascripts/translations.js
This is someone who made it work, but in a really complicated way: https://engineering.klarna.com/migrating-from-rails-asset-pipeline-to-node-s-webpack-684230e3a93a

david-yz-liu:
Note: at some point we can try to remove the `i18n-js` gem, but that's not a high priority.

PR #3553, opened 2018-09-19
Unit to rspec tests

hl197:
Converted tests from test/unit for assignment, flexible criterion, grace period submission rule, and grouping to rspec tests at spec/models.
Removed the old test files.
Note: Assignment tests only partially converted. New failing rspec tests are marked with # FAILING.


spec/models/flexible_criterion_spec.rb
```diff
@@ -3,6 +3,111 @@
...
+  end
+
+  context 'An assignment, of type flexible criteria' do
+    before :each do
+      @assignment = create(:assignment)
+    end
+
+    it 'overwrite criterion from a 2 element row with no description' do
+      criterion = FlexibleCriterion.create_or_update_from_csv_row(['name', 10.0], @assignment)
+      expect(criterion).not_to be_nil
```
mishaschwartz: each of these `expect`s should be in their own test case

mishaschwartz:
Thanks for converting these tests, this is really useful. Also thank you for marking the failing tests, I will look into why they are failing and update the code if needed. 

As we convert tests to rspec tests (and also from old rspec tests to updated ones) we are trying to stick as much as possible with the 'better spec guidelines' outlined here: http://www.betterspecs.org/

I think it is especially important to follow the ['Single expectation test' guideline](http://www.betterspecs.org/#single) which basically states that there should be one expect statement for each test. As a first priority could you please refactor the tests so that there is a single expect statement for each test; I made a comment above showing a few examples that should be refactored.

spec/models/grace_period_submission_rule_spec.rb
```diff
@@ -429,28 +432,29 @@ class GracePeriodSubmissionRuleTest < ActiveSupport::TestCase
...
     end
   end
 
   private
 
   def submit_files_before_due_date
     pretend_now_is(Time.parse('July 20 2009 5:00PM')) do
-      assert Time.now < @assignment.due_date
-      assert Time.now < @assignment.submission_rule.calculate_collection_time
+      expect(Time.now).to be < @assignment.due_date
```
mishaschwartz: There should be no `expect` statements in helper functions. Instead, write some tests that call this helper and then check that each of your `expect`s pass.  That way, if the helper fails for any reason we see why in exactly one place instead of in every test that calls the helper. 

Also, helper functions should be put in `spec/support/helpers.rb` if they are a generic helper method (could be used by multiple specs).

spec/support/helpers.rb
```diff
@@ -22,4 +22,60 @@ def destroy_repos
   def extract_text(string)
     Nokogiri::HTML(string).text.strip.gsub(/\s+/, ' ')
   end
+
+  def submit_files_before_due_date
+    pretend_now_is(Time.parse('July 20 2009 5:00PM')) do
```
mishaschwartz: Do not use hardcoded dates (or hardcoded anything really) in helper functions.  Consider removing this function and using `submit_file_at_time` instead (below)

spec/support/helpers.rb
```diff
@@ -22,4 +22,60 @@ def destroy_repos
...
+      # expect(Time.now).to be < @assignment.due_date
+      # expect(Time.now).to be < @assignment.submission_rule.calculate_collection_time
+      # expect(Time.now).to be < @assignment.submission_rule.calculate_grouping_collection_time(@membership.grouping)
+
+      @group.access_repo do |repo|
+        txn = repo.get_transaction('test')
+        txn = add_file_helper(@assignment, txn, 'TestFile.java', 'Some contents for TestFile.java')
```
mishaschwartz: Same thing here... no hardcoded file names and also instance variables (like `@assignment` and `@group` should not be used in helper methods. This is because, this helper method could be called from any spec and there is no guarantee that this instance variable would necessarily be defined.

david-yz-liu:
Thanks, @hl197!

PR #3561, opened 2018-09-20
Add a script to reset a markus+autotesting installation

adisandro:


script/reset.sh
```diff
@@ -0,0 +1,41 @@
...
+THIS_SCRIPT_DIR=$(dirname "${THIS_SCRIPT}")
+MARKUS_ROOT=$(readlink -f "${THIS_SCRIPT_DIR}/..")
+AUTOTEST_ROOT=$(readlink -f "${MARKUS_ROOT}/../markus-autotesting")
+
+echo "[MARKUS] Shutting down existing MarkUs processes"
+pkill -f rails
+pkill -f resque
+pkill -KILL -f webpack-dev-server # shuts down only with a -KILL
+
+echo "[MARKUS] Shutting down existing autotest processes"
```
mishaschwartz: We might want to note that we should always try to kill supervisor with a -TERM signal because killing it with a -KILL signal stops supervisor but the rq workers keep running.

script/reset.sh
```diff
@@ -0,0 +1,41 @@
...
+rm -rf "${MARKUS_ROOT}"/public/javascripts
+rm -rf "${MARKUS_ROOT}"/public/packs
+rm -rf "${MARKUS_ROOT}"/tmp/cache
+
+echo "[MARKUS] Removing old autotest files"
+rm -rf "${AUTOTEST_ROOT}"/server/workspace/scripts/*
+rm -rf "${AUTOTEST_ROOT}"/server/workspace/results/*
+
+echo "[MARKUS] Removing old redis entries"
+redis-cli --raw DUMP autotest:workers | head -c-1 >| /tmp/.redis.tmp # make the dump play nice with shell commands
```
mishaschwartz: Instead of hardcoding `autotest:workers` we should probably try to read the `config.py` script in the autotester since this key may not always be named the same thing

script/reset.sh
```diff
@@ -0,0 +1,41 @@
...
+redis-cli FLUSHDB
+cat /tmp/.redis.tmp | redis-cli -x RESTORE autotest:workers 0
+
+echo "[MARKUS] Resetting MarkUs database"
+pushd "${MARKUS_ROOT}" > /dev/null
+bundle exec rake db:reset
+popd > /dev/null
+
+echo "[MARKUS] Restarting MarkUs"
+"${MARKUS_ROOT}"/script/start-autotest-workers.sh
```
mishaschwartz: We should probably run this line before we do the `bundle exec rake db:reset` since that will send jobs to rq (like updating test scripts)

PR #3562, opened 2018-09-20
display all tests run by admins in batch table if current user is an admin

mishaschwartz:


david-yz-liu:
I don't think it's a good idea to only show collected submissions. It may be good to access test results from previously collected versions in this table.

A future addition could be to pass along a boolean specifying whether the submission is collected, enabling filtering to be done client-side. But that also doesn't belong in this pull request.

david-yz-liu:
@mishaschwartz I will discuss this with you further. But what you should do now is remove the commit you made so that the other one can go through. (Incidentally, this is why small pull requests are preferred.)

PR #3564, opened 2018-09-20
indicate errors in test batch table

mishaschwartz:
* add new status for test runs: completed_with_errors
* in the submissions table, show a mark of 0

### DIFF
--- a/app/assets/javascripts/Components/submission_table.jsx
+++ b/app/assets/javascripts/Components/submission_table.jsx
@@ -163,7 +163,7 @@ class RawSubmissionTable extends React.Component {
     {
       Header: I18n.t('activerecord.attributes.result.total_mark'),
       accessor: 'final_grade',
-      Cell: ({value}) => value ? value + ' / ' + this.props.max_mark : '',
+      Cell: ({value}) => value + ' / ' + this.props.max_mark,
       className: 'number',
       minWidth: 80,
       filterable: false,
--- a/app/assets/javascripts/Components/submission_table.jsx
+++ b/app/assets/javascripts/Components/submission_table.jsx
@@ -163,7 +163,7 @@ class RawSubmissionTable extends React.Component {
     {
       Header: I18n.t('activerecord.attributes.result.total_mark'),
       accessor: 'final_grade',
-      Cell: ({value}) => value + ' / ' + this.props.max_mark,
+      Cell: ({value}) => (value === undefined ? '-' : value) + ' / ' + this.props.max_mark,
       className: 'number',
       minWidth: 80,
       filterable: false,
--- a/app/models/test_run.rb
+++ b/app/models/test_run.rb
@@ -17,7 +17,7 @@ class TestRun < ApplicationRecord
   }.freeze
 
   def status
-    unless test_script_results.empty?
+    if test_script_results.exists?
       if test_script_results.joins(:test_results).where('test_results.completion_status': 'error').count&.positive?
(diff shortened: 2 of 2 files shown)


app/assets/javascripts/Components/submission_table.jsx
```diff
@@ -163,7 +163,7 @@ class RawSubmissionTable extends React.Component {
     {
       Header: I18n.t('activerecord.attributes.result.total_mark'),
       accessor: 'final_grade',
-      Cell: ({value}) => value ? value + ' / ' + this.props.max_mark : '',
+      Cell: ({value}) => value + ' / ' + this.props.max_mark,
```
david-yz-liu: The string concatenation of `undefined` with `' / ' + this.props.max_mark` would lead to something like `"undefined / 10"`, wouldn't it? This isn't great. We could do something like `(value === undefined ? '-' : value) + ' / ' + ...`

app/models/test_run.rb
```diff
@@ -12,23 +12,33 @@ class TestRun < ApplicationRecord
...
     in_progress: 'in_progress',
-    cancelled: 'cancelled'
+    cancelled: 'cancelled',
+    complete_with_errors: 'complete_with_errors'
   }.freeze
 
   def status
-    return STATUSES[:complete] unless test_script_results.empty?
+    unless test_script_results.empty?
+      if test_script_results.joins(:test_results).where('test_results.completion_status': 'error').count&.positive?
```
david-yz-liu: Hmm can't you use `exists?` or `any?` here?

app/models/test_run.rb
```diff
@@ -12,23 +12,33 @@ class TestRun < ApplicationRecord
...
     complete: 'complete',
     in_progress: 'in_progress',
-    cancelled: 'cancelled'
+    cancelled: 'cancelled',
+    complete_with_errors: 'complete_with_errors'
   }.freeze
 
   def status
-    return STATUSES[:complete] unless test_script_results.empty?
+    unless test_script_results.empty?
```
david-yz-liu: The structure here is pretty funky. It would be good to change this to a flatter conditional:

```
if test_script_results.(with an error).exists?
  STATUSES[:complete_with_errors]
elsif test_script_results.exists?
  STATUSES[:complete]
elsif time_to_service&.negative?
  STATUSES[:cancelled]
else
  STATUSES[:in_progress]
end
```

app/models/test_run.rb
mishaschwartz: I like `empty?` because it returns false for an array of null values but you're not going to get that with an active record relation. I'll change it.

app/models/test_run.rb
mishaschwartz: I actually thing the nested if statements make sense in this context. The `complete_with_errors` status is only ever assigned to test runs we would otherwise give a `complete` status to, so it makes sense to treat one as a subset of the other in the logic.  

Also this mirrors the way I have structured the logic in the `TestRun.status` method.  If you want, I can do something like


      if test_script_results_id
        if completion_status == 'error' || status_hash[id] == STATUSES[:complete_with_errors]
          status_hash[id] = STATUSES[:complete_with_errors]
          next
        end
        status_hash[id] = STATUSES[:complete]
      elsif time_to_service&.negative?
        status_hash[id] = STATUSES[:cancelled]
      else
        status_hash[id] = STATUSES[:in_progress]
      end

To make it a little more clear that the `complete` status is the default in this case and the `complete_with_errors` is a special case.

PR #3584, opened 2018-09-29
submission-file: Fix get-file encoding and subdirectory issues

david-yz-liu:
@adisandro I assumed you didn't have time to do this last week, so I took a stab at it today. Please review, and try out in svn.

Two fixes:
1. Using String#force_encoding (doesn't modify the bytes, does modify what encoding is interpreted as). This has the plus side that actual UTF-8 files in svn *should* be actually rendered properly. This has the downside that non-UTF-8 encoded files won't be. If you have time to look into other options that would be good, otherwise I'm inclined to merge this in for now.

2. Fixes loading of plaintext files in subdirectories (before, they wouldn't load). This is a bug that should have affected both git and svn, so I'm surprised this hasn't been reported yet.

I also changed the error-reporting to something that would actually show up on the page. Couldn't get the flash message to work with this AJAX request, maybe because we're using `fetch` rather than `$.ajax` for this one. Something to look into later.

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -478,18 +478,15 @@ class SubmissionsController < ApplicationController
     elsif file.is_pdf?
       render json: { type: 'pdf' }
     else
-      path = params[:path] || '/'
       grouping.group.access_repo do |repo|
         revision = repo.get_revision(submission.revision_identifier)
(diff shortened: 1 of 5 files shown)


adisandro:
All good on my dev machine, both git and svn.

PR #3585, opened 2018-10-01
Unit tests to Rspec conversion

nkhi:
student_spec passes 7 of 27, some tests hidden.
ta_spec passes 1 of 2.
submission_rule_s…pec passes 54/63. Grace period id context and Grace Period ids contexts hidden.

### DIFF
--- a/spec/models/student_spec.rb
+++ b/spec/models/student_spec.rb
@@ -16,161 +16,189 @@ require 'spec_helper'
 require 'shoulda'
 
 describe Student do
-
-  # Tests for a generic Student instantiation
   context 'A good Student model' do
 
-    it { should have_many(:accepted_groupings).through(:memberships) }
-    it { should have_many(:pending_groupings).through(:memberships) }
-    it { should have_many(:rejected_groupings).through(:memberships) }
-    it { should have_many :student_memberships }
-    it { should have_many :grace_period_deductions }
-    it { should belong_to :section }
+    it 'will have many accepted groupings' do
+      is_expected.to have_many(:accepted_groupings).through(:memberships)
+    end
 
-    it { should validate_numericality_of :grace_credits }
+    it 'will have many pending groupings' do
+      is_expected.to have_many(:pending_groupings).through(:memberships)
+    end
+
+    it 'will have many rejected groupings' do
+      is_expected.to have_many(:rejected_groupings).through(:memberships)
+    end
+
+    it 'will have many student memberships' do
+      is_expected.to have_many :student_memberships
+    end
+
+    it 'will have many grace period deductions available' do
+      is_expected.to have_many :grace_period_deductions
+    end
+
+    it 'will belong to a section' do
+      is_expected.to belong_to :section
+    end
+
+    it 'will have some number of grace credits' do
+      is_expected.to validate_numericality_of :grace_credits
+    end
   end
 
-  # context 'A pair of students in the same group' do
-  #     @membership1 = StudentMembership.new(membership_status: StudentMembership::STATUSES[:inviter])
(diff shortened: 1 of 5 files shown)


david-yz-liu:
Hello @nkhi, the main issue with the tests is that you want to use factory_bot to create instances of the models, rather than calling `new` everywhere. Using, e.g., `create(:student)` is the simplest way to get an instance of the class that also has an id.

You can read more about the basics of factory bot at https://www.rubydoc.info/gems/factory_bot/file/GETTING_STARTED.md#Using_factories. (Really, only the first two methods `build` and `create` should be necessary for now.)

spec/models/student_spec.rb
```diff
@@ -0,0 +1,326 @@
...
+  #
+  #         #You have to find the students to get the updated values
+  #         students = Student.find(@student_id_list)
+  #
+  #         expect(item[:expected]).eql?(students[0].grace_credits)
+  #         expect(item[:expected]).eql?(students[1].grace_credits)
+  #       end
+  #     end
+  #   end
+  # end
```
mishaschwartz: If code has been replaced/updated somewhere else, please remove commented out old code.  If you are saving this as a reference for later please either indicate that with a comment somewhere or add 'Work in Progress' or 'WIP' to the description of this pull request so we know you will come back to this later and the pull request should not be merged right away.

spec/models/student_spec.rb
```diff
@@ -0,0 +1,326 @@
...
+
+    context 'and a grouping' do
+      it 'should be invited to a grouping' do
+        grouping = Grouping.new
+        @student.invite(grouping.id)
+
+        pending_memberships = @student.pending_memberships_for(grouping.assignment_id)
+
+        expect(pending_memberships).not_to be_nil
+        expect(pending_memberships.length).to eq(1)
```
mishaschwartz: Please use a single `expect` statement per test (one `expect` for every `it` context). It makes it easier to determine exactly why a test has passed or failed.

spec/models/student_spec.rb
```diff
@@ -0,0 +1,326 @@
...
+  #
+  #       pending_memberships = @student.pending_memberships_for(grouping.assignment_id)
+  #
+  #       expect(pending_memberships).not_to be_nil
+  #       expect(pending_memberships.length).to eq(0)
+  #     end
+  #   end
+
+  context 'A Student' do
+    @student = Student.new
```
mishaschwartz: Please use [FactoryBot](https://github.com/thoughtbot/factory_bot) as much as possible to create objects for testing.  Also, it is preferable to use [rspec's let and let!](https://relishapp.com/rspec/rspec-core/v/2-5/docs/helper-methods/let-and-let) methods instead of setting instance variables inside a context block.

spec/models/student_spec.rb
```diff
@@ -0,0 +1,326 @@
...
+#  - with no duplicates and no sections
+#  - with duplicates and no sections
+#  - with no duplicates and sections
+#  - with duplicates and sections
+#  - with no duplicates and one section
+#  - with duplicates and sections and update of a section
+#  - with an invalid file
+
+require 'spec_helper'
+require 'shoulda'
```
mishaschwartz: We're trying to move away from using `shoulda`. Please write all new tests using the `expect` syntax.  Also, any additional `require`s for your tests should be in the `spec_helper` file, not in each individual spec file.  

For a general overview of rspec best practices please see:

http://www.betterspecs.org

david-yz-liu:
Merged in 619cd930ac7d7b4368d3dcf319d8e4ce19913a82.

PR #3599, opened 2018-10-05
Migrate marked to webpack

adisandro:


david-yz-liu:
Thanks @adisandro. I should point out that I found one lingering instance of [importing marked in a JS file](https://github.com/MarkUsProject/Markus/commit/3c0470275b0acdaab0884095aa05d104b6baae84#diff-32acce37941092c979d72ff91eeda3dfL2). I removed it, but just FYI it came up when I searched for `'marked'` in the project files.

adisandro:
Nice catch, I must have missed it.

Your merge commit though has reintroduced an old version of the main.js file:
https://github.com/MarkUsProject/Markus/commit/3c0470275b0acdaab0884095aa05d104b6baae84#diff-975dcbe34a69a8c368dd76b1b5d5f6ae

david-yz-liu:
@adisandro lol yeah, you're totally right! Must have got confused between "added both" and "deleted both". I'll remove this, or you can if you have time before me.

PR #3607, opened 2018-10-09
fix bug caused by newline in annotation description

mishaschwartz:
causes syntax error if there is a newline in the content


david-yz-liu:
Eh, I think space is fine, at least for now. Is there a reason this is still WIP?

mishaschwartz:
Because I forgot to change the title after the last commit. Sorry

PR #3609, opened 2018-10-09
Tighten Assignments table schema

hl197:
**Summary**: Tightened Assignment table schema to not allow null values for booleans and added defaults.

**Problem**: Some boolean fields in the assignments table allowed for null values which could lead to a [Three-state Boolean Problem](https://robots.thoughtbot.com/avoid-the-threestate-boolean-problem). This is an issue since having NULL when a boolean is expected can lead to unexpected results.

**Changes**: Defaults and not allowing NULL values were added for all boolean fields through the migration `ChangeAssignmentsDetailsNotNull`. The new schema was generated and added.


db/schema.rb
```diff
@@ -80,19 +79,19 @@
...
     t.datetime "updated_at"
-    t.boolean "student_form_groups"
-    t.boolean "group_name_autogenerated"
-    t.boolean "group_name_displayed"
+    t.boolean "student_form_groups", default: false, null: false
+    t.boolean "group_name_autogenerated", default: true, null: false
+    t.boolean "group_name_displayed", default: false, null: false
     t.string "repository_folder", null: false
-    t.boolean "invalid_override"
+    t.boolean "invalid_override", default: false, null: false
```
adisandro: @david-yz-liu What is this flag for?

mishaschwartz:
@david-yz-liu should this close PR #3638 ?

PR #3616, opened 2018-10-12
Upgrade jquery

adisandro:
The usual suspect pages work, but I checked it quickly, I may still miss something.


david-yz-liu:
@adisandro I'll probably sit on this one for a while until I read more about it myself. But thanks for kicking this off.

If you want to keep doing the webpacker migration off of this branch that's find with me.

PR #3647, opened 2018-10-24
Assignment Controller method refactoring

nkhi:
Modified methods in Assignment controller for download_assignment_list and upload_assignment_list to call methods within the assignment model.


app/controllers/assignments_controller.rb
```diff
@@ -616,45 +616,28 @@ def view_summary
   end
 
   def download_assignment_list
-    assignments = Assignment.all
-
+    @output = Assignment.get_assignment_list(params[:file_format], DEFAULT_FIELDS)
```
david-yz-liu: Make `output` a *local variable* (don't put the `@` in front of it). In general, the `@` should be used rarely, only when the variable needs to be passed to the view.

Also a reminder to move `DEFAULT_FIELDS` into the model.

app/controllers/assignments_controller.rb
```diff
@@ -616,45 +616,28 @@ def view_summary
...
-                  type: 'text/csv', disposition: 'attachment',
-                  filename: 'assignment_list.csv')
-        return
-      else
-        flash_message(:error, t(:incorrect_format))
-        redirect_to action: 'index'
-        return
+    when 'yml'
+      send_data(@output,
+                filename: "assignments_#{Time.
```
david-yz-liu: Reminder to unify the two calls to `send_data`.

app/controllers/assignments_controller.rb
```diff
@@ -616,45 +616,28 @@ def view_summary
...
+                type: 'text/yml', disposition: 'inline')
+    when 'csv'
+      send_data(@output,
+                type: 'text/csv', disposition: 'attachment',
+                filename: 'assignment_list.csv')
+    # else should never be reached under normal use
+    else
+      flash_message(:error, t(:incorrect_format))
+      redirect_to action: 'index'
+      return
```
david-yz-liu: You don't need the `return` here at the method's end.

app/controllers/assignments_controller.rb
```diff
@@ -664,49 +647,25 @@ def upload_assignment_list
...
-            row[:non_regenerating_tokens] = false
-            row[:unlimited_tokens] = false
-            update_assignment!(row)
-          end
-        rescue ActiveRecord::ActiveRecordError, ArgumentError => e
-          flash_message(:error, e.message)
-          redirect_to action: 'index'
-          return
+    when 'csv'
+      @result = Assignment.upload_assignment_list('csv', assignment_list, DEFAULT_FIELDS)
```
david-yz-liu: Similar to above, make `result` a local variable.

app/controllers/assignments_controller.rb
```diff
@@ -664,49 +647,25 @@ def upload_assignment_list
...
+
+    when 'yml'
+        @result = Assignment.upload_assignment_list('yml', assignment_list, DEFAULT_FIELDS)
+        if @result.is_a?(error)
+          flash_message(:error, @result.message)
         end
-      else
-        return
+    else
+      return
```
david-yz-liu: Don't do the `return` here. (Redirecting to `index` is still good.)

But as in the download method, add a `flash_message` to report an incorrect format.

app/models/assignment.rb
```diff
@@ -1464,4 +1464,68 @@ def same_membership_as_csv_row?(row, existing_grouping)
...
+      map = {}
+      map[:assignments] = []
+      assignments.map do |assignment|
+        m = {}
+        default_fields.length.times do |i|
+          m[default_fields[i]] = assignment.send(default_fields[i])
+        end
+        map[:assignments] << m
+      end
+      output = map.to_yaml
```
david-yz-liu: Ruby's functions automatically return the last evaluated value.

This means that you can remove the `output` variable altogether, and simply write `map.to_yaml` here, and that's what will be returned. (Similarly in the "csv" branch, simply write `MarkusCSV.generate(assignments) do ...` and the value will be returned.

app/models/assignment.rb
```diff
@@ -1464,4 +1464,68 @@ def same_membership_as_csv_row?(row, existing_grouping)
...
+  def self.get_assignment_list(file_format, default_fields)
+    assignments = self.all
+
+    case file_format
+    when 'yml'
+      map = {}
+      map[:assignments] = []
+      assignments.map do |assignment|
+        m = {}
+        default_fields.length.times do |i|
```
david-yz-liu: Use `default_fields.map` to loop over elements rather than indexes.

Also `map` *returns* a list, so you can actually do `map[:assignments] = default_fields.map ...`, and remove the `<< m` part in the loop body.

app/models/assignment.rb
```diff
@@ -1464,4 +1464,68 @@ def same_membership_as_csv_row?(row, existing_grouping)
...
+        default_fields.map do |f|
+          assignment.send(f.to_s)
+        end
+      end
+    end
+    return output
+  end
+
+  def self.upload_assignment_list(file_format, assignment_list, default_fields)
+
```
david-yz-liu: Remove this blank line. Also, the indentation below is off (should be indented inside the `def`).

app/models/assignment.rb
```diff
@@ -1464,4 +1464,68 @@ def same_membership_as_csv_row?(row, existing_grouping)
...
+      output = MarkusCSV.generate(assignments) do |assignment|
+        default_fields.map do |f|
+          assignment.send(f.to_s)
+        end
+      end
+    end
+    return output
+  end
+
+  def self.upload_assignment_list(file_format, assignment_list, default_fields)
```
david-yz-liu: Change the `assignment_list` name to `assignment_data`.

app/models/assignment.rb
```diff
@@ -1464,4 +1464,68 @@ def same_membership_as_csv_row?(row, existing_grouping)
...
+
+  case file_format
+  when 'csv'
+    result = MarkusCSV.parse(assignment_list) do |row|
+    assignment = self.find_or_create_by(short_identifier: row[0])
```
david-yz-liu: The indentation for this block is funny, please take a careful look.

app/models/assignment.rb
```diff
@@ -1464,4 +1464,68 @@ def same_membership_as_csv_row?(row, existing_grouping)
...
+    attrs = Hash[default_fields.zip(row)]
+    attrs.delete_if { |_, v| v.nil? }
+      if assignment.new_record?
+         assignment.submission_rule = NoLateSubmissionRule.new
+         assignment.assignment_stat = AssignmentStat.new
+         assignment.token_period = 1
+         assignment.non_regenerating_tokens = false
```
david-yz-liu: These may be default values. Check with @hl197 to see if they can be removed.

app/models/assignment.rb
```diff
@@ -1464,4 +1464,68 @@ def same_membership_as_csv_row?(row, existing_grouping)
...
+    return result
+
+  when 'yml'
+    begin
+      map = YAML::load(assignment_list)
+      map[:assignments].map do |row|
+        row[:submission_rule] = NoLateSubmissionRule.new
+        row[:assignment_stat] = AssignmentStat.new
+        row[:token_period] = 1
+        row[:non_regenerating_tokens] = false
```
david-yz-liu: See above comment about default values.

app/models/assignment.rb
hl197:
```suggestion
         assignment.non_regenerating_tokens = false
```
can be removed since it is the default.

spec/models/submission_rule_spec.rb
```diff
@@ -602,4 +620,109 @@
...
+  context 'Assignment with a past due date' do
+    before(:each) do
+      @assignment = create(:assignment, due_date: 2.days.ago)
+    end
+
+    it 'can collect submission files' do
+      expect(@assignment.due_date.eql?(@assignment.submission_rule.get_collection_time))
+
+      # due date is two days ago, so it can be collected
+      expect(@assignment.submission_rule.can_collect_all_now?).to be true
```
mishaschwartz: Please make this a separate test case

spec/models/submission_rule_spec.rb
```diff
@@ -602,4 +620,109 @@
...
+    end
+
+    it 'should sort in ascending order' do
+      # Loop through periods for this SubmissionRule and verify the ids are sorted in ascending order
+      previous_id = @submission_rule.periods[0][:id]
+      for i in (1..4) do
+        expect(@submission_rule.periods[i][:id]).to be > previous_id
+        previous_id = @submission_rule.periods[i][:id]
+      end
+    end
```
mishaschwartz: Please use one `expect` statement per `it` context. See https://stackoverflow.com/a/34858160/5992438 for suggestions of how to create `it` contexts in a loop

app/controllers/assignments_controller.rb
```diff
@@ -441,97 +439,44 @@ def view_summary
...
   end
 
   def upload_assignment_list
     assignment_list = params[:assignment_list]
-
+    file_format = params[:file_format]
     if assignment_list.blank?
       flash_message(:error, I18n.t('csv.invalid_csv'))
       redirect_to action: 'index'
-      return
```
david-yz-liu: This `return` is necessary (this is the reason for the third test failure).

app/models/assignment.rb
```diff
@@ -135,7 +135,7 @@ class Assignment < ApplicationRecord
   after_create :build_repository
 
   before_save :reset_collection_time
-  after_save :update_permissions_if_vcs_changed
+  before_save :update_permissions_if_vcs_changed
```
david-yz-liu: This shouldn't be changed (it is the reason for the first test failure).

spec/controllers/assignments_controller_spec.rb
```diff
@@ -99,20 +99,15 @@
...
       # for some reason, assignments aren't always cleared from the db
       # before these tests
-      Assignment.all.each do |asn|
-        asn.delete
-      end
-      @assignment = FactoryBot.create(:assignment)
+    #  Assignment.all.each do |asn|
+    #    asn.delete
+    #  end
     end
```
david-yz-liu: You need to comment this out as well, and uncomment the end at the bottom of this file. (I think this is the cause of the second error, with `let(:assignment)` ending its scope here.

david-yz-liu:
Merged in eeb8bff.

PR #3654, opened 2018-10-28
Fixed N+1 query problems when loading student's assignment view

hl197:
**Summary:** Fixed N+1 query problem when loading student assignment view by pre-caching results using `.includes`.

**Problem:** N+1 queries are made when finding the accepted grouping for each assignment and current submission used for each accepted grouping when loading the student's assignment view.

**Changes:**
1. In AssignmentsController#index, all `accepted_groupings` for the student are loaded at once with `assignment` and `current_submission_used` pre-cached. `@a_id_results` is filled in by looping over the the accepted groupings instead of assignments.
2. In CourseSummariesController#populate, the helper function `get_mark_for_all_assignments_for_student` now pre-caches `current_submission_used` for each accepted grouping.


app/controllers/assignments_controller.rb
```diff
@@ -208,16 +208,14 @@ def index
...
-          if grouping.has_submission?
-            submission = grouping.current_submission_used
-            if submission.has_remark? && submission.remark_result.released_to_students
-              @a_id_results[a.id] = submission.remark_result
-            elsif submission.has_result? && submission.get_original_result.released_to_students
-              @a_id_results[a.id] = submission.get_original_result
-            end
+      accepted_groupings = current_user.accepted_groupings.includes(:assignment, :current_submission_used)
+      accepted_groupings.each do |grouping|
+        if grouping.has_submission?
```
david-yz-liu: I realized because of this change, there should be an extra check here for `!grouping.assignment.id_hidden` (if the assignment is hidden, this result should not be shown to students yet).

app/controllers/assignments_controller.rb
```diff
@@ -208,16 +208,14 @@ def index
...
-          if grouping.has_submission?
-            submission = grouping.current_submission_used
-            if submission.has_remark? && submission.remark_result.released_to_students
-              @a_id_results[a.id] = submission.remark_result
-            elsif submission.has_result? && submission.get_original_result.released_to_students
-              @a_id_results[a.id] = submission.get_original_result
-            end
+      accepted_groupings = current_user.accepted_groupings.includes(:assignment, :current_submission_used)
+      accepted_groupings.each do |grouping|
+        if !grouping.assignment.id_hidden && grouping.has_submission?
```
david-yz-liu: There's a typo on this line with the assignment attribute. (The fact that this was not caught means you didn't test the code in all possible cases here; not a big deal, but please keep in mind for next time.)

david-yz-liu: (I realize I made the original typo, but still good to double-check.)

hl197: Yes, sorry I should have caught this! I realized I deployed the wrong version when I tested.

PR #3656, opened 2018-10-29
Hide autotest extra_info from student views

adisandro:


david-yz-liu:
Eh, it's a bit more work but this should really be done server-side. There's a method `Grouping#test_runs_students` that looks promising.

adisandro:
I wanted to potentially keep it available for scripts run by students, but I can do that too server side.

PR #3660, opened 2018-10-30
display warning in file viewer if file is binary

mishaschwartz:


david-yz-liu:
@mishaschwartz you can target master for this one. Also, would be good to change the type to `'error'`, and also modify the comment on line 493 (right above the changed line).

PR #3667, opened 2018-11-02
display error messages when adding new students

mishaschwartz:


david-yz-liu:
@mishaschwartz I looked at this and found an underlying issue, which has been fixed in #3680.

PR #3678, opened 2018-11-09
display warning if file is binary

mishaschwartz:
PR #3660 but moved to target master instead of 1.7.0

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -495,8 +495,8 @@ class SubmissionsController < ApplicationController
         end
 
         if SubmissionFile.is_binary?(file_contents)
-          # If the file appears to be binary, send it as a download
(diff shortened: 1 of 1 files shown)


app/controllers/submissions_controller.rb
```diff
@@ -495,8 +495,8 @@ def get_file
         end
 
         if SubmissionFile.is_binary?(file_contents)
-          # If the file appears to be binary, send it as a download
-          render plain: 'not a plaintext file'
+          # If the file appears to be binary, display a warning
+          render json: { content: I18n.t('submission_file.error.binary_file_message').to_json, type: 'error' }
```
david-yz-liu: Sorry, I was wrong about this. type should be `'unknown'`, otherwise the syntax highlighter won't be able to handle it.

PR #3680, opened 2018-11-11
repsonders: Fix flash message configuration for responders gem

david-yz-liu:
@adisandro it looks like the setting `config.responders.flash_keys` didn't work in the initializer file. Maybe you know why?


adisandro:
@david-yz-liu I don't know exactly what the responders gem is supposed to do, let me look at it.
It's indeed weird that an initializer won't work.

Issue #3577, opened 2018-09-26
Review all authorizations and migrate them to use action_policy


mishaschwartz:
possibly update to handle permissions with pundit?

mishaschwartz:
We have decided to go with action_policy instead of pundit for authorizations

david-yz-liu:
Closed in #4865.

PR #3683, opened 2018-11-14
Authorize running tests with action_policy

adisandro:
(i18n integration is not released yet in the gem, and it has a few quirks, but the code is much better with it)

### DIFF
--- a/app/controllers/automated_tests_controller.rb
+++ b/app/controllers/automated_tests_controller.rb
@@ -80,6 +80,7 @@ class AutomatedTestsController < ApplicationController
       @grouping.refresh_test_tokens
       # authorization
       begin
+        authorize! @assignment, to: :run_tests? # TODO: Remove it when reasons will have the dependent policy details
         authorize! @grouping, to: :run_tests?
         @authorized = true
       rescue ActionPolicy::Unauthorized => e
@@ -96,6 +97,7 @@ class AutomatedTestsController < ApplicationController
       assignment = Assignment.find(params[:id])
       grouping = current_user.accepted_grouping_for(assignment.id)
       grouping.refresh_test_tokens
+      authorize! assignment, to: :run_tests? # TODO: Remove it when reasons will have the dependent policy details
       authorize! grouping, to: :run_tests?
       grouping.decrease_test_tokens
       test_scripts = assignment.select_test_scripts(current_user)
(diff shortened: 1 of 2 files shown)


adisandro:
@david-yz-liu I think I'll go with the flattening, to make the policy functionality easier to understand rather than mimic the policy logic exactly.
The text will be a bit repeated and christmas tree-like, that's the con side.

app/controllers/automated_tests_controller.rb
```diff
@@ -78,7 +78,16 @@ def student_interface
 
     unless @grouping.nil?
       @grouping.refresh_test_tokens
+      # authorization
+      begin
+        authorize! @grouping, to: :run_tests?
+        @authorized = true
+      rescue ActionPolicy::Unauthorized => e
+        @authorized = false
+        @reason = e.result.reasons.full_messages.join(' ')
```
david-yz-liu: I'd prefer keeping flash messages in the controller rather than the view.

spec/policies/assignment_policy_spec.rb
```diff
@@ -0,0 +1,110 @@
+describe AssignmentPolicy do
+  include PolicyHelper
+
+  describe '#run_tests?' do
+
+    context 'when the user is an admin' do
+      context 'if enable_test is false' do
+        it do
+          user = build(:admin)
```
david-yz-liu: While I think the overall structure and choice to use local variables is totally fine, I think that the setting of `user` is static and independent enough that it should be moved into a `let` at the top of each top-level context block.

app/controllers/automated_tests_controller.rb
adisandro: This change made me discover a regression in the messages displayed, a bit more work to do :(

spec/policies/assignment_policy_spec.rb
adisandro: I may be wrong, but with the custom matcher approach and without the stubbing, the let syntax for everything may be feasible again. Let me do a quick test.

adisandro:
@david-yz-liu all done, but I'll have to file a bug report with the gem for the reasons not properly tracked between dependent policies

mishaschwartz:
This is the first step to resolving Issue #3577

Issue #3639, opened 2018-10-21
Annotation removes code from a line


reidka:
![syntax-highlightbug](https://user-images.githubusercontent.com/420794/47271157-d2be1d80-d543-11e8-818a-7279e1d03254.jpeg)

Line 150 is longer in the student's submission.  I selected most of the text on this line and then added an annotation.  The annotation doesn't show, but also the text I annotated disappeared.

mishaschwartz:
I'm still unable to recreate this bug so I'm going to close this for now. If this continues to happen, please re-open this issue.

Issue #3306, opened 2018-03-02
codeviewer: Fix annotations not being displayed


adisandro:
(and possibly update to a newer version of it, or a better viewer)

PR #3692, opened 2018-11-21
Edit or remove annotation from annotations table

mishaschwartz:


mishaschwartz:
This allows users to remove a problematic annotation but doesn't fully solve the issues in #3639 #3306

PR #3702, opened 2018-11-27
Use a single repository for assignments' starter code

adisandro:


app/lib/repository.rb
```diff
@@ -84,7 +84,7 @@ def self.open(connect_string)
 
     # Static method: Creates a new repository at given location; returns
     # an AbstractRepository instance, with the repository opened.
-    def self.create(connect_string)
+    def self.create(connect_string, with_hooks:)
```
david-yz-liu: This colon thing really bugs me. Can we set a default here?

app/models/assignment.rb
```diff
@@ -1094,53 +1094,42 @@ def create_peer_review_assignment_if_not_exist
...
-      repo.commit(txn)
-    end
-    true
   end
 
-  def repo_loc
-    repo_loc = File.join(MarkusConfigurator.markus_config_repository_storage, repository_name)
+  def starter_code_repo_path
+    repo_loc = File.join(MarkusConfigurator.markus_config_repository_storage, STARTER_CODE_REPO_NAME)
     unless Repository.get_class.repository_exists?(repo_loc)
```
david-yz-liu: Since you're changing this method, I'd prefer putting the `exists?` check into `starter_code_repo` and `access_starter_code_repo` to keep this method pure. It would allow you to use the method up in `build_starter_code_repo`, for example.

app/models/assignment.rb
```diff
@@ -1094,53 +1094,42 @@ def create_peer_review_assignment_if_not_exist
...
-      txn = repo.get_transaction('Markus')
-      txn.add_path(repository_folder)
-      repo.commit(txn)
-    end
-    true
   end
 
-  def repo_loc
-    repo_loc = File.join(MarkusConfigurator.markus_config_repository_storage, repository_name)
+  def starter_code_repo_path
```
david-yz-liu: It would be nice to make this, `starter_code_repo`, and `access_starter_code_repo` class methods.

This, combined with the my comment below, should allow you to use `starter_code_repo` in the view to display the repo url.

adisandro:
@david-yz-liu I will address the comments next week, unless you want it urgently.

app/models/assignment.rb
david-yz-liu: Yes you're right, my bad.

PR #3706, opened 2018-11-28
report bonuses and deductions in assignment summary

### LINKED ISSUE #3368: Downloading the Detailed Report Should Include Raw Mark
When I download the Detailed CSV Report for an assignment I would like there to be a column in the CSV that corresponds to the total marks the student received before any late penalties were applied.

Special cases sometimes require late penalties to be removed of reduced, and as it is right now this cannot be done in the MarkUs web interface without completely removing the TA's grades.

Instead, I would like to do this manually. Currently the CSV has a column for the late penalty that was applied. Please include a column totaling the marks before that penalty. This allows instructors to manually calculate percent score and apply late penalties if we need.

Thanks

mishaschwartz:
closes #3368

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -688,7 +688,8 @@ class Assignment < ApplicationRecord
         final_grade: result && result.total_mark,
         criteria: result.nil? ? {} : result.mark_hash,
         result_id: result && result.id,
-        submission_id: result && result.submission_id
+        submission_id: result && result.submission_id,
+        total_extra_marks: result && result.get_total_extra_marks
       }
     end
     criteria_columns = self.get_criteria(:ta).map do |crit|
@@ -721,7 +722,7 @@ class Assignment < ApplicationRecord
                     .where('memberships.user_id': user.id)
     end
 
-    headers = [['User name', 'Group', 'Final grade'], ['', 'Out of', self.max_mark]]
+    headers = [['User name', 'Group', 'Final grade', 'Bonus/Deductions'], ['', 'Out of', self.max_mark, '']]
     criteria = self.get_criteria(:ta)
     criteria.each do |crit|
       headers[0] << crit.name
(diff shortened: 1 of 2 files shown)


david-yz-liu:
This is good, but it looks like `get_total_extra_marks` is recalculating the assignment max mark for each grouping. Can you work on speeding this up before I merge it?

app/models/assignment.rb
```diff
@@ -677,6 +677,7 @@ def summary_json(user)
                     .where('memberships.user_id': user.id)
     end
 
+    maximum_mark = max_mark(user.admin? ? :all : :ta)
```
david-yz-liu: You want `:ta` here, which is the default (the `:all` and `:peer` are only used for peer reviews).

app/models/result.rb
```diff
@@ -90,15 +90,21 @@ def get_subtotal(user_visibility: :ta, assignment: nil)
 
   # The sum of the bonuses deductions and late penalties
   #
-  # See the documentation for update_total_mark for information about when to explicitly
-  # pass the +assignment+ variable and associated warnings.
-  def get_total_extra_marks(user_visibility: :ta, assignment: nil)
+  # If the +max_mark+ value is nil, its value will be determined dynamically.
```
david-yz-liu: remove the period

app/models/result.rb
```diff
@@ -90,15 +90,21 @@ def get_subtotal(user_visibility: :ta, assignment: nil)
...
+  # based on the max_mark value of the associated assignment.
+  # However, passing the +max_mark+ value explicitly is more efficient if we are
+  # repeatedly calling this method where the max_mark doesn't change, such as when
+  # all the results are associated with the same assignment.
+  #
+  # +user_visibility+ is passed to the Assignment.max_mark method to determin the
```
david-yz-liu: there's an "e" missing on "determine"

app/models/result.rb
```diff
@@ -90,15 +90,21 @@ def get_subtotal(user_visibility: :ta, assignment: nil)
...
+  # max_mark value only if the +max_mark+ argument is nil
```
david-yz-liu: There should be a period at the end of this sentence.

david-yz-liu:
@mishaschwartz I merged this in but there's still an N+1 query on retrieving extra marks for each result. This should be fixed soon; probably the best approach is to compute all the deductions in bulk (in a separate Hash), and the merge this Hash into the result data hash. So the line

```ruby
        total_extra_marks: result&.get_total_extra_marks(max_mark: maximum_mark)
```

would become something like

```ruby
        total_extra_marks: total_extra_marks_hash[result.id]
```

PR #3708, opened 2018-11-29
add authenticity token to form in students table

mishaschwartz:


mishaschwartz:
This resolves one of the suggestions from the security audit

david-yz-liu:
This is interesting. Rails 5 introduced [per-form CSRF tokens](https://github.com/rails/rails/pull/22275), and it would be nice to use those here. 

The easiest way to get them is to use Rails helpers like `form_with`, which isn't possible in this context. However, you should be able to pass in the token as a React prop to the table, using [per_form_csrf_token](https://api.rubyonrails.org/classes/ActionController/RequestForgeryProtection.html#method-i-per_form_csrf_token).

config/application.rb
```diff
@@ -47,6 +47,9 @@ class Application < Rails::Application
     # NOTE: This didn't work when put in config/initializers/responders.rb.
     config.responders.flash_keys = [:success, :error]
 
+    # Use per form CSRF tokens for all controllers
+    Rails.configuration.action_controller.per_form_csrf_tokens = true
```
adisandro: This is already loaded by default on line 17 of this file (try looking at the source of load_defaults), it's not necessary.

PR #3721, opened 2018-12-05
Eliminate n+1 queries for calculating bonus/deductions for results

mishaschwartz:
This was prompted by a comment by @david-yz-liu on PR #3706

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -666,7 +666,7 @@ class Assignment < ApplicationRecord
                               :accepted_students,
                               :inviter,
                               :tas,
-                              current_result: [:marks, :extra_marks_points, :extra_marks_percentage])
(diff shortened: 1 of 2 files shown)


app/models/assignment.rb
```diff
@@ -732,7 +733,8 @@ def summary_csv(user)
     headers[0] << 'Bonus/Deductions'
     headers[1] << ''
 
-    maximum_mark = max_mark(:all)
+    result_ids = groupings.pluck('results.id').uniq.compact
+    extra_marks_hash = Result.get_total_extra_marks(result_ids, max_mark: max_mark(:all))
```
david-yz-liu: I just noticed this in the diff, but we should use `max_mark`, not `max_mark(:all)` here. (This was a bug from before, but is good to fix here anyway.)

PR #3722, opened 2018-12-05
Allow updating starter code after groups are created

adisandro:
I have one question for you @david-yz-liu:

I can't get this new starter code form to work with 'post' in react because of some authenticity token quirk (but 'get' just works with it instead), @mishaschwartz was looking at it too and didn't find anything obvious.

Should we extract everything that is not managing the files out of react?

### DIFF
--- a/app/views/assignments/_file_manager.html.erb
+++ b/app/views/assignments/_file_manager.html.erb
@@ -7,8 +7,9 @@
           repo_url: '<%= File.join(MarkusConfigurator.markus_config_repository_external_base_url,
                                    Assignment::STARTER_CODE_REPO_NAME) %>',
           groupsExist: <%= @assignment.groups.size == 0 ? 'false' : 'true' %>,
-          authenticity_token: '<%= form_authenticity_token(form_options: { action: update_starter_code_assignment_url,
-                                                                           method: 'post' }) %>'
+          authenticity_token:
+            '<%= form_authenticity_token(form_options: { action: update_starter_code_assignment_url(locale: nil),
+                                                         method: 'post' }) %>'
         }
       );
     });


david-yz-liu:
@adisandro can you push your work with the the POST and I can take a look.

adisandro:
@david-yz-liu done (and the code will not work now)

app/views/assignments/_file_manager.html.erb
```diff
@@ -6,7 +6,9 @@
           assignment_id: <%= @assignment.id %>,
           repo_url: '<%= File.join(MarkusConfigurator.markus_config_repository_external_base_url,
                                    Assignment::STARTER_CODE_REPO_NAME) %>',
-          readOnly: <%= @assignment.can_upload_starter_code? ? 'false' : 'true' %>
+          groupsExist: <%= @assignment.groups.size == 0 ? 'false' : 'true' %>,
+          authenticity_token: '<%= form_authenticity_token(form_options: { action: update_starter_code_assignment_url,
```
david-yz-liu: The problem is caused by a discrepancy between the Rails path helpers and the `js-routes` path helpers. Looks like by default, the former uses the current locale (`en`), but the js-routes version doesn't. I guess this normally doesn't matter, but here it does because these are the strings used to generate the authenticity tokens.

You can fix this here by passing `locale: nil` to the path helper. I guess a longer-term solution would be to make `js-routes` use the current locale when generate paths. For some reason I thought it already did that, but I guess now.

mishaschwartz: @david-yz-liu Good catch. I have implemented the `locale: nil` solution for now and made the Issue #3725 so that we can make the long-term change later on.

PR #3727, opened 2018-12-06
Don't display flash message if not authorized to run tests in results view

mishaschwartz:
If tests are not enabled for an assignment then the "Test Results" tab in hidden anyway. This makes a flash message about not being able run tests potentially confusing. 

If there are additional reasons why you are not authorized to run tests then these will be triggered by the `run_tests` method in the `ResultsController` when the user actually attempts to run the tests


david-yz-liu:
I think sometimes the "Run Tests" button doesn't appear at all, and this was one reason to use a global flash message.

PR #3728, opened 2018-12-07
Show the first file in the code viewer on page load 

mishaschwartz:
Previously this was enabled for the instructor/grader view but not for the student view or if a grader was marking an assignment (exam) file with both a criteria and assignment file association.


david-yz-liu:
Thanks @mishaschwartz. The only thing I noticed is that there's some `console.log` calls that should be removed.

mishaschwartz:
@david-yz-liu Oops my bad. They were left over from debugging. I've removed them

PR #3731, opened 2018-12-10
Fix internationalization string discovery in marks_graders table

mishaschwartz:
It looks like I18n can't discover the correct string properly if it is called from within a static structure


adisandro:
@mishaschwartz targeting 1.7.1 if bugfix?

david-yz-liu:
@mishaschwartz @adisandro can one of you please look into fixing this globally using an approach like https://gist.github.com/bazzel/ecdff4718962e57c2d5569cf01d332fe?

mishaschwartz:
@david-yz-liu yes I can look into implementing a global solution

PR #3766, opened 2019-01-04
Fixed calculating subtotals and totals in the markers pane for peer reviews

hl197:
**Summary:** If the current user is a peer review, correctly calculate subtotal and total by referring to peer review assignment.

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -281,7 +281,11 @@ class Assignment < ApplicationRecord
   def max_mark(user_visibility = :ta)
     # TODO: sum method does not work with empty arrays. Consider updating/replacing gem:
     #       see: https://github.com/thirtysixthspan/descriptive_statistics/issues/44
-    max_marks = get_criteria(user_visibility).map(&:max_mark)
+    if user_visibility == :peer
+      max_marks = pr_assignment.get_criteria(user_visibility).map(&:max_mark)
+    else
+      max_marks = get_criteria(user_visibility).map(&:max_mark)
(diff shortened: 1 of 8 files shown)


app/models/assignment.rb
```diff
@@ -281,7 +281,11 @@ def display_for_note
   def max_mark(user_visibility = :ta)
     # TODO: sum method does not work with empty arrays. Consider updating/replacing gem:
     #       see: https://github.com/thirtysixthspan/descriptive_statistics/issues/44
-    max_marks = get_criteria(user_visibility).map(&:max_mark)
+    if user_visibility == :peer
```
david-yz-liu: Don't do it like this. Look for where `max_mark` is called, and call it on either the assignment or `pr_assignment` as appropriate.

app/views/results/marker/_marker_summary.html.erb
```diff
@@ -16,7 +16,7 @@
 
   <div class='summary_block'>
     <span><strong><%= t('results.subtotal') %></strong>:</span>
-    <span id='current_subtotal_div'> <%= result.get_subtotal %></span>
+    <span id='current_subtotal_div'> <%= result.get_subtotal(user_visibility: @result.is_a_review? ? :peer : :ta) %></span>
```
david-yz-liu: It's a bit better to move the `is_a_review` check directly into the `Result#get_subtotal` method. This might mean we can remove the `user_visibility` parameter from this method altogether (check other usages).

PR #3774, opened 2019-01-07
Migrate result controller tests to rspec

mishaschwartz:
This moves the tests for the results controller from rails tests to rspec tests. Currently it is a simple translation of the tests from one framework to the other (no new tests were added).


david-yz-liu:
@mishaschwartz can you please try adding this to the rubocop config: https://stackoverflow.com/a/41689335

PR #3789, opened 2019-01-16
Count number of submitted files properly for student view

mishaschwartz:
* include files in subdirectories in the count
* don't include git internal files (such as .gitkeep) in the count

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -615,7 +615,7 @@ class AssignmentsController < ApplicationController
       @last_modified_date = @revision&.server_timestamp
       files = @revision.tree_at_path(assignment.repository_folder)
                        .select do |_, obj|
-                         obj.is_a?(Repository::RevisionFile) && !GitRepository.internal_file_names.include?(obj.name)
+                         obj.is_a?(Repository::RevisionFile) && !Repository.get_class.internal_file_names.include?(obj.name)
                        end
       @num_submitted_files = files.length
       missing_assignment_files = grouping.missing_assignment_files(@revision)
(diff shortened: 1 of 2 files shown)


app/lib/git_repository.rb
```diff
@@ -585,7 +585,7 @@ def changes_at_path?(path)
 
   # Gets all entries at directory +path+ of a specified +type+ (:blob, :tree, or nil for both), as
   # Repository::RevisionFile and Repository::RevisionDirectory.
-  def entries_at_path(path, type=nil)
+  def entries_at_path(path, type=nil, ignore_internal=true)
```
adisandro: I think you should filter the .gitkeep files externally, instead of making the api more complex.
(i.e. get all the entries, then drop the ones you don't need)

app/controllers/assignments_controller.rb
```diff
@@ -613,7 +613,10 @@ def set_repo_vars(assignment, grouping)
     grouping.group.access_repo do |repo|
       @revision = repo.get_revision_by_timestamp(Time.current, assignment.repository_folder)
       @last_modified_date = @revision&.server_timestamp
-      files = @revision.files_at_path(assignment.repository_folder)
+      files = @revision.tree_at_path(assignment.repository_folder)
+                       .select do |_, obj|
+                         obj.is_a?(Repository::RevisionFile) && !GitRepository.internal_file_names.include?(obj.name)
```
david-yz-liu: It seems like you should only check for this when MarkUs is actually using git.

adisandro: Good catch, it should be Repository.get_class.internal_file_names to keep it repo-agnostic.

mishaschwartz: Thanks for catching that. I will update this but please don't pull this in @david-yz-liu until PR #3790 is finished so I can rebase on top of that one.

david-yz-liu: @mishaschwartz ready for you to rebase.

PR #3804, opened 2019-01-22
RSpec Tests for Rubric_Criterion 

ashih2018:
Finished converting RSpec tests for 'A good rubric criterion model' and 'A rubric criterion assigning a TA'


spec/models/rubric_criterion_spec.rb
```diff
@@ -2,4 +2,138 @@
...
+      expect(@criterion.criterion_ta_associations.count).to be == 0, 'Got unexpected TA membership count'
+      @criterion.add_tas(@ta)
+      expect(@criterion.criterion_ta_associations.count).to be == 1, 'Got unexpected TA membership count'
+      @ta.reload
+      @criterion.remove_tas(@ta)
+      expect(@criterion.criterion_ta_associations.count).to be == 0, 'Got unexpected TA membership count'
+    end
+
+    it 'assign multiple TAs' do
+      @ta1 = create(:ta)
```
david-yz-liu: For local variables like `ta1`, don't use the `@` symbol (that's for instance attributes).

spec/models/rubric_criterion_spec.rb
```diff
@@ -2,4 +2,138 @@
...
+  end
+
+  context 'A rubric criterion assigning a TA' do
+    before(:each) do
+      @criterion = create(:rubric_criterion)
+      @ta = create(:ta)
+    end
+
+    it 'assign a TA by id' do
+      expect(@criterion.criterion_ta_associations.count).to be == 0, 'Got unexpected TA membership count'
```
david-yz-liu: Use `.to eq 0` (and similarly below)

spec/models/rubric_criterion_spec.rb
```diff
@@ -2,4 +2,233 @@
...
+        end
+        # ...containing commas and quotes in the descriptions
+        (0..RubricCriterion::RUBRIC_LEVELS - 1).each do |i|
+          @row << 'description' + i.to_s + ' with comma (,) and ""quotes""'
+        end
+        @csv_base_row = @row
+      end
+    end
+
+    context 'be able to create a new instance without level descriptions' do
```
david-yz-liu: Your problem here is that the `before(:each)` above isn't triggering because it's in a separate `context` block. You need to make this test a `should` rather than a `context`, and make sure it's nested within the `context` block above. Same thing with the next test below (starting on line 221).

PR #3830, opened 2019-02-12
Sync with the markus-autotesting development

adisandro:
The basic 'custom' tester to execute client scripts now works with the new autotesting infrastructure. Better test run table details and other testers will come later.

(The custom tester is the one needed to run the seeded autotesting assignment, and it's equivalent to uploading a custom test script)


app/assets/javascripts/Components/test_run_table.jsx
```diff
@@ -176,7 +174,7 @@ class TestScriptResultTable extends React.Component {
   ];
 
   render() {
-    const extraInfo = this.props.data[0]['test_script_results.extra_info'] || '';
```
david-yz-liu: Why did you change the `const` here?

app/controllers/api/test_results_controller.rb
```diff
@@ -134,9 +134,7 @@ def update
 
     # User params for create & update
     def test_result_params
-      params.permit(:name, :completion_status, :marks_earned,
-                    :input, :actual_output, :expected_output, :created_at,
-                    :updated_at)
+      params.permit(:name, :status, :marks_earned, :output, :created_at, :updated_at)
```
david-yz-liu: Happened to notice the `:created_at` and `:updated_at`. They'd only be necessary if the autotesting server were passing these.

app/controllers/automated_tests_controller.rb
```diff
@@ -59,15 +44,12 @@ def update
   # Manage is called when the Automated Test UI is loaded
   def manage
     @assignment = Assignment.find(params[:assignment_id])
-    @assignment.test_scripts.build(
+    @assignment.test_groups.build(
       # TODO: make these default values
```
david-yz-liu: Since we're doing a migration, why don't we add these default values as well.

app/controllers/automated_tests_controller.rb
```diff
@@ -114,43 +96,17 @@ def execute_test_run
...
-      else
-        flash_message(:error, I18n.t('automated_tests.download_wrong_place_or_unreadable'))
-        redirect_to action: 'manage'
-      end
+    assignment = Assignment.find(params[:assignment_id])
+    file_path = File.join(AutomatedTestsClientHelper::ASSIGNMENTS_DIR, assignment.short_identifier, params[:filename])
+    if File.exist?(file_path)
+      file_contents = IO.read(file_path)
+      send_file file_path,
+                type: ( SubmissionFile.is_binary?(file_contents) ? 'application/octet-stream':'text/plain' ),
```
david-yz-liu: Not sure why this isn't a Rubocop thing, but I'd prefer having spaces around the `:` in the ternary if.

app/models/test_group.rb
```diff
@@ -0,0 +1,56 @@
+class TestGroup < ApplicationRecord
+  TO_INSTRUCTORS = 'instructors'.freeze
```
david-yz-liu: Would be nice to use enums. https://api.rubyonrails.org/v5.2.2/classes/ActiveRecord/Enum.html

app/models/test_group_result.rb
```diff
@@ -0,0 +1,73 @@
+class TestGroupResult < ApplicationRecord
+  has_many :test_results, dependent: :destroy
+  belongs_to :test_group
+  belongs_to :test_run
+
+  validates :marks_earned, :marks_total, presence: true, numericality: { greater_than_or_equal_to: 0 }
+  validates :time, presence: true, numericality: { greater_than_or_equal_to: 0, only_integer: true }
+
+  def create_test_result(name:, output: '', marks_earned: 0.0, marks_total: 0.0, status: 'error', time: nil)
```
david-yz-liu: I actually think it's better to remove this method altogether and stick with the Rails-conventional `test_results.create(...)`. You can pick default values for the attributes you want, but otherwise I'd just require passing in things like `status` explicitly.

db/schema.rb
```diff
@@ -517,18 +517,44 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "test_group_results", id: :serial, force: :cascade do |t|
+    t.integer "test_group_id"
```
david-yz-liu: While we're changing the schema, it would be great to make this non-null.

db/schema.rb
```diff
@@ -517,18 +517,44 @@
...
+    t.float "marks_earned", default: 0.0, null: false
+    t.datetime "created_at"
```
david-yz-liu: And the timestamps non-null as well.

db/migrate/20190130143545_change_test_run_errors_to_problems.rb
```diff
@@ -0,0 +1,6 @@
+class ChangeTestRunErrorsToProblems < ActiveRecord::Migration[5.2]
```
david-yz-liu: Since I've asked for a few schema changes anyway, I'd prefer consolidating all the migrations into one file.

app/controllers/api/test_results_controller.rb
adisandro: You're right, didn't think about that.

app/assets/javascripts/Components/test_run_table.jsx
adisandro: I didn't want it to convey some meaning that is really not there.
Are we into using const and let consistently? If we're not, I'd rather just use let for simplicity.

app/models/test_group.rb
adisandro: Cool, I'll try them

db/schema.rb
adisandro: It should be nullable, for the case when the test group (previously script) gets deleted while tests are being done on the autotester. It actually really happened in production.

db/migrate/20190130143545_change_test_run_errors_to_problems.rb
adisandro: Should I rewrite history for this? (otherwise, individual commits will be broken)

app/assets/javascripts/Components/test_run_table.jsx
david-yz-liu: I'm trying to use `const` more now (and consistently), so I'd prefer changing this back.

adisandro:
@david-yz-liu this should be good now

PR #3835, opened 2019-02-18
Update rspec test for annotation_category controller

YolandaZIK:
**There are 6 failed test cases. 2 failed test cases for csv_upload, 4 failed test cases for yml_upload**

- csv_upload 

1. Failed test case `csv_upload does not accept files with invalid columns`
`Line 66`: if a .csv file contains a row of annotation texts without an annotation category name, then the file should not be uploaded. 

2. Failed test case `csv_upload does not accept any file without .csv extension`
`Line 136`: if a file does not contain .csv as an extension, the file should not be uploaded. However, the current implementation parses the file and update parsed information as annotation categories. (This test case mimics uploading a .yml file via `csv_upload`)

The current implementation of `csv_upload` does not allow users to upload .xls file and non-csv file with .csv extension. In both cases, `csv_upload` prompts the same error suggesting that "The selected file is not a CSV file (even though it may have a .csv extension)".

- yml_upload

1. Failed test case `yml_upload does not accept fileless submission`
`Line 188`: this test case mimics calling `yml_upload` without a given file to upload, but `yml_upload` did not prompt an error as `csv_upload` does when being called without a selected file.

2. Fail test case 
`yml_upload does not accept files with empty annotation category name`
`Line 181`: I tried testing whether the webpage is redirected to an expected page. Then it fails. 
(I suggest that `Line 180` and `Line 181` test the same thing, but why one of them fails? )

Samp problems in 
`yml_upload does not accept a .xls file` on `Line 223`
`yml_upload does not accept a non-yml file with .yml extension` on `Line 206`

**Problems with prompted errors on `yml_upload`**

When `yml_upload` is called with a _non-yml file with .yml extension_ or a _.xls file_, it reports there is error in the file. I suppose it is better to tell users that they select a non-yml file even though the file has a .yml extension as `csv_upload` prompts `unparseable_csv` error.


david-yz-liu:
Merged in 98375115a438318c997d9ca9ee00ac3a13f08b9b.

PR #3843, opened 2019-02-25
Upload YML file

ashih2018:
AssignmentsController upload_assignment_list accepts a valid YAML file
     Failure/Error: update_assignment!(row)

     NoMethodError:
       undefined method `update_assignment!' for #<Class:0x000000000780c5f8>
       Did you mean?  update_counters

I cannot tell whether this is an error with my YML file or the assignment.rb file as when I debugged with byebug, it would call method missing when update assignment is supposed to be called.


ashih2018:
Fixed the problem before which was that the method I tried to call was private. Now I am running into another error when I try to assignment.update_attributes!(map) when update_attribute is called and I am not sure what the error is.

app/controllers/assignments_controller.rb
```diff
@@ -659,19 +672,6 @@ def get_file_info(file_name, file, assignment_id, path)
     }
   end
 
-    def update_assignment!(map)
```
david-yz-liu: @ashih2018 okay I finally understand the problem. It was an error introduced by a previous refactoring that was never caught because we didn't test yml uploads here (so it's great that you're adding to this).

Basically you should delete this method entirely (i.e., don't just move it, delete it), and do the following:

1. Move the code that deals with creating and updating the assignment directly into the body of the loop in the `Assignment.upload_assignment_list` method. Note that the `submission_rule` and `assignment_stat` attributes seem to be set both in that loop and in this method here, so you only need to keep one.

2. *Don't* worry about keeping the `flash_message` call, you can just delete that.

3. You can delete the `@testing_controller` attribute you added in the `upload_assignment_list` method, since you'll no longer need a controller instance. We never should need to use controller methods inside a model.

ashih2018:
The test at line 205 does not work. It seems that I pass in the information as Assignment and it is the value they are looking for. However, when get_assignment_list[yml] is called in the Assignment class, when it does map.to_yaml, it adds many "\n" characters and edits the assignment data in a way that is not supposed to.

david-yz-liu:
Squashed and merged in 235ceda4df1a131c7deac8be9bcd92c881343f8e.

PR #3869, opened 2019-03-04
Handle corrupt git repos

mishaschwartz:
When a git repository is corrupted for any reason, fix it by checking out again before giving access. This also fixes the problem where an error is raised when trying to close an uninitialized repository object (in the `ensure` block).

This problem was causing background jobs that accessed failed repositories to fail silently. For example, required files were not being updated properly for all students if any student repos were corrupted. 

We still need to figure out exactly why student repositories are being corrupted


adisandro:
Looks good.

PR #3873, opened 2019-03-07
Documented criteria_controller's upload/download functions

YolandaZIK:
Under this branch, there should not be any duplicate from commits on branch `annotation_category_file_template`.

I upload a YML file template for users to upload rubric/flexible/checkbox grading criteria.

I upload a template YML file to show what files users will have when they download all grading criteria on MarkUs.

Changes in test cases:
      - Pending test case: When users upload files with undefined keys, we'd better show a flash message that some grading criteria are not successfully uploaded.
      - Pending test case: Markus will round up a maximum mark to one digit after decimal point.


doc/imported_samples/marking_criteria.yml
```diff
@@ -0,0 +1,36 @@
+Quality of Writing:
```
david-yz-liu: Specify a `type:` here. You can also leave a comment in the YML file saying that rubric is the default.

doc/imported_samples/marking_criteria.yml
```diff
@@ -0,0 +1,36 @@
...
+    level_0:
+        name: Beginner
+        description: The essay is very poorly organized structure and gives no new information.
+    level_1:
+        name: Capable
+        description: The essay is poorly organized but gives new information.
+    level_2:
+        name: Accomplished
+        description: The essay is well-structure and conveys new information clearly.    
+    ta_visible: False
```
david-yz-liu: For the first entry, show default values (`ta_visible` is true and `peer_visible` is false by default.

doc/imported_samples/marking_criteria.yml
```diff
@@ -0,0 +1,36 @@
+Quality of Writing:
+    max_mark: 24.0
+    level_0:
+        name: Beginner
+        description: The essay is very poorly organized structure and gives no new information.
+    level_1:
+        name: Capable
+        description: The essay is poorly organized but gives new information.
+    level_2:
```
david-yz-liu: Also, your example should show all four levels for each rubric criterion.

doc/imported_samples/marking_criteria.yml
```diff
@@ -0,0 +1,56 @@
+Quality of Writing:
+    type: Rubric
```
david-yz-liu: Use lowercase words for the type (so `rubric` instead of `Rubric`). Same applies to `Checkbox` below.

doc/imported_samples/marking_criteria.yml
```diff
@@ -0,0 +1,56 @@
...
+    level_2:
+        name: Accomplished
+        description: The essay is well-structure and conveys new information clearly.
+    level_3:
+        name: level 3 criterion
+        description: level 3 description
+    level_4:
+        name: level 4 criterion
+        description: level 4 description
+    ta_visible: True
```
david-yz-liu: Similarly, prefer lowercase `true` and `false`.

PR #3877, opened 2019-03-11
Don't collect if a released submission already exists

mishaschwartz:
You can still manually collect an individual submission even if the result is released from the repository view

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -203,25 +203,36 @@ class SubmissionsController < ApplicationController
     end
     assignment = Assignment.includes(:groupings).find(params[:assignment_id])
     groupings = assignment.groupings.find(params[:groupings])
-    partition = groupings.partition do |grouping|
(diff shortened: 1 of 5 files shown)


app/controllers/submissions_controller.rb
```diff
@@ -203,25 +203,36 @@ def collect_submissions
     end
     assignment = Assignment.includes(:groupings).find(params[:assignment_id])
     groupings = assignment.groupings.find(params[:groupings])
-    partition = groupings.partition do |grouping|
+    collectable = []
+    some_before_due = false
+    some_released = Grouping.joins(current_submission_used: :results)
+                            .where('results.released_to_students': true)
+                            .where(id: groupings)
```
david-yz-liu: I'd add a `.pluck(:id).to_set` here. It's not a big deal in most cases, but would matter somewhat if the instructor accidentally selects a whole bunch of released submissions. (I just tried it one ~200 submissions from our seed data.)

PR #3879, opened 2019-03-12
Unify exporting student files for autotesting

adisandro:
Don't export the whole repo for student tests any more, only the assignment files and only the required files. This unifies the behavior with the instructor tests, allowing cross-caching of exported files as well.


david-yz-liu:
This is great. Two comments:

1. Seems like the database schema should enforce that a test run's `revision_identifier` is not null? I feel like maybe we've had this conversation before, but I can't remember. 
2. I think the changes to the file `submission_file.rb` are independent of this, and perhaps you meant to do them separate (e.g., as part of #3880). I'd rather remove them here and put them there.

adisandro:
1) It can be null in the submissions table (and so it can be null in the test_runs table too): it signals a submission without commits.
2) If you then merge 1.7.1 into master, sure, this is totally a duplicate.

david-yz-liu:
1. Ah yes, thanks for reminding me. Always forget about that.
2. Thanks.

PR #3880, opened 2019-03-12
Don't fetch extended revision info when retrieving a submission file

adisandro:
This is a very simple optimization extracted from #3879:
When we just want to get the submission file contents, fetching extended revision info is useless (e.g. the push date in git, which requires walking the commit history and the reflog)

### DIFF
--- a/app/models/submission_file.rb
+++ b/app/models/submission_file.rb
@@ -117,7 +117,7 @@ class SubmissionFile < ApplicationRecord
       close_repo = true
     end
     revision = repo.get_revision(revision_identifier)
-    revision_file = revision.files_at_path(self.path)[self.filename]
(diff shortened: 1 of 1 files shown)


david-yz-liu:
This is all good, but I'd rather be conservative and target master rather than 1.7.1, unless this is currently an issue? Or if the problem is what commits this is based on, I'll merge 1.7.1 into master soon, so we can wait until then.

adisandro:
There isn't an issue, but I was thinking that this is trivial enough that the benefits far outweigh the risks.
If there really is a concurrency problem when accessing git repos, this drastically reduces the window in which a repo being tested is open.
In any case, up to you, I'm happy to rebase it on master.

david-yz-liu:
Alright, thanks for the info. I'd prefer putting this on master then.

PR #3889, opened 2019-03-21
Replace thin server with puma in development

mishaschwartz:
Puma gives us a server that can handle true multiprocessing in development. This is useful for uncovering concurrency related bugs before they make their way to production.


mishaschwartz:
@david-yz-liu. Confirmed that this works with the current vagrant setup.

PR #3890, opened 2019-03-22
Bugfix: git repository recovery

mishaschwartz:
This builds on #3887 and #3869 so that git reclones local repos from the remote if the local repo does not exist or if the local repo is corrupted (the previous behaviour did not reclone if the local repo was missing entirely). 

Repos were (likely) missing because they were removed prior to being recloned (if an error was raised while trying to access the repo) and then if there was an error that occurred during recloning due to a concurrency conflict, the repo would never be recloned at all.

With this PR, if a git repo does not exist the user will initially receive a 500 error but then the repo will be recloned so the next time that page is revisited (refreshed) the page will load as normal. In the case where there are multiple successive failures, the user may have to refresh the page several times but eventually the repo should be recloned properly and exist in a stable state. 

This is meant as a temporary fix so that there is only a minor disruption for the user as opposed to a total failure which requires manually recloning the missing repo to fix.

### DIFF
--- a/app/lib/git_repository.rb
+++ b/app/lib/git_repository.rb
@@ -46,22 +46,22 @@ class GitRepository < Repository::AbstractRepository
   end
 
   def self.reclone_repo(connect_string)
-    begin
-      if File.exist? connect_string
-        bad_repo_path = "#{connect_string}.bad"
-        FileUtils.rm_rf(bad_repo_path)
-        FileUtils.mv(connect_string, bad_repo_path)
-      end
-      repo_path, _sep, repo_name = connect_string.rpartition(File::SEPARATOR)
-      bare_path = File.join(repo_path, 'bare', "#{repo_name}.git")
-      @repos = Rugged::Repository.clone_at(bare_path, connect_string)
-      m_logger = MarkusLogger.instance
-      m_logger.log "Recloned corrupted or missing git repo: #{connect_string}"
-    rescue StandardError
-      msg = "Failed to clone corrupted or missing git repo: #{connect_string}"
-      m_logger.log msg
-      raise msg
+    repo_path, _sep, repo_name = connect_string.rpartition(File::SEPARATOR)
+    bare_path = File.join(repo_path, 'bare', "#{repo_name}.git")
+    raise "Cannot reclone from non-existant bare repository #{bare_path}" unless Dir.exist?(bare_path)
+    if Dir.exist?(connect_string)
+      bad_repo_path = "#{connect_string}.bad"
+      FileUtils.rm_rf(bad_repo_path)
+      FileUtils.mv(connect_string, bad_repo_path)
     end
+    @repos = Rugged::Repository.clone_at(bare_path, connect_string)
+    m_logger = MarkusLogger.instance
+    m_logger.log "Recloned corrupted or missing git repo: #{connect_string}"
+  rescue StandardError
+    msg = "Failed to clone corrupted or missing git repo: #{connect_string}"
+    m_logger = MarkusLogger.instance
+    m_logger.log msg
+    raise
   end
 
   def self.do_commit(repo, author, message)
@@ -140,10 +140,12 @@ class GitRepository < Repository::AbstractRepository
 
   # static method that should yield to a git repo and then close it
   def self.access(connect_string)
-    repo = GitRepository.open(connect_string)
-    yield repo
-  ensure
-    repo&.close
+    begin
+      repo = GitRepository.open(connect_string)
+      yield repo
+    ensure
+      repo&.close
+    end
   end
 
   # static method that deletes the git repo
--- a/app/lib/git_repository.rb
+++ b/app/lib/git_repository.rb
@@ -140,7 +140,6 @@ class GitRepository < Repository::AbstractRepository
 
   # static method that should yield to a git repo and then close it
   def self.access(connect_string)
-    repo = nil
     repo = GitRepository.open(connect_string)
     yield repo
   ensure
--- a/app/lib/git_repository.rb
+++ b/app/lib/git_repository.rb
@@ -48,7 +48,7 @@ class GitRepository < Repository::AbstractRepository
   def self.reclone_repo(connect_string)
     repo_path, _sep, repo_name = connect_string.rpartition(File::SEPARATOR)
     bare_path = File.join(repo_path, 'bare', "#{repo_name}.git")
-    raise "Cannot reclone from non-existant bare repository #{bare_path}" unless Dir.exist?(bare_path)
+    raise 'Repository does not exist' unless Dir.exist?(bare_path)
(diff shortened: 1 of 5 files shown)


app/lib/git_repository.rb
```diff
@@ -38,17 +38,29 @@ def initialize(connect_string)
...
-    else
-      raise "Repository does not exist at path \"#{@repos_path}\""
+      repo_path, _sep, repo_name = connect_string.rpartition(File::SEPARATOR)
+      bare_path = File.join(repo_path, 'bare', "#{repo_name}.git")
+      @repos = Rugged::Repository.clone_at(bare_path, connect_string)
+      m_logger = MarkusLogger.instance
+      m_logger.log "Recloned corrupted or missing git repo: #{connect_string}"
+    rescue StandardError
+      msg = "Failed to clone corrupted or missing git repo: #{connect_string}"
+      m_logger.log msg
```
adisandro: You need to get MarkusLogger.instance again here, could be uninitialized.

app/lib/git_repository.rb
```diff
@@ -38,17 +38,29 @@ def initialize(connect_string)
...
                            #      we need to prevent the corruption instead/as well
+        GitRepository.reclone_repo(connect_string)
+      end
+    else
+      GitRepository.reclone_repo(connect_string)
+    end
+  end
+
+  def self.reclone_repo(connect_string)
+    begin
```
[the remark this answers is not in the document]
adisandro: I guess you could do like rubocop suggests: https://www.rubydoc.info/gems/rubocop/0.27.0/RuboCop/Cop/Style/RedundantBegin

app/lib/subversion_repository.rb
```diff
@@ -57,15 +57,21 @@ def self.create(connect_string, with_hooks: false)
   # Static method: Opens an existing Subversion repository
   # at location 'connect_string'
   def self.open(connect_string)
+    unless SubversionRepository.repository_exists?(location)
+      raise "Could not open repository at location #{location}"
+    end
     return SubversionRepository.new(connect_string)
```
adisandro: You could throw away the return while you're there

app/lib/subversion_repository.rb
```diff
@@ -57,15 +57,21 @@ def self.create(connect_string, with_hooks: false)
   # Static method: Opens an existing Subversion repository
   # at location 'connect_string'
   def self.open(connect_string)
+    unless SubversionRepository.repository_exists?(location)
+      raise "Could not open repository at location #{location}"
+    end
```
adisandro: Can you make it a one-liner just like for access?

app/lib/subversion_repository.rb
```diff
@@ -57,15 +57,21 @@ def self.create(connect_string, with_hooks: false)
...
   end
 
   # Static method: Yields an existing Subversion repository and closes it afterwards
   def self.access(connect_string)
-    repository = self.open(connect_string)
-    yield repository
-  ensure
-    repository.close
+    raise 'Respository does not exist' unless SubversionRepository.repository_exists? connect_string
+    begin
```
adisandro: You can remove the begin and do `repository&.close` in the ensure.

app/lib/memory_repository.rb
```diff
@@ -192,10 +192,13 @@ def get_all_revisions
 
   # Static method: Yields an existing Memory repository and closes it afterwards
   def self.access(connect_string)
-    repository = MemoryRepository.open(connect_string)
-    yield repository
-  ensure
-    repository.close
+    raise 'Respository does not exist' unless MemoryRepository.repository_exists? connect_string
```
adisandro: Can you port the one-liner to memory#open too?

app/lib/memory_repository.rb
```diff
@@ -192,10 +192,13 @@ def get_all_revisions
...
+    begin
```
adisandro: You can remove the begin and do `repository&.close` in the ensure.

app/lib/git_repository.rb
```diff
@@ -38,17 +38,29 @@ def initialize(connect_string)
         @repos.reset('origin/master', :hard) # align to whatever is in origin/master
       rescue StandardError # TODO this shouldn't be necessary. It catches the case when a local repo is corrupted,
                            #      we need to prevent the corruption instead/as well
+        GitRepository.reclone_repo(connect_string)
+      end
+    else
+      GitRepository.reclone_repo(connect_string)
```
adisandro: This is the only relevant comment, everything else is about style:

This makes it asymmetric with svn and memory, where we raise an exception instead.
I know what you're trying to do, but it would mean that even with an invalid connect_string (i.e. a repository that is not supposed to exist) you would try to reclone the repo over and over. Is this supposed to go away after we land a proper concurrency fix?

mishaschwartz: To your last question: yes (see the last paragraph of the original comment on the PR)

However, I'll update it so that it doesn't try to reclone if the bare repository doesn't exist for now as well.

app/lib/subversion_repository.rb
mishaschwartz: I think we need the begin in this case because we don't want the ensure to trigger if the RuntimeError is raised (before the `repository` variable is initialized)

app/lib/memory_repository.rb
mishaschwartz: same comment as above

app/lib/git_repository.rb
adisandro: Perfect, that'll do!

app/lib/subversion_repository.rb
adisandro: The `&.close` would just do nothing in that case, but again this is only style, both versions are valid. I was just trying to be consistent with Rubocop.

app/lib/git_repository.rb
```diff
@@ -38,20 +38,32 @@ def initialize(connect_string)
...
     else
-      raise "Repository does not exist at path \"#{@repos_path}\""
+      GitRepository.reclone_repo(connect_string)
     end
   end
 
+  def self.reclone_repo(connect_string)
+    repo_path, _sep, repo_name = connect_string.rpartition(File::SEPARATOR)
+    bare_path = File.join(repo_path, 'bare', "#{repo_name}.git")
+    raise "Cannot reclone from non-existant bare repository #{bare_path}" unless Dir.exist?(bare_path)
```
adisandro: Use the same message that is used in svn and memory, which should really be a single I18n entry in the future:
Respository does not exist

app/lib/subversion_repository.rb
adisandro: Disregard

mishaschwartz:
@david-yz-liu I have tested this offline and every time a repo is corrupted or missing, the re-clone results in a successful page load after a single refresh

PR #3896, opened 2019-03-29
Don't display deadline including grace tokens until the group has been created

mishaschwartz:
Before this, the modified deadline would always be the same as the original deadline regardless of the number of grace tokens the student currently has. 

Since the minimum of the grace tokens of the members of the grouping is what counts it doesn't make sense to display this information until the grouping is created.

### DIFF
--- a/app/views/assignments/_read.html.erb
+++ b/app/views/assignments/_read.html.erb
@@ -25,10 +25,12 @@
       <% acc = 0 %>
       <% if @penalty.type == "GracePeriodSubmissionRule" %>
         <%# Relative remaining grace credits for the grouping %>
-        <% unless @grouping.nil? %>
-          <span class='prop_label'>
+        <span class='prop_label'>
             <%= t('grace_period_submission_rules.deadline_html') %>
-          </span>
+        </span>
+        <% if @grouping.nil? %>
+          <%= t('grace_period_submission_rules.no_group_yet') %>
+        <% else %>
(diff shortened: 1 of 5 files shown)


david-yz-liu:
I'd prefer to keep the header there but give a message like "Form a group to determine how many grace tokens you have remaining" or something.

app/views/assignments/_read.html.erb
```diff
@@ -24,26 +24,29 @@
     <div class ='sub_block'>
       <% acc = 0 %>
       <% if @penalty.type == "GracePeriodSubmissionRule" %>
+        <%# Relative remaining grace credits for the grouping %>
         <span class='prop_label'>
-          <%= t('grace_period_submission_rules.deadline_html') %>
+            <%= t('grace_period_submission_rules.deadline_html') %>
```
david-yz-liu: Delete the new space here.

config/locales/views/submission_rules/en.yml
```diff
@@ -19,6 +19,7 @@ en:
       other: "%{count} credits"
     deadline_html: "Deadline using all <br> remaining grace <br> credits"
     group_credits_html: "Your group has <strong>%{available_grace_credits}</strong> available grace credit(s)."
+    no_group_yet: "Form a group to determine how many grace tokens are available for this assignment."
```
david-yz-liu: Let's say "...how many grace credits you have available..." here.

PR #3916, opened 2019-04-01
Let file viewer handle case where there are no submitted files

mishaschwartz:
Before the page crashed since it was trying to load a submission_file with a null id. 
Now it doesn't try to load a non-existant submission_file and displays a message instead indicating there are no student files.

### DIFF
--- a/app/assets/javascripts/Components/Result/file_viewer.jsx
+++ b/app/assets/javascripts/Components/Result/file_viewer.jsx
@@ -19,12 +19,14 @@ export class FileViewer extends React.Component {
   componentDidMount() {
     if (this.props.selectedFile !== null) {
       this.set_submission_file(this.props.selectedFile);
+    } else {
+      this.setState({loading: false})
     }
   }
 
   // Manually manage a change of selectedFile, as this requires fetching new file data.
   shouldComponentUpdate(nextProps) {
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/Components/Result/file_viewer.jsx
```diff
@@ -19,12 +19,14 @@ export class FileViewer extends React.Component {
   componentDidMount() {
     if (this.props.selectedFile !== null) {
       this.set_submission_file(this.props.selectedFile);
+    } else {
+      this.setState({loading: false, type: null})
```
david-yz-liu: Keep the `loading: false` but remove the `type: null` (see comment below).

app/assets/javascripts/Components/Result/file_viewer.jsx
```diff
@@ -86,6 +88,8 @@ export class FileViewer extends React.Component {
     };
     if (this.state.loading) {
       return I18n.t('working');
+    } else if (this.state.type === null) {
```
david-yz-liu: You can remove this (and the strings). There's already a message that should be displayed at `submission_file_panel.jsx:218`, but it seems that it's not being displayed properly. Please fix the condition in that file to make that message display in the selector.

PR #3917, opened 2019-04-01
Use test_runs.problems instead of creating fake test results when autotest global errors happen

adisandro:
Didn't care about formatting in the test runs table, since we want to revamp it anyway.


david-yz-liu:
Okay I was thinking about this a bit during the meeting, and two comments:

1. It would be nice to have `test_runs.problems` added to the `Grouping.group_hash_list` list, so that the top-level component can access it directly. 

2. Instead of putting the problems where you put it, can you put it into the `SubComponent` of the `TestRunTable` (lines 95-99).

adisandro:
@david-yz-liu Done!

PR #3920, opened 2019-04-03
Upgrade webpacker to v4, webpack-dev-server to v3 and babel to v7

adisandro:


config/webpacker.yml
```diff
@@ -33,11 +54,15 @@ development:
...
   compile: true
 
+  # Verifies that correct packages and versions are installed by inspecting package.json, yarn.lock, and node_modules
+  check_yarn_integrity: true
+
   # Reference: https://webpack.js.org/configuration/dev-server/
   dev_server:
     https: false
-    host: 0.0.0.0
+    host: localhost
```
david-yz-liu: This change seems to stop webpack-dev-server from working from the vagrant VM. Does the old version make it not work for you locally?

adisandro: Sorry about that, I didn't realize. I just updated to the most recent default configs, can easily turn it back to 0.0.0.0.

PR #3933, opened 2019-04-08
Serialize repository access in repo_browser component

mishaschwartz:
Get file data only after the revision data has been collected to prevent multiple concurrent requests to access the same repo object. 
This was happening because the calls to the server to fetch various repo information that were made in each components' `componentDidMount` calls were overlapping. Subsequent calls to update that data once the components have mounted should happen in sequence since the file data is only updated if the revision identifier has been changed.

### DIFF
--- a/app/assets/javascripts/Components/repo_browser.jsx
+++ b/app/assets/javascripts/Components/repo_browser.jsx
@@ -6,16 +6,14 @@ import { SubmissionFileManager } from './submission_file_manager'
 class RepoBrowser extends React.Component {
   constructor(props) {
     super(props);
-    this.submissionFileManagerRef = React.createRef();
     this.state = {
-      revision_identifier: props.collected_revision_id,
+      revision_identifier: undefined,
       revisions: []
     };
   }
 
   componentDidMount() {
     this.fetchRevisions();
-    this.submissionFileManagerRef.current.fetchData();
   }
 
   // Update the list of revisions and set the currently-viewed revision to the most recent one.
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/Components/repo_browser.jsx
```diff
@@ -6,6 +6,7 @@ import { SubmissionFileManager } from './submission_file_manager'
 class RepoBrowser extends React.Component {
   constructor(props) {
     super(props);
+    this.submissionFileManagerRef = React.createRef();
     this.state = {
       revision_identifier: props.collected_revision_id,
```
david-yz-liu: Did some more thinking about this one. This line is interesting; I think my original intention was to load the files for the collected revision, while the other other revisions loaded. But this is also overridden when the revisions are loaded to be the most recent revision. You can change this to `undefined`.

app/assets/javascripts/Components/repo_browser.jsx
```diff
@@ -14,6 +15,7 @@ class RepoBrowser extends React.Component {
 
   componentDidMount() {
     this.fetchRevisions();
+    this.submissionFileManagerRef.current.fetchData();
```
david-yz-liu: This should no longer be necessary if you make the change I suggested above. The file manager already has a `componentDidUpdate` hook that calls `fetchData` when its revision identifier prop changes (which should be triggered by a state change in `fetchRevisions`).

PR #3940, opened 2019-04-18
Add test cases for downloading a list of graders in XML format

YolandaZIK:
In MarkUs, `Users-->Graders`, admins are able to download a list of graders they upload in YML/XML format. However, the front end webpage shows `Download in YML format`.


YolandaZIK:
The function is fixed. Users can download a list of graders in YML format.

PR #3953, opened 2019-04-24
Speed up remark count

mishaschwartz:
Also update remark count when releasing from the submission table as well.


david-yz-liu:
@mishaschwartz it would be nice if we could use the association `submitted_remark` from the `Submission` model. Can you try something like `groupings.joins(submissions: :submitted_remark).where(result.state == incomplete)`?

PR #3954, opened 2019-04-26
Autotest form validation

mishaschwartz:
- pass schema to the autotester so that it can validate the form data
- when parsing the `extra_info`, don't assume defaults will be set
- use `extra_info` instead of `extra_data` exclusively


mishaschwartz:
should be tested with: https://github.com/MarkUsProject/markus-autotesting/pull/182

PR #3987, opened 2019-05-29
Install jdbc and sql testers when installing the autotester

mishaschwartz:
Also makes sure jdk-12 is installed (used by java and jdbc)


david-yz-liu:
@mishaschwartz sorry I merged this one too quickly. When provisioning, the installation hangs on the Oracle license agreement. Could you please look into that?

PR #3989, opened 2019-05-30
Add rake task for archiving submissions for an entire instance

mishaschwartz:


lib/tasks/archive_submissions.rake
```diff
@@ -0,0 +1,82 @@
...
+    target_dir = Pathname.new(args[:target_dir])
+    raise "Directory #{target_dir} does not exist" unless Dir.exist?(target_dir)
+
+    relative_root = Rails.application.config.action_controller.relative_url_root
+    if relative_root
+      target_dir += relative_root
+      FileUtils.makedirs target_dir
+    end
+
+    total_groupings = Grouping.joins(current_submission_used: :submission_files).uniq.count
```
david-yz-liu: My understanding here is that the intention with the `:submission_files` is to ignore groups who didn't submit anything. That's good, but in this case there should be another check in the `Grouping` loop below that does something like `next unless submission.submission_files`, otherwise the count might be off.

lib/tasks/archive_submissions.rake
```diff
@@ -0,0 +1,82 @@
...
+    processed_groupings = 0
+    print "\r"
+    CSV.open(target_dir + 'group_memberships.csv', 'w') do |csv|
+      Assignment.all.pluck(:id, :short_identifier).each do |aid, short_id|
```
david-yz-liu: Ignore peer review assignments here (`.where(parent_assignment_id: nil)`).

lib/tasks/archive_submissions.rake
```diff
@@ -0,0 +1,82 @@
...
+          submission.submission_files.each do |file|
+            file_content = file.retrieve_file
+            rel_path = Pathname.new(file.path).relative_path_from(short_id_dir)
+            file_dir = group_dir + rel_path
+            FileUtils.makedirs(file_dir)
+            mode = SubmissionFile.is_binary?(file_content) ? 'wb' : 'w'
+            File.write(file_dir + file.filename, file_content, mode: mode)
+          end
+          processed_groupings += 1
+          print "written files for #{processed_groupings}/#{total_groupings} groupings\r"
```
david-yz-liu: Let's use the word "archived" instead of "written" here.

PR #4006, opened 2019-06-03
Don't install jdk-12 because it hangs on license

mishaschwartz:
- in response to final comment on #3987
- should be pulled in with https://github.com/MarkUsProject/markus-autotesting/pull/198

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -2,7 +2,7 @@
 VAGRANTFILE_API_VERSION = "2"
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
-  config.vm.box = "bento/ubuntu-18.04"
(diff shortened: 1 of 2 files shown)


adisandro:
@mishaschwartz you should just use openjdk-12-jdk. OpenJDK is the official reference implementation of Java.

mishaschwartz:
@adisandro any recommendations for installing openjdk-12 for 18.04 in a way that does not require you to manually agree to the license agreement? (so that it can be installed in a script in our vagrant dev environment)

adisandro:
@mishaschwartz OpenJDK does not ask you to accept a licence, it's open source, while oracle's jdk is not. Don't know though if version 12 is available on 18.04, or if it needs a ppa (I think there's 11 there, but it will be equivalent).

PR #4011, opened 2019-06-13
Show info about extension in student view

mishaschwartz:


app/views/assignments/_read.html.erb
```diff
@@ -17,7 +17,15 @@
   </div>
   <div class='sub_block'>
     <span class='prop_label'><%= Assignment.human_attribute_name(:due_date) %></span>
-    <span class="prop_paragraph"><%= l(@due_date) %></span>
+    <span class="prop_paragraph">
+      <% if !@grouping.extension.nil? %>
```
david-yz-liu: @grouping might be `nil` here as well

config/locales/en.yml
```diff
@@ -17,6 +17,7 @@ en:
         peer_review_results: "Peer Review Results"
         your_reviews: "Your Reviews"
         review: "Peer Review"
+        deadline_with_extension: "%{deadline} (you have an extension until %{extension_deadline})"
```
david-yz-liu: Please add this under `locales/views/assignments/en.yml` (the key will be `assignments.deadline_with_extension)`.

PR #4012, opened 2019-06-13
Seed database with section due dates and grouping extensions

mishaschwartz:


db/data/students.csv
```diff
@@ -1,114 +1,114 @@
-c8mahler,Mahler,Gustav
-c9magnar,Magnard,Alberic
+c8mahler,Mahler,Gustav,section_b
```
david-yz-liu: The section names should be user facing. You can use the Toronto convention, `LEC0101` and `LEC0201`.

mishaschwartz:
@david-yz-liu what if we just don't collect any of the assignments in the seed file?

We don't really have a good way of uncollecting submissions so this would put all the submissions in an uncollected state by default and the developer can then choose to collect them or not based on what they want to test. 

Collecting takes two clicks so it shouldn't be too much of a hassle to do it manually.

mishaschwartz:
@david-yz-liu as discussed, this is now only creating section due dates and extensions for A3 and A4 (the assignments that are due in the future)

PR #4018, opened 2019-06-20
Manage subdirectories in submissions file manager

mishaschwartz:
- enables creation and deletion of subdirectories
- enables uploading files to subdirectories
- only enabled for the submission_file_manager (others will come in a future PR)
- introduces a new react modal for uploading files
- introduces a new policy to check whether a user can create, remove or change subdirectories
- introduces new props to the raw file manager component that selectively prevents certain action bar elements from being created

### DIFF
--- a/app/assets/javascripts/Components/Modals/submission_file_upload_modal.jsx
+++ b/app/assets/javascripts/Components/Modals/submission_file_upload_modal.jsx
@@ -1,0 +1,49 @@
+import React from 'react';
+import Modal from 'react-modal';
+
+class SubmissionFileUploadModal extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      newfiles: []
+    }
+  }
+
+  componentDidMount() {
+    Modal.setAppElement('body');
+  }
+
+  onSubmit = (event) => {
+    event.preventDefault();
+    this.props.onSubmit(this.state.newfiles);
+  };
+
+  handleFileUpload = (event) => {
+    this.setState({newfiles: event.target.files})
+  };
+
+  render() {
+    return (
+      <Modal
+        className="react-modal"
+        isOpen={this.props.isOpen}
+        onRequestClose={this.props.onRequestClose}
+      >
+        <h2>{I18n.t('add_new')}</h2>
+        <form onSubmit={this.onSubmit}>
+          <div className={'modal-container-vertical'}>
+            <div className={'modal-container'}>
+              <input type={'file'} name={'newfiles'} multiple={true} onChange={this.handleFileUpload}/>
+            </div>
+            <div className={'modal-container'}>
+              <input type='submit' value={I18n.t('save')} />
+            </div>
+          </div>
+        </form>
+      </Modal>
+    )
+  }
+}
+
+export default SubmissionFileUploadModal;
--- a/app/assets/javascripts/Components/Modals/submission_file_upload_modal.jsx
+++ b/app/assets/javascripts/Components/Modals/submission_file_upload_modal.jsx
@@ -6,8 +6,8 @@ class SubmissionFileUploadModal extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      newfiles: []
-    }
+      newFiles: []
+    };
   }
 
   componentDidMount() {
@@ -16,11 +16,11 @@ class SubmissionFileUploadModal extends React.Component {
 
   onSubmit = (event) => {
     event.preventDefault();
-    this.props.onSubmit(this.state.newfiles);
+    this.props.onSubmit(this.state.newFiles);
   };
 
   handleFileUpload = (event) => {
-    this.setState({newfiles: event.target.files})
+    this.setState({newFiles: event.target.files})
   };
 
   render() {
@@ -30,11 +30,11 @@ class SubmissionFileUploadModal extends React.Component {
         isOpen={this.props.isOpen}
         onRequestClose={this.props.onRequestClose}
       >
-        <h2>{I18n.t('add_new')}</h2>
+        <h2>{I18n.t('upload')}</h2>
         <form onSubmit={this.onSubmit}>
           <div className={'modal-container-vertical'}>
             <div className={'modal-container'}>
-              <input type={'file'} name={'newfiles'} multiple={true} onChange={this.handleFileUpload}/>
+              <input type={'file'} name={'new_files'} multiple={true} onChange={this.handleFileUpload}/>
             </div>
             <div className={'modal-container'}>
               <input type='submit' value={I18n.t('save')} />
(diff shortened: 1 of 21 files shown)


app/assets/javascripts/Components/Modals/submission_file_upload_modal.jsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import Modal from 'react-modal';
+
+class SubmissionFileUploadModal extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      newfiles: []
```
david-yz-liu: Let's use camel case here (`newFiles`).

app/assets/javascripts/Components/Modals/submission_file_upload_modal.jsx
```diff
@@ -0,0 +1,49 @@
...
+    }
```
david-yz-liu: Please put a semicolon here.

app/assets/javascripts/Components/Modals/submission_file_upload_modal.jsx
```diff
@@ -0,0 +1,49 @@
...
+  };
+
+  render() {
+    return (
+      <Modal
+        className="react-modal"
+        isOpen={this.props.isOpen}
+        onRequestClose={this.props.onRequestClose}
+      >
+        <h2>{I18n.t('add_new')}</h2>
```
david-yz-liu: Oh man let's use this opportunity to get rid of this misnamed key. I'd be happy with `I18n.t('upload')` ("Upload"), or you can use `'upload_the'` and create a new generic key for `"File"` if you want.

app/assets/javascripts/Components/Modals/submission_file_upload_modal.jsx
```diff
@@ -0,0 +1,49 @@
...
+        <form onSubmit={this.onSubmit}>
+          <div className={'modal-container-vertical'}>
+            <div className={'modal-container'}>
+              <input type={'file'} name={'newfiles'} multiple={true} onChange={this.handleFileUpload}/>
```
david-yz-liu: The "newfiles" here is a bit tricky. I guess to make the name realistic you can use `'new_files'` (this is the name of the param that's sent back).

app/assets/javascripts/Components/markus_file_manager.jsx
```diff
@@ -10,13 +10,21 @@ import { RawFileBrowser, Headers, FileRenderers, BaseFileConnectors } from 'reac
 
 
 class RawFileManager extends RawFileBrowser {
-  handleActionBarAddFileClick = (event) => {
+  handleActionBarAddFileClick = (event, selected) => {
     event.preventDefault();
-    window.modal_addnew.open();
+    let uploadTarget = '';
+    if (selected) {
+      if (selected.children) {
```
david-yz-liu: Does this work if a folder is selected, but it has no children? (Is `selected.children` an empty list in this case?)

I believe the react-keyed-file-browser way of detecting whether something's a folder or not is whether its name ends with a `/`. Kind of weird, but I think maybe we should just be consistent with that.

app/assets/javascripts/Components/markus_file_manager.jsx
```diff
@@ -10,13 +10,21 @@ import { RawFileBrowser, Headers, FileRenderers, BaseFileConnectors } from 'reac
...
+        uploadTarget = selected.relativeKey;
+      } else {
+        uploadTarget = selected.relativeKey.substring(0, selected.relativeKey.lastIndexOf(selected.name));
+      }
+    }
+    this.props.onActionBarAddFileClick(uploadTarget)
```
david-yz-liu: Please put a semicolon here.

app/assets/javascripts/Components/markus_file_manager.jsx
```diff
@@ -10,13 +10,21 @@ import { RawFileBrowser, Headers, FileRenderers, BaseFileConnectors } from 'reac
...
   };
 
   renderActionBar(selectedItem) {
-    const selectionIsFolder = selectedItem && !selectedItem.size;
+    const selectionIsFolder = selectedItem && selectedItem.children;
```
david-yz-liu: See above comment.

app/assets/javascripts/Components/submission_file_manager.jsx
```diff
@@ -1,14 +1,16 @@
...
 
 
 class SubmissionFileManager extends React.Component {
 
   constructor(props) {
     super(props);
     this.state = {
-      files: []
+      files: [],
+      showModal: false
```
david-yz-liu: Seems like you need to add an `uploadTarget` state as well?

app/assets/javascripts/Components/submission_file_manager.jsx
```diff
@@ -105,17 +134,32 @@ class SubmissionFileManager extends React.Component {
     });
   };
 
+  uploadFiles = (uploadTarget) => {
```
david-yz-liu: I think this name is a bit misleading. You could do something like `openUploadModal`.

app/controllers/submissions_controller.rb
```diff
@@ -684,18 +702,22 @@ def get_all_file_data(revision, grouping, path)
...
+        dirname, basename = File.split(file_name)
+        dirname = '' if dirname == '.'
+        data = get_file_info(basename, file_obj, grouping.assignment.id,
+                             revision.revision_identifier, dirname, grouping.id)
+        next if data.nil?
+        data[:key] = file_name
+        data[:modified] = data[:last_revised_date]
+        data
+      else
+        {key: "#{file_name}/"}
```
david-yz-liu: Put spaces inside the `{` and `}` (not sure why Hound didn't complain?)

app/assets/javascripts/Components/markus_file_manager.jsx
mishaschwartz: An empty directory still has a `children` attribute but I agree that keeping it consistent with the package behaviour makes sense.

app/assets/javascripts/Components/submission_file_manager.jsx
mishaschwartz: it's not necessary but it's good to add so the reader is aware of all possible state keys.

app/assets/javascripts/Components/markus_file_manager.jsx
```diff
@@ -57,27 +81,30 @@ class RawFileManager extends RawFileBrowser {
...
             <li key="action-add-folder">
               <a
-                onClick={this.handleActionBarAddFolderClick}
+                onClick={(event) => this.handleActionBarAddFolderClickSetSelection(event, target)}
                 href="#"
                 role="button"
               >
                 <i className="fa fa-folder-o" aria-hidden="true"/>
-                &nbsp;Add Subfolder
+                &nbsp;{target ? "Add Subolder" : "Add Folder"}
```
david-yz-liu: There's a typo on "Subolder"

PR #4046, opened 2019-07-02
Update for autotest file manager

mishaschwartz:
Applies the same changes made in #4018 but for the file manager on the Test Framework page


david-yz-liu:
The old modal (`#addnew`) should be removed from `app/views/automated_tests/_form.html.erb`.

PR #4049, opened 2019-07-03
Confirm that submissions should be re-collected

mishaschwartz:
- introduces a modal where the user can choose to override existing submissions (default: don't override) when collecting submissions.

- [x] write rspec test for the new functionality of the collect_submissions method


david-yz-liu:
@mishaschwartz I merged this in, but please make a note that if "override" is not selected and _only_ collected groups are selected, no flash message is displayed to the user. This should be fixed later, but not a high priority.

BTW I also updated the styling and message a little bit.

PR #4081, opened 2019-07-25
Test Framework form improvements

mishaschwartz:
- only allow the user to select files, not folders in form components where files are selected
- give more informative strings for output display options

### DIFF
--- a/config/locales/views/automated_tests/en.yml
+++ b/config/locales/views/automated_tests/en.yml
@@ -19,10 +19,11 @@ en:
       in_progress: "in progress"
       problems: "error"
     title: "Automated Testing"
+    display_output_title: "Display test output to students?"
     display_output:
-      instructors: "Never display"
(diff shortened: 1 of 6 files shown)


config/locales/views/automated_tests/en.yml
```diff
@@ -19,6 +19,10 @@ en:
       in_progress: "in progress"
       problems: "error"
     title: "Automated Testing"
+    display_output:
+      instructors: "Never display"
```
david-yz-liu: I'd like to make these messages much more explicit. Why don't we rename the field `title` to "Display test output to students?" and use the following texts:

- `instructors`: Never display test output to students.
- `instructors_and_students`: Always display test output to students.
- `instructors_and_student_tests`: Only display test output to students for student-run tests.

We can add a note to the documentation that instructors _always_ see test outputs, which is good with me.

PR #4101, opened 2019-08-01
Simplify Submission Files API controller

mishaschwartz:
- reduce number of parameters (don't pass in group name and short identifier since they can be inferred from the required group id and assignment id)
- remove any previous temporary zip files since the `Zip::File::CREATE` mode for `Zip::File.open` does not overwrite a zip file, instead it either creates or appends (if the file exists).

### DIFF
--- a/app/controllers/api/submission_files_controller.rb
+++ b/app/controllers/api/submission_files_controller.rb
@@ -26,8 +26,7 @@ module Api
         return
       end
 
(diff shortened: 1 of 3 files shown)


david-yz-liu:
@mishaschwartz do you mind also removing the method `Submission.get_submission_by_group_and_assignment`? I don't think it's used anymore, other than a few tests.

PR #4114, opened 2019-08-29
Upgrade to Rails 6

david-yz-liu:


mishaschwartz:
Just so I understand, the changes made here are:
- update rails version and related gems
- change `update_attributes` everywhere to just `update`
- rename `MarkusCSV` to `MarkusCsv`
- change `CSVInvalidLineError` to `CsvInvalidLineError`
- change `.utf8_encode` to `.encode(Encoding::UTF_8, ...)`
- remove `.utf8_encode` methods
- move repository revision classes into their own files
- rename `SplitPdfJob` to `SplitPDFJob`
- make the mathjax config file a `.html.erb` file not a `.js.erb` file
- use `Rails.cache` instead of redis for storing activejob statuses
- remove `updated_at` and `created_at` column creating commands from a migration from 2016 (??)
- set `BINDING` environment variable in markus install script instead of `HOST`
- make pdf job tests pending
- render views when running rspec tests

Is there anything I'm missing? 
Are these all required to update to Rails 6?

david-yz-liu:
Thanks for summarizing this!

- `update_attributes` will be deprecated in Rails 6.1 (emits a warning in 6.0)
- Rails 6 uses the [zeitwerk](https://github.com/fxn/zeitwerk) code loader by default. The class renaming was to conform to its default name handling; it also expects one class per file. There's still another deprecation warning related to how the MarkUs configuration is being loaded at initialization time, so we'll need to look at that in more detail later. I'll make an issue for it.
- [`redis-rails`](https://github.com/redis-store/redis-rails) had a dependency conflict, and isn't necessary any more because Rails 5.2+ has built-in Redis support, at least for our purposes I think.
- The `HOST` variable is deprecated in Rails 6.
- The mathjax file couldn't be found: default search was for `.html` only, because this partial was included in an HTML file). I couldn't find a reference for whether this changed in Rails 6, so I'm not sure about this one.

Getting tests to work:

- The rspec-rails default template rendering is [incompatible with Rails 6](https://github.com/rspec/rspec-rails/issues/2086), but this should be fixed when we upgrade to rspec-rails v4.0.0. I'll make an issue for that.
- The migration change was prompted by https://travis-ci.org/david-yz-liu/Markus/builds/578120082, but again I couldn't find a reference for why this started happening.
- The SplitPdfJob tests seemed to cause a segfault (https://travis-ci.org/david-yz-liu/Markus/builds/578128110), but I couldn't reproduce this locally. I can't remember if this happened before, but I'll look into this later.

PR #4140, opened 2019-09-20
Create a table for LEVELS

YolandaZIK:
I created a wrong table with migration file `db/migrate/20190920210713_create_levels.rb`. Then I dropped the table and created a new table with `db/migrate/20190920221738_create_levels.rb`. The latter file created a table for LEVELS with columns: `assignment_id`, `rubric_criterion_id`, `level_num`, `level_name`, `level_description` and `mark`(which stands for the points assigned to a level under a rubric criterion).


YolandaZIK:
The changes should be done under a new branch instead of `sti-criteria`.

PR #4142, opened 2019-09-24
Create a table for levels of rubric criteria, and update the schema

YolandaZIK:
I did `rails delete migration create_levels` first and then `rake db:rollback`. It just deleted the migration file but there still was the wrong table for levels in the `schema.rb`. So I deleted the entire branch and built the vagrant box again. 

I think the right way to delete a wrong migration file and update the schema is to do `rake db:rollback` first to change the `schema.rb` back to the version without the wrong table and then do `rails delete migration <wrong_migration_name>` to delete the wrong migration file.

### DIFF
--- a/db/migrate/20190924061752_create_levels.rb
+++ b/db/migrate/20190924061752_create_levels.rb
@@ -1,10 +1,10 @@
 class CreateLevels < ActiveRecord::Migration[6.0]
   def change
     create_table :levels do |t|
-      t.belongs_to :rubric_criterion, foreign_key: true
+      t.belongs_to :rubric_criterion, foreign_key: true, null: false
       t.string :name, null: false
       t.integer :number, null: false
-      t.string :description
+      t.string :description, null: false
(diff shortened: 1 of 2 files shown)


db/migrate/20190924061752_create_levels.rb
```diff
@@ -0,0 +1,12 @@
+class CreateLevels < ActiveRecord::Migration[6.0]
+  def change
+    create_table :levels do |t|
+      t.belongs_to :rubric_criterion, foreign_key: true
```
david-yz-liu: Good, but also add `null: false` here.

db/migrate/20190924061752_create_levels.rb
```diff
@@ -0,0 +1,12 @@
...
+      t.string :name, null: false
+      t.integer :number, null: false
+      t.string :description
```
david-yz-liu: I'd also like a `null: false` here as well.

PR #4145, opened 2019-09-27
Code-viewer and annotation bugfixes

mishaschwartz:
Bugs:

- caching the submission file id in `localStorage` meant that a pdf or image file from a previous student was being displayed when changing between student submissions in the results page. 
- the `annotationPanel` component stores the current overall comment as both a prop and in the component state. This state was not being updated after the initial render which meant that a previously saved overall comment was never visible. 

Fixes:

- only cache the file name in `localStorage` and update the `getFirstFile` function to find either the first visible file or the first file with the cached filename. 
- add a `componentDidUpdate` method to the `annotationPanel` that makes sure the state updates to reflect the props when the props change (ie. when new data is sent from the server).


david-yz-liu:
@mishaschwartz looks good to me, please feel free to merge & send patch.

PR #4147, opened 2019-09-30
Level Model

ashih2018:
Created a Level model that belongs to the Rubric Criterion model, as well made small changes to the Rubric Criterion controller and Yolanda's migration file. 
Possible future concern: Should we look to removing the number column in the Levels migration. If we are sorting models in the future by the mark, then the number column may just get in the way. For example, if we create levels 1, 2, 3, 4 and remove level 3 and add level 5, we would have levels 1, 2, 4, 5 in the database.

### DIFF
--- a/app/controllers/rubric_criteria_controller.rb
+++ b/app/controllers/rubric_criteria_controller.rb
@@ -2,7 +2,7 @@ class RubricCriteriaController < ApplicationController
 
   before_action :authorize_only_for_admin
 
-  def csv_download
+  def download_csv
     @assignment = Assignment.find(params[:assignment_id])
     file_out = MarkusCsv.generate(@assignment.get_criteria(:all, :rubric)) do |criterion|
       criterion_array = [criterion.name, criterion.max_mark]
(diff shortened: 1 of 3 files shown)


app/controllers/rubric_criteria_controller.rb
```diff
@@ -2,7 +2,7 @@ class RubricCriteriaController < ApplicationController
 
   before_action :authorize_only_for_admin
 
-  def download_csv
+  def csv_download
```
david-yz-liu: Don't rename this in this pull request. (You can rename it later when you're actually working on this method.)

db/migrate/20190924061752_create_levels.rb
```diff
@@ -4,8 +4,9 @@ def change
       t.belongs_to :rubric_criterion, foreign_key: true, null: false
       t.string :name, null: false
       t.integer :number, null: false
-      t.string :description, null: false
+      t.text :description, null: false
```
david-yz-liu: In general you shouldn't make changes to an existing migration, but also you shouldn't make changes to a migration without also committing the new `schema.rb` file.

In this particular case, I'd say that you don't need to make these changes here, and instead should revert them and focus on the `Level` model file.

PR #4150, opened 2019-09-30
Fix spreadsheet render error

mishaschwartz:
Fix bug that stopped spreadsheet table from rendering if a section is not null and not in the sections hash.


david-yz-liu:
Looks good to me @mishaschwartz.

PR #4167, opened 2019-10-02
Grade-entry-students do not have the same id as users

mishaschwartz:
- select `grade_entry_students` based on their `id` not their `user_id` (which was causing the wrong objects to be updated/created)
- use `upsert_all` to do bulk updates instead of looping which takes forever!

### DIFF
--- a/app/helpers/grade_entry_forms_helper.rb
+++ b/app/helpers/grade_entry_forms_helper.rb
@@ -2,26 +2,6 @@
 
 module GradeEntryFormsHelper
 
-  # Release/unrelease the marks for a grade entry form for the given
-  # grade_entry_students
-  # Return the number of GradeEntryStudents that have been updated successfully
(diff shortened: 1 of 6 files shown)


app/helpers/grade_entry_forms_helper.rb
```diff
@@ -2,26 +2,6 @@
...
-  # Release/unrelease the marks for a grade entry form for the given
-  # grade_entry_students
-  # Return the number of GradeEntryStudents that have been updated successfully
-  def set_release_on_grade_entry_students(grade_entry_students, release, errors)
-    numGradeEntryStudentsChanged = 0
-    grade_entry_students.each do |grade_entry_student|
-      begin
-        grade_entry_student.released_to_student = release
-        unless grade_entry_student.save
-          raise I18n.t('grade_entry_forms.grades.update_error',
```
david-yz-liu: @mishaschwartz can you remove this key as well? (This is causing the failing test.)

PR #4172, opened 2019-10-07
Don't crash when annotation is out of bounds

mishaschwartz:


app/assets/javascripts/SourceCodeGlower/SourceCodeLineAnnotations.js
```diff
@@ -48,6 +48,10 @@ SourceCodeLineAnnotations.prototype.annotateLine = function(
 
   // Glow the Source Code Line
   var line = this.getLineManager().getLine(lineNum);
+  if (!line) {
+    console.log("Annotation line out of bounds");
```
david-yz-liu: Let's use `console.error` and also report the line number.

PR #4177, opened 2019-10-07
1.8.0: Validate sections before associating them with users

mishaschwartz:
- copy of #4176 for 1.8.0 branch

### DIFF
--- a/app/models/student.rb
+++ b/app/models/student.rb
@@ -25,6 +25,8 @@ class Student < User
   belongs_to :section, optional: true
   accepts_nested_attributes_for :section
 
+  validate :section_nil_or_exists?
+
   validates_numericality_of :grace_credits,
                             only_integer: true,
                             greater_than_or_equal_to: 0
@@ -248,5 +250,12 @@ class Student < User
     grouping&.due_date ||
       section&.section_due_date_for(assignment)&.due_date ||
       assignment&.due_date
+  end
+
+  private
+
+  # make sure that the section_id refers to a valid section if it is not nil
+  def section_nil_or_exists?
+    errors.add(:base, 'Section must be nil or exist') unless section_id.nil? || !section.nil?
   end
 end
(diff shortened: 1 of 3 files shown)


app/models/student.rb
```diff
@@ -25,6 +25,8 @@ class Student < User
   belongs_to :section, optional: true
   accepts_nested_attributes_for :section
 
+  validate :section_nil_or_exists?
```
david-yz-liu: Hey @mishaschwartz, so what I was wondering here is whether there's a better way to use built-in validations, as this seemed like a fairly common use case.

Looking at https://guides.rubyonrails.org/active_record_validations.html#validates-associated, I think something like

```
validates :section, presence: true, allow_nil: true
```

will work?

mishaschwartz: I agree that there should be a better way to do it. However, we have the issue that if we set `allow_nil` then the validation checks for a section with `id=0` which returns `nil` and the validation passes. We need the validation to fail when the foreign key value itself is invalid (ie. it isn't null but the `id` it refers to doesn't actually exist). 

However, if you want it to be in one line we can do something like:

```ruby
validates_presence_of :section, unless: -> { section_id.nil? }
```

which I just checked also works

mishaschwartz: If you want some background reading on this. 

- There was a PR that proposed these sort of foreign key integrity checks (https://github.com/rails/rails/pull/26572) when `belongs_to` is `optional` but it was rejected because they felt that this sort of check should be done at the db level or in the application code.
- the same author of that PR suggested a workaround that is similar to my original solution on a related issue (https://github.com/rails/rails/issues/26557#issuecomment-248586909).

david-yz-liu: Okay great, thanks. Please change the validation to the code you suggested. In the PR for `master`, please also add a migration to add the foreign key constraint.

PR #4181, opened 2019-10-10
Fix test data generation for assignments model

robert-tan:
mini-milestone :)


david-yz-liu:
@robert-tan please do a merge from the `sti-assignment` branch.

PR #4192, opened 2019-10-20
Create from csv

ashih2018:
Updated create_from_csv_row method, such that it utilizes the newly implemented level association between the Level model and the RubricCriterion model. Furthermore, I created/edited the existing tests so that they function properly with the new method. For example, the csv rows that are used in the rspec tests have name, number, description, and mark to match the required fields of level.
Notes and possible shortcomings: 
- we are still using the constants RUBRIC_LEVELS, DEFAULT_MAX_MARK, and MAX_LEVEL. I will remove these and update their respective tests in my next PR.
- I have added an add level test and an edit level test. I believe these would cover the only cases for the update_csv_row method, but if there are edge cases I should test that I have not thought about, I will create more tests to satisfy those cases.

### DIFF
--- a/app/models/rubric_criterion.rb
+++ b/app/models/rubric_criterion.rb
@@ -70,10 +70,9 @@ class RubricCriterion < Criterion
   # ===Params:
   #
   # row::         An array representing one CSV file row. Should be in the following
-  #               format: [name, weight, _names_, _descriptions_] where the _names_ part
-  #               must contain RUBRIC_LEVELS elements representing the name of each
-  #               level and the _descriptions_ part (optional) can contain up to
-  #               RUBRIC_LEVELS description (one for each level).
+  #               format: [name, weight, _levels_ ] where the _levels part contains
+  #               the following information about each level in the following order:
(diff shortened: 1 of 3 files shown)


app/models/rubric_criterion.rb
```diff
@@ -101,19 +100,27 @@ def self.create_or_update_from_csv_row(row, assignment)
...
-    (0..RUBRIC_LEVELS - 1).each do |i|
-      criterion['level_' + i.to_s + '_description'] = working_row.shift
-    end
-    unless criterion.save
-      raise CsvInvalidLineError
-    end
 
-    criterion
+    # create/update the levels
+    (0..RUBRIC_LEVELS - 1).each do
```
david-yz-liu: @ashih2018 here, instead of using the constant `RUBRIC_LEVELS` (which we should remove altogether eventually), use the length of `working_row` to calculate how many levels there arel.

ashih2018: for my next PR I will remove the constants and refactor the tests

PR #4204, opened 2019-10-29
Constant refactor

ashih2018:
Removed usages of RUBRIC_LEVEL, MAX_LEVEL, and MAX_MARK constants within rubric_criterion.rb as well as within rubric_criterion_spec.rb. Also edited how csv_rows are inputted as we do not need a 'weight' of marks as we needed before, so I removed the second entry in every csv_row which originally was a weight of 1.0, as levels keep track of marks now. 
Possible shortcomings and additional notes:
- I have kept the constants within the Rubric Criterion class as there are other references to them within Markus to the constants. I will work on removing those next.
- This PR should be merged after create_or_update from_csv PR gets merged as this branch is branched off of that one.

### DIFF
--- a/app/models/rubric_criterion.rb
+++ b/app/models/rubric_criterion.rb
@@ -81,7 +81,9 @@ class RubricCriterion < Criterion
   #                      does not evaluate to a float, or if the criterion is not
   #                      successfully saved.
   def self.create_or_update_from_csv_row(row, assignment)
-    if row.length < RUBRIC_LEVELS + 2
+    # get number of required fields
+    required_fields = Level.validators.grep(ActiveRecord::Validations::PresenceValidator).length
+    if row.length < required_fields + 2
       raise CsvInvalidLineError, I18n.t('upload_errors.invalid_csv_row_format')
     end
 
@@ -146,8 +148,10 @@ class RubricCriterion < Criterion
     criterion.name = name
     criterion.max_mark = criterion_yml[1]['max_mark']
 
+    # get number of required fields
+    required_fields = Level.validators.grep(ActiveRecord::Validations::PresenceValidator).length
     # Next comes the level names.
-    (0..RUBRIC_LEVELS - 1).each do |i|
+    (0..required_fields - 1).each do |i|
       if criterion_yml[1]['level_' + i.to_s]
         criterion['level_' + i.to_s + '_name'] =
           criterion_yml[1]['level_' + i.to_s]['name']
--- a/app/models/rubric_criterion.rb
+++ b/app/models/rubric_criterion.rb
@@ -85,7 +85,6 @@ class RubricCriterion < Criterion
     if row.length < 1
       raise CsvInvalidLineError, I18n.t('upload_errors.invalid_csv_row_format')
     end
-
     working_row = row.clone
     name = working_row.shift
     # If a RubricCriterion of the same name exits, load it up.  Otherwise,
@@ -115,6 +114,7 @@ class RubricCriterion < Criterion
       else
         criterion.levels.create(name: name, number: number, description: description, mark: mark)
       end
+
       unless criterion.save
         raise CsvInvalidLineError
       end
@@ -177,7 +177,7 @@ class RubricCriterion < Criterion
   end
 
   def weight
-    max_mark / MAX_LEVEL
+    self.max_mark / (self.levels.length - 1)
   end
 
   def round_max_mark
--- a/app/models/rubric_criterion.rb
+++ b/app/models/rubric_criterion.rb
@@ -116,12 +116,10 @@ class RubricCriterion < Criterion
         criterion.levels.create(name: name, number: number, description: description, mark: mark)
       end
       unless criterion.save
-        byebug
         raise CsvInvalidLineError
       end
     end
-    # criterion.update
-    criterion.max_mark = Float(criterion.levels.maximum('mark'))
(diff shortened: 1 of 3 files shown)


david-yz-liu:
@ashih2018 great, please do a merge from `sti-criteria`.

app/models/rubric_criterion.rb
```diff
@@ -82,7 +81,9 @@ def set_default_levels
   #                      does not evaluate to a float, or if the criterion is not
   #                      successfully saved.
   def self.create_or_update_from_csv_row(row, assignment)
-    if row.length < RUBRIC_LEVELS + 2
+    # get number of required fields
+    required_fields = Level.validators.grep(ActiveRecord::Validations::PresenceValidator).length
```
david-yz-liu: We can allow the user to upload a rubric criterion with no levels. So really the only thing to check here is that `working_row` has at least *one* element (the name of the criterion). So don't worry about the Level required fields here.

app/models/rubric_criterion.rb
```diff
@@ -92,28 +93,35 @@ def self.create_or_update_from_csv_row(row, assignment)
...
-    # the rest of the values are level descriptions.
-    (0..RUBRIC_LEVELS - 1).each do |i|
-      criterion['level_' + i.to_s + '_description'] = working_row.shift
-    end
-    unless criterion.save
-      raise CsvInvalidLineError
-    end
 
-    criterion
+    # there are 5 fields for each level
```
david-yz-liu: Looking at the loop below, it seems like there are 4 fields per level, not 5. (`name`, `number`, `description`, `mark`).

app/models/rubric_criterion.rb
```diff
@@ -136,8 +144,10 @@ def self.load_from_yml(criterion_yml)
     criterion.name = name
     criterion.max_mark = criterion_yml[1]['max_mark']
 
+    # get number of required fields
+    required_fields = Level.validators.grep(ActiveRecord::Validations::PresenceValidator).length
     # Next comes the level names.
-    (0..RUBRIC_LEVELS - 1).each do |i|
+    (0..required_fields - 1).each do |i|
```
david-yz-liu: This change is incorrect (you're mixing up the number of levels a criterion has by default with the number of required fields for a single level). Please revert this change and instead as @YolandaZIK to handle this, as she is working on this method.

spec/models/rubric_criterion_spec.rb
```diff
@@ -143,7 +143,9 @@
     context 'when parsing a CSV file' do
       describe 'raise csv line error on rows with elements without names for every criterion' do
         row = %w[name 1.0]
-        (0..RubricCriterion::RUBRIC_LEVELS - 2).each do |i|
+        # get number of required fields
+        required = Level.validators.grep(ActiveRecord::Validations::PresenceValidator).length
```
david-yz-liu: You should change this test (see my comment about the CSV row format above).

spec/models/rubric_criterion_spec.rb
```diff
@@ -177,65 +179,65 @@
       before(:each) do
         # we'll need a valid assignment for those cases.
         @assignment = create(:assignment)
-        row = ['criterion 5', '1.0']
-        (0..RubricCriterion::RUBRIC_LEVELS - 1).each do |i|
+        row = ['criterion 5']
+        # get number of required fields
+        required_fields = Level.validators.grep(ActiveRecord::Validations::PresenceValidator).length
```
david-yz-liu: See my comment about what `RUBRIC_LEVELS` really refers to. Here you should be able to choose in the test case how many levels you want to create, it doesn't depend on the number of required fields for a Level. You have the right idea in a test you wrote further down.

app/models/rubric_criterion.rb
```diff
@@ -170,10 +180,6 @@ def to_yml
         'peer_visible' => self.peer_visible } }
   end
 
-  def weight
```
david-yz-liu: Don't delete this method unless you've first checked that it's not used anywhere. Remember that in Ruby, it's possible to call a method without using parentheses, so look for a `<rubric criterion object>.weight`.

spec/models/rubric_criterion_spec.rb
ashih2018: Because we are only checking to see if the first input raises an error, can I hard code a number into the test?

app/models/rubric_criterion.rb
```diff
@@ -82,7 +81,8 @@ def set_default_levels
   #                      does not evaluate to a float, or if the criterion is not
   #                      successfully saved.
   def self.create_or_update_from_csv_row(row, assignment)
-    if row.length < RUBRIC_LEVELS + 2
+    # we only require the user to upload a single entry, as blank rubric criterion can be uploaded
+    if row.length < 1
```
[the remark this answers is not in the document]
david-yz-liu: @ashih2018 you should do this, use `row.empty?`

spec/models/rubric_criterion_spec.rb
```diff
@@ -177,9 +178,11 @@
       before(:each) do
         # we'll need a valid assignment for those cases.
         @assignment = create(:assignment)
-        row = ['criterion 5', '1.0']
+        row = ['criterion 5']
+        rubric_levels = 5
         # order is name, number, description, mark
-        (0..RubricCriterion::RUBRIC_LEVELS - 1).each do |i|
+        # .weight
```
david-yz-liu: Remove this line

PR #4207, opened 2019-10-31
1.8.0: Use multiple count for I18n string properly

mishaschwartz:
Be consistent with interpolated variables in I18n strings

This prevents the dashboard (admin landing page) from loading at all if:

- default language is English
- there at least two remark requests for the "featured assignment" (`Assignment.get_current_assignment`)


mishaschwartz:
Note for posterity: travis tests fail because of a test that doesn't take daylight saving time into account. This is fixed on the master branch with https://github.com/MarkUsProject/Markus/pull/4205/commits/94337c24684276c566328229d8a0a2b0c5d17416

PR #4229, opened 2019-11-08
refactored updating levels for cleaner code

ashih2018:
Made a simple change where I changed the way I previously updated levels model so that the code would be easier to understand, and gets rid of redundant code.


app/models/rubric_criterion.rb
```diff
@@ -102,20 +103,17 @@ def self.create_or_update_from_csv_row(row, assignment)
     end
 
     # there are 5 fields for each level
```
david-yz-liu: @ashih2018 please update this comment as well

PR #4230, opened 2019-11-09
saved criterion before adding levels to avoid parent not saved error

ashih2018:
There was an error trying to save a rubric criteria before hand because we were trying to create and save levels before the rubric criteria itself was saved. I made a small change where we first update the attributes in the criteria and then save it, and set the default levels at the very end so we will not get this error.

### DIFF
--- a/app/controllers/criteria_controller.rb
+++ b/app/controllers/criteria_controller.rb
@@ -28,7 +28,7 @@ class CriteriaController < ApplicationController
     end
     criterion_class = params[:criterion_type].constantize
     @criterion = criterion_class.new
-    @criterion.set_default_levels if params[:criterion_type] == 'RubricCriterion'
+
     if @criterion.update(name: params[:new_criterion_prompt],
                              assignment_id: @assignment.id,
                              max_mark: params[:max_mark_prompt],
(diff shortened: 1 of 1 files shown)


app/controllers/criteria_controller.rb
```diff
@@ -41,6 +41,8 @@ def create
       end
       head :unprocessable_entity
     end
+    @criterion.save!
+    @criterion.set_default_levels if params[:criterion_type] == 'RubricCriterion'
```
david-yz-liu: Don't do this here, move this to inside the `if` statement above. This eliminates the redundant save.

app/controllers/criteria_controller.rb
```diff
@@ -28,11 +28,13 @@ def create
     end
     criterion_class = params[:criterion_type].constantize
     @criterion = criterion_class.new
-    @criterion.set_default_levels if params[:criterion_type] == 'RubricCriterion'
+
     if @criterion.update(name: params[:new_criterion_prompt],
                              assignment_id: @assignment.id,
                              max_mark: params[:max_mark_prompt],
                              position: @assignment.next_criterion_position)
+      @criterion.save!
```
david-yz-liu: Calling `update` already saves the criterion, so there's no reason to save it again inside here (this is what I meant by it's redundant)

PR #4236, opened 2019-11-14
Change to_yml in rubric criterion

YolandaZIK:
Change to_yml in rubric criterion so that when users download all rubric criterions into a YAML file, the downloaded file will no longer include an index for each level. Instead, the name of a level is a key for its corresponding values (level description and mark).

### DIFF
--- a/app/models/rubric_criterion.rb
+++ b/app/models/rubric_criterion.rb
@@ -158,21 +158,17 @@ class RubricCriterion < Criterion
 
   # Returns a hash containing the information of a single rubric criterion.
   def to_yml
-    { self.name =>
-      { 'type' => 'rubric',
-        'max_mark' => self.max_mark.to_f,
-        'level_0' => { 'name' => self.level_0_name,
-                       'description' => self.level_0_description },
-        'level_1' => { 'name' => self.level_1_name,
-                       'description' => self.level_1_description },
-        'level_2' => { 'name' => self.level_2_name,
-                       'description' => self.level_2_description },
-        'level_3' => { 'name' => self.level_3_name,
-                       'description' => self.level_3_description },
-        'level_4' => { 'name' => self.level_4_name,
-                       'description' => self.level_4_description },
-        'ta_visible' => self.ta_visible,
-        'peer_visible' => self.peer_visible } }
+    levels_to_yml = { self.name => { 'type' => 'rubric',
+                                     'max_mark' => self.max_mark.to_f } }
+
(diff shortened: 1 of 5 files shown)


david-yz-liu:
@YolandaZIK you're on the right track here, but for the YML format I want the level data to be nested within a separate key, rather than at the top level of each criterion. So the examples should look like thing like

```
cr10:
  type: rubric
  max_mark: 10
  levels:
    level0:
      description: ...
      mark: ...
    level1:
      ...
    ...
```

app/models/rubric_criterion.rb
```diff
@@ -158,21 +158,18 @@ def self.load_from_yml(criterion_yml)
...
-                       'description' => self.level_3_description },
-        'level_4' => { 'name' => self.level_4_name,
-                       'description' => self.level_4_description },
-        'ta_visible' => self.ta_visible,
-        'peer_visible' => self.peer_visible } }
+    levels_to_yml = { self.name => { 'type' => 'rubric',
+                                     'max_mark' => self.max_mark.to_f,
+                                     'levels' => {} } }
+    index = 0
+    while index < self.levels.size - 1
```
david-yz-liu: @YolandaZIK you're on the right track, but can make this loop more idiomatic Ruby by doing

```ruby
self.levels.each do |level|
  ...
```

app/models/rubric_criterion.rb
```diff
@@ -158,21 +158,18 @@ def self.load_from_yml(criterion_yml)
...
+      levels_to_yml[self.name]['levels'][self.levels[index].name] = { 'description': self.levels[index].description,
+                                                                      'mark': self.levels[index].mark }
+      index += 1
+    end
+    levels_to_yml[self.name].store('ta_visible', self.ta_visible)
```
david-yz-liu: Move these two lines into the hash literal at lines 161-163 above.

doc/import_samples/criteria.yml
```diff
@@ -1,41 +1,41 @@
 Quality of Writing:
     type: rubric
     max_mark: 24.0
-    level_0:
-        name: Beginner
+    Beginner:
```
david-yz-liu: @YolandaZIK This is better, but not quite right: you should create a new key called `levels`, and nested all of this data under that key.

david-yz-liu: (You did this correctly in the examples below, so I think you just forgot this one.)

PR #4237, opened 2019-11-14
Change imported sample for criteria.yml

YolandaZIK:
Users should follow this new template in order to have levels as many as they want for each rubric criterion.


david-yz-liu:
@YolandaZIK you can update the examples in #4236 directly, since that is the pull request in which we are discussing the new format.

PR #4240, opened 2019-11-15
Enable error messages and progress messages for all jobs

mishaschwartz:
- added new message strings for reporting errors and progress messages for jobs
- moved default job status methods into the `ApplicationJob` class so that they can be inherited by subclasses
- added job status methods to all jobs classes that should override the defaults
- report error messages by capturing `StandardError` for all `perform` messages
- change the logic (order of `if` statements) in `JobMessagesController.get` in order to make sure that the correct message is displayed throughout the lifecycle of the job
- use the `:error_message` status key to indicate whether an error occurred during the job lifecycle
     - the `:failed` key should get set when `ActiveJob::Status` catches an `Exception` however this does not appear to happen consistently (unsure why)

### DIFF
--- a/app/controllers/job_messages_controller.rb
+++ b/app/controllers/job_messages_controller.rb
@@ -14,9 +14,9 @@ class JobMessagesController < ApplicationController
       flash_message(:notice, t('poll_job.queued'))
     elsif status.working?
       flash_progress_message(status)
-    elsif status[:job_class]&.show_error_message(status).present?
+    elsif status[:error_message].present?
       flash_progress_message(status)
-      flash_message(:error, status[:job_class].show_error_message(status))
+      flash_message(:error, status[:error_message])
     elsif status.completed?
       status[:progress] = status[:total]
       flash_progress_message(status)
@@ -33,10 +33,6 @@ class JobMessagesController < ApplicationController
   private
 
   def flash_progress_message(status)
-    if status[:job_class]
-      flash_message(:notice, status[:job_class].show_status(status))
-    else # default x out of y message
-      flash_message(:notice, t('poll_job.working_message', progress: status[:progress], total: status[:total]))
-    end
+    flash_message(:notice, status[:job_class].show_status(status))
   end
 end
--- a/app/controllers/job_messages_controller.rb
+++ b/app/controllers/job_messages_controller.rb
@@ -33,6 +33,7 @@ class JobMessagesController < ApplicationController
   private
 
   def flash_progress_message(status)
-    flash_message(:notice, status[:job_class].show_status(status))
+    current_status = status[:job_class].show_status(status)
+    flash_message(:notice, current_status) unless current_status.nil?
   end
 end
--- a/app/controllers/job_messages_controller.rb
+++ b/app/controllers/job_messages_controller.rb
@@ -1,39 +1,32 @@
 class JobMessagesController < ApplicationController
   before_action :authorize_only_for_admin
 
-  def show
-   @job_message = Rails.cache.fetch(params[:job_id])
-   respond_to do |format|
-     format.json { render 'job_messages/show'}
-   end
-  end
-
   def get
     status = ActiveJob::Status.get(params[:job_id])
-    if status.queued?
-      flash_message(:notice, t('poll_job.queued'))
-    elsif status.working?
-      flash_progress_message(status)
-    elsif status[:error_message].present?
-      flash_progress_message(status)
-      flash_message(:error, status[:error_message])
+    if status.failed?
+      flash_message(:error, t('poll_job.failed')) unless status[:error_message].present?
+      session[:job_id] = nil
     elsif status.completed?
       status[:progress] = status[:total]
-      flash_progress_message(status)
       flash_message(:success, t('poll_job.completed'))
-    else
-      flash_message(:error, t('poll_job.failed'))
-    end
-    if status.completed? || status.failed?
       session[:job_id] = nil
+    elsif status.read.empty?
+      flash_message(:error, t('poll_job.failed'))
+      session[:job_id] = nil
+      render 'shared/http_status', locals: { code: '404', message: 'No background job was enqueued' }, status: 404
+      return
     end
+    flash_job_messages(status)
     render json: status.read
   end
 
   private
 
-  def flash_progress_message(status)
-    current_status = status[:job_class].show_status(status)
-    flash_message(:notice, current_status) unless current_status.nil?
+  def flash_job_messages(status)
+    flash_message(:error, status[:error_message]) if status[:error_message].present?
+    current_status = status[:job_class]&.show_status(status)
+    return if current_status.nil?
+
+    status.queued? ? flash_message(:notice, t('poll_job.queued')) : flash_message(:notice, current_status)
   end
 end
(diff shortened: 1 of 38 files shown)


app/jobs/create_groups_job.rb
```diff
@@ -36,72 +28,69 @@ def log_creation
...
-            member_hash = StudentMembership.joins(:grouping)
-                                           .joins(:user)
-                                           .where('groupings.group_id': group.id)
-                                           .where(membership_status: [:inviter, :accepted])
-                                           .pluck('groupings.assignment_id', 'users.user_name')
-                                           .group_by(&:first)
-                                           .transform_values { |g| Set.new g.map(&:second) }
+    Repository.get_class.update_permissions_after(only_on_request: true) do
+      ApplicationRecord.transaction do
+        data.each do |group_name, repo_name, *members|
```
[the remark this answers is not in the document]
mishaschwartz: vote to ignore

app/jobs/create_groups_job.rb
```diff
@@ -36,72 +28,69 @@ def log_creation
...
-            group = log_creation { Group.find_or_create_by(group_name: group_name, repo_name: repo_name) }
-            member_hash = StudentMembership.joins(:grouping)
-                                           .joins(:user)
-                                           .where('groupings.group_id': group.id)
-                                           .where(membership_status: [:inviter, :accepted])
-                                           .pluck('groupings.assignment_id', 'users.user_name')
-                                           .group_by(&:first)
-                                           .transform_values { |g| Set.new g.map(&:second) }
+    Repository.get_class.update_permissions_after(only_on_request: true) do
+      ApplicationRecord.transaction do
```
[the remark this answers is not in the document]
mishaschwartz: vote to ignore

app/jobs/create_groups_job.rb
```diff
@@ -36,72 +28,69 @@ def log_creation
...
-          data.each do |group_name, repo_name, *members|
-            group = log_creation { Group.find_or_create_by(group_name: group_name, repo_name: repo_name) }
-            member_hash = StudentMembership.joins(:grouping)
-                                           .joins(:user)
-                                           .where('groupings.group_id': group.id)
-                                           .where(membership_status: [:inviter, :accepted])
-                                           .pluck('groupings.assignment_id', 'users.user_name')
-                                           .group_by(&:first)
-                                           .transform_values { |g| Set.new g.map(&:second) }
+    Repository.get_class.update_permissions_after(only_on_request: true) do
```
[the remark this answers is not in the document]
mishaschwartz: vote to ignore

app/controllers/job_messages_controller.rb
```diff
@@ -10,29 +10,30 @@ def show
 
   def get
     status = ActiveJob::Status.get(params[:job_id])
-    if status.working?
-      if status[:job_class]
-        flash_now(:notice, status[:job_class].show_status(status))
-      else #default x out of y message
-        flash_now(:notice, t('poll_job.working_message', progress: status[:progress], total: status[:total]))
-      end
+    if status.queued?
```
david-yz-liu: I'm wondering if before this we should actually be checking for `status.nil?`; I don't think this has come up before, but it seems like this might be a good thing to check just to be safe.

app/jobs/autotest_cancel_job.rb
```diff
@@ -1,6 +1,12 @@
 class AutotestCancelJob < ApplicationJob
   queue_as MarkusConfigurator.autotest_cancel_queue
 
+  def self.on_complete_js(_status)
+    'window.BatchTestRunTable.fetchData'
```
david-yz-liu: I don't see how this could work right now? I think you need to bind the component to a variable in https://github.com/MarkUsProject/Markus/blob/master/app/views/assignments/batch_runs.html.erb#L4.

app/jobs/generate_job.rb
```diff
@@ -3,61 +3,61 @@ class GenerateJob < ApplicationJob
...
-      Rails.logger.error e.message
-      raise e
+      combine_pdf_qr = CombinePDF.parse(pdf.render)
+      template_pdf.pages.zip(combine_pdf_qr.pages) do |template_page, qr_page|
+        generated_pdf << (qr_page << template_page)
+      end
+      progress.increment
+    rescue StandardError => e
+      status.update(error_message: e.to_s)
+      raise
```
david-yz-liu: I might be missing something here, but is this `raise` actually going to be caught anywhere? The pattern seems different than all the other jobs, where there's a top-level `rescue` that doesn't raise an error.

app/controllers/job_messages_controller.rb
mishaschwartz: Yes, why not? Better safe than sorry 👍

app/jobs/generate_job.rb
mishaschwartz: There is a `rescue_from` in active job status that should catch all remaining raised errors. The reason this re-raises the error is actually to appease rubocop which doesn't like plain return statements from within a loop. Re-raising the error seemed to be the cleanest way to exit the job early while keeping rubocop happy. 

Alternatives are to return here or to break and then to include some logic afterwards to prevent an incomplete pdf from being created.

mishaschwartz:
> Specifically: did you try defining using rescue_from in ApplicationJob? (e.g. https://stackoverflow.com/questions/29085458/how-to-access-perform-parameters-in-activejob-rescue) It would be great if that works.

I did try adding an additional `rescue_from` block in `ApplicationJob` at some point but I got rid of it because it seemed to be being ignored. Possibly because there is a second `rescue_from` that's included from `ActiveJob::Status` already but it didn't seem to make sense to me that you couldn't have multiple `rescue_from`s. I'll go back and investigate further.

app/controllers/job_messages_controller.rb
mishaschwartz: Actually, `status` will never be `nil` because `ActiveJob::Status.get` calls its `new` method which will always create a new object. If a job with `:job_id` doesn't exist already then no data will be read from `Storage` and the status will be empty instead of `nil`. In this case, it will be handled by the `else` block.

app/jobs/create_groups_job.rb
```diff
@@ -36,74 +28,67 @@ def log_creation
...
```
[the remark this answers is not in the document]
mishaschwartz: vote to ignore

app/jobs/generate_job.rb
mishaschwartz: this has been resolved because error catching has been moved to the `rescue_from` block in `ApplicationJob`

app/controllers/job_messages_controller.rb
mishaschwartz: This has been resolved by adding the ability for this method to render a 404 error if there is no status messages (it is empty not nil).

app/controllers/job_messages_controller.rb
```diff
@@ -1,38 +1,32 @@
...
+      flash_message(:error, t('poll_job.failed')) unless status[:error_message].present?
+      session[:job_id] = nil
+    elsif status.completed?
+      status[:progress] = status[:total]
+      flash_message(:success, t('poll_job.completed'))
+      session[:job_id] = nil
+    elsif status.read.empty?
+      flash_message(:error, t('poll_job.failed'))
+      session[:job_id] = nil
+      render 'shared/http_status', locals: { code: '404', message: 'No background job was enqueued' }, status: 404
```
david-yz-liu: Please internationalize this string.

app/jobs/autotest_cancel_job.rb
david-yz-liu: I don't think you actually added this.

PR #4247, opened 2019-11-25
Criteria yml upload: better warning messages

mishaschwartz:
- make it clear that grades will be deleted when uploading a yml file after marking has already started
- make sure the total mark is updated for all results after deleting marks

### DIFF
--- a/config/locales/views/assignments/en.yml
+++ b/config/locales/views/assignments/en.yml
@@ -3,6 +3,7 @@ en:
     due_date:
       final_due_date_passed: "The deadline, including late or grace periods, has passed."
       marking_started_warning: "Marking has already started for this assignment."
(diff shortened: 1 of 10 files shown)


config/locales/views/assignments/en.yml
```diff
@@ -3,6 +3,7 @@ en:
     due_date:
       final_due_date_passed: "The deadline, including late or grace periods, has passed."
       marking_started_warning: "Marking has already started for this assignment."
+      marking_overwrite_warning: "<strong>Any previous grading work will be lost.</strong>"
```
david-yz-liu: @mishaschwartz sorry I forgot about this PR. Everything looks good, but for this string can you use the suffix `_html` for the key? When you do this, you won't need to use `.html_safe` to render the string (https://guides.rubyonrails.org/i18n.html#using-safe-html-translations).

PR #4251, opened 2019-12-01
Don't override native text range selection keybindings

Xyene:
Oftentimes while marking, I'll select some code to annotate by double-clicking a token, which might select more (or less) text on the right end than I wanted to, at which point I'll instinctively type Shift+Left/Right to adjust my range selection.

Currently, Shift+Left/Right are rebound to go to the previous/next submissions, which is frustrating (muscle memory makes it so that this happens almost every submission, at least once).

This PR proposes disabling the Shift-modifier submission-navigation keybindings if there is currently an active range selection, so that existing workflows are (hopefully) not affected.

### DIFF
--- a/app/assets/javascripts/Results/keybinding.js
+++ b/app/assets/javascripts/Results/keybinding.js
@@ -3,14 +3,23 @@ Mousetrap.bind('?', function() {
   modal_help.open();
 });
 
+function is_text_selected() {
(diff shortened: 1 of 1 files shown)


david-yz-liu:
@Xyene that's awesome, thanks for contributing this. Do you mind if I add your name to the contributors list?

Xyene:
@david-yz-liu sure, that's fine 😄

PR #4273, opened 2019-12-03
ui: vertically align flash dismiss button

Xyene:
This PR proposes a minor UI tweak to the alignment of the flash dismissal button.

Before:
image

After:
image


Xyene:
Whoops, doesn't actually work in not-Chrome.

PR #4274, opened 2019-12-03
Assignment create/update bugfix

mishaschwartz:
- the repo folder for a peer review assignment is set in the `Assignment.create_peer_review_assignment_if_not_exist` method and so should not be updated by this form.

- Update methods for checking if an attribute was changed after save (as of Rails 5.2). See this table (https://github.com/rails/rails/pull/25337#issuecomment-225166796) for reference.

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -655,14 +655,12 @@ class AssignmentsController < ApplicationController
   def process_assignment_form(assignment)
     num_files_before = assignment.assignment_files.length
     assignment.assign_attributes(assignment_params)
-    assignment.repository_folder = assignment_params[:short_identifier]
-    new_required_files = assignment.only_required_files_changed? ||
-                         assignment.is_hidden_changed? ||
-                         assignment.assignment_files.any? { |file| file.changed? }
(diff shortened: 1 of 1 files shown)


app/controllers/assignments_controller.rb
```diff
@@ -655,14 +655,12 @@ def get_file_info(file_name, file, assignment_id, path)
...
-    assignment.repository_folder = assignment_params[:short_identifier]
-    new_required_files = assignment.only_required_files_changed? ||
-                         assignment.is_hidden_changed? ||
-                         assignment.assignment_files.any? { |file| file.changed? }
+    assignment.repository_folder = assignment_params[:short_identifier] unless assignment.is_peer_review?
     assignment.save!
-    new_required_files = new_required_files ||
+    new_required_files = assignment.saved_change_to_only_required_files? ||
+                         assignment.saved_change_to_is_hidden? ||
+                         assignment.assignment_files.any?(&:changed?) ||
```
david-yz-liu: Doesn't `changed?` need to be renamed as well?

mishaschwartz: good catch, I thought I had updated that one as well.

PR #4322, opened 2020-01-13
docker: Make data/dev a separate docker volume.

david-yz-liu:


mishaschwartz:
1. This seems to make the `data/dev/repos` subdirectory unwritable from within the docker container.
I believe this is actually the intended consequence but I'm investigating further.
2. Is there a reason why we need a separate volume for this subdir (the entire `app` directory is a volume/bind mount already and we don't need to share this volume with other containers as far as I can tell)?

mishaschwartz:
Ok I believe the initial issue I was having was more to do with my docker setup than any issue with your changes itself. 

This now seems to be running properly and when I bring up a new service and the group seeding seems to be going at a decent rate. I didn't do a time test but my guess is that this is in fact improving the performance as intended. 

Normal group creation happens without any noticeable slowness as well now.

PR #4326, opened 2020-01-16
Notes policy

ANUSHIYAPRIYA:
1. Created new policy called NotePolicy to manage user permissions in notes controller
2. Added new test file for note_policy.rb.
3. Made notes_dialog a collection method from member method. previously it was passing the note_id for an assignment, so now changed the note_id into assignment_id. Because an assignment can have multiple notes.

### DIFF
--- a/app/controllers/notes_controller.rb
+++ b/app/controllers/notes_controller.rb
@@ -1,12 +1,17 @@
 class NotesController < ApplicationController
-  before_action :authorize_for_ta_and_admin
-  before_action :ensure_can_modify, only: [:edit, :update]
-
+  before_action do |_|
+    if params[:id].nil?
+      authorize! with: NotePolicy
+    else
+      note = Note.find(params[:id])
+      authorize! note, with: NotePolicy
+    end
+  end
   responders :flash, :collection
 
   # TODO this method needs explaining ! What is return_id ?
   def notes_dialog
-    @return_id = params[:id]
+    @return_id = params[:assignment_id]
     @cls = params[:noteable_type]
     @noteable = Kernel.const_get(@cls).find_by_id(params[:noteable_id])
     @cont = params[:controller_to]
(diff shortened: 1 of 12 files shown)


app/policies/note_policy.rb
```diff
@@ -0,0 +1,19 @@
+class NotePolicy < ApplicationPolicy
+  default_rule :manage?
+
+  def manage?
+    user.ta? || user.admin?
+  end
+
+  def ensure_can_modify?
```
david-yz-liu: Rename this to `modify?`

app/policies/note_policy.rb
```diff
@@ -0,0 +1,19 @@
...
+    user.admin? || user.id == record.creator_id
+  end
+
+  def edit?
```
david-yz-liu: Why don't you use aliasing here for `edit?` and `update?`

app/controllers/notes_controller.rb
```diff
@@ -106,11 +113,12 @@ def update
...
-    if @note.user_can_modify?(current_user)
+    begin
+      authorize! @note, to: :ensure_can_modify?
       @note.destroy
-      respond_with @note
-    else
-      flash_message(:error, I18n.t('notes.delete.error_permissions'))
+      respond_with(@note)
+    rescue ActionPolicy::Unauthorized
+      flash_now(:error, t('action_policy.policy.note.ensure_can_modify?'))
```
mishaschwartz: Can we use the actual error message here. This will make things easier if we decide to change the error message in the future.

app/controllers/notes_controller.rb
```diff
@@ -1,12 +1,17 @@
 class NotesController < ApplicationController
-  before_action :authorize_for_ta_and_admin
-  before_action :ensure_can_modify, only: [:edit, :update]
-
+  before_action do |_|
+    if params[:id].nil?
+      authorize! with: NotePolicy
```
david-yz-liu: Can you and Misha check if you and just do `authorize!` without the `with: NotePolicy` part (I think this should be looked up automatically).

PR #4328, opened 2020-01-17
Added GroupPolicy for groups_controller

ANUSHIYAPRIYA:
1. Added group policy to manage user permissions in groups_controller
2. Added test file for group_policy.rb


app/controllers/groups_controller.rb
```diff
@@ -3,24 +3,7 @@
...
-  before_action      :authorize_for_student,
-                     only: [:create,
-                            :destroy,
-                            :delete_rejected,
-                            :disinvite_member,
-                            :invite_member,
-                            :accept_invitation,
-                            :decline_invitation]
-
+  before_action { authorize! with: GroupPolicy }
```
david-yz-liu: @ANUSHIYAPRIYA the only comment I have is to just do `authorize!` here.

PR #4330, opened 2020-01-21
Edit levels

ashih2018:
Was able to fix bug with multiple levels of one instance showing up. Currently working on having levels save with nested attributes as the current error that occurs is 'unpermitted parameter :level'.
Also changed self.levels.build to self.levels.create in rubric_criterion.rb as it allows the level to be saved (as far as I know, build is an alias to new, other than slight changes but we want to use create in this case).

### DIFF
--- a/app/controllers/criteria_controller.rb
+++ b/app/controllers/criteria_controller.rb
@@ -36,7 +36,6 @@ class CriteriaController < ApplicationController
       @criterion.set_default_levels if params[:criterion_type] == 'RubricCriterion'
       flash_now(:success, t('flash.actions.create.success',
                             resource_name: criterion_class.model_name.human))
-      @criterion
     else
       @criterion.errors.full_messages.each do |message|
         flash_message(:error, message)
--- a/app/controllers/criteria_controller.rb
+++ b/app/controllers/criteria_controller.rb
@@ -74,7 +74,6 @@ class CriteriaController < ApplicationController
       return
     end
     if criterion_type == 'RubricCriterion'
-      byebug
       properly_updated = @criterion.update(rubric_criterion_params.except(:assignment_files))
       unless rubric_criterion_params[:assignment_files].nil?
         assignment_files = AssignmentFile.find(rubric_criterion_params[:assignment_files].select { |id| !id.empty? })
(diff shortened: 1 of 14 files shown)


app/controllers/criteria_controller.rb
```diff
@@ -224,8 +224,16 @@ def rubric_criterion_params
                                              :position,
                                              :ta_visible,
                                              :peer_visible,
+                                             :max_mark,
+                                             levels_attributes: [:id, :name, :mark, :description],
                                              assignment_files: []).to_h.deep_merge(params.require(:rubric_criterion)
-                                                                           .permit(:max_mark).to_h)
+                                                                           .permit(:max_mark,
```
david-yz-liu: Can you try deleting the `to_h` and everything after it on the previous line? I'm not sure this entire section is still necessary.

app/views/rubric_criteria/_rubric_criterion_level.html.erb
```diff
@@ -1,25 +1,25 @@
...
-<div>
-  <%= f.label :mark, Level.human_attribute_name(:mark) + " (#{level.mark}/#{criterion.max_mark})",
-              class: 'required' %>
-  <%= f.number_field :mark, value: "#{level.mark}", step: 0.1%>
-</div>
-<div>
-  <%= f.label :description, Level.human_attribute_name(:description), class: 'description required' %>
-  <%= f.text_area :description, value: "#{level.description}"%>
-</div>
+<% i=0 %>
```
david-yz-liu: This use of an index is okay. In terms of style, always put spaces around `=`, and also put spaces after the `<%`/`<%=` and before the `%>` (you need to fix this throughout this file).

app/views/rubric_criteria/_rubric_criterion_level.html.erb
```diff
@@ -1,25 +1,25 @@
...
+<%= criterion_form.fields_for :levels do |level_form|%>
+  <div class="rubric_level_header bold_inline_label">
```
david-yz-liu: Don't use a separate class `bold_inline_label`. Instead, just add the styles you want in the CSS to `rubric_level_header`.

app/views/rubric_criteria/_rubric_criterion_level.html.erb
```diff
@@ -1,25 +1,25 @@
...
+    Level <%=i%>
```
david-yz-liu: Use `I18n` rather than the literal string "Level".

PR #4331, opened 2020-01-21
Anonymous grading

mishaschwartz:
- add columns in `assignments` table to keep track of whether group information should be hidden from graders as well as whether non-assigned criteria should be displayed
- add checkboxes to the graders page to toggle these new table values
- added controller methods to process changes to these new values
- added tests for these controller methods, new routes
- removed tests for old routes that have been removed 
- update views to not display anonymous group info or non-assigned criteria
- update controller methods not to pass anonymous group information

#### TODO:

- [x] bulk set complete/incomplete for admins
- [x] add rspec tests for new controller functionality
- [x] update changelog


app/assets/javascripts/Components/graders_manager.jsx
```diff
@@ -156,11 +160,51 @@ class GradersManager extends React.Component {
   toggleAssignGradersToCriteria = () => {
     const assign = !this.state.assign_graders_to_criteria;
     $.post({
-      url: Routes.set_assign_criteria_assignment_graders_path(this.props.assignment_id),
-      data: {value: assign},
+      url: Routes.set_boolean_graders_options_assignment_path(this.props.assignment_id),
+      data: {
+        value: assign,
+        attribute: 'assign_graders_to_criteria'
```
david-yz-liu: Good idea unifying the routes here. One comment I have is you should be able to mimic the typical attribute update params, e.g. `data: {properties: {assign_graders_to_criteria: assign}}`, and then in the `GradersController` first do a `params.permit` and then simply pass the whitelisted params to `Assignment.update` directly.

app/assets/javascripts/Components/assignment_summary_table.jsx
```diff
@@ -82,7 +82,14 @@ class AssignmentSummaryTable extends React.Component {
     {
       Header: I18n.t('activerecord.attributes.result.total_mark'),
       accessor: 'final_grade',
-      Cell: ({value}) => value ? value + ' / ' + this.props.max_mark : '',
+      Cell: row => {
+        if (row.original.final_grade) {
+          const max_mark = Math.round(row.original.max_mark * 100) / 100;
+          return row.original.final_grade + ' / ' + max_mark
```
david-yz-liu: Please insert a semicolon here (and line 90 below).

app/assets/javascripts/Components/graders_manager.jsx
```diff
@@ -156,11 +160,51 @@ class GradersManager extends React.Component {
...
+      return <div style={{marginBottom: '1em'}}>
+        <label>
+          <input type="checkbox"
+                 checked={this.state.hide_unassigned_criteria}
+                 onChange={this.toggleHideUnassignedCriteria}
+                 style={{marginRight: '5px'}}
+          />
+          {I18n.t('graders.hide_unassigned_criteria')}
+        </label>
+      </div>
```
david-yz-liu: Please insert a semicolon here.

app/assets/javascripts/Components/submission_table.jsx
```diff
@@ -137,7 +137,11 @@ class RawSubmissionTable extends React.Component {
     {
       Header: I18n.t('activerecord.attributes.result.total_mark'),
       accessor: 'final_grade',
-      Cell: ({value}) => (value === undefined ? '-' : value) + ' / ' + this.props.max_mark,
+      Cell: row => {
+        const value = row.original.final_grade;
+        const max_mark = Math.round(row.original.max_mark * 100) / 100;
+        return (value === undefined ? '-' : value) + ' / ' + max_mark
```
david-yz-liu: Please insert a semicolon here.

app/assets/javascripts/Components/submission_table.jsx
```diff
@@ -192,6 +196,16 @@ class RawSubmissionTable extends React.Component {
     });
   };
 
+  setMarkingStates = (marking_state) => {
```
david-yz-liu: This is nice, but actually can you pull out this change into a separate PR? I'd like to put this into the next minor release separately.

app/controllers/assignments_controller.rb
```diff
@@ -581,6 +581,23 @@ def switch_assignment
     end
   end
 
+  def set_boolean_graders_options
+    unless %w[assign_graders_to_criteria
```
david-yz-liu: (This is the part I was referring to above where you should be able to use `params.permit`.

app/controllers/results_controller.rb
```diff
@@ -139,7 +139,11 @@ def show
           data[:notes_count] = submission.grouping.notes.count
           data[:num_marked] = assignment.get_num_marked(current_user.admin? ? nil : current_user.id)
           data[:num_assigned] = assignment.get_num_assigned(current_user.admin? ? nil : current_user.id)
-          data[:group_name] = submission.grouping.get_group_name
+          if current_user.ta? && assignment.anonymize_groups
+            data[:group_name] = "Group #{submission.grouping.id}"
```
david-yz-liu: Internationalize the `"Groups"` string (this appears in a few places).

app/controllers/results_controller.rb
```diff
@@ -139,7 +139,11 @@ def show
           data[:notes_count] = submission.grouping.notes.count
           data[:num_marked] = assignment.get_num_marked(current_user.admin? ? nil : current_user.id)
           data[:num_assigned] = assignment.get_num_assigned(current_user.admin? ? nil : current_user.id)
-          data[:group_name] = submission.grouping.get_group_name
+          if current_user.ta? && assignment.anonymize_groups
```
david-yz-liu: Elsewhere you use `!current_user.admin?`, so I think you should be consistent here.

app/controllers/results_controller.rb
```diff
@@ -176,26 +180,31 @@ def show
...
+        if original_result.nil?
+          old_marks = {}
+        else
+          old_marks = original_result.mark_hash
+        end
 
-          marks_map = marks_map.partition { |m| assigned_criteria.include? "#{m[:criterion_type]}-#{m[:id]}" }
-                               .flatten
+        if assignment.assign_graders_to_criteria && current_user.ta?
+          assigned_crit = current_user.criterion_ta_associations
```
david-yz-liu: Please change the variable name back to `assigned_criteria`.

app/controllers/results_controller.rb
david-yz-liu: Or better, use `ActionPolicy` for this? Something like `allowed_to? :view_group_membership, assignment`.

app/views/submissions/browse.html.erb
```diff
@@ -8,7 +8,6 @@
           show_sections: <%= Section.exists? %>,
           is_admin: <%= @current_user.admin? %>,
           can_run_tests: <%= @current_user.admin? && @assignment.enable_test %>,
-          max_mark: <%= @assignment.max_mark %>,
```
david-yz-liu: I get what you're doing here, but I wonder if there's a better way here. For a given TA and assignment, the "`max_mark`" is the same for all submissions (depends only on assigned criteria), so it's still possible to pass in a single value here as a prop instead of of including it in every row. I don't feel too strongly about this, but maybe something you could think about for a bit?

app/assets/javascripts/Components/Result/marks_panel.jsx
```diff
@@ -32,7 +32,7 @@ export class MarksPanel extends React.Component {
     //   2) A mark has not yet been given, and the current user can give the mark.
     let expanded = new Set();
     this.props.marks.forEach(data => {
-      const key = `${data.criterion_type}-${data.id}`;
+      const key = `criterion_${data.criterion_type}_${data.id}`;
```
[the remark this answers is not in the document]
mishaschwartz: This is the same string format that is used in `Result.mark_hash` and `Assignment.summary_json`. I can change it back to the old one if you would like but then I would recommend updating the other two for consistency.

app/assets/javascripts/Components/Result/summary_panel.jsx
```diff
@@ -195,10 +196,20 @@ export class SummaryPanel extends React.Component {
   };
 
   renderGraceTokenDeductions = () => {
+    let rows;
     if (this.props.is_reviewer || this.props.graceTokenDeductions.length === 0) {
       return '';
+    } else if (this.props.graceTokenDeductions[0]['users.user_name'] === undefined) {
```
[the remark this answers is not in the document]
mishaschwartz: It's not going to be a column of numbers, it's only a single line indicating the number of grace tokens deducted. There is no duplication even if there are multiple group members.

app/controllers/results_controller.rb
mishaschwartz: Looking at this now I think we should actually change the rest to `current_user.ta?`. It's always clearer to make the positive check. Thoughts?

And yes, eventually I want to do a lot of action policy changes to these files but I figure that's a separate PR.

app/assets/javascripts/Components/submission_table.jsx
mishaschwartz: #4336

app/views/submissions/browse.html.erb
david-yz-liu: @mishaschwartz I merged in the PR, but can you please make a separate PR to address this comment? I think you missed it the last round.

mishaschwartz: @david-yz-liu  Yes I thought we agreed that this would be a separate PR

PR #4334, opened 2020-01-25
config: Remove MarkusConfigurator.

david-yz-liu:


app/controllers/assignments_controller.rb
```diff
@@ -244,7 +244,7 @@ def update
         @assignment, new_required_files = process_assignment_form(@assignment)
         @assignment.save!
       end
-      if new_required_files && !MarkusConfigurator.markus_config_repository_hooks.empty?
+      if new_required_files && !Rails.configuration.x.repository.hooks.empty?
```
mishaschwartz: I'm assuming you're using the `x` here to not conflict with an existing configuration setting called `repository`. Would it not be better to just put all of our custom MarkUs settings in their own namespace.

So make the following changes for example:
```
Rails.configuration.x.repository.hooks -> Rails.configuration.markus.repository.hooks
```

but also change these ones:

```
Rails.configuration.logout_redirect -> Rails.configuration.markus.logout_redirect
```

To make it clear which are the MarkUs settings

david-yz-liu: This is what is recommended by the Rails guide: https://guides.rubyonrails.org/configuring.html#custom-configuration.

PR #4335, opened 2020-01-27
Fixed all tests

robert-tan:
As of right now, all rspec tests are passing with the sti-assessments integration. Further testing would be required until it is ready to merge into master.


app/models/assessment.rb
```diff
@@ -13,8 +13,9 @@ class Assessment < ApplicationRecord
   validates_presence_of :short_identifier
   validate :short_identifier_unchanged, on: :update
   validates_presence_of :description
-  validates_presence_of :due_date
+  validates_presence_of :due_date, if: ->(assessment) { assessment.type == 'Assignment' }
```
david-yz-liu: I just realized that you can move this validation into the `Assignment` model rather than keep it here. That way you should be able to remove the `if`.

app/models/assignment.rb
```diff
@@ -1266,15 +1283,16 @@ def self.upload_assignment_list(file_format, assignment_data)
     case file_format
     when 'csv'
       result = MarkusCsv.parse(assignment_data) do |row|
-        assignment = self.find_or_create_by(short_identifier: row[0],
-                                            assignment_properties: { repository_folder: row[0] })
+        assignment = self.find_or_create_by(short_identifier: row[0])
         attrs = Hash[DEFAULT_FIELDS.zip(row)]
         attrs.delete_if { |_, v| v.nil? }
         if assignment.new_record?
+          assignment.assignment_properties = AssignmentProperties.new
```
david-yz-liu: Pass in options `token_period: 1, etc.` to the call to `new`, rather than having the three separate statements on lines 1293-1295.

app/models/assignment.rb
```diff
@@ -1284,13 +1302,14 @@ def self.upload_assignment_list(file_format, assignment_data)
...
-          assignment = self.find_or_create_by(short_identifier: row[:short_identifier],
-                                              assignment_properties: { repository_folder: row[:short_identifier] })
+          assignment = self.find_or_create_by(short_identifier: row[:short_identifier])
           if assignment.new_record?
+            row[:assignment_properties] = AssignmentProperties.new
             row[:submission_rule] = NoLateSubmissionRule.new
             row[:assignment_stat] = AssignmentStat.new
-            row[:token_period] = 1
-            row[:unlimited_tokens] = false
+            row[:assignment_properties][:repository_folder] = row[:short_identifier]
```
david-yz-liu: Similar to the above comment, pass these three attribute values directly to the `AssignmentProperties.new` call above.

config/.rubocop.yml
```diff
@@ -7,7 +7,7 @@ AllCops:
     - 'Vagrantfile'
 
 # Multi-line method chaining should be done with leading dots.
-Layout/DotPosition:
+Metrics/DotPosition:
```
david-yz-liu: Please revert the changes in this file. (We can discuss these changes separately, but not in this PR.)

db/migrate/20190914204954_replace_assignments_and_grade_entry_forms_with_assessments.rb
```diff
@@ -30,7 +30,7 @@ def change
       t.string :short_identifier, null: false
       t.string :type, null: false
       t.string :description, null: false
-      t.text :message, null: false
+      t.text :message
```
david-yz-liu: Keep the `null: false`, but you can have this be an empty string as a default value.

db/schema.rb
```diff
@@ -545,7 +545,7 @@
 
   create_table "test_results", id: :serial, force: :cascade do |t|
     t.text "name", null: false
-    t.text "status", null: false
+    t.text "status", null: falsel
```
david-yz-liu: Revert this change.

spec/models/assignment_properties_spec.rb
```diff
@@ -3,7 +3,6 @@
     it { is_expected.to validate_presence_of(:repository_folder) }
     it { is_expected.to validate_presence_of(:group_min) }
     it { is_expected.to validate_presence_of(:group_max) }
-    it { is_expected.to validate_presence_of(:notes_count) }
```
david-yz-liu: This is okay, but then add it back to `assessment_spec.rb`.

PR #4336, opened 2020-01-27
Bulk set to complete/incomplete on submissions table

mishaschwartz:
TODO:
- [x] update changelog
- [x] write spec tests for new routes, controller method, model method

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -16,6 +16,7 @@ class SubmissionsController < ApplicationController
                          :download_groupings_files,
                          :manually_collect_and_begin_grading,
                          :repo_browser,
+                         :set_result_marking_state,
                          :update_submissions,
                          :populate_submissions_table,
                          :populate_peer_submissions_table]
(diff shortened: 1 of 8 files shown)


app/controllers/submissions_controller.rb
```diff
@@ -611,6 +612,27 @@ def server_time
...
+      flash_now(:error, t('groups.select_a_group'))
+      head 400
+      return
+    end
+    results = Result.where(id: Grouping.joins(:current_result).where(id: params[:groupings]).select('results.id'))
+    Result.transaction do
+      results.each do |result|
+        result.marking_state = params[:marking_state]
+        unless result.save
+          raise result.errors.full_messages.first
```
david-yz-liu: It would be nice here if we just record that a submission could not be set to complete, but allow the loop to continue. (I think this is what we're doing for releasing/unreleasing assignments.) I could imagine a standard use case would be for an instructor to select all submissions and try to mark them as complete, and then see the remaining "In Progress" submissions are the ones that actually have marks missing.

mishaschwartz: I believe the behaviour for release and unrelease is all or nothing currently. If you select a grouping that cannot be released, the transaction will roll back entirely for every grouping you selected on the table (see `SubmissionHelper.set_release_on_results`)

The current behaviour for setting bulk in/complete mirrors this. So I think we should either change both or leave it as is. 

We discussed the pros and cons for both a little while ago and we decided that it was better to display an error indicating why a bulk release couldn't happen instead of requiring the instructor to scroll through the table after every click to check if there were one or two rows which didn't change.

david-yz-liu: Ah you're right, thanks!

PR #4342, opened 2020-02-01
Delete levels

ashih2018:
Added functionality so that a user is able to delete levels of a rubric criterion.

Possible shortcomings and improvements: 
- Currently, users delete levels through checkboxes. In the future, we could try and implement buttons that the user can use to delete levels.
- The styling/placement of the checkbox is kind of weird right now(maybe?) Any suggestion on where to place it would be greatly appreciated! 
image

### DIFF
--- a/app/views/rubric_criteria/_rubric_criterion_level.html.erb
+++ b/app/views/rubric_criteria/_rubric_criterion_level.html.erb
@@ -7,7 +7,7 @@
   </div>
   <div class="destroy_checkbox">
     <%= level_form.check_box :_destroy%>
-    <%= level_form.label :_destroy, "Destroy Level " + level_form.index.to_s %>
+    <%= level_form.label :_destroy, "Delete Level " + level_form.index.to_s %>
   </div>
   <div>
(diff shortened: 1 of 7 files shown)


app/views/rubric_criteria/_rubric_criterion_level.html.erb
```diff
@@ -1,12 +1,13 @@
...
 <%= criterion_form.fields_for :levels do |level_form| %>
   <div class="rubric_level_header">
-    <%= I18n.t("rubric_criteria.level.name")%>
-    <%= i %>
-    <% i += 1 %>
+    <%= I18n.t("rubric_criteria.level.name") + " " + level_form.index.to_s %>
+  </div>
+  <div class="destroy_checkbox">
+    <%= level_form.check_box :_destroy%>
+    <%= level_form.label :_destroy, "Delete Level " + level_form.index.to_s %>
```
david-yz-liu: @ashih2018 everything looks good, except here you should internationalize this string.

PR #4360, opened 2020-02-06
Added new method for removing folders

ANUSHIYAPRIYA:
Added a new method for removing folders and its contents
Added a route in routes.rb


app/controllers/api/submission_files_controller.rb
```diff
@@ -114,8 +114,7 @@ def create
                                                       filename: params[:filename],
                                                       type: params[:mime_type])
         success, messages = grouping.group.access_repo do |repo|
-          file_path = File.dirname(params[:filename]).gsub(%r{^/}, '')
-          path = Pathname.new(grouping.assignment.repository_folder).join(file_path)
+          path = Pathname.new(grouping.assignment.repository_folder)
```
mishaschwartz: Can you remind me why we don't need the file_path here anymore?

app/controllers/api/submission_files_controller.rb
```diff
@@ -168,5 +170,37 @@ def remove_file
...
+      end
+
+      if has_missing_params?([:folder_path])
+        # incomplete/invalid HTTP params
+        render 'shared/http_status', locals: { code: '422', message:
+            HttpStatusHelper::ERROR_CODE['message']['422'] }, status: 422
+        return
+      end
+      success, messages = grouping.group.access_repo do |repo|
+        folders = params[:folder_path]
```
mishaschwartz: we probably want to use the singular `folder` here since there will only be one of them

app/helpers/repository_helper.rb
```diff
@@ -90,7 +90,7 @@ def add_files(files, user, repo, path: '/', txn: nil, check_size: true, required
   # nil, the boolean indicates whether any errors were encountered which mean the caller should not
   # commit the transaction. The array contains any error or warning messages as arrays of arguments
   # (each of which can be passed directly to flash_message from a controller).
-  def remove_files(files, user, repo, path: '/', txn: nil)
+  def remove_files(files, user, repo, path: '/', txn: nil, option: 'File')
```
mishaschwartz: Usually with this sort of parameter that can either be one thing or another (in this case 'file' or 'directory') it is better to use a boolean value and name the parameter something less generic. "option" doesn't clearly describe what we're using this parameter for. What are some other names we could use?

app/helpers/repository_helper.rb
```diff
@@ -169,19 +169,19 @@ def remove_folders(folders, user, repo, path: '/', txn: nil)
...
-    success, file_messages = remove_files(files, user, repo, path: '', txn: txn)
+    success, file_messages = remove_files(files.reverse, user, repo, path: '', txn: txn, option: 'Folder')
 
     return [success, file_messages] unless success
 
     # folders are removed in git if their contents are removed
-    unless repo.is_a? GitRepository
-      dirs.each do |dir, revision|
-        txn.remove(dir, revision)
+    if repo.is_a? GitRepository
```
mishaschwartz: This will remove directories _only_ if we're using git repositories won't it? What will happen when we use subversion or memory repositories?

app/helpers/repository_helper.rb
```diff
@@ -169,19 +169,19 @@ def remove_folders(folders, user, repo, path: '/', txn: nil)
...
+      dirs.each do |dir|
```
mishaschwartz: Isn't the order that the directories are deleted important? You're reversing the order that the files are deleted in (line 177), does that matter for directories as well?

app/helpers/repository_helper.rb
```diff
@@ -169,19 +169,19 @@ def remove_folders(folders, user, repo, path: '/', txn: nil)
...
         else
-          dirs[path] = current_revision
+          dirs << path
         end
       end
-      dirs[folder_path] = current_revision
+      dirs << folder_path
     end
-    success, file_messages = remove_files(files, user, repo, path: '', txn: txn)
+    success, file_messages = remove_files(files.reverse, user, repo, path: '', txn: txn, option: 'Folder')
```
mishaschwartz: Why is the order that files are deleted in important? Why reverse the order of the list of files here?

app/lib/repository.rb
```diff
@@ -456,8 +456,12 @@ def add(path, file_data=nil, mime_type=nil)
       @jobs.push(action: :add, path: path, file_data: file_data, mime_type: mime_type)
     end
 
-    def remove(path, expected_revision_identifier)
-      @jobs.push(action: :remove, path: path, expected_revision_identifier: expected_revision_identifier)
+    def remove(path, expected_revision_identifier, option)
```
mishaschwartz: You've changed the signature of this method by adding a required argument called `option`. Have you made sure that every piece of code that calls `remove` will provide an argument for `option`? What might be a better way of passing this argument to the `remove` method?

app/lib/git_repository.rb
```diff
@@ -469,15 +475,51 @@ def add_directory(path)
...
     end
+
     File.unlink(File.join(@repos_path, path))
     @repos.index.remove(path)
+    a = Pathname.new(File.join(@repos_path, path))
+    b = Pathname.new(path)
+    c = Pathname.new(b.dirname)
+    d = Pathname.new(File.join(@repos_path, c))
+    return if File.exist?(File.join(a.dirname, DUMMY_FILE_NAME))
+    if b.basename.to_s.start_with? '.'
```
mishaschwartz: Why does it matter if this starts with a dot?

app/lib/git_repository.rb
```diff
@@ -469,15 +475,51 @@ def add_directory(path)
...
       raise Repository::FileDoesNotExist.new(path)
     end
+
     File.unlink(File.join(@repos_path, path))
     @repos.index.remove(path)
+    a = Pathname.new(File.join(@repos_path, path))
+    b = Pathname.new(path)
+    c = Pathname.new(b.dirname)
+    d = Pathname.new(File.join(@repos_path, c))
+    return if File.exist?(File.join(a.dirname, DUMMY_FILE_NAME))
```
mishaschwartz: I think that this and everything below it can be simplified. Remember we have three cases here:

1. we remove a file and the directory it was in is now empty -> add .gitkeep to that directory
2. we remove a file and the directory it was in is not empty -> do nothing
3. we remove a file and the directory it was in is now empty and the `option` argument is `Folder` -> do nothing

app/lib/git_repository.rb
```diff
@@ -469,15 +475,51 @@ def add_directory(path)
...
+  def remove_directory(path, expected_revision_identifier)
+    if @repos.last_commit.oid != expected_revision_identifier
+      raise Repository::FileOutOfSyncConflict.new(path)
+    end
+    unless get_latest_revision.path_exists?(path)
+      raise Repository::FileDoesNotExist.new(path)
+    end
+
+    a = Pathname.new(File.join(@repos_path, path))
+    FileUtils.remove_dir(a, force = true)
```
mishaschwartz: What will happen here if we call `FileUtils.remove_dir(a, force = true)` before we have removed all the files from `a`? (Hint: look at what `remove_file` does at line [487](https://github.com/MarkUsProject/Markus/pull/4360/files#diff-016bd76a1aa7dbe633f06d6be0087442R487))

app/lib/repository.rb
```diff
@@ -456,8 +456,12 @@ def add(path, file_data=nil, mime_type=nil)
...
+      @jobs.push(action: :remove, path: path, expected_revision_identifier: expected_revision_identifier, option: option)
+    end
+
+    def remove_directory(path, expected_revision_identifier)
```
mishaschwartz: If `GitRepository` is the only class that will really use `remove_repository`, could we instead make `GitRepository.commit` a bit smarter so that we don't need `remove_repository` and we could just make `remove` work for both files and folders (like it does for `MemoryRepository` and `SubversionRepository`)?

app/helpers/repository_helper.rb
```diff
@@ -168,8 +168,6 @@ def remove_folders(folders, user, repo, path: '/', txn: nil)
         path = File.join obj.path, obj.name
         if obj.is_a? Repository::RevisionFile
           files << path
-        else
-          dirs << path
```
mishaschwartz: This will work well for the GitRepository. Can you double check that it works for subversion and memory repositories as well.

app/lib/git_repository.rb
```diff
@@ -482,31 +482,18 @@ def remove_file(path, expected_revision_identifier, option)
     unless get_latest_revision.path_exists?(path)
       raise Repository::FileDoesNotExist.new(path)
     end
-
-    File.unlink(File.join(@repos_path, path))
-    @repos.index.remove(path)
     a = Pathname.new(File.join(@repos_path, path))
```
mishaschwartz: use descriptive variable names

app/lib/git_repository.rb
```diff
@@ -482,31 +482,18 @@ def remove_file(path, expected_revision_identifier, option)
...
     b = Pathname.new(path)
-    c = Pathname.new(b.dirname)
-    d = Pathname.new(File.join(@repos_path, c))
-    return if File.exist?(File.join(a.dirname, DUMMY_FILE_NAME))
-    if b.basename.to_s.start_with? '.'
-      unless File.exist?(File.join(d.dirname, DUMMY_FILE_NAME))
-        gitkeep_filename = File.join(c.dirname, DUMMY_FILE_NAME)
+    if option == 'File'
+      File.unlink(File.join(@repos_path, path))
```
mishaschwartz: duplicate of lines at 495

app/lib/git_repository.rb
```diff
@@ -517,9 +504,12 @@ def remove_directory(path, expected_revision_identifier)
     unless get_latest_revision.path_exists?(path)
       raise Repository::FileDoesNotExist.new(path)
     end
-
     a = Pathname.new(File.join(@repos_path, path))
```
mishaschwartz: use descriptive variable names

app/lib/git_repository.rb
mishaschwartz: maybe include explanation in docstring that the folder must be empty

app/controllers/api/submission_files_controller.rb
mishaschwartz: reminder: this is necessary so that dir isn't added twice

app/lib/repository.rb
mishaschwartz: Nevermind

app/helpers/repository_helper.rb
```diff
@@ -109,7 +109,7 @@ def remove_files(files, user, repo, path: '/', txn: nil)
       basename = sanitize_file_name(basename)
       file_path = current_path.join(subdir_path).join(basename)
       file_path = file_path.to_s
-      txn.remove(file_path, current_revision)
+      txn.remove(file_path, current_revision, keep_folder)
```
mishaschwartz: Please change the `remove` call here to also pass the keep_folder value as a keyword argument:

```
txn.remove(file_path, current_revision, keep_folder: keep_folder)
```

app/lib/repository.rb
```diff
@@ -456,8 +456,13 @@ def add(path, file_data=nil, mime_type=nil)
       @jobs.push(action: :add, path: path, file_data: file_data, mime_type: mime_type)
     end
 
-    def remove(path, expected_revision_identifier)
-      @jobs.push(action: :remove, path: path, expected_revision_identifier: expected_revision_identifier)
+    def remove(path, expected_revision_identifier, keep_folder)
```
mishaschwartz: make `keep_folder` a keyword argument here too:

```
def remove(path, expected_revision_identifier, keep_folder: true)
```

app/lib/git_repository.rb
```diff
@@ -395,7 +395,13 @@ def commit(transaction)
         end
       when :remove
         begin
-          remove_file(job[:path], job[:expected_revision_identifier])
+          remove_file(job[:path], job[:expected_revision_identifier], keep_folder: job[:keep_folder])
+        rescue Repository::Conflict => e
+          transaction.add_conflict(e)
+        end
+      when :remove_directory
```
mishaschwartz: Please add this case to the `commit` method for both the memory and subversion repository as well.

app/lib/git_repository.rb
```diff
@@ -469,15 +475,38 @@ def add_directory(path)
...
+    if @repos.last_commit.oid != expected_revision_identifier
+      raise Repository::FileOutOfSyncConflict.new(path)
+    end
+    unless get_latest_revision.path_exists?(path)
+      raise Repository::FileDoesNotExist.new(path)
+    end
+    absolute_path = Pathname.new(File.join(@repos_path, path))
+    relative_path = Pathname.new(path)
+    FileUtils.remove_dir(absolute_path)
+    return if File.exist?(File.join(absolute_path.dirname, DUMMY_FILE_NAME))
```
mishaschwartz: I think it's better to use the `Revision.path_exists?` method here as well. That way we know for sure that the file is there and that it has been committed to the repository in a previous commit.

app/helpers/repository_helper.rb
```diff
@@ -168,21 +168,17 @@ def remove_folders(folders, user, repo, path: '/', txn: nil)
...
+    success, file_messages = remove_files(files, user, repo, path: '', txn: txn, keep_folder: false)
 
     return [success, file_messages] unless success
 
     # folders are removed in git if their contents are removed
-    unless repo.is_a? GitRepository
-      dirs.each do |dir, revision|
-        txn.remove(dir, revision)
-      end
+    dirs.each do |dir|
```
mishaschwartz: Please double check if the order that directories are deleted in matters or not for subversion and memory repositories.

app/lib/git_repository.rb
```diff
@@ -469,15 +475,38 @@ def add_directory(path)
...
+  def remove_directory(path, expected_revision_identifier)
+    if @repos.last_commit.oid != expected_revision_identifier
+      raise Repository::FileOutOfSyncConflict.new(path)
+    end
+    unless get_latest_revision.path_exists?(path)
+      raise Repository::FileDoesNotExist.new(path)
+    end
+    absolute_path = Pathname.new(File.join(@repos_path, path))
+    relative_path = Pathname.new(path)
+    FileUtils.remove_dir(absolute_path)
```
mishaschwartz: This is better but I still think we should raise a conflict error if this directory is not empty and we're trying to delete it.

app/lib/memory_repository.rb
```diff
@@ -126,6 +126,12 @@ def commit(transaction)
         rescue Repository::Conflict => e
           transaction.add_conflict(e)
         end
+      when :remove_directory
```
ANUSHIYAPRIYA: will add :remove, :remove_directory in the same line both in memory and subversion. Forgot to modify

app/helpers/repository_helper.rb
```diff
@@ -161,33 +161,49 @@ def remove_folders(folders, user, repo, path: '/', txn: nil)
...
     end
-    success, file_messages = remove_files(files, user, repo, path: '', txn: txn)
 
-    return [success, file_messages] unless success
+    if files.length > 0
+      success, file_messages = remove_files(files, user, repo, path: '', txn: txn, keep_folder: false)
+      return [success, file_messages] unless success
+    end
 
     # folders are removed in git if their contents are removed
```
mishaschwartz: please remove this comment now that we don't need it anymore

app/lib/memory_repository.rb
mishaschwartz: please do the same for subversion_repository as well

config/locales/en.yml
```diff
@@ -29,6 +29,7 @@ en:
         empty_file_warning: "%{file_name} is empty"
         file_too_large: "The size of the uploaded file %{file_name} exceeds the maximum of %{max_size} MB."
         exist: "The Folder is already exist"
```
mishaschwartz: "The folder already exists"
"The folder does not exist"

mishaschwartz: Also, please add translations for es, fr, and pt.

You can just use the string "UPDATE ME" and we can add them later but they should exist for all of them

mishaschwartz:
I tested this with subversion repos and it appears to work as intended

app/helpers/repository_helper.rb
```diff
@@ -90,7 +90,7 @@ def add_files(files, user, repo, path: '/', txn: nil, check_size: true, required
   # nil, the boolean indicates whether any errors were encountered which mean the caller should not
   # commit the transaction. The array contains any error or warning messages as arrays of arguments
   # (each of which can be passed directly to flash_message from a controller).
-  def remove_files(files, user, repo, path: '/', txn: nil)
+  def remove_files(files, user, repo, path: '/', txn: nil, keep_folder: true)
```
david-yz-liu: You should add some documentation for `keep_folder` here.

app/helpers/repository_helper.rb
```diff
@@ -161,33 +161,48 @@ def remove_folders(folders, user, repo, path: '/', txn: nil)
...
+      txn.remove_directory(dir, current_revision.to_s)
+    end
 
-    # folders are removed in git if their contents are removed
-    unless repo.is_a? GitRepository
-      dirs.each do |dir, revision|
-        txn.remove(dir, revision)
+    if repo.is_a? GitRepository
+      if relative_path.dirname == Pathname.new('.')
+        unless repo.get_latest_revision.path_exists?(File.join(current_path, '.gitkeep'))
```
david-yz-liu: You should be able to access the constant name (`.gitkeep`) rather than hard-coding it again here.

config/routes.rb
```diff
@@ -28,7 +28,9 @@
           resources :submission_files, except: [:new, :edit] do
             collection do
               delete 'remove_file'
+              delete 'remove_folder'
               post 'create_folders'
+
```
david-yz-liu: Remove this line.

app/controllers/api/submission_files_controller.rb
```diff
@@ -183,7 +183,10 @@ def remove_file
       end
 
       begin
-        success, messages = grouping.remove_files([params[:filename]], @current_user)
+        success, messages = grouping.group.access_repo do |repo|
```
david-yz-liu: Can you explain the purpose of this change? I'm not sure how this part is related to removing folders.

app/lib/git_repository.rb
```diff
@@ -470,15 +476,32 @@ def add_directory(path)
   end
 
   # Removes a file from the repository.
-  def remove_file(path, expected_revision_identifier)
+  def remove_file(path, expected_revision_identifier, keep_folder: true)
```
david-yz-liu: Please document the new option here as well.

app/controllers/api/submission_files_controller.rb
ANUSHIYAPRIYA: when I was working on remove_folders it was using remove_files method, so I changed it here too so that it will send all the arguments to remove_files. Should I make a separate PR  for this?

app/helpers/repository_helper.rb
david-yz-liu: I see you changed it in the two lines below, which is good, but you should also change the occurrence of `.gitkeep` on this line.

Also, rather than use `internal_file_names`, please use `DUMMY_FILE_NAME` instead.

app/controllers/api/submission_files_controller.rb
david-yz-liu: Ah okay, yes ideally this would be a separate commit and pull request, include a new test that illustrates why this change is good/necessary.

PR #4366, opened 2020-02-10
Remove upsert gem

mishaschwartz:
- replace upsert gem by using `upsert_all` in all cases
- fix bug where criterion marks were not being scaled due to a change in ActiveModel::Dirty (we wanted to look at changes made after a save not before)


mishaschwartz:
Bug fix has been pulled out into a separate PR (#4369)

PR #4368, opened 2020-02-11
Fix bug where uploading tag csv only sent the filename

mishaschwartz:
Previously, the form in the upload modal did not send all the file data, just the filename.

- [x] update changelog


david-yz-liu:
This is good, but I'm thinking we use the `html:` option in other forms, and if these are all bugs? If so we should fix them all at once, I think.

mishaschwartz:
Yes we do use this in other forms but I think this is only an issue for forms built using the `form_tag` method (not the `form_for` method which takes different arguments).

Basically, `form_for` uses `html: { multipart: true }` and `form_tag` uses `multipart: true`. We were using the wrong version for the `form_tag` which is what this PR addresses.  

The real issue is that the functionality for both `form_tag` and `form_for` has been merged into the `form_with` method (as of rails 5.1: https://guides.rubyonrails.org/form_helpers.html#using-form-for-and-form-tag) so if anything we should be converting everything over to use `form_with`.

PR #4373, opened 2020-02-12
Bug fix and corresponding tests

ANUSHIYAPRIYA:
This is regarding the RT ticket id:7666
1. Whenever the students did not submit any assignment files the Submission dates field value will be set to nil.
2. When the Instructor is collecting submissions with no assignment files, it is getting collected (by setting is_collected=true) but it is showing the following errors
undefined method `<=' for nil:NilClass
undefined method `-' for nil:NilClass
Fix:
1.Now the error message wont appear in the page and also the instructor can collect even if the students didn't submit any files.
2. Added tests for the corresponding changes made.


app/lib/memory_repository.rb
```diff
@@ -177,6 +177,7 @@ def get_revision_by_timestamp(at_or_earlier_than, path = nil, later_than = nil)
     end
 
     (@revision_history + [@current_revision]).reverse_each do |revision|
+      return revision if revision.server_timestamp.nil?
```
mishaschwartz: Can you explain why we need to return early here?

ANUSHIYAPRIYA: Correct me if i am wrong
In the next 2 lines we have to compare **revision.server_timestamp** value with other values. If *revision.server_timestamp* value is nil it will raise error.

spec/models/grace_period_submission_rule_spec.rb
```diff
@@ -5,6 +5,28 @@
     expect(rule.save).to be_truthy
   end
 
+  context 'When the student did not submit any assignment files' do
+    describe '#apply_submission_rule' do
+      before :each do
+        @group = create(:group)
```
mishaschwartz: Instead of setting all of these new objects as instance attributes in the `before :each` method, can we write these as `let` statements?

spec/models/grace_period_submission_rule_spec.rb
```diff
@@ -5,6 +5,28 @@
...
+                             grouping: @grouping,
+                             membership_status: StudentMembership::STATUSES[:inviter])
+        @assignment = @grouping.assignment
+        @rule = GracePeriodSubmissionRule.new
+        @assignment.replace_submission_rule(@rule)
+        GracePeriodDeduction.destroy_all
+        @submission = create(:submission, revision_timestamp: nil)
+        @rule.save
+      end
+      it 'should not apply any submission rule' do
```
mishaschwartz: I think this is a good start but there are more important things you should be testing here as well. Think about what applying each submission rule actually does (to the students, the results, the marks, etc.) and write some tests to check those as well.

spec/models/grace_period_submission_rule_spec.rb
ANUSHIYAPRIYA: Yes we can but i thought If we are going add more test under 'when the student did not submit any assignment files' it would be useful to set everything under before :each method.

app/models/grace_period_submission_rule.rb
```diff
@@ -25,6 +25,7 @@ def assignment_valid?
   end
 
   def apply_submission_rule(submission)
+    return submission if submission.revision_timestamp.nil?
```
mishaschwartz: @david-yz-liu made a good point that we should be checking for `submission.is_empty` here instead. 

That will prevent us from applying a submission rule to students who simply create a grouping during the penalty period but still don't submit anything on time.

mishaschwartz:
This is looking good! 
I think we need an additional test for each submission rule type (grace period, penalty decay, etc.) that tests the case where the submission is empty and the submission was created during the penalty period.

spec/models/grace_period_submission_rule_spec.rb
```diff
@@ -5,6 +5,41 @@
     expect(rule.save).to be_truthy
   end
 
+  context 'When the student did not submit any assignment files' do
+    let(:student) { create(:student, grace_credits: 7) }
+    let(:grouping) { create(:grouping_with_inviter) }
+    let(:submission) { create(:submission, is_empty: true) }
```
david-yz-liu: My comment about these `create` calls is that they're actually creating a bunch of unrelated objects. For example, the submission that's created on this line doesn't belong to the grouping created on line 10, nor does it belong to the assignment created on line 13. The Submission factory (`factories/submissions.rb`) automatically creates a separate grouping (with a brand-new assignment as well). This doesn't affect what happens in the test, because `rule.apply_submission_rule(submission)` works even when the submission is for a different assignment than `rule`, but it's still kind of misleading.

In fact, the existing test cases also test `apply_submission_rule`. So rather than create a new block here, my recommendation would be to take the logic from the tests you've written and add them as new cases inside the block on line 43; this way you can take advantage of the objects already created there. You can create a new nested context that tests for the case when a grouping has no files submitted.

spec/models/penalty_decay_period_submission_rule_spec.rb
```diff
@@ -65,7 +33,56 @@
...
+          @assignment.due_date = Time.now + 2.days
+          # Add two 24 hour penalty decay periods
+          # Overtime begins in two days.
+          add_period_helper(@assignment.submission_rule, 24, 10, 12)
+          add_period_helper(@assignment.submission_rule, 24, 10, 12)
+          # Collection date is in 4 days.
+          @assignment.save
+          @grouping.create_grouping_repository_folder
+        end
+        it 'should not apply submission rule' do
```
david-yz-liu: @ANUSHIYAPRIYA I really like your use of shared examples for the grace tokens. It seems that you can also use a new shared example for the other two submission rules (for penalty periods) since both rely on checking that there are no extra marks created.

spec/models/grace_period_submission_rule_spec.rb
```diff
@@ -40,6 +40,60 @@
       destroy_repos
     end
 
+    context 'when the student did not submit any files' do
+      # Before the due_date and before collection_time
+      describe 'when assignment is collected before the due date' do
+        before :each do
+          @submission = create(:submission, grouping: @grouping, is_empty: true)
```
david-yz-liu: After looking at this a bit more, I'm worried that just by creating the submission here, `apply_submission_rule` isn't actually being applied. Look at how it's done in the existing tests on lines 119-120; that's probably what you should do here.

ANUSHIYAPRIYA: The rule is being applied for the corresponding submission at no_submission_file.rb.

david-yz-liu: Okay that's fine (and it's good you added it to the other shared example).

david-yz-liu:
@ANUSHIYAPRIYA last thing: can you also update the changelog?

PR #4376, opened 2020-02-13
Update hook commit message 

mishaschwartz:
- remind students to push their commits


lib/repo/git_hooks/client/multihook_client.py
```diff
@@ -27,4 +27,4 @@
             if hook.returncode != 0:
                 exit(hook.returncode)
 
-        print('[MarkUs] Commit looks good!')
+        print("[MarkUs] Commit looks good! Don't forget to push")
```
david-yz-liu: Let's be a bit more explicit here. Something like "Don't forget to push your work to the MarkUs server." (Remember to include the period as well.)

PR #4390, opened 2020-02-24
Results levels

ashih2018:
Updated show method in results_controller so that it uses information passed in by each rubric criterion's associated level's model rather than hard coded level_x_name and level_x_description. To do this I created levels_information as an array and merged it into the preexisting marks_info hash.
It is being passed into the view now, where we can access level information in files such as result.jsx and marks_panel.jsx.

### DIFF
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -151,13 +151,6 @@ class ResultsController < ApplicationController
         common_fields = [:id, :name, :position, :max_mark]
         marks_map = [CheckboxCriterion, FlexibleCriterion, RubricCriterion].flat_map do |klass|
           if klass == RubricCriterion
-            # fields = common_fields + [
-            #   :level_0_name, :level_0_description,
-            #   :level_1_name, :level_1_description,
-            #   :level_2_name, :level_2_description,
-            #   :level_3_name, :level_3_description,
-            #   :level_4_name, :level_4_description
-            # ]
             fields = common_fields
           end
           if klass != RubricCriterion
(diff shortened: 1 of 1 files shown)


app/controllers/results_controller.rb
```diff
@@ -166,11 +160,22 @@ def show
                                  peer_visible: is_review)
           criteria_info = criteria.pluck_to_hash(*fields)
           marks_info = criteria.joins(:marks)
-                               .where('marks.result_id': result.id)
-                               .pluck_to_hash(*fields, 'marks.mark')
-                               .group_by { |h| h[:id] }
-          criteria_info.map do |h|
-            info = marks_info[h[:id]]&.first || h.merge('marks.mark': nil)
+                              .where('marks.result_id': result.id)
```
david-yz-liu: You should indent this line and the next three by 1 space (the diff is saying you changed these lines, but you didn't really, just the indentation is off).

app/controllers/results_controller.rb
```diff
@@ -166,11 +160,22 @@ def show
...
+                              .pluck_to_hash(*fields, 'marks.mark')
+                              .group_by { |h| h[:id] }
+          # adds an extra levels field to the marks info hash with the same rubric criterion id
+          levels_info = []
+          if klass == RubricCriterion
```
david-yz-liu: Move this check to inside the `criteria_info.map` loop (this avoids having to repeat the loop structure)

app/controllers/results_controller.rb
```diff
@@ -166,11 +160,22 @@ def show
...
+            criteria_info.map do |cr|
+              levels_info = Level.where(rubric_criterion_id: cr[:id])
+                                  .pluck_to_hash(:name, :description, :mark)
```
david-yz-liu: this should be out-dented by 1 space

app/controllers/results_controller.rb
```diff
@@ -166,11 +160,22 @@ def show
...
+              rubric_marks = marks_info[cr[:id]]&.first
+              rubric_marks.merge!(levels: levels_info)
```
david-yz-liu: Note that `&.` (on the previous line) can return `nil`. So you can just do this call in one line, e.g. `...&.first&.merge!(levels: ...)` to avoid having a temp variable `rubric_marks`.

app/controllers/results_controller.rb
```diff
@@ -166,11 +160,18 @@ def show
...
-            info = marks_info[h[:id]]&.first || h.merge('marks.mark': nil)
+                                 .where('marks.result_id': result.id)
+                                 .pluck_to_hash(*fields, 'marks.mark')
+                                 .group_by { |h| h[:id] }
+          # adds a criterion type to each of the marks info hashes
+          criteria_info.map do |cr|
+            # adds an extra levels field to the marks info hash with the same rubric criterion id
+            if klass == RubricCriterion
+              levels_info = Level.where(rubric_criterion_id: cr[:id])
+                                .pluck_to_hash(:name, :description, :mark)
```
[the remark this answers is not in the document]
ashih2018: I'm confused about the indenting hound comments. Should I be adding a space to this?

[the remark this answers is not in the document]
david-yz-liu: Yes, add exactly 1 space. (The periods should be aligned)

app/controllers/results_controller.rb
david-yz-liu: @ashih2018 it looks like the indentation is still off here. You need to align the periods with the line above.

PR #4391, opened 2020-02-24
Marks react

ashih2018:
Updated the marks_panel.jsx file so that it uses the levels information that is being passed in from its props, rather than using the hardcoded data it used previously.
Possible shortcomings/other comments:
- This branch is branched off of results levels so it should be merged in after results levels gets merged in

### DIFF
--- a/app/assets/javascripts/Components/Result/marks_panel.jsx
+++ b/app/assets/javascripts/Components/Result/marks_panel.jsx
@@ -322,12 +322,12 @@ class RubricCriterionInput extends React.Component {
   // The parameter `level` is the level number selected.
   handleChange = (level) => {
     this.props.updateMark(
-      this.props.criterion_type, this.props.id, level * this.props.max_mark / 4
+      this.props.criterion_type, this.props.id, this.props.levels[level].mark
     );
   };
 
   renderRubricLevel = (i) => {
-    const levelMark = (i * this.props.max_mark / 4).toFixed(2);
+    const levelMark = this.props.levels[i].mark.toFixed(2);
     let selectedClass = '';
     let oldMarkClass = '';
     if (this.props.mark !== undefined &&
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/Components/Result/marks_panel.jsx
```diff
@@ -361,7 +361,11 @@ class RubricCriterionInput extends React.Component {
   };
 
   render() {
-    const levels = [0, 1, 2, 3, 4].map(this.renderRubricLevel);
+    const level_marks = [];
+    for (let i = 0; i < this.props.levels.length; i++) {
+      level_marks.push(this.props.levels[i].mark);
+    }
+    const levels = level_marks.map(this.renderRubricLevel);
```
david-yz-liu: Try changing `this.renderRubricLevel` to accept a "level" object rather than just a single number. This way you should be able to just do `this.props.levels.map(this.renderRubricLevel)`.

app/assets/javascripts/Components/Result/marks_panel.jsx
```diff
@@ -343,13 +344,13 @@ class RubricCriterionInput extends React.Component {
 
     return (
       <tr
-        data-level-index={i} onClick={() => this.handleChange(i)}
-        key={`${this.props.id}-${i}`}
+        data-level-index={levelMark} onClick={() => this.handleChange(levelMark)}
```
david-yz-liu: You need to be careful about how you're calling `handleChange`. It expects an *index* and you're giving it a *mark*.

Similar issue with `data-level-index`. Try looking for where `data-level-index` is used to determine how to fix this mismatch. Note that JavaScript provides a way of accessing `data-*` attributes (https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes), so just doing a string search for 'data-level-index' won't necessarily find where it's being used.

app/assets/javascripts/Components/Result/marks_panel.jsx
```diff
@@ -343,16 +345,16 @@ class RubricCriterionInput extends React.Component {
 
     return (
       <tr
-        data-level-index={i} onClick={() => this.handleChange(i)}
-        key={`${this.props.id}-${i}`}
+        onClick={() => this.handleChange(levelIndex)}
```
david-yz-liu: @ashih2018 let's change `this.handleChange` so that it *also* accepts a level object. This will allow you to simplify the code and remove the calculation of `levelIndex` as well.

PR #4396, opened 2020-02-25
Levels Migration

ashih2018:
Generated migration to remove level_x_name and description columns from rubric criteria. 

Possible shortcomings and other notes:

- This should be merged in after results-levels gets merged in, because that branch removes all uses of level_x_name and description in results controller.

### DIFF
--- a/spec/factories/levels.rb
+++ b/spec/factories/levels.rb
@@ -1,8 +1,7 @@
 FactoryBot.define do
   factory :level do
-    association :rubric_criteria
     name { Faker::Lorem.word }
     description { Faker::Lorem.word }
-    mark
+    rubric_criteria
   end
 end
--- a/spec/factories/levels.rb
+++ b/spec/factories/levels.rb
@@ -1,0 +1,8 @@
+FactoryBot.define do
+  factory :level do
+    association :rubric_criteria
+    name { Faker::Lorem.word }
+    description { Faker::Lorem.word }
+    mark
+  end
+end
--- a/spec/factories/levels.rb
+++ b/spec/factories/levels.rb
@@ -2,6 +2,6 @@ FactoryBot.define do
   factory :level do
     name { Faker::Lorem.word }
     description { Faker::Lorem.word }
-    rubric_criteria
+    rubric_criterion
   end
 end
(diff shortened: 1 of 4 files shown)


david-yz-liu:
This is a good start, but you need to update the spec factories as well.

spec/factories/levels.rb
```diff
@@ -0,0 +1,7 @@
+FactoryBot.define do
+  factory :level do
+    name { Faker::Lorem.word }
+    description { Faker::Lorem.word }
+    rubric_criteria
```
david-yz-liu: Be careful with spelling here. You want the singular form, `rubric_criterion`, for the association.

spec/factories/rubric_criteria.rb
```diff
@@ -5,16 +5,12 @@
...
-    level_2_name { 'Good' }
-    level_2_description { 'This criterion was satisfied well.' }
-    level_3_name { 'Great' }
-    level_3_description { 'This criterion was satisfied very well.' }
-    level_4_name { 'Excellent' }
-    level_4_description { 'This criterion was satisfied exceptionally well.' }
     sequence(:position)
   end
+
+  factory :rubric_criteria_with_levels do
```
david-yz-liu: This isn't bad, but I don't think this really deserves its own factory, since the original rubric criterion factory used to have levels. Why don't you put the `after(:create)` block into the above factory.

spec/factories/rubric_criteria.rb
```diff
@@ -5,16 +5,12 @@
...
+    after(:create) do |criteria|
+      5.times.each { |i| create(:level, rubric_criteria: rubric_criteria, mark: i) }
```
david-yz-liu: Two issues here. First, rename the **key** from `rubric_criteria:` to `rubric_criterion:`. Second, right now you're passing in `rubric_criteria`, but what you really want is the current criterion value, which you've named `criteria` on Line 12 (i.e., the current object is passed to the block in `after(:create)`. But also you should rename the variable to `criterion` rather than `criteria`.

david-yz-liu:
@ashih2018 I merged in your other PR, can you update this one by doing a pull from `upstream sti-criteria`?

PR #4418, opened 2020-03-03
Scale marks

ashih2018:
Created a callback method that scales levels when the max mark is updated. Each level's mark is scaled appropriately, multiplied by the ratio (previous max level mark / new max mark).
Example with marks 0, 1, 2, 3, 4 with original max mark of 4 scaled to a max mark of 12. Thus each level's mark will be tripled.

**EDIT: This PR should be merged in only after levels-migration gets merged in.**

image
image

### DIFF
--- a/app/models/rubric_criterion.rb
+++ b/app/models/rubric_criterion.rb
@@ -41,10 +41,8 @@ class RubricCriterion < Criterion
   def scale_marks_if_max_mark_changed
     return unless self.changed.include?('max_mark')
     return if self.changes['max_mark'][0].nil?
-    
     old_max = self.changes['max_mark'][0]
     new_max = self.changes['max_mark'][1]
-
     scale = new_max / old_max
     self.levels.each do |level|
       level.update(mark: (level.mark * scale).round(2))
--- a/app/models/rubric_criterion.rb
+++ b/app/models/rubric_criterion.rb
@@ -14,6 +14,7 @@ class RubricCriterion < Criterion
 
   has_many :levels, -> { order(:mark) }, inverse_of: :rubric_criterion, dependent: :destroy
   accepts_nested_attributes_for :levels, allow_destroy: true
+  after_save :scale_marks_if_max_mark_changed
 
   belongs_to :assignment, counter_cache: true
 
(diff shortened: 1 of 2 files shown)


app/models/rubric_criterion.rb
```diff
@@ -37,6 +38,17 @@ def update_assigned_groups_count
     self.assigned_groups_count = result.uniq.length
   end
 
+  def scale_marks_if_max_mark_changed
+    max_level_mark = self.levels.max_by(&:mark).mark
+    new_max = self.max_mark
+    return if max_level_mark == new_max
```
david-yz-liu: Rather than doing this, just check if `self`'s `max_mark` attribute was changed. ActiveModel can actually do this (even after saving), check out https://api.rubyonrails.org/classes/ActiveModel/Dirty.html.

app/models/rubric_criterion.rb
```diff
@@ -37,6 +38,17 @@ def update_assigned_groups_count
...
+
+    scale = new_max / max_level_mark
+    self.levels.each do |level|
+      level.mark = (level.mark * scale).round(1)
```
david-yz-liu: You'll want to do an `level.update` here to make sure the changes are saved. Also round to 2 decimal places.

david-yz-liu:
BTW it would be great if you wrote some tests to check this behaviour too. :)

spec/models/rubric_criterion_spec.rb
```diff
@@ -241,4 +241,29 @@
       end
     end
   end
+
+  context 'A rubric criteria with levels' do
+    before(:each) do
+      @criterion = create(:rubric_criterion)
+      @levels = @criterion.levels
```
ashih2018: would it make sense to make this as an instance variable instead of continuously defining levels = @criterion.levels within each test block?

david-yz-liu: that's fine with me

david-yz-liu:
@ashih2018 it would be great to add a test for the case that the criterion max mark changes and one of the level marks changes (a new special case).

PR #4434, opened 2020-03-11
Updated the remove_file api controller method

ANUSHIYAPRIYA:
The remove_file method was using an old method from grouping. we deleted it but remove_file method was still looking for the same method from grouping.

Modified the remove_file method so that now it will use a helper method to remove the file.

### DIFF
--- a/app/helpers/repository_helper.rb
+++ b/app/helpers/repository_helper.rb
@@ -108,9 +108,6 @@ module RepositoryHelper
     current_revision = repo.get_latest_revision.revision_identifier
 
     files.each do |file_path|
-      relative_path = Pathname.new(file_path)
-      abs_path = File.join(current_path, relative_path)
-      return [false, [:file_not_exist]] unless repo.get_latest_revision.path_exists?(abs_path)
       subdir_path, basename = File.split(file_path)
       basename = sanitize_file_name(basename)
       file_path = current_path.join(subdir_path).join(basename)
--- a/app/helpers/repository_helper.rb
+++ b/app/helpers/repository_helper.rb
@@ -105,11 +105,11 @@ module RepositoryHelper
     end
 
     current_path = Pathname.new path
-    relative_path = Pathname.new(files[0])
-    abs_path = File.join(current_path, relative_path)
-
-    return [true, [:file_not_exist]] unless repo.get_latest_revision.path_exists?(abs_path)
-
+    files.each do |file_name|
+      relative_path = Pathname.new(file_name)
+      abs_path = File.join(current_path, relative_path)
+      return [true, [:file_not_exist]] unless repo.get_latest_revision.path_exists?(abs_path)
+    end
     current_revision = repo.get_latest_revision.revision_identifier
 
     files.each do |file_path|
--- a/app/helpers/repository_helper.rb
+++ b/app/helpers/repository_helper.rb
@@ -105,6 +105,10 @@ module RepositoryHelper
     end
 
     current_path = Pathname.new path
+    relative_path = Pathname.new(files[0])
+    abs_path = File.join(current_path, relative_path)
+
+    return [true, [:file_not_exist]] unless repo.get_latest_revision.path_exists?(abs_path)
 
     current_revision = repo.get_latest_revision.revision_identifier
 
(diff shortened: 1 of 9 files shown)


app/helpers/repository_helper.rb
```diff
@@ -105,6 +105,10 @@ def remove_files(files, user, repo, path: '/', txn: nil, keep_folder: true)
     end
 
     current_path = Pathname.new path
+    relative_path = Pathname.new(files[0])
+    abs_path = File.join(current_path, relative_path)
+
+    return [true, [:file_not_exist]] unless repo.get_latest_revision.path_exists?(abs_path)
```
mishaschwartz: This is only checking if the first file in `files` doesn't exist. What happens if there are multiple files in `files` and the second, third, etc. file doesn't exist?

ANUSHIYAPRIYA: Good point.Here in remove_folder (submission_controller)we can delete the files one by one not as a list. So if any other method which access remove_folder in repository_helper sends multiple files that time it wont work. I will make sure this works for all the methods which is accessing remove_folder method. Thanks for pointing this.

app/helpers/repository_helper.rb
```diff
@@ -105,7 +105,11 @@ def remove_files(files, user, repo, path: '/', txn: nil, keep_folder: true)
     end
 
     current_path = Pathname.new path
-
+    files.each do |file_name|
+      relative_path = Pathname.new(file_name)
+      abs_path = File.join(current_path, relative_path)
+      return [true, [:file_not_exist]] unless repo.get_latest_revision.path_exists?(abs_path)
```
david-yz-liu: Overall, you should merge this logic into the loop below, which already loops through each file and constructs a `file_path` (which you should be able to pass into `repo.get_latest_revision.path_exists?`).

Also, it appears from the documentation of this method that the first return value (the boolean) means different things depending on whether `txn` is passed in or not. It's counter-intuitive, but it seems to indicate success/failure in opposite ways depending on whether `txn` is passed in. You should maintain the same behaviour for the new `return` statement as well.

Then, you should change the `success` variable name in the `submission_files_controller.rb` (I don't think it actually means "success")

app/helpers/repository_helper.rb
ANUSHIYAPRIYA: *remove_file

app/helpers/repository_helper.rb
```diff
@@ -105,10 +105,12 @@ def remove_files(files, user, repo, path: '/', txn: nil, keep_folder: true)
     end
 
     current_path = Pathname.new path
-
     current_revision = repo.get_latest_revision.revision_identifier
 
     files.each do |file_path|
+      relative_path = Pathname.new(file_path)
```
david-yz-liu: Okay, this is a good start but it would be better if you are able to use the variables that are *already* defined in this loop (see `subdir_path`, `basename`, `file_path` below).

app/helpers/repository_helper.rb
```diff
@@ -105,10 +105,12 @@ def remove_files(files, user, repo, path: '/', txn: nil, keep_folder: true)
...
+      abs_path = File.join(current_path, relative_path)
+      return [false, [:file_not_exist]] unless repo.get_latest_revision.path_exists?(abs_path)
```
david-yz-liu: The `false` is now correct when `txn` is `nil`, but not when `txn` is not nil. Look at how the return value is being computed at the bottom of this method, and try to use the same approach for this return value as well.

ANUSHIYAPRIYA: Removed this check because we already checking this in repository file(git, memory and svn).
I thought it will be useful in the future If more files are accessing this method. In add_folder method we wanted to differentiate if the user trying to add folder which is already exists by adding this check and displaying 304 response code. Now we decided to display this as an error. So now 201 for success and 500 for failure with the custom failure message.

app/helpers/repository_helper.rb
```diff
@@ -245,6 +244,8 @@ def flash_repository_messages(messages, suppress: nil)
         flash_message(:warning, I18n.t('student.submission.exist'))
       when :not_exist
         flash_message(:warning, I18n.t('student.submission.not_exist'))
+      when :file_not_exist
```
david-yz-liu: It seems that this symbol is no longer used (nor the string `'student.submission.file_not_exist'`).

PR #4438, opened 2020-03-14
Add additional assignments controller tests

david-yz-liu:
Also did some code cleanup and refactoring for the relevant routes.


david-yz-liu:
@mishaschwartz the `AssignmentsController#show` tests still failed, so I marked them as pending.for now. It's not urgent, though.

PR #4439, opened 2020-03-17
Levels validation

ashih2018:
Added validation to make levels unique by their mark, such that no two levels in a single rubric criterion can have the same level mark.


david-yz-liu:
@ashih2018 this is a good start, but check out the failing tests. The first three look like they're caused by your change, so you should fix those tests here.

PR #4447, opened 2020-03-19
Annotation specs

mishaschwartz:
- added/updated specs for image/pdf/text annotations


david-yz-liu:
@mishaschwartz I remember what I wanted to mention here! The `creator_type` attribute shouldn't be necessary because all of the user subclasses are in the same table. We should update the `belongs_to :creator` association to just refer to the `User` class, I think.

PR #4456, opened 2020-03-20
Criterion ta association: add tests and made validation method private

mishaschwartz:
- add tests for criterion ta associations
- made validation method private

### DIFF
--- a/app/models/criterion_ta_association.rb
+++ b/app/models/criterion_ta_association.rb
@@ -11,10 +11,6 @@ class CriterionTaAssociation < ApplicationRecord
 
   before_validation       :add_assignment_reference, on: :create
 
-  def add_assignment_reference
-    self.assignment = criterion.assignment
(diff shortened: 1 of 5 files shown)


david-yz-liu:
@mishaschwartz I merged this in, but I'm wondering for this one is we can remove the `belongs_to: assignment` altogether and replace it with a `has_one :assignment, through: :criterion`? Not sure if the `has_one` is compatible with the polymorphic association, but if so I feel like this would be an improvement to our db schema.

mishaschwartz:
Yeah I was experimenting with that in both directions and I agree that it would be better but it belongs in a different PR since it would require a db change (since we'd be removing the assignment_id column). 
Also, it gets tricky because the association goes the other way too (assignment has many criterion_ta_associations) which I'm not sure can be done polymorphically. Maybe it would have to be done with a scope?

PR #4463, opened 2020-03-25
Background submission file download

### LINKED ISSUE #4283: Cannot download all submissions for an assignment
If I select all the submissions and click "Download Submissions", I get an HTTP 500 from MarkUs. However, if I select a smaller group of submissions, it works fine. (I tried 1 submission and 5 submissions.)

Also, each submission contains 10 relatively small cpp files (none should be longer than 100 lines of code), and there are 83 submissions in total.

mishaschwartz:
Downloading bulk submissions now happens in the background.
The behaviour is now:

- click the download submissions button from the submission page
- wait for the zip file to be prepared in the background (flash messages keep the user updated on the status)
- a flash message is displayed indicating that the zip file is ready
- this flash message contains a link which triggers the file download when clicked

Resolves #4283

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -488,7 +488,7 @@ class SubmissionsController < ApplicationController
   def zip_groupings_files
     assignment = Assignment.find(params[:assignment_id])
 
-    groupings = Grouping.includes(:group, :current_submission_used).where(id: params[:groupings]&.map(&:to_i))
+    groupings = assignment.groupings.where(id: params[:groupings]&.map(&:to_i))
 
     zip_path = zipped_grouping_file_name(assignment)
 
@@ -507,7 +507,13 @@ class SubmissionsController < ApplicationController
   def download_zipped_file
     assignment = Assignment.find(params[:assignment_id])
     zip_path = zipped_grouping_file_name(assignment)
-    send_file zip_path, disposition: 'inline', filename: File.basename(zip_path)
+    zip_file = File.basename(zip_path)
+    begin
+      send_file zip_path, disposition: 'inline', filename: zip_file
+    rescue ActionController::MissingFile
+      flash_message(:error, I18n.t('submissions.download_zipped_file.file_missing', zip_file: zip_file))
+      redirect_back(fallback_location: root_path)
+    end
   end
 
   ##
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -14,7 +14,8 @@ class SubmissionsController < ApplicationController
                          :get_feedback_file,
                          :download,
                          :downloads,
-                         :download_groupings_files,
+                         :zip_groupings_files,
+                         :download_zipped_file,
                          :manually_collect_and_begin_grading,
                          :repo_browser,
                          :set_result_marking_state,
@@ -27,7 +28,8 @@ class SubmissionsController < ApplicationController
                        :manually_collect_and_begin_grading,
                        :revisions,
                        :repo_browser,
-                       :download_groupings_files,
+                       :zip_groupings_files,
+                       :download_zipped_file,
                        :update_submissions]
   before_action :authorize_for_student,
                 only: [:file_manager]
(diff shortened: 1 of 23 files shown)


app/controllers/submissions_controller.rb
```diff
@@ -481,41 +483,31 @@ def download
...
-    ## create the zip name with the user name to have less chance to delete
-    ## a currently downloading file
-    short_id = assignment.short_identifier
-    zip_name = Pathname.new(short_id + '_' + current_user.user_name + '.zip')
-    zip_path = Pathname.new('tmp') + zip_name
-
-    ## delete the old file if it exists
-    File.delete(zip_path) if File.exist?(zip_path)
-
     groupings = Grouping.includes(:group, :current_submission_used).where(id: params[:groupings]&.map(&:to_i))
```
david-yz-liu: This is kind of awkward now that we're just passing in grouping ids to the job.

1. You can definitely remove the `.includes` here and move it into the job itself.
2. You can `pluck(:id)`.
3. Since it seems this query would just be acting as a filter (e.g. removing invalid ids), I think it would be nice to do `assignment.groupings.where(...)` rather than `Grouping.where(...)`. In general I'm trying to make a change to do more association querying for these nested routes.

(Also, this comment affects the update on line 496 too.)

app/controllers/submissions_controller.rb
```diff
@@ -481,41 +483,31 @@ def download
...
+  end
 
-    ## Send the Zip file
-    send_file zip_path, disposition: 'inline', filename: zip_name.to_s
+  # download a zip file previously prepared by calling the
+  # zip_groupings_files method
+  def download_zipped_file
+    assignment = Assignment.find(params[:assignment_id])
+    zip_path = zipped_grouping_file_name(assignment)
+    send_file zip_path, disposition: 'inline', filename: File.basename(zip_path)
```
david-yz-liu: Would be good to check if `zip_path` exists (not sure what `send_file` does if the path doesn't exist).

config/locales/en.yml
```diff
@@ -56,6 +56,9 @@ en:
       create_groups_job_extra_info: "failed for group: %{group_name}, repo: %{repo_name}, members: %{members}"
       uncollect_submissions_job: "Uncollecting assignments ... "
       submissions_job: "%{progress}/%{total} Submissions collected"
+      download_submissions_job: "%{progress}/%{total} Submissions prepared for download"
+      download_submissions_finalizing: "Finalizing submissions for download"
+      download_submissions_complete: "Submissions are ready. Click here to complete the download."
```
david-yz-liu: Let's change "Submissions are ready." to "Submissions are ready for download."

mishaschwartz:
> [...] it would be nice if the `:notice` flash div was removed after the job is completed (so that only the `:success` download link shows). 

I remember looking into this before, the main issue is that we currently don't have a way of removing just the one `:notice` flash div that shows the progress without removing any other `:notice` flash messages that might be on the page for other reasons.

That's not to say that it can't be done, just that it might require a reconfigure of how we are currently keeping track of flash messages.

> [...] but I think it's worth it.

I agree and I'll start working on it. I'm going to do it so that it applies to all background jobs though, not just this one. So I think I'll do it in a separate PR.

app/controllers/submissions_controller.rb
mishaschwartz: for 2. I'm going to stick with `.ids` instead of `pluck(:id)` since it does the exact same thing and I like the more compact syntax if that's ok with you:

```ruby
irb(main):005:0> a.groupings.pluck(:id)
   (1.1ms)  SELECT "groupings"."id" FROM "groupings" WHERE "groupings"."assessment_id" = $1  [["assessment_id", 5]]
=> [463, 464, 465, 466, 467, 468, 469, 470, 471, 472, 473, 474, 475, 476, 477]
irb(main):006:0> a.groupings.ids
   (1.0ms)  SELECT "groupings"."id" FROM "groupings" WHERE "groupings"."assessment_id" = $1  [["assessment_id", 5]]
=> [463, 464, 465, 466, 467, 468, 469, 470, 471, 472, 473, 474, 475, 476, 477]
```

PR #4469, opened 2020-03-27
Tests for background jobs

mishaschwartz:
- added spec file for all background jobs
- added tests for enqueuing the jobs themselves


mishaschwartz:
Job status is now tested indirectly by testing the JobMessagesController (see #4519)

spec/support/job_shared_examples.rb
```diff
@@ -0,0 +1,80 @@
+shared_examples 'background job' do
+  include ActiveJob::TestHelper
+  ActiveJob::Base.queue_adapter = :test
```
david-yz-liu, line 3: This is good, but I'd prefer it if all the test configuration was in `spec_helper.rb`.

PR #4470, opened 2020-03-28
Grouping invite email

ignaspa:

### DIFF
--- a/app/controllers/groups_controller.rb
+++ b/app/controllers/groups_controller.rb
@@ -411,12 +411,7 @@ class GroupsController < ApplicationController
       to_invite = params[:invite_member].split(',')
       errors = @grouping.invite(to_invite)
       if errors.blank?
-        if !to_invite.instance_of?(Array)
-          invitees = [to_invite]
-        else
-          invitees = to_invite
-        end
-        invitees.each do |i|
+        to_invite.each do |i|
           i = i.strip
           invited_user = Student.where(hidden: false).find_by(user_name: i)
           NotificationMailer.with(inviter: current_user,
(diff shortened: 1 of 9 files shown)


config/locales/mailers/en.yml
```diff
@@ -15,3 +15,11 @@ en:
       button_text: 'View Results'
       explicit_link: "If the above button doesn't work, copy-and-paste this link into your web browser: %{link}"
       disclaimer: 'This is an automated email. Please do not reply.'
+
+    grouping_invite_email:
+      subject: 'MarkUs Notification (%{course}) You have been invited to a group!'
+      greeting: 'Hello %{first_name} %{last_name},'
+      text_body: '%{first_name} %{last_name} has invited you to join a group for %{assignment}.'
```
david-yz-liu: Please add a second sentence here, "If you wish to accept this invitation, you need to do so *before* the assignment due date."

app/controllers/groups_controller.rb
```diff
@@ -411,6 +411,18 @@ def invite_member
       to_invite = params[:invite_member].split(',')
       errors = @grouping.invite(to_invite)
       if errors.blank?
+        if !to_invite.instance_of?(Array)
```
david-yz-liu, line 414: I'm actually not sure if `Grouping#invite` is called from other places (you could check this) but in this case `.split(',')` always returns a list of strings, so you don't need this check here.

spec/controllers/groups_controller_spec.rb
```diff
@@ -326,5 +328,19 @@
         is_expected.to respond_with(:redirect)
       end
     end
+
+    describe 'POST #invite_member' do
+      before :each do
+        @grouping = create(:grouping_with_inviter, assignment: @assignment, inviter: @current_student)
```
david-yz-liu, line 334: You aren't actually using the `@grouping` variable, so you can just call `create(...)` without assigning it to a variable. You can remove the `@grouping.reload` line below this one too.

PR #4472, opened 2020-03-29
Update marks on level update

ashih2018:
When a rubric criterion level is updated, all marks given out corresponding with that level should be updated as well. For example, if a level with the mark 1 is updated to 2, all marks that were given out with the mark 1 should also be updated to 2. Furthermore, if a level with a mark 5 is destroyed, all marks with that mark 5 that are associated with that rubric criterion should have their mark set to nil.

Currently there is an error in the rspec tests when I am creating marks. When I run @marks.reload, @marks gets set to [].


app/models/level.rb
```diff
@@ -8,4 +8,28 @@ class Level < ApplicationRecord
 
   validates_uniqueness_of :mark, scope: :rubric_criterion_id
   validates_numericality_of :mark, greater_than_or_equal_to: 0
+
+  before_destroy :destroy_associated_marks
+  before_update :update_associated_marks
+
+  def destroy_associated_marks
+    marks = self.rubric_criterion.marks
```
david-yz-liu, line 16: @ashih2018 change this to the "one line" version you were referring to in our meeting yesterday. I want to see exactly what happens.

spec/models/rubric_criterion_spec.rb
```diff
@@ -249,6 +249,46 @@
       @levels = @criterion.levels
     end
 
+    context 'editing levels edits marks' do
+      before(:each) do
+        @marks = @criterion.marks
+        @result = create(:result, marking_state: Result::MARKING_STATES[:incomplete])
+        @marks.create(mark: 0, result: @result)
+        @marks.create(mark: 1, result: @result)
```
david-yz-liu, line 257: @ashih2018 the problem is you're trying to create three different marks for the same criterion and result. You need to create three different results, and associate one mark to each one.

Also, you should not be making `@result` an instance variable, as it isn't used in the tests later on.

spec/models/rubric_criterion_spec.rb
```diff
@@ -293,15 +293,13 @@
           expect(@levels[1].mark).to eq(2.0)
         end
       end
-
```
david-yz-liu, line 296: Revert these changes. In general put 1 blank line between `describe` blocks.

PR #4474, opened 2020-03-31
Rubric yml

ashih2018:
Fixed test on line 880 in criteria_controller_spec.rb file. To do this, I changed some old rubric criterion yml files so that they follow the new yml format for a rubric criterion and updated the load_from_yml method to be able to create multiple levels through a yml format, which originally was causing an error.
Possible shortcomings/other notices:

- I changed the return type for the RubricCriterion method load_from_yml from returning a new rubric criterion to returning an array whose first index is the created criterion and second index are the levels that are created. This is because levels cannot be created within the load_from_yml file as the rubric_criterion is saved within the controller, not within the model.

### DIFF
--- a/app/controllers/criteria_controller.rb
+++ b/app/controllers/criteria_controller.rb
@@ -160,8 +160,8 @@ class CriteriaController < ApplicationController
             type = criterion_yml[1]['type']
             begin
               if type.casecmp('rubric') == 0
-                data = RubricCriterion.load_from_yml(criterion_yml)
-                criterion, levels = data[0], data[1]
+                criterion, levels = RubricCriterion.load_from_yml(criterion_yml)[0],
+                                    RubricCriterion.load_from_yml(criterion_yml)[1]
               elsif type.casecmp('flexible') == 0
                 criterion = FlexibleCriterion.load_from_yml(criterion_yml)
               elsif type.casecmp('checkbox') == 0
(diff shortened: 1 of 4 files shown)


app/models/rubric_criterion.rb
```diff
@@ -167,14 +171,16 @@ def self.load_from_yml(criterion_yml)
...
     criterion.ta_visible = criterion_yml[1]['ta_visible'] unless criterion_yml[1]['ta_visible'].nil?
     criterion.peer_visible = criterion_yml[1]['peer_visible'] unless criterion_yml[1]['peer_visible'].nil?
-
+    levels = []
     criterion_yml[1]['levels'].each do |name, level_yml|
-      criterion.levels.build(rubric_criterion: criterion,
-                             name: name,
-                             description: level_yml['description'],
-                             mark: level_yml['mark'])
+      level = Level.new
```
david-yz-liu, line 176: You should be able to keep the exact same code as before but just append it directly to `levels` using `<<`, i.e. `levels << criterion.levels.build(...)`.

app/controllers/criteria_controller.rb
```diff
@@ -160,7 +160,9 @@ def upload
             type = criterion_yml[1]['type']
             begin
               if type.casecmp('rubric') == 0
-                criterion = RubricCriterion.load_from_yml(criterion_yml)
+                data = RubricCriterion.load_from_yml(criterion_yml)
```
david-yz-liu, line 163: Use unpacking assignment here: `criterion, levels = ...`.

david-yz-liu: @ashih2018 I actually meant directly on line 163 (so no need for the `data` variable).

PR #4475, opened 2020-03-31
Change minor things about assignment_properties

robert-tan:
1. Added callback so when new Assignment is created, AssignmentProperties also gets created
2. Automatically save associated AssignmentProperties object when Assignment is saved


spec/factories/assignments.rb
```diff
@@ -17,12 +17,12 @@
 
     after(:build) do |assignment, evaluator|
       if evaluator.assignment_properties_attributes
-        assignment.assignment_properties ||= build(:assignment_properties,
+        assignment.assignment_properties = build(:assignment_properties,
                                                    assignment: assignment,
```
david-yz-liu, line 21: You should fix the indentation in this line and the lines below (and note the same is true with the other `build` call).

app/models/assignment.rb
```diff
@@ -889,16 +895,14 @@ def has_peer_review_assignment?
...
-    peerreview_assignment_properties.token_period = 1
-    peerreview_assignment_properties.non_regenerating_tokens = false
-    peerreview_assignment_properties.unlimited_tokens = false
-    peerreview_assignment_properties.repository_folder = repository_folder
     peerreview_assignment = Assignment.new
     peerreview_assignment.parent_assignment = self
     peerreview_assignment.submission_rule = NoLateSubmissionRule.new
     peerreview_assignment.assignment_stat = AssignmentStat.new
-    peerreview_assignment.assignment_properties = peerreview_assignment_properties
+    peerreview_assignment.assignment_properties.token_period = 1
```
david-yz-liu, line 902: Because of the proxying, you should be able to just do `peerreview_assignment.token_period = 1` etc. here and below.

PR #4476, opened 2020-03-31
Improve test coverage

robert-tan:
Add tests for:
1. models/assignment.rb
2. controllers/groups_controller.rb


spec/controllers/groups_controller_spec.rb
```diff
@@ -287,13 +287,114 @@
...
+        }
+        expect(grouping.reload.admin_approved).to be true
+      end
+    end
+
+    describe '#invalidate_groupings' do
+      let(:grouping) { create :grouping_with_inviter }
+
+      it 'should invalidate groupings' do
+        post :global_actions, params: {
```
mishaschwartz, line 308: Try not to call the method that you're testing in the setup for your test. This either duplicates your test at line 294 or it assumes that the method must work properly in order to test if the method works properly. 

Instead, please set the attributes you want to change when you create the model object using the factory. So something like:

```ruby
let(:grouping) { create :grouping_with_inviter, admin_approved: false }
``` 

and then just check that it gets swapped to `true` after you call the `global_actions` method.

The same for the other tests below.

spec/controllers/groups_controller_spec.rb
```diff
@@ -287,13 +287,114 @@
...
+            global_actions: 'invalid'
+        }
+
+        expect(grouping.reload.admin_approved).to be false
+      end
+    end
+
+    describe '#delete_groupings' do
+      before :each do
+        @grouping = create :grouping_with_inviter
```
mishaschwartz, line 327: please use `let` or `let!` instead of setting instance attributes for tests whenever possible

spec/models/assignment_spec.rb
```diff
@@ -1808,4 +1808,123 @@ def grouping_count(groupings)
...
+
+      context 'with assigned students' do
+        it 'has criteria columns' do
+          expect(@assignment.summary_json(admin)[:criteriaColumns]).not_to be_empty
+        end
+
+        it 'has correct criteria information' do
+          criteria_info = @assignment.summary_json(admin)[:criteriaColumns][0]
+          expect(criteria_info).is_a? Hash
+          expect(criteria_info.key? :Header)
```
mishaschwartz, line 1859: Instead of checking for the existence of each key one by one, we should be checking for them all at once. That way, if we are missing multiple keys, our test will report that they are all missing at the same time. 

Have a look at rspec's [contain_exactly](https://relishapp.com/rspec/rspec-expectations/v/3-9/docs/built-in-matchers/contain-exactly-matcher) matcher

spec/models/assignment_spec.rb
```diff
@@ -1808,4 +1808,123 @@ def grouping_count(groupings)
...
+      end
+
+      context 'with assigned students' do
+
+        it 'has student data' do
+          summary_string = @assignment.summary_csv(admin)
+          summary = CSV.parse(summary_string)
+
+          expect(summary).to_not be_empty
+          expect(summary.length()).to be > 1
```
mishaschwartz, line 1921: this check makes the one above redundant

spec/controllers/groups_controller_spec.rb
```diff
@@ -280,13 +280,99 @@
...
+          groupings: [grouping.id],
+          global_actions: 'invalid'
+        }
+
+        expect(grouping.reload.admin_approved).to be false
+      end
+    end
+
+    describe '#delete_groupings' do
+      let(:grouping) { create :grouping_with_inviter }
```
david-yz-liu, line 312: @robert-tan use `let!` instead of let for these two variables. This should cause them to be evaluated before every test, which should fix the failing tests.

PR #4494, opened 2020-04-01
Download test scripts

mishaschwartz:
- add method to zip test script files
- add functionality in file manager to download all files as a zip file
- add api route to download all files as a zip file


app/controllers/automated_tests_controller.rb
```diff
@@ -2,7 +2,9 @@ class AutomatedTestsController < ApplicationController
   include AutomatedTestsHelper
 
   before_action      :authorize_only_for_admin,
-                     only: [:manage, :update]
+                     only: [:manage, :update,
+                            :download_file, :download_files,
```
david-yz-liu, line 6: Everything look great; my only comment here is that if you're going to update the permissions, we should use an `except:` option rather than `only:` like other controllers.

PR #4498, opened 2020-04-02
Spec file upload download

mishaschwartz:
- new api route to download spec file
- new api route to update spec file
- new controller routes to upload/download spec files
- added links in automated testing view to upload/download spec files


spec/controllers/automated_tests_controller_spec.rb
```diff
@@ -1,6 +1,78 @@
...
+        expect(response.body).to eq '{}'
+      end
+      it 'should respond with a success' do
+        expect(response.status).to eq 200
+      end
+    end
+  end
+  context 'POST upload_specs' do
+    before :each do
+      FileUtils.mkdir_p assignment.autotest_path
```
david-yz-liu, line 34: This part is interesting! I did some digging and it looks like the only (?) place that creates this directory is under `AutomatedTestsController#manage`. This strikes me as not great in general, and especially now since you're adding to the API.

One possibility is to create the directory when we create the assignment. Another possibility is to do an existence check in the `Assignment#autotest_path` method itself. Or you might have other suggestions.

app/assets/javascripts/Components/Modals/autotest_file_upload_modal.jsx
```diff
@@ -16,7 +16,12 @@ class AutotestFileUploadModal extends React.Component {
 
   onSubmit = (event) => {
     event.preventDefault();
-    this.props.onSubmit(this.state.newFiles);
+    if (!!this.state.newFiles) {
+      this.props.onSubmit(this.state.newFiles);
+      this.setState({newFiles: []})
```
david-yz-liu, line 21: please put a semicolon here (and throughout)

app/assets/javascripts/Components/Modals/autotest_specs_upload_modal.jsx
```diff
@@ -0,0 +1,57 @@
...
+    if (!!this.state.newFile) {
+      this.props.onSubmit(this.state.newFile);
+      this.setState({newFile: null})
+    } else {
+      this.props.onRequestClose()
+    }
+  };
+
+  handleFileUpload = (event) => {
+    this.setState({newFile: event.target.files[0]})
```
david-yz-liu, line 28: I could be wrong, but I think it's possible for `onChange` to be triggered if you already have a file selected, but then go to select another file and press "Cancel", which results in no file being selected.

So I'd add an emptiness check on `event.target.files` and if it's empty, update `newFile` to be `null`.

PR #4499, opened 2020-04-03
API: API for adding and removing extra marks

### LINKED ISSUE #3670: Feature Request: API for adding/removing extra marks

ANUSHIYAPRIYA:
Added new API for adding and removing extra marks
Added new routes in routes.rb
Added corresponding tests

Resolves #3670


app/controllers/api/groups_controller.rb
```diff
@@ -161,6 +161,46 @@ def update_marks
         HttpStatusHelper::ERROR_CODE['message']['200'] }, status: 200
     end
 
+    def update_extra_marks
```
mishaschwartz, line 164: "update" implies that we're changing an existing extra mark. In this case we're creating an extra mark. We should also have the ability to remove extra marks (probably as another route).

app/controllers/api/groups_controller.rb
```diff
@@ -161,6 +161,46 @@ def update_marks
...
+              .has_submission?
+        result = group.grouping_for_assignment(params[:assignment_id])
+                      .current_submission_used
+                      .get_latest_result
+      else
+        render 'shared/http_status', locals: { code: '404', message:
+            'No submissions exist for that group' }, status: 404
+        return
+      end
+      extra_mark = ExtraMark.new
```
mishaschwartz, line 188: This line until line 193 can all be done in one step

app/controllers/api/groups_controller.rb
```diff
@@ -161,6 +161,86 @@ def update_marks
         HttpStatusHelper::ERROR_CODE['message']['200'] }, status: 200
     end
 
+    def add_extra_marks
```
mishaschwartz, line 164: I should have been more specific in the last comment.
Please change this to `create_extra_mark`

app/controllers/api/groups_controller.rb
```diff
@@ -161,6 +161,86 @@ def update_marks
...
+      end
+      extra_mark = ExtraMark.new(result_id: result.id, extra_mark: params[:extra_marks],
+                                 description: params[:description], unit: ExtraMark::POINTS)
+      unless extra_mark.save
+        # Some error occurred
+        render 'shared/http_status', locals: { code: '500', message:
+        extra_mark.errors.full_messages.first }, status: 500
+        return
+      end
+      extra_mark.save
```
mishaschwartz, line 196: you've already saved this on line 190

app/controllers/api/groups_controller.rb
```diff
@@ -161,6 +161,86 @@ def update_marks
...
+                      .current_submission_used
+                      .get_latest_result
+      else
+        render 'shared/http_status', locals: { code: '404', message:
+            'No submissions exist for that group' }, status: 404
+        return
+      end
+      extra_mark = ExtraMark.find_by(result_id: result.id,
+                                     description: params[:description],
+                                     extra_mark: params[:extra_marks])
```
mishaschwartz, line 227: Don't you think it's overkill to have to specify three attributes in order to find a single extra mark?

app/controllers/api/groups_controller.rb
```diff
@@ -161,6 +161,86 @@ def update_marks
...
+      if extra_mark.nil?
+        render 'shared/http_status', locals: { code: '404', message:
+            'No such Extra Mark exist for that result' }, status: 404
+      elsif extra_mark.destroy
```
mishaschwartz, line 231: I'm pretty sure destroy will raise an error if it fails, not that it will return false

app/controllers/api/groups_controller.rb
ANUSHIYAPRIYA: 1. A result can have more than 1 extra mark 
2. If 2 records have same result id and same extra mark but different description
3. If 2 records have same result id and same description but different extra mark

example:
result_id: 20, description: bonus, extra_mark: 10
result_id: 20, description: bonus for good score, extra_mark: 10
result_id: 20, description: bonus, extra_mark: 2

I think we need to specify these three fields to find a single extra mark. If there is any other way or any other specific reason not to include, I am ready to work on that too.

app/controllers/api/groups_controller.rb
ANUSHIYAPRIYA: Got it. I will update it

app/controllers/api/groups_controller.rb
mishaschwartz: Ok good point. In general we should be specifying extra_marks based on the `id`. Getting that data is going to require another API call though. Let's leave this for now and I'll think about how to reconfigure things after this PR gets pulled in.

app/controllers/api/groups_controller.rb
```diff
@@ -161,6 +161,86 @@ def update_marks
...
+        render 'shared/http_status', locals: { code: '404', message:
+            'No such Extra Mark exist for that result' }, status: 404
+      end
+      begin
+        extra_mark.destroy
+        result.update_total_mark
+        # Successfully deleted the Extra Mark; render success
+        render 'shared/http_status', locals: { code: '200', message:
+            HttpStatusHelper::ERROR_CODE['message']['200'] }, status: 200
+      rescue ActiveRecord::RecordNotDestroyed
```
mishaschwartz, line 237: please report the error that occurred as the message

spec/controllers/api/groups_controller_spec.rb
```diff
@@ -270,6 +270,64 @@
         end
       end
     end
+    context 'POST add_extra_marks' do
+      let(:submission) { create(:version_used_submission, grouping: grouping) }
+      context 'add extra_mark' do
+        before :each do
+          submission
```
mishaschwartz, line 277: no need to call submission here

spec/controllers/api/groups_controller_spec.rb
```diff
@@ -270,6 +270,64 @@
...
+          @old_mark = submission.get_latest_result.total_mark
```
mishaschwartz, line 278: please use `let` variables not instance variables as much as possible

mishaschwartz:
If this is ready, please remove the WIP tag, and make sure you've completed everything in the checklist in the description of this PR.

spec/controllers/api/groups_controller_spec.rb
```diff
@@ -270,6 +270,109 @@
...
+    end
+    context 'DELETE remove_extra_marks' do
+      describe 'when the arguments are invalid' do
+        it_behaves_like 'Invalid_arguments'
+      end
+      describe 'when the arguments are valid' do
+        let(:submission) { create(:version_used_submission, grouping: grouping) }
+        before :each do
+          submission
+          post :create_extra_marks, params: { assignment_id: grouping.assignment.id,
```
mishaschwartz, line 330: please create the extra mark before the tests using factories. If you create a mark by calling this route and there is a bug in the `create_extra_marks` method, these tests will fail for the wrong reasons.

PR #4511, opened 2020-04-06
Validation to prevent levels and rubric criteria to be updated if results are released

ashih2018:
Added a validation to prevent levels and rubric criteria to be updated if results are released, in both the rubric criterion and the level model. 

Note:
- When a criteria is updated, both the validations for the levels and the criteria are run, which may cause convoluted error messages in cases like the following:
image


app/models/level.rb
```diff
@@ -8,4 +8,12 @@ class Level < ApplicationRecord
 
   validates_uniqueness_of :mark, scope: :rubric_criterion_id
   validates_numericality_of :mark, greater_than_or_equal_to: 0
+
+  validate :only_update_if_results_unreleased
+
+  def only_update_if_results_unreleased
+    return if self.rubric_criterion.marks.empty?
```
david-yz-liu, line 15: Rather than duplicate the logic, reuse the validation method in `RubricCriterion`.

app/models/rubric_criterion.rb
```diff
@@ -38,6 +39,12 @@ def validate_max_mark
     errors.add(:max_mark, 'Max mark of rubric criterion should not be greater than max level mark')
   end
 
+  def only_update_if_results_unreleased
```
david-yz-liu, line 42: Change the name of the method. Look at how other validation method underneath this one is named, and follow the same idea.

app/models/rubric_criterion.rb
```diff
@@ -38,6 +39,12 @@ def validate_max_mark
...
+    return if self.marks.empty?
+    return unless self.marks[0].result.released_to_students
```
david-yz-liu, line 44: Checking the first mark isn't enough, there might be a mixture of released and unreleased results. Use a join on `self.marks` and results and filter by `released_to_students`. See https://guides.rubyonrails.org/active_record_querying.html#specifying-conditions-on-the-joined-tables.

app/models/rubric_criterion.rb
```diff
@@ -25,6 +25,7 @@ class RubricCriterion < Criterion
 
   has_many :test_groups, as: :criterion
 
+  validate :only_update_if_results_unreleased
```
david-yz-liu, line 28: This is good, but you should actually move it into the `Criterion` superclass so that it affects all three criteria types. And you can add tests for the other two as well.

spec/models/rubric_criterion_spec.rb
```diff
@@ -326,6 +326,33 @@
           expect(@levels[0].update(mark: 1)).to be false
         end
       end
+      context 'when a result is released' do
+        before do
+          @marks = @criterion.marks
+          results = []
+          for i in 0..3
```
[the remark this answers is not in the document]
ashih2018: what's the best way to fix this hound comment?

app/models/checkbox_criterion.rb
```diff
@@ -9,6 +9,7 @@ class CheckboxCriterion < Criterion
   has_many :test_groups, as: :criterion
 
   validate :visible?
+  validate :results_unreleased?
```
david-yz-liu, line 12: Oh, you can move the validation line into `criterion.rb` as well.

app/models/criterion.rb
```diff
@@ -165,4 +165,17 @@ def visible?
       false
     end
   end
+
+  public
```
david-yz-liu, line 169: Generally you shouldn't use this keyword. You just need to move this method to before the `private` keyword above.

PR #4519, opened 2020-04-08
Background jobs: hide progress flash messages once the job has finished

mishaschwartz:
- adds ability to hide flash messages based on the message type (:notice, :error, etc.)
- only available for ajax requests currently
- used here to hide the progress messages for background jobs once the job has finished, either successfully or due to an error/failure. 

TODO:

- [x] update changelog
- [x] update tests for `flash_job_messages`

### DIFF
--- a/app/controllers/job_messages_controller.rb
+++ b/app/controllers/job_messages_controller.rb
@@ -13,7 +13,7 @@ class JobMessagesController < ApplicationController
     elsif status.read.empty?
       flash_message(:error, t('poll_job.failed'))
       session[:job_id] = nil
-      render 'shared/http_status', locals: { code: '404', message: t('poll_job.not_enqueued') }, status: 404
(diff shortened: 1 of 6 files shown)


app/controllers/job_messages_controller.rb
```diff
@@ -13,7 +13,7 @@ def get
     elsif status.read.empty?
       flash_message(:error, t('poll_job.failed'))
       session[:job_id] = nil
-      render 'shared/http_status', locals: { code: '404', message: t('poll_job.not_enqueued') }, status: 404
+      render json: { code: '404', message: t('poll_job.not_enqueued') }, status: 404
```
david-yz-liu, line 16: Just to confirm, did you change this so that it's the same type as the normal case (Line 20)?

mishaschwartz: Yes. Also for some (unknown?) reason, the previous version caused any rspec test that ran that line to hang indefinitely.

PR #4535, opened 2020-04-14
Support arbitrary number of rubric levels

david-yz-liu:
This pull request introduces the ability for a rubric criterion to have an arbitrary number of levels, with arbitrary mark values associated with each level.

This also changes the format for exporting criteria to yml and csv files.


lib/tasks/demo.rake
```diff
@@ -87,22 +87,17 @@ namespace :markus do
...
+    attributes = []
+    5.times do |number|
+      lvl = { name: random_words(1), description: random_sentences(5), mark: number }
+      attributes.push(lvl)
+    end
+    params = { rubric: {
+      name: 'Mark3', assessment_id: a.id,
+      position: 2, max_mark: 4, levels_attributes: attributes
+    } }
+    RubricCriterion.create!(params[:rubric])
```
david-yz-liu, line 100: As above, make use of nested attributes here to just create a single hash, and pass it to `RubricCriterion.create`.

lib/tasks/results.rake
```diff
@@ -41,14 +41,21 @@ namespace :markus do
...
+      rubric_criteria = [{ name: 'Uses Conditionals', max_mark: 4 }, { name: 'Code Clarity', max_mark: 8 },
+                         { name: 'Code Is Documented', max_mark: 12 }, { name: 'Uses For Loop', max_mark: 4 }]
+      default_levels = [
+          { name: 'Quite Poor', description: 'This criterion was not satisfied whatsoever', mark: 0 },
+          { name: 'Satisfactory', description: 'This criterion was satisfied', mark: 1 },
+          { name: 'Good', description: 'This criterion was satisfied well', mark: 2 },
+          { name: 'Great', description: 'This criterion was satisfied really well!', mark: 3 },
+          { name: 'Excellent', description: 'This criterion was satisfied excellently', mark: 4 }
+      ]
       rubric_criteria.each do |rubric_criteria|
```
david-yz-liu, line 53: Change the loop variable to `rubric_criterion`.

lib/tasks/results.rake
```diff
@@ -41,14 +41,21 @@ namespace :markus do
...
-        rc = RubricCriterion.new
-        rc.update(rubric_criteria)
-        rc.update(default_levels)
-        rc.assignment = a1
-        rc.save
+        params = { rubric: {
```
david-yz-liu, line 54: You don't need the `rubric` key here. Just set `params = { assignment: ... }` and similarly below as well.

spec/controllers/criteria_controller_spec.rb
```diff
@@ -890,8 +877,8 @@
       end
 
       it 'creates criteria with rounded (up to first digit after decimal point) maximum mark' do
-        post_as admin, :upload, params: { assignment_id: assignment.id, upload_file: round_max_mark_file }
-
+        post_as admin, :upload, params: { assignment_id: assignment.id,
```
david-yz-liu, line 880: revert this change

spec/controllers/criteria_controller_spec.rb
```diff
@@ -792,40 +793,26 @@
...
-        expect(cr2.level_2_description).to be_nil
-        expect(cr2.level_3_name).to be_nil
-        expect(cr2.level_3_description).to be_nil
-        expect(cr2.level_4_name).to be_nil
-        expect(cr2.level_4_description).to be_nil
+        expect(cr2.levels.size).to eq(5)
         expect(cr2.ta_visible).to be true
         expect(cr2.peer_visible).to be false
       end
-
```
david-yz-liu, line 828: Revert this change

app/assets/stylesheets/criteria.scss
```diff
@@ -80,20 +80,38 @@
...
+      min-width: 12em;
+    }
+
     textarea {
       height: 150px;
       margin-bottom: 1.5em;
       width: 100%;
     }
+
+    .description {
```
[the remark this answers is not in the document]
ashih2018: should I be changing this because I have the same depth as previous code in this scss file?

lib/tasks/demo.rake
ashih2018: I'm don't quite understand, what should I change about the way I'm using nested attributes here?

ashih2018: I think I understand now, you can ignore this comment above

david-yz-liu: Oh no you're right, you're using nested attributes well here. Just remove the wrapping `rubric:` key from `params`, you don't need it. I think I commented on this somewhere else too.

Issue #3471, opened 2018-08-21
Don't reset max mark when changing criteria type


mishaschwartz:
When adding marking scheme criterion, if the "Maximum Mark" field is filled in first and then the criteria type is changed (eg. from "Rubric" to "Flexible"), then the "Maximum Mark" is reset to 1.
Changing the criteria type should not change the maximum mark

PR #4539, opened 2020-04-20
Max mark does not reset when changing the criteria type

ANUSHIYAPRIYA:
This resolves issue #3471

### DIFF
--- a/app/views/criteria/new.js.erb
+++ b/app/views/criteria/new.js.erb
@@ -6,13 +6,13 @@ document.getElementById('new_criterion_prompt').focus();
 
 $(document).ready(function() {
   $('input[name=criterion_type]:radio').change(function () {
-    var current = document.getElementById('max_mark_prompt')
+    const max_mark = document.getElementById('max_mark_prompt');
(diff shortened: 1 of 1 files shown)


app/views/criteria/new.js.erb
```diff
@@ -6,13 +6,13 @@ document.getElementById('new_criterion_prompt').focus();
 
 $(document).ready(function() {
   $('input[name=criterion_type]:radio').change(function () {
-
+    var current = document.getElementById('max_mark_prompt')
```
david-yz-liu, line 9: Use `const` instead of `var` here. (See https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75 for further reading.)

Also, choose a better variable name than `current`.

And finally, add a semicolon to the end of the line.

mishaschwartz:
If you add the following in the description:

Resolves #3471

The issue will be resolved automatically when this is pulled in. See here: https://help.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords

PR #4541, opened 2020-04-21
Refactored scaling marks callback

ashih2018:
Refactored scale_marks_if_max_mark_changed callback in rubric criterion model. Originally if a level mark would scale and have the same mark as another level temporarily, then an error would be raised as a result of uniqueness of level marks within a rubric criterion.
 For example, if a rubric has levels with marks 0, 1, 2 and the max mark was doubled, an error would be raised as 1 would double to 2 so there would be two duplicate level marks temporarily. To fix this, when saving the level, I disable the validation check when scaling.

### DIFF
--- a/app/models/rubric_criterion.rb
+++ b/app/models/rubric_criterion.rb
@@ -55,6 +55,8 @@ class RubricCriterion < Criterion
     self.levels.each do |level|
       # don't scale levels that the user has manually changed
       unless (level.changed.include? 'mark') || level.mark.nil?
+        # use update_attribute to skip validatation in case updating level mark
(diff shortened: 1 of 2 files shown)


david-yz-liu:
@ashih2018 this is good, please add a new test to verify this fix.

ashih2018:
@david-yz-liu there already is a preexisting test that doubles/halves level marks to demonstrate this fix. I updated the messages for those tests, is that sufficient?

david-yz-liu:
@ashih2018 as far as I can tell, this test was passing before. So what you should be doing is adding a new test that *failed* before the changed you made, but now passes.

PR #4545, opened 2020-04-22
[WIP] Refactor sti criteria branch. 

ashih2018:
This PR addresses many of the small changes to the sti-criteria merge into master PR, which includes hound comments, as well as some styling changes and some of the comments made by @david-yz-liu


app/models/rubric_criterion.rb
```diff
@@ -25,15 +25,16 @@ class RubricCriterion < Criterion
 
   has_many :test_groups, as: :criterion
 
+  DEFAULT_MAX_MARK = 4
```
ashih2018, line 28: DEFAULT MAX MARK is currently being used in the new criterion erb file as a default mark for the maximum. I think this should be the only constant we should keep

david-yz-liu: Okay we can keep this.

app/controllers/rubric_criteria_controller.rb
```diff
@@ -6,10 +6,10 @@ def download_csv
     @assignment = Assignment.find(params[:assignment_id])
     file_out = MarkusCsv.generate(@assignment.get_criteria(:all, :rubric)) do |criterion|
       criterion_array = [criterion.name, criterion.max_mark]
-      (0..criterion.levels.length - 1).each do |i|
+      criterion.levels.length.times do |i|
```
david-yz-liu, line 9: @ashih2018 Please delete the extra space at the end of this line.

PR #4551, opened 2020-04-23
assignment: Add inverse_of option to associations.

david-yz-liu:
This fixes a bug when creating assignments where associated object passed
as nested attributes were failing a validation (no assignment id).

I noticed this on the assignment form when attempting to create an
assignment with required files.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,7 @@
 - Fixed sorting in annotation table in results view (#4542)
 - Fixed N+1 queries in Assignment repo list methods (#4543)
 - Fixed submission download_repo_list file extension (#4543)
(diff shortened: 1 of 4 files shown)


mishaschwartz:
Do we also want to do this for:

- each of the criterion types for assignment (`has_many :rubric_criteria, inverse_of: :assignment` ...)
- grade_entry_students for the grade_entry_form model (`has_many :grade_entry_students, inverse_of: :grade_entry_form` ...)

david-yz-liu:
BTW I added a test case that fails without the `inverse_of`, as the behaviour in this case seemed pretty subtle to me.

PR #4557, opened 2020-04-30
Assignment penalty bugfix

mishaschwartz:
Fixes a bug where submission rules could not be created along with an assignment or updated for an existing assignment through the UI.
Also fixes a bug where the first submission period couldn't be created if all submission periods for a given rule had previously been deleted.

## Motivation and Context

Fixes two bugs which prevent users from creating or updating assignments that include submission rules. We also had no tests for the `AssignmentsController.update` route so this PR introduces some basic tests as well.

## Your Changes

**Description**:

- Removed `submission_rule_attributes` and sub-nested params from `assignment_params`. `assignment_params` is used to update nested attributes but the `submission_rule_attributes` cannot be updated directly since the classes they belong to are polymorphic.
- Added an extra conditional in the view that generates the submission rule form on the assignment page. This now generates a new empty form for a submission rule if the rule doesn't already exist OR if the rule has no associated penalty periods (before it only checked for the existence of the rule)

## Testing

- Added additional tests for the `update` method
- Tested in UI to ensure that assignments could be created/updated with any submission rule

## Checklist


spec/controllers/assignments_controller_spec.rb
```diff
@@ -414,4 +414,101 @@
       end
     end
   end
+
+  context '#update' do
+    let(:assignment) { create :assignment }
+    let(:admin) { create :admin }
+    let(:submission_rule) { create :penalty_decay_period_submission_rule, assignment: assignment }
+    let(:params) do
```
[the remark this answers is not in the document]
mishaschwartz: vote to ignore

david-yz-liu:
@mishaschwartz can do you another merge with master? I'm hoping I fixed some flaky tests in #4559.

app/controllers/assignments_controller.rb
```diff
@@ -611,9 +611,6 @@ def process_assignment_form(assignment)
       # Is the instructor forming groups?
       if assignment_params[:assignment_properties_attributes][:student_form_groups] == '0'
         assignment.invalid_override = true
-        # Increase group_max so that create_all_groups button is not displayed
-        # in the groups view.
-        assignment.group_max = 2
```
david-yz-liu, line 616: Can you explain the purpose of this change?

mishaschwartz: If groups are enabled but students can form their own group is disabled, the group_max value cannot be changed from 2. This means that an instructor who wants to manage their own groups, but wants to have groups that are larger than 1 student each, will have to manually validate every group they make on the groups page (because MarkUs will think that the group size is invalid).

This was here originally because we didn't want instructors to be able to click the "Create all groups" button if they said that students could form groups. However, this button is only visible if they set the max group size to 1. In that case, it would actually be very convenient for instructors to be able to create all remaining groups for the rest of the students with a click of a button.

david-yz-liu: Great, thanks!

PR #4593, opened 2020-05-14
Deductive annotation model changes

ignaspa:
## Motivation and Context

Changing the models for the requirements of deductive annotations. This is required so that changes to the controllers/views can be applied.

## Your Changes

**Description**:
Changing validations on the Mark, AnnotationText, and AnnotationCategory models.
Adding a calculate_deduction method to the mark model

- [ ] Other (please specify): 

## Checklist

### DIFF
--- a/app/models/annotation_category.rb
+++ b/app/models/annotation_category.rb
@@ -53,17 +53,14 @@ class AnnotationCategory < ApplicationRecord
   end
 
   def update_annotation_text_deductions
-    return unless changes_to_save.key?('flexible_criterion_id')
+    return if !changes_to_save.key?('flexible_criterion_id') || self.annotation_texts.empty?
 
     if marks_released?
       errors.add(:base, 'Cannot update annotation category flexible criterion once results are released.')
       throw(:abort)
     end
-
-    return if self.annotation_texts.nil?
     prev_criterion = FlexibleCriterion.find_by_id(changes_to_save['flexible_criterion_id'].first)
     new_criterion = FlexibleCriterion.find_by_id(changes_to_save['flexible_criterion_id'].second)
-    return unless prev_criterion != new_criterion
     if new_criterion.nil?
       self.annotation_texts.each do |text|
         text.update!(deduction: nil)
--- a/app/models/annotation_category.rb
+++ b/app/models/annotation_category.rb
@@ -8,10 +8,8 @@ class AnnotationCategory < ApplicationRecord
 
   belongs_to :flexible_criterion, required: false
 
-  before_update :check_if_released_deductions
+  before_update :update_annotation_text_deductions
   before_destroy :check_if_deductions_exist
-
-  after_update :update_annotation_text_deductions
 
   # Takes an array of comma separated values, and tries to assemble an
   # Annotation Category, and associated Annotation Texts
@@ -35,15 +33,6 @@ class AnnotationCategory < ApplicationRecord
     end
   end
 
-  def check_if_released_deductions
-    return unless changes_to_save.key?('flexible_criterion_id')
-
-    return if marks_not_released?
-    byebug
-    errors.add(:base, 'Cannot update annotation category flexible criterion once results are released.')
-    throw(:abort)
-  end
-
   def marks_not_released?
     return true if self.flexible_criterion.nil?
     self.flexible_criterion.marks.joins(:result).where('results.released_to_students' => true).empty?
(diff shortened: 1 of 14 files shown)


app/models/mark.rb
```diff
@@ -18,13 +18,29 @@ class Mark < ApplicationRecord
...
+  end
+
   def scale_mark(curr_max_mark, prev_max_mark, update: true)
     return if mark.nil?
     return 0 if prev_max_mark == 0 || mark == 0 # no scaling occurs if prev_max_mark is 0 or mark is 0
     if markable.is_a? RubricCriterion
       new_mark = (mark * (curr_max_mark / prev_max_mark)).round(1)
     elsif markable.is_a? FlexibleCriterion
-      new_mark = (mark * (curr_max_mark.to_f / prev_max_mark)).round(2)
+      new_mark = (mark * (curr_max_mark.to_f / prev_max_mark)).round(2) - :calculate_deduction
```
david-yz-liu, line 43: Let's not do it like this. This method is called when the flexible criterion's `max_mark` changes. Look for that callback, and then scale the criterion's annotation deductions there as well.

app/models/mark.rb
```diff
@@ -18,13 +18,29 @@ class Mark < ApplicationRecord
   validates_uniqueness_of :markable_id,
                           scope: [:result_id, :markable_type]
 
+  validates_inclusion_of :override, in: [true, false]
+
+  def calculate_deduction
+    return 0 if self.markable_type != 'FlexibleCriterion' || !self.override?
+    total_deduction = 0
+    applied_annotations = mark.result.annotations
+    applied_annotations.each do |annotation|
```
david-yz-liu, line 27: This is good, but you can speed it up with a smarter database query. https://guides.rubyonrails.org/active_record_querying.html#calculations.

app/models/annotation_category.rb
```diff
@@ -27,4 +31,21 @@ def self.add_by_row(row, assignment, current_user)
...
+    prev_criterion = previous_changes["flexible_criterion_id"].first
+    new_criterion = previous_changes["flexible_criterion_id"].second
+    if prev_criterion != new_criterion
+      if new_criterion.nil?
+        self.annotation_texts.each do |text|
+          text.update!(deduction: nil)
+        end
+      else
+        self.annotation_texts.each do |text|
+          text.update!(deduction: 0)
```
david-yz-liu, line 46: This is interesting, let's calculate this by scaling the deduction according to the ratio of the previous criterion's `max_mark` and the new `max_mark` (unless the previous criterion was `nil`).

app/models/annotation_text.rb
```diff
@@ -3,15 +3,30 @@ class AnnotationText < ApplicationRecord
...
+  after_update :update_mark_deductions
+
   # An AnnotationText has many Annotations that are destroyed when an
   # AnnotationText is destroyed.
   has_many :annotations, dependent: :destroy
 
   belongs_to :annotation_category, optional: true, counter_cache: true
   validates_associated :annotation_category, on: :create
 
+  validates_numericality_of :deduction,
```
david-yz-liu, line 15: You might need to add an `if:` option to guard when the `deduction` or the associated `flexible_criterion` is `nil`, I can't remember how these validates work.

app/models/mark.rb
ignaspa: I actually have this wrong I think, I did a sum after your comment, but now I realize that this will add all the deductions to the mark, even the ones for other criteria if there are.

app/models/annotation_category.rb
```diff
@@ -27,4 +34,41 @@ def self.add_by_row(row, assignment, current_user)
...
+    return unless prev_criterion != new_criterion
+    if new_criterion.nil?
+      self.annotation_texts.each do |text|
+        text.update!(deduction: nil)
+      end
+    elsif prev_criterion.nil?
+      text.update!(deduction: 0)
+    else
+      self.annotation_texts.each do |text|
+        text.scale_deduction(new_criterion.max_mark.to_f / prev_criterion.max_mark.to_f)
```
david-yz-liu, line 70: You shouldn't have to do `to_f` here.

app/models/criterion.rb
```diff
@@ -154,6 +154,13 @@ def scale_marks
     unless updated_marks.empty?
       Mark.upsert_all(all_marks.pluck_to_hash.map { |h| { **h.symbolize_keys, mark: updated_marks[h['id'].to_i] } })
     end
+
+    AnnotationCategory.where(flexible_criterion_id: self.id).each do |category|
```
david-yz-liu, line 158: It's not enough to just check `id` here; guard this by checking whether `self` is a `FlexibleCriterion`

david-yz-liu: (In which case you can do `self.annotation_categories`)

app/models/mark.rb
```diff
@@ -18,6 +18,27 @@ class Mark < ApplicationRecord
...
+                                .where(annotation_texts: { annotation_categories:
+                                                              { flexible_criteria: { id: self.markable_id } } })
+                                .sum(:deduction)
+
+    deductive_annotations
+  end
+
+  def update_deduction
+    # this might need to be update_attributes()
+    self.mark = self.markable.max_mark - calculate_deduction
```
david-yz-liu, line 39: Right now this won't trigger a database save.

app/models/mark.rb
```diff
@@ -18,6 +18,27 @@ class Mark < ApplicationRecord
   validates_uniqueness_of :markable_id,
                           scope: [:result_id, :markable_type]
 
+  validates_inclusion_of :override, in: [true, false]
+
+  def calculate_deduction
+    return 0 if self.markable_type != 'FlexibleCriterion' || self.override?
+
+    # use joins to fix this statement to be query
+    deductive_annotations = self.result
```
david-yz-liu, line 27: You don't need a variable here, you can return the result of the query directly.

spec/factories/annotation_texts.rb
```diff
@@ -6,4 +6,10 @@
     creator { FactoryBot.create(:admin) }
     association :annotation_category, factory: :annotation_category
   end
+
+  factory :annotation_text_with_deduction, parent: :annotation_text do
+    after(:create) do |annotation_text|
```
david-yz-liu, line 11: Since you're just setting an attribute here you don't need an `after(:create)`. You can just do `deduction { 1.0 }` as with the other attributes (see above factory).

PR #4601, opened 2020-05-20
Enable variable permissions for graders

ANUSHIYAPRIYA:
## Motivation and Context
This allows admin to enable/disable various permissions for graders.

## Your Changes

**Description**:
 Created a new table that contains all the configurable grader actions and admin can enable/disable them. Enforced those permissions using Action policy for various controllers. There will be structural changes in the user interface of the application. Also creating a new user interface for the purpose of this PR.

## Testing
Manually tested this PR through the web interface

## Checklist

- [ ] I have described any required documentation changes below, if applicable.


app/assets/javascripts/Components/submission_table.jsx
```diff
@@ -246,6 +246,7 @@ class RawSubmissionTable extends React.Component {
           ref={(r) => this.actionBox = r}
           disabled={this.props.selection.length === 0}
           is_admin={this.props.is_admin}
+          is_ta={this.props.is_ta}
```
mishaschwartz, line 249: Please use 'grader' instead of 'ta' wherever possible. I know that we use ta in a bunch of places but I'm trying to get rid of that language where we can.

app/assets/javascripts/Components/submission_table.jsx
```diff
@@ -357,6 +358,16 @@ class SubmissionsActionBox extends React.Component {
         </button>
       );
     }
+    if (this.props.is_ta) {
```
mishaschwartz, line 361: Shouldn't this be conditional on whether they have permission to collect submissions (not just if they are a grader or not)?

app/controllers/annotations_controller.rb
```diff
@@ -10,46 +10,50 @@ def add_existing_annotation
...
+          type: 'TextAnnotation',
+          line_start: params[:line_start],
+          line_end: params[:line_end],
+          column_start: params[:column_start],
+          column_end: params[:column_end],
+          **base_attributes
+        )
+      end
+      render :create
+    rescue ActionPolicy::Unauthorized => e
```
mishaschwartz, line 54: rescuing from an `ActionPolicy::Unauthorized` error should go immediately after the call to `authorize!` if possible

app/controllers/assignments_controller.rb
```diff
@@ -144,6 +150,7 @@ def index
 
   # Called on editing assignments (GET)
   def edit
+    authorize! with: AssignmentPolicy
```
mishaschwartz, line 153: I think in this case you want to render a 403 error if this fails, not flash a message. Please handle this sort of authorization in a `before_action` context.

app/controllers/grade_entry_forms_controller.rb
```diff
@@ -2,24 +2,8 @@
...
-                only: [:grades,
-                       :populate_grades_table,
-                       :download,
-                       :upload,
-                       :update_grade]
-  before_action :authorize_for_student,
-                only: [:student_interface]
-
+  before_action { authorize! with: GradeEntryFormPolicy }
+  rescue_from ActionPolicy::Unauthorized, with: :not_authorized
```
mishaschwartz, line 6: Please use the default `rescue_from` in `ApplicationController`

app/controllers/grade_entry_forms_controller.rb
```diff
@@ -233,4 +217,9 @@ def upload
     end
     redirect_to action: 'grades', id: @grade_entry_form.id
   end
+
+  def not_authorized
```
mishaschwartz, line 221: This should not be necessary (see above)

app/controllers/job_messages_controller.rb
```diff
@@ -1,5 +1,5 @@
 class JobMessagesController < ApplicationController
-  before_action :authorize_only_for_admin
+  before_action :authorize_for_ta_and_admin
```
mishaschwartz, line 2: Are you going to make the authorization for TAs configurable afterwards?

app/controllers/submissions_controller.rb
```diff
@@ -197,43 +202,48 @@ def collect_submissions
...
-    groupings.each do |grouping|
-      unless is_scanned_exam
-        collect_now = collection_dates[grouping.id] <= Time.current
-        some_before_due = true unless collect_now
-        next unless collect_now
-      end
-      next if params[:override] != 'true' && grouping.is_collected?
-      next if some_released.include?(grouping.id)
+    begin
+      authorize! with: SubmissionPolicy
```
mishaschwartz, line 206: For each authorization, please think carefully about how to report an unauthorized error message.
The main thing to consider is: will the user be able to see a button/link/etc. that will call this action? If yes, then a flash message might be appropriate. If no, then any attempt to access this route should give a 403 error.

This applies not just to this route but to every route that you are changing/will change as part of this PR.

app/models/grader_permission.rb
```diff
@@ -0,0 +1,5 @@
+# Contains the grader permissions for a particular grader
+class GraderPermission < ApplicationRecord
+  self.table_name = 'grader_permission'
```
mishaschwartz, line 3: The table name should be 'grader_permissions'. Please stick with the default naming conventions. You will have to change this in your migration files as well.

app/policies/annotation_policy.rb
```diff
@@ -0,0 +1,5 @@
+class AnnotationPolicy < ApplicationPolicy
+  def add_existing_annotation?
+    user.admin? || (user.ta? && GraderPermission.find_by(user_id: user.id).create_delete_annotations)
```
mishaschwartz, line 3: You can probably simplify this syntax by making a GraderPermissionPolicy class and defining a policy for each configuration option.

app/policies/submission_policy.rb
```diff
@@ -27,4 +27,12 @@ def not_a_student?
   def before_release?
     !record.current_result.released_to_students
   end
+
+  def manually_collect_and_begin_grading?
+    user.admin? || (user.ta? && GraderPermission.find_by(user_id: user.id).manually_collect_and_begin_grading)
```
mishaschwartz, line 32: Use an alias when policies are identical

app/views/assignments/index.html.erb
```diff
@@ -1,6 +1,6 @@
 <% content_for :title, t('assignments.manage_course_work') %>
 
-<% if @current_user.admin? %>
+<% if @current_user.admin? || @current_user.ta? %>
```
mishaschwartz, line 3: Shouldn't this be conditional on whether they have permission to edit assignments or not (not just if they are a grader or not)?

app/views/submissions/browse.html.erb
```diff
@@ -7,6 +7,7 @@
           show_grace_tokens: <%= @assignment.submission_rule.type == 'GracePeriodSubmissionRule' %>,
           show_sections: <%= Section.exists? %>,
           is_admin: <%= @current_user.admin? %>,
+          is_ta: <%= @current_user.ta? %>,
```
mishaschwartz, line 10: A user can't be both a grader and an admin so it might be better to combine the `is_admin` and `is_ta` props somehow.

db/migrate/20200506170130_create_grader_permission.rb
```diff
@@ -0,0 +1,12 @@
+class CreateGraderPermission < ActiveRecord::Migration[6.0]
```
mishaschwartz, line 1: Please combine all the migration files in to a single file

spec/controllers/notes_controller_spec.rb
```diff
@@ -104,13 +104,24 @@
       expect(response).to render_template('notes/index')
     end
 
-    it 'get :new' do
+    it 'when TA is allowed to get :new' do
+      GraderPermission.create(user_id: @ta.id, create_notes: true)
```
mishaschwartz, line 108: please use factories to create objects in tests whenever possible

app/assets/javascripts/Components/submission_table.jsx
ANUSHIYAPRIYA: The collect submissions button will be visible to all graders. I am checking whether they have permission or not in submission_policy.rb. If they are not allowed, when they click the button, it will raise an error and won't allow them to collect. If we want to display the button based on their permission I can modify according to that.

mishaschwartz: Yes, please only display pages/button/links/etc. to users who have permission to use them

app/views/layouts/menus/_ta_sub_menu.html.erb
```diff
@@ -1,27 +1,72 @@
...
-    controller.controller_name == 'results' ||
-    controller.action_name == 'summary' %>
-    <li <% if controller.controller_name == 'submissions' ||
-      controller.controller_name == 'results' %>
-        class='active'
-        <% end %>>
-      <%= link_to Submission.model_name.human.pluralize,
-                  browse_assignment_submissions_path(@assignment || @grouping.assignment) %>
+
+  <% if assessment.is_a?(GradeEntryForm) && !assessment.new_record? %>
```
mishaschwartz, line 25: A lot of the code below is copied from the admin view. Instead of copying code. Please find a way to combine the two views or factor out the shared code.

app/views/submissions/browse.html.erb
```diff
@@ -7,6 +12,8 @@
...
```
mishaschwartz, line 15: A user can't be both a grader and an admin so it might be better to combine the is_admin and is_ta props somehow.

db/migrate/20200602012602_add_column_to_graders_permissions.rb
```diff
@@ -0,0 +1,5 @@
+class AddColumnToGradersPermissions < ActiveRecord::Migration[6.0]
```
mishaschwartz, line 1: *one* migration file please

spec/controllers/results_controller_spec.rb
```diff
@@ -520,6 +520,33 @@ def self.test_unauthorized(route_name)
     end
     include_examples 'shared ta and admin tests'
 
+    describe '#delete_grace_period_deduction' do
+      let(:deduction) do
+        create :grace_period_deduction, membership: grouping.accepted_student_memberships.first, deduction: 1
+      end
+      context 'When TA is allowed to delete_grace_period_deduction' do
+        before do
+          GraderPermissions.create(user_id: ta.id, delete_grace_period_deduction: true)
```
mishaschwartz, line 529: create objects with factories

spec/controllers/results_controller_spec.rb
```diff
@@ -520,6 +520,33 @@ def self.test_unauthorized(route_name)
...
+        it 'should delete grace period deduction' do
+          delete :delete_grace_period_deduction,
+                 params: { assignment_id: assignment.id, submission_id: submission.id,
+                           id: complete_result.id, deduction_id: deduction.id }
+          expect(grouping.grace_period_deductions.exists?).to be false
+        end
+      end
+      context 'When TA is not allowed to delete_grace_period_deduction' do
+        before do
+          GraderPermissions.create(user_id: ta.id, delete_grace_period_deduction: false)
```
mishaschwartz, line 540: create objects with factories

yarn.lock
```diff
@@ -4558,13 +4558,25 @@ mime-db@1.40.0, "mime-db@>= 1.40.0 < 2":
   resolved "https://registry.yarnpkg.com/mime-db/-/mime-db-1.40.0.tgz#a65057e998db090f732a68f6c276d387d4126c32"
```
mishaschwartz, line 4558: Why are there changes in your yarn.lock file?

app/views/submissions/browse.html.erb
```diff
@@ -7,7 +7,8 @@
           show_grace_tokens: <%= @assignment.submission_rule.type == 'GracePeriodSubmissionRule' %>,
           show_sections: <%= Section.exists? %>,
           is_admin: <%= @current_user.admin? %>,
-          can_run_tests: <%= @current_user.admin? && @assignment.enable_test %>,
+          is_allowed_to_collect: <%= allowed_to? :collect? %>,
+          can_run_tests: <%= (allowed_to? :allowed_to_run_tests?) && @assignment.enable_test %>,
```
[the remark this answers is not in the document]
mishaschwartz: To follow up on this... ideally this should only call one policy not two. Please see my second point here: https://github.com/MarkUsProject/Markus/pull/4601#pullrequestreview-428081772

app/views/automated_tests/_test_runs.html.erb
```diff
@@ -6,7 +6,7 @@
         submission_id: <%= @result&.submission_id || 'null' %>,
         assignment_id: <%= @assignment.id %>,
         grouping_id: <%= @grouping.id %>,
-        instructor_run: <%= ran_by == 'instructor' %>,
+        instructor_run: <%= ran_by == 'instructor' || (allowed_to? :can_run_tests?) %>,
```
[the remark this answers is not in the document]
mishaschwartz: Please change this is a separate PR

david-yz-liu:
@ANUSHIYAPRIYA okay, I've started working through this PR now. I will focus first on simplifying the design of the permissions themselves, and also mention opportunities to break out smaller parts.

1. `create_delete_annotations`: I want you to *remove* this permission altogether. TAs should always be able to add/modify annotations in the `results/edit` view, and for `annotation_controllers/index` we will just use the `manage_annotations` setting. Please pull out the changes related to `AnnotationPolicy` (and the corresponding changes in the results view files) into a separate PR. (And so they should no longer appear here.) You can keep the `AnnotationCategoriesPolicy` here, though, since that will rely on the `manage_annotations` setting.
2. `create_notes`: You can remove this one altogether. TAs can always create notes.
3. `manage_exam_templates`: You can remove this one, and replace it with `manage_assignments` (i.e., if TAs can manage assignment, then they can manage exam templates).
4. `manage_reviewers`: You can remove this one, and replace it with `manage_assignments` (i.e., if TAs can manage assignment, then they can assign reviewers).
5. `manage_assignments`/`manage_grade_entry_forms`: Merge these into a single permission, `manage_assessments`.

app/views/layouts/_sub_menu.html.erb
```diff
@@ -25,8 +25,11 @@
 <% elsif @current_user.ta? && (controller.controller_name == 'assignments' ||
```
mishaschwartz, line 25: instead of checking @current_user.ta? (and @current_user.admin? above) we should be using your new policies

app/views/layouts/_sub_sub_menu.html.erb
```diff
@@ -15,43 +15,53 @@
...
+                class='active'
+                <% end %>>
+              <%= link_to t('menu.properties'), edit_assignment_path(@assignment) %>
             </li>
           <% end %>
-          <% if @assignment.scanned_exam? %>
-            <li class='<%= "active" if controller.controller_name == 'exam_templates' %>'>
-              <%= link_to ExamTemplate.model_name.human.pluralize,
-                          assignment_exam_templates_path(@assignment) %>
+          <% if @current_user.admin? %>
```
mishaschwartz, line 25: We should try to avoid checking the type of user in views and instead try to check the policies instead.

This applies in a few other places as well

app/views/submissions/browse.html.erb
```diff
@@ -8,7 +8,9 @@
           show_sections: <%= Section.exists? %>,
           is_admin: <%= @current_user.admin? %>,
```
mishaschwartz, line 9: please double check this but I don't think the `is_admin` and `is_ta` props are being used anymore so they should be removed.

db/migrate/20200805172907_create_grader_permissions.rb
```diff
@@ -0,0 +1,19 @@
+class CreateGraderPermissions < ActiveRecord::Migration[6.0]
+  def self.up
+    create_table :grader_permissions do |t|
+      t.column :user_id, :int, :unique => true
```
mishaschwartz, line 4: make this a reference (and a foreign key):

```ruby
t.references :user, null: false, foreign_key: true, index: { unique: true }
```

db/migrate/20200805172907_create_grader_permissions.rb
```diff
@@ -0,0 +1,19 @@
...
+      t.column :manage_extensions, :boolean, :default => false
```
mishaschwartz, line 5: for all of these boolean columns, set `null: false`

spec/controllers/automated_tests_controller_spec.rb
```diff
@@ -240,98 +234,93 @@
...
+  end
+  describe 'an authenticated admin' do
+    let!(:user) { create(:admin) }
+    include_examples 'An authorized admin and grader managing automated testing'
+  end
+
+  describe 'When the grader is allowed to download grades report and populate course summary' do
+    let!(:user) { create(:ta) }
+    before do
+      user.grader_permission.manage_assessments = true
```
mishaschwartz, line 311: it would be nice to change the ta factory so that you could create tas with grader permissions already set (you could pass in the permissions you want when you call `create` instead of having to update the permissions in a `before` block)

spec/controllers/job_messages_controller_spec.rb
```diff
@@ -74,12 +74,13 @@
       end
     end
   end
-  context 'as an admin' do
-    let(:user) { create :admin }
-    it_behaves_like 'job messages'
+  describe 'When user is an authenticated admin' do
+    let!(:user) { create(:admin) }
```
mishaschwartz, line 78: Is there a reason you're using `let!` instead of `let` (here and in a bunch of other places)? It seems like in most cases `let` works just as well. And we prefer to use `let` whenever possible because it lazily creates objects which means it's more efficient in cases where the object is not required.

spec/controllers/notes_controller_spec.rb
```diff
@@ -109,50 +109,53 @@
...
-        expect(flash.empty?).to be_truthy
-        expect(assigns :assignments).not_to be_nil
-        expect(assigns :groupings).not_to be_nil
-      end
+      describe 'When grader is trying to create notes' do
+        it 'be able to create with empty note' do
+          post_as @ta,
+                  :create,
+                  params: { noteable_type: 'Grouping', note: { noteable_id: @grouping.id } }
+          expect((assigns :note)).not_to be_nil
```
mishaschwartz, line 117: just a general comment: 

`assigns(:note)` is preferable to `(assigns :note)`. I'm actually surprised houndbot doesn't complain about this one.

spec/controllers/notes_controller_spec.rb
```diff
@@ -109,50 +109,53 @@
...
-      it 'be able to create with empty note' do
-        post_as @ta,
-                :create,
-                params: { noteable_type: 'Grouping', note: { noteable_id: @grouping.id } }
-        expect(assigns :note).not_to be_nil
-        expect(flash.empty?).to be_truthy
-        expect(assigns :assignments).not_to be_nil
-        expect(assigns :groupings).not_to be_nil
-      end
+      describe 'When grader is trying to create notes' do
```
mishaschwartz, line 112: These changes should be removed since we're not using permissions to manage notes any more

spec/factories/grader_permissions.rb
```diff
@@ -0,0 +1,5 @@
+FactoryBot.define do
+  factory :grader_permission, class: GraderPermission do
+    association :user, factory: :ta
```
mishaschwartz, line 3: change this to `association :ta` when you make the changes to the association in the model

app/models/grouping.rb
```diff
@@ -711,7 +711,7 @@ def self.group_hash_list(hash_list)
   end
 
   def test_runs_instructors(submission)
-    filtered = filter_test_runs(filters: { 'users.type': 'Admin', 'test_runs.submission': submission })
+    filtered = filter_test_runs(filters: { 'users.type' => %w[Admin Ta], 'test_runs.submission' => submission })
```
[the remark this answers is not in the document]
mishaschwartz: Please remove these changes not that you've fixed the issue in automated_tests_helper.rb

app/views/layouts/_sub_sub_menu.html.erb
mishaschwartz: Example of "a few other places" that you should look at changing:

- app/views/layouts/menus/_admin_ta_sub_menu_assessments.html.erb

app/views/submissions/browse.html.erb
ANUSHIYAPRIYA: We are using is_admin props but not is_ta. I removed is_ta props.

app/policies/submission_policy.rb
```diff
@@ -27,4 +28,12 @@ def not_a_student?
   def before_release?
     !record.current_result.released_to_students
   end
+
+  def collect?
+    allowed_to?(:manage_submissions?, with: GraderPermissionPolicy)
+  end
+
+  def allowed_to_run_tests?
```
david-yz-liu, line 36: We shouldn't use "allowed_to" in policy names. (Since we call `allowed_to` on the policy)

Rename this to `run_tests?`

app/policies/grader_permission_policy.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def manage_submissions?
+    user.admin? || user.grader_permission.manage_submissions
+  end
+
+  def manage_assessments?
+    user.admin? || user.grader_permission.manage_assessments
+  end
+
+  def grader_run_tests?
```
david-yz-liu, line 13: Rename this to `run_tests?` (no need to refer to graders here).

app/policies/exam_template_policy.rb
```diff
@@ -0,0 +1,16 @@
...
+  alias_rule :index?, :create?, :destroy?, :update?, :generate?,
+             :split?, :add_fields?, :view_logs?, :download?, :download_generate?,
+             :download_raw_split_file?, :show_cover?, to: :modify?
+
+  def manage?
+    user.admin?
+  end
+
+  # Only admin and authorized grader can manage or modify exam templates
+  def modify?
```
david-yz-liu, line 13: You can remove this rule (and the `alias_rule` above), and have the default rule `manage` refer to the `manage_assessments?` permission.

(Like `AnnotationCategoryPolicy`.)

app/policies/grade_entry_form_policy.rb
```diff
@@ -0,0 +1,22 @@
+# Policy for Grade entry forms controller
+class GradeEntryFormPolicy < ApplicationPolicy
+  default_rule :grading?
+  alias_rule :new?, :create?, :edit?, :update?, to: :manage?
```
david-yz-liu, line 4: In general, make `manage?` the default rule. This matches the previous behaviour, where we had `before_action :authorize_only_for_admin, except: ...`. This will ensure that any new routes that are added are by default restricted to admins and "manage" TAs.

app/policies/peer_review_policy.rb
```diff
@@ -0,0 +1,9 @@
+# Policy for Peer reviews controller
+class PeerReviewPolicy < ApplicationPolicy
+  alias_rule :index?, :populate?, :assign_groups?, to: :assign_reviewers?
```
david-yz-liu, line 3: You can change this one to a `default_rule`, and also rename the policy to the more general `manage?`.

app/policies/assignment_policy.rb
```diff
@@ -53,4 +51,28 @@ def not_yet_in_group?
...
+
+  def manage?
+    allowed_to?(:manage_assessments?, with: GraderPermissionPolicy)
+  end
+
+  def can_run_tests?
+    allowed_to?(:grader_run_tests?, with: GraderPermissionPolicy) || user.student?
+  end
+
+  def view?
```
david-yz-liu, line 63: I liked the name you gave in `GradeEntryFormPolicy`, `grading?`. Please use that here as well.

app/controllers/submissions_controller.rb
```diff
@@ -170,19 +178,25 @@ def populate_file_manager
   end
 
   def manually_collect_and_begin_grading
+    begin
```
david-yz-liu, line 181: It seems like we should have `:manually_collect_and_begin_grading` in the `before_action` you added (line 44). If you're worried about the flash message not appearing, that's something we can solve separately for all the `authorize!` filters, so it's better for now to treat them all the same.

app/controllers/submissions_controller.rb
```diff
@@ -20,14 +20,19 @@ class SubmissionsController < ApplicationController
...
                          :populate_submissions_table,
                          :populate_peer_submissions_table]
   before_action :authorize_for_ta_and_admin,
                 only: [:index,
                        :browse,
                        :manually_collect_and_begin_grading,
+                       :set_result_marking_state,
                        :revisions,
                        :repo_browser,
+                       :collect_submissions,
```
david-yz-liu, line 34: No need to add this here, since it is handled by the `authorize!` call down below. 

In fact, it seems to me like you should also add `run_tests` and `manually_collect_and_begin_grading` to the line 44 list, and remove them from this `only` list.

app/controllers/peer_reviews_controller.rb
```diff
@@ -4,9 +4,12 @@ class PeerReviewsController < ApplicationController
 
   before_action :set_peer_review, only: [:show, :edit, :update, :destroy]
 
-  before_action :authorize_only_for_admin, except: [:list_reviews, :show_reviews, :show_result]
+  before_action :authorize_only_for_admin, except: [:list_reviews, :index, :populate, :assign_groups,
+                                                    :show_reviews, :show_result]
   before_action :authorize_for_user, only: [:list_reviews, :show_reviews, :show_result]
-
+  before_action only: [:index, :populate, :assign_groups] do
```
david-yz-liu, line 10: It's actually okay to authorize all methods here except for the original three exceptions. So you can do the same `before_action except:`, but now call `authorize!` instead of `authorize_only_for_admin`.

app/controllers/notes_controller.rb
```diff
@@ -1,12 +1,16 @@
 class NotesController < ApplicationController
-  before_action do
+  before_action except: :add_note do
```
david-yz-liu, line 2: Please revert the changes in this file. (It's not related to the current grader permissions, and l'm not entirely sure why this change is necessary anyway. You could open up a separate pull request to discuss this change in more detail later.)

app/controllers/automated_tests_controller.rb
```diff
@@ -1,15 +1,14 @@
...
-                     except: [:student_interface,
-                              :get_test_runs_students,
-                              :execute_test_run]
-
-  before_action      :authorize_for_student,
-                     only: [:student_interface,
-                            :get_test_runs_students,
-                            :execute_test_run]
+  before_action do
+    if params[:id].nil?
```
david-yz-liu, line 5: It looks like none of the policies in `AutomatedTestPolicy` require a target (they're all based on the current user). This means you should be able to simplify this code by just doing an `authorize!` call, like in other files.

app/controllers/annotation_categories_controller.rb
```diff
@@ -1,10 +1,8 @@
 class AnnotationCategoriesController < ApplicationController
   include AnnotationCategoriesHelper
-
```
david-yz-liu, line 3: Revert this change.

app/controllers/peer_reviews_controller.rb
```diff
@@ -4,9 +4,12 @@ class PeerReviewsController < ApplicationController
 
   before_action :set_peer_review, only: [:show, :edit, :update, :destroy]
 
-  before_action :authorize_only_for_admin, except: [:list_reviews, :show_reviews, :show_result]
+  before_action :authorize_only_for_admin, except: [:list_reviews, :index, :populate, :assign_groups,
+                                                    :show_reviews, :show_result]
   before_action :authorize_for_user, only: [:list_reviews, :show_reviews, :show_result]
-
```
david-yz-liu, line 9: Keep this blank line (before the `layout` call).

app/assets/javascripts/Components/submission_table.jsx
```diff
@@ -312,25 +311,35 @@ class SubmissionsActionBox extends React.Component {
...
+    if (this.props.can_run_tests) {
+      runTestsButton = (
+        <button onClick={this.props.runTests}
+                disabled={this.props.disabled}
         >
-          {I18n.t('results.set_to_incomplete')}
+          {I18n.t('submissions.run_tests')}
         </button>
       );
-
```
david-yz-liu, line 333: Please do not delete blank lines in existing code. Revert this change, and the one below.

app/assets/javascripts/Components/submission_table.jsx
```diff
@@ -312,25 +311,35 @@ class SubmissionsActionBox extends React.Component {
...
+    incompleteButton = (
+      <button
+        onClick={this.props.incompleteResults}
+        disabled={this.props.disabled}
+      >
+        {I18n.t('results.set_to_incomplete')}
+      </button>
+    );
+
+    if (this.props.can_run_tests) {
```
david-yz-liu, line 333: Put this block of code back in its original place.

app/assets/javascripts/Components/marks_spreadsheet.jsx
```diff
@@ -260,6 +260,7 @@ class RawMarksSpreadsheet extends React.Component {
           ref={(r) => this.actionBox = r}
           toggleRelease={this.toggleRelease}
           showHidden={this.state.show_hidden}
+          is_allowed_to_release_unrelease={this.props.is_allowed_to_release_unrelease}
```
david-yz-liu, line 263: In general. prefer the prop name "can_*" rather than "is_allowed_to_*" since the former is shorter. So for example, here use `can_release_unrelease`. But also in this particular case, I think `can_release` is a better name (basically we're "merging" the permission of release/unrelease into a single name).

You should go through all of the new props you added and do this renaming.

app/assets/javascripts/Components/submission_table.jsx
```diff
@@ -297,7 +297,6 @@ class RawSubmissionTable extends React.Component {
 
 let SubmissionTable = withSelection(RawSubmissionTable);
 SubmissionTable.defaultProps = {
-  is_admin: false,
```
david-yz-liu, line 300: Since this component uses defaultProps, we should specify a new default value for the prop you added. (Set it to `false`)

app/policies/assignment_policy.rb
```diff
@@ -1,16 +1,14 @@
 class AssignmentPolicy < ApplicationPolicy
-
```
david-yz-liu, line 2: Keep this blank line before `def run_tests?`.

app/policies/submission_policy.rb
ANUSHIYAPRIYA: We already have run_tests? in the same policy file. So I renamed this into can_run_tests?

app/controllers/automated_tests_controller.rb
ANUSHIYAPRIYA: Yes, the policies do not require a target but there is no specific model and table for automated tests. We should pass a target explicitly. I tried without the target like **"authorize!, with: :AutomatedTestPolicy"** but it is not looking up the policy correctly.

david-yz-liu:
@ANUSHIYAPRIYA okay everything looks good! I forgot to ask you to update the `Changelog`, so please update that file (and refer to this PR), and then I'll merge this in.

PR #4612, opened 2020-05-27
Toggle annotation category

ignaspa:
## Motivation and Context

For deductive annotations to be configured, it is necessary that an annotation category can be linked to a Flexible Criterion. This pull request should enable admins to edit that setting.

## Your Changes

**Description**:
Changes to the annotation_categories controller update function, and changes to the edit view for the annotation category settings.

- [ ] Other (please specify): 

## Testing

Integration tests for the annotation_category update route.

## Checklist

### DIFF
--- a/app/views/annotation_categories/_annotation_category_edit.html.erb
+++ b/app/views/annotation_categories/_annotation_category_edit.html.erb
@@ -19,12 +19,13 @@
         <%= f.label :annotation_category_name,
                     AnnotationCategory.human_attribute_name(:annotation_category_name) %>
         <%= f.text_field :annotation_category_name, rows: '3' %>
-      </div>
-      <br>
-      <div class='inline-labels'>
         <%= f.label :flexible_criterion_id,
                     t('annotation_categories.flexible_criterion_choice') %>
-        <%= f.collection_select :flexible_criterion_id, FlexibleCriterion.all, :id, :name, { include_blank: true } %>
+        <%= f.collection_select :flexible_criterion_id,
+                                FlexibleCriterion.all.where(assignment: annotation_category.assignment),
+                                :id,
+                                :name,
+                                { include_blank: true } %>
       </div>
 
       <p><%= f.submit t(:save), data: { disable_with: t('working') } %></p>
--- a/app/views/annotation_categories/_annotation_category_edit.html.erb
+++ b/app/views/annotation_categories/_annotation_category_edit.html.erb
@@ -11,6 +11,7 @@
     <% end %>
   </div>
   <h2><%= annotation_category.annotation_category_name %></h2>
+  <% released = annotation_category.marks_released? %>
 
   <fieldset>
     <%= form_for [annotation_category.assignment, annotation_category],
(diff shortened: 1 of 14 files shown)


app/views/annotation_categories/_annotation_category_edit.html.erb
```diff
@@ -20,6 +20,12 @@
                     AnnotationCategory.human_attribute_name(:annotation_category_name) %>
         <%= f.text_field :annotation_category_name, rows: '3' %>
       </div>
+      <br>
+      <div class='inline-labels'>
+        <%= f.label :flexible_criterion_id,
+                    t('annotation_categories.flexible_criterion_choice') %>
+        <%= f.collection_select :flexible_criterion_id, FlexibleCriterion.all, :id, :name, { include_blank: true } %>
```
david-yz-liu, line 27: You should only be passing in the flexible criteria associated with the current assignment.

ignaspa: So I fixed this (and for some reason it auto displays the current flexible_criterion when the page loads), however I wonder if this should be another aspect of validation in the annotation_category?

app/views/annotation_categories/_annotation_category_edit.html.erb
```diff
@@ -19,6 +19,13 @@
         <%= f.label :annotation_category_name,
                     AnnotationCategory.human_attribute_name(:annotation_category_name) %>
         <%= f.text_field :annotation_category_name, rows: '3' %>
+        <%= f.label :flexible_criterion_id,
+                    t('annotation_categories.flexible_criterion_choice') %>
+        <%= f.collection_select :flexible_criterion_id,
+                                FlexibleCriterion.all.where(assignment: annotation_category.assignment),
```
david-yz-liu, line 25: Generally prefer using associations when applicable, so `annotation_category.assignment.flexible_criteria`. Also use a `where` to filter by `ta_visible: true`.

config/locales/views/annotation_categories/en.yml
```diff
@@ -25,3 +25,4 @@ en:
     no_annotations: "There are no annotations in this category"
     none: "There are no annotation categories yet."
     one_time_only: "One Time Only"
+    flexible_criterion_choice: 'Associated Criterion (Flexible type only): '
```
david-yz-liu, line 28: This is the right string, but don't add it here. Instead, add it as an _attribute_ to the _model_ locale file (look at how this is specified for the annotation category name).

spec/models/mark_spec.rb
```diff
@@ -155,6 +155,7 @@
     end
 
     it 'does not take into account deductions related to other criteria' do
+      assignment.reload
```
david-yz-liu, line 158: I'm not sure why this is necessary?

ignaspa: So this is one of the things I was unsure about. Yesterday, when I added the validation that a category cannot belong to a criterion of another assignment, this test failed. When I went through it with byebug it seems that the call to create immediately after it is troublesome, because of the execution. The `assignment` that is declared in the `let` before the examples is referenced and thus created within this other create for a flexible_criterion, and when the category created within the creation of the new flexible criterion does not seem to be aware the the criterion it belongs to is in the list of the assignment's criteria, because the criterion is still being processed.

ignaspa: So this is one of the things I was unsure about. Yesterday, when I added the validation that a category cannot belong to a criterion of another assignment, this test failed. When I went through the test with byebug, the call to `create(:flexible_criterion_with_annotation_category,  assignment: assignment)` immediately after my addition stops the test due to failed validation of an annotation_category (before I added the `reload`) I believe this happens because of the order of execution within that create. In the create, the assignment that is declared in the `let` (before the all the examples, including this test) is referenced and thus created. When the annotation category created as a consequence of the creation of the new flexible criterion validates, the category's predicate that checks the list of criteria it can belong to does not find that the criteria it belongs is in the assignment's list of criteria, because the criterion is still being processed.

ignaspa: So this is one of the things I was unsure about. Yesterday, when I added the validation that a category cannot belong to a criterion of another assignment, this test failed. When I went through the test with byebug, the call to `create(:flexible_criterion_with_annotation_category,  assignment: assignment)` immediately after my addition stops the test due to failed validation of an annotation_category (before I added the `reload`). I believe this happens because of the order of execution within that create. In the create, the assignment that is declared in the `let` (before the all the examples, including this test) is referenced and thus created. When the annotation category created as a consequence of the creation of the new flexible criterion validates, the category's predicate that checks the list of criteria the category can belong to does not find that the criteria the category belongs to is in the assignment's list of criteria. Due to, I believe, the criterion is still being processed in memory alongside the assignment.

PR #4618, opened 2020-05-28
Annotation callback fix

ignaspa:
## Motivation and Context

In creating seed data, it became clear that when an annotation updates a mark with a deduction if it has one, it should query marks with not only the same criterion id, but the same criterion type as well. This is to prevent applying a deduction to the wrong criterion.

## Your Changes

**Description**:
Changing the query in the modify_mark_with_deduction callback.

- [ ] Other (please specify): 

Should I write additional tests? I think I should, since this did not seem to be caught before.
EDIT: done.

## Checklist

### DIFF
--- a/spec/models/annotation_spec.rb
+++ b/spec/models/annotation_spec.rb
@@ -41,7 +41,7 @@ describe Annotation do
       expect(mark.mark).to eq(3.0)
     end
 
-    it 'correctly updates the mark for it\'s criterion\'s mark when a different type of criterion has the same id' do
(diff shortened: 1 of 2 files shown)


spec/models/annotation_spec.rb
```diff
@@ -40,5 +40,23 @@
       result.annotations.find_by(annotation_text: annotation_text).destroy
       expect(mark.mark).to eq(3.0)
     end
+
+    it 'correctly updates the mark for it\'s criterion\'s mark when a different type of criterion has the same id' do
```
david-yz-liu, line 44: `it\'s` should be `'its'

PR #4619, opened 2020-05-29
Fix flaky job tests

mishaschwartz:
## Motivation and Context

Tests in `spec/jobs/submissions_job_spec.rb` fail occasionally but only on travis for some reason. This is an attempt to fix the flakiness.

## Your Changes

**Description**:

Reload everything! 

The `SubmissionsJob` is somewhat unique in that its `perform` method takes actual objects as arguments as opposed to object ids. This may make it susceptible to keeping saved but not reloaded changes in memory when `perform_now` is called in tests (as opposed to other tests that take ids and so must reload objects whether `perform_now` or `perform_later` are called).

Why this is only happening occasionally and only on travis (can't reproduce locally) ... I have no idea. 

## Testing

- ran tests locally but couldn't reproduce flakiness
- ran tests on travis and reran them multiple times to observe the flakiness
- re-ran tests on travis multiple times after making the changes to confirm that the flakiness was no longer happening. After 5+ successful runs I made this PR. This is NOT a guarantee that it won't fail (flakiness is flaky) but I'm pretty confident that this has solved the issue 🤞 .

## Checklist

- [ ] I have fixed any Hound bot comments (check after opening pull request).
- [ ] I have verified that the TravisCI tests have passed (check after opening pull request).
- [ ] I have reviewed the test coverage changes reported on Coveralls (check after opening pull request).

### DIFF
--- a/spec/jobs/submissions_job_spec.rb
+++ b/spec/jobs/submissions_job_spec.rb
@@ -89,8 +89,8 @@ describe SubmissionsJob do
     end
   end
   context 'when applying a late penalty' do
-    let!(:period) { create :period, submission_rule: submission_rule, hours: 2 }
     before :each do
+      create :period, submission_rule: submission_rule, hours: 2
       groupings.each do |g|
         submit_file_at_time(g.assignment, g.group, 'test', (g.due_date + 1.hour).to_s, 'test.txt', 'aaa')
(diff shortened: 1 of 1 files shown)


mishaschwartz:
😠

PR #4658, opened 2020-06-09
Fix file download bugs for graders downloading submission files

mishaschwartz:
## Motivation and Context

Graders can't view job messages which are now required to download zip files.
Also, file downloads for individual submissions fail if a file in a repository is not a submission file (for example this happens when using git repos when downloading .gitkeep files)

## Your Changes

**Description**:

- Allow graders to view job messages (and update tests accordingly)
- Let non-submission files be written as empty files in the zip archive (this is a temporary fix until we get rid of submission files model entirely in favour of using the repo as the source-of-truth for submission files). 

## Checklist

### DIFF
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -498,7 +498,7 @@ class ResultsController < ApplicationController
         revision = repo.get_revision(revision_identifier)
         repo.send_tree_to_zip(assignment.repository_folder, zip_file, zip_name, revision) do |file|
           submission_file = files.find_by(filename: file.name, path: file.path)
-          submission_file.retrieve_file(params[:include_annotations] == 'true' && !submission_file.is_supported_image?)
+          submission_file&.retrieve_file(params[:include_annotations] == 'true' && !submission_file.is_supported_image?)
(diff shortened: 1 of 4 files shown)


app/controllers/results_controller.rb
```diff
@@ -498,7 +498,7 @@ def download_zip
         revision = repo.get_revision(revision_identifier)
         repo.send_tree_to_zip(assignment.repository_folder, zip_file, zip_name, revision) do |file|
           submission_file = files.find_by(filename: file.name, path: file.path)
-          submission_file.retrieve_file(params[:include_annotations] == 'true' && !submission_file.is_supported_image?)
+          submission_file&.retrieve_file(params[:include_annotations] == 'true' && !submission_file.is_supported_image?)
```
david-yz-liu, line 501: Do we need to modify `send_tree_to_zip` to handle a `nil` value? You've probably tested this, I just wanted to double-check.

mishaschwartz: `nil` values are written as the empty string which in this case in the intended outcome

PR #4665, opened 2020-06-12
Timed assignments

mishaschwartz:
## Motivation and Context

This PR creates a new assignment type, the Timed Assignment. A timed assignment has the following properties:

- a start time: a student can start the assignment after this date
- an end time: as student must submit their assignment before this date
- a duration: a time period in hours and minutes. Once the student has started the assignment they must finish within this time period.

Start times, and end times can be set differently for different sections.

An instructor can choose to extend the duration in two ways:

- submission policy: a student may submit after the duration has finished but a penalty will be added to their final mark (or grace tokens will be added)
- group extension: an individual group may have extra time added to their duration. 

## Your Changes

**Description**:

- added columns to three database tables to keep track of start time, duration, end time, etc.
- an assignment is timed if its `is_timed` value in the `assignment_properties` table is `true`
- made changes to the assignment creation/update form to allow for the creation/update of a timed assignment
- updated the student's assignment view (`GET show`) for timed assignments to show start time, end time, duration, etc. as well as a button to start the assignment (if available) and information about how long they have to complete the assignment (if started already).
- updated the repository permission update code to only allow students to have access to their repositories for an assignment once they have started working on it (this also means that repositories for timed assignments cannot be reused for other assignments). 

## Checklist

### Additional checklist:

- [x] write tests for new/updated models, controllers, routes, etc.
- [x] add timed assignment example to seed data
- [x] update/add docstrings for all new methods
- [x] add last few internationalization strings for error messages

- Instructor info:
    - how to set up a timed assignment
        - start time and end time
        - duration
    - how extensions and late policies work with timed assignments
        - they get added to the due date (start time + duration) but a student cannot submit after the end date under any circumstances
- Student info:
    - how to interpret the info on the assignment page
    - when you can and cannot start a timed assignment
    - how to start and complete a timed assignment


app/assets/javascripts/Components/Helpers/table_helpers.jsx
```diff
@@ -16,6 +16,19 @@ export function dateSort(a, b) {
   return (Date.parse(a) || 0) > (Date.parse(b) || 0);
 };
 
+export function durationSort(a, b) {
+  /** Sort values as a duration in weeks, days, hours, etc. */
+  a = [a.weeks || -1, a.days || -1, a.hours || -1, a.minutes || -1, a.seconds || -1];
+  b = [b.weeks || -1, b.days || -1, b.hours || -1, b.minutes || -1, b.seconds || -1];
+  if (a < b) {
+    return 1
```
david-yz-liu, line 24: Please add semicolons here.

app/assets/javascripts/Components/Modals/extension_modal.jsx
```diff
@@ -83,47 +87,54 @@ class ExtensionModal extends React.Component {
...
+          name={time}
+          onChange={this.handleModalInputChange}
+        /> {this.getLabel(time)}
+      </label>
+    )
+  };
+
+  renderExtraInfo = () => {
+    if (!!this.props.extra_info) {
+      return <div className={'modal-container'}>{this.props.extra_info}</div>
```
david-yz-liu, line 119: Please add a semicolon here.

app/assets/javascripts/Components/Modals/extension_modal.jsx
```diff
@@ -83,47 +87,54 @@ class ExtensionModal extends React.Component {
...
+        <input
+          type="number"
+          value={this.state[time]}
+          max={time === 'minutes' ? 60 : 999}
+          min={0}
+          name={time}
+          onChange={this.handleModalInputChange}
+        /> {this.getLabel(time)}
+      </label>
+    )
```
david-yz-liu, line 114: https://github.com/pyta-uoft/pyta/pulls

app/controllers/assignments_controller.rb
```diff
@@ -50,8 +52,7 @@ def show
         end
         @grouping = @current_user.accepted_grouping_for(@assignment.id)
       end
-    end
-    unless @grouping.nil?
+    else
```
david-yz-liu, line 55: I'm not sure why you changed this, my understanding is `@grouping` can be assigned on line 53.

app/models/assignment_properties.rb
```diff
@@ -51,6 +51,47 @@ class AssignmentProperties < ApplicationRecord
 
   validate :minimum_number_of_groups
 
+  attribute :duration, :duration
```
david-yz-liu, line 54: This is interesting, what is the purpose of this line?

app/models/grouping.rb
```diff
@@ -489,15 +489,22 @@ def missing_assignment_files(revision = nil)
...
+  #
+  # If the assignment is a timed assignment, the due date is this grouping's start time plus
+  # the duration plus any extension. If this due date is after the assignment due date or if
+  # the grouping hasn't started the timed assignment yet, the assignment due date is returned
+  # instead.
   def due_date
-    if use_section_due_date?
-      assignment_due_date = assignment.section_due_dates.find_by(section_id: inviter.section.id).due_date
+    a_due_date = assignment.section_due_dates.find_by(section_id: inviter&.section)&.due_date || assignment.due_date
+    extension_time_delta = (extension&.time_delta || 0)
```
david-yz-liu, line 499: You can remove the parentheses around the RHS

app/models/grouping.rb
```diff
@@ -524,6 +531,14 @@ def past_collection_date?
     collection_date < Time.current
   end
 
+  # Return the duration of this grouping's assignment or the time between now and the collection_date whichever is
+  # less. If the collection date has already passed, return 0 seconds.
+  def adjusted_duration
+    add = assignment.submission_rule.periods.pluck(:hours).sum.hours + (extension&.time_delta || 0)
```
david-yz-liu, line 537: I'm thinking the submission rule periods shouldn't be counted as part of the reported duration, but rather listed separately.

app/views/assignments/_timed_due_info.html.erb
```diff
@@ -0,0 +1,40 @@
+<% due_date = @assignment.section_due_date(@current_user&.section)
+   start_time = @assignment.section_start_time(@current_user&.section) %>
+<% duration = @grouping.nil? ?
+                @assignment.assignment_properties.adjusted_duration(due_date, [Time.current, start_time].max) :
+                @grouping.adjusted_duration %>
+<h3><%= I18n.t('activerecord.models.timed_assignment.one') %></h3>
+<p>
+  <strong><%= I18n.t('assignments.timed.start_time_instructions') %></strong>
+  <%= @assignment.section_start_time(@current_user&.section).strftime('%Y-%m-%d %l:%M %p') %>
```
david-yz-liu, line 9: In general in the views it would be preferable to use `I18n.l` rather than `strftime` I think.

app/views/assignments/_timed_due_info.html.erb
```diff
@@ -0,0 +1,40 @@
...
+  <%= due_date.strftime('%Y-%m-%d %l:%M %p') %>
+</p>
+<% if @grouping&.start_time.nil? %>
+  <% if Time.current <= due_date && Time.current >= start_time %>
+    <p>
+      <% parts = AssignmentProperties.duration_parts(duration)
+         duration_string = ["#{parts[:hours]} #{I18n.t('durations.hours', count: parts[:hours])}",
+                            "#{parts[:minutes]} #{I18n.t('durations.minutes', count: parts[:minutes])}"].join(', ')
+      %>
+      <%= I18n.t('assignments.timed.before_start_instructions', duration_string: duration_string)%>
```
david-yz-liu, line 21: Insert a space before the `%>`

app/views/assignments/_timed_due_info.html.erb
```diff
@@ -0,0 +1,40 @@
...
+    </p>
+    <%= form_with url: start_timed_assignment_assignment_path(@assignment.id), method: 'put', local: true do |f| %>
+      <%= f.hidden_field :grouping_id, value: @grouping.id %>
```
david-yz-liu, line 24: There's something to investigate here. The condition on line 14 allows for the case that `@grouping` is `nil`, but here this isn't allowed. Perhaps this form needs to be guarded by an `if @grouping.nil?` and separate instructions given to students to form a (valid?) grouping first.

db/migrate/20200605183957_add_start_time_access_to_section_due_dates.rb
```diff
@@ -0,0 +1,5 @@
+class AddStartTimeAccessToSectionDueDates < ActiveRecord::Migration[6.0]
+  def change
+    add_column :section_due_dates, :start_time, :datetime, null: true
```
david-yz-liu, line 3: Should this also have a `default: nil`?

config/locales/views/assignments/en.yml
```diff
@@ -28,6 +29,18 @@ en:
       description: "Starter code files you upload will be automatically added to every new group's repositories."
       overwrite: "Overwrite group files"
       title: "Starter Code"
+    timed:
+      start_button: "Start"
+      end_time: "End time"
+      start_time_instructions: "You may start any time after: "
+      end_time_instructions: "You must complete the assignment before the final due date: "
```
david-yz-liu, line 36: For this section use the term "assessment" instead of "assignment".

app/controllers/assignments_controller.rb
```diff
@@ -157,7 +158,7 @@ def edit
         flash_now(:notice, t('assignments.due_date.past_due_date_notice') + past_date.join(', '))
       end
     end
-
+    @assignment_duration_parts = @assignment.assignment_properties.duration_parts if @assignment.is_timed
```
david-yz-liu, line 161: I prefer not creating another instance variable here. We setup a proxy on `assignment_properties`, so you'll be able to do `@assignment.duration_parts` where this is used in the views. Also, the only place I could see this variable being used is guarded by an `if @assignment.is_timed` already.

app/assets/javascripts/Components/Modals/extension_modal.jsx
mishaschwartz: Can you explain this please?

app/controllers/assignments_controller.rb
mishaschwartz: I'm not sure either

app/models/assignment_properties.rb
mishaschwartz: This tells the model that the `duration` attribute (which is stored as a string in the database) should be interpreted as a [duration](https://api.rubyonrails.org/classes/ActiveSupport/Duration.html) data type, which we have registered previously in `config/initializers/types.rb` for use with the `extensions` model.

app/models/grouping.rb
mishaschwartz: Ok I see that. I think we should discuss separately exactly what info we should give the students in their view. That will inform what these various methods do.

app/views/assignments/_timed_due_info.html.erb
mishaschwartz: Oh, good catch. I rewrote this view so many times now, I definitely just missed this.

db/migrate/20200605183957_add_start_time_access_to_section_due_dates.rb
mishaschwartz: I don't see why not. The current behaviour is for datetimes with `null: true` to default to `nil` but I can't find any rails documentation to back that up (though this is the [correct behaviour for postgres](https://www.postgresql.org/docs/10/ddl-default.html)) so maybe better safe than sorry if that gets changed from under us.

app/assets/javascripts/Components/Modals/extension_modal.jsx
david-yz-liu: Whoops sorry, I think I meant to copy-and-paste a comment but didn't copy properly.

There should be a semicolon on this line.

spec/controllers/assignments_controller_spec.rb
```diff
@@ -415,100 +436,276 @@
     end
   end
 
+  context '#new' do
+    context 'as an admin' do
+      shared_examples 'assignment_new_success' do
+        it 'responds with a success' do
+          expect(response).to have_http_status :success
```
david-yz-liu, line 443: I've been trying to use [`assert_response`](https://apidock.com/rails/v6.0.0/ActionDispatch/Assertions/ResponseAssertions/assert_response) more in testing status codes.

app/views/assignments/_read.html.erb
```diff
@@ -33,11 +37,13 @@
...
   <% unless @assignment.submission_rule.type == "NoLateSubmissionRule" || @grouping&.extension&.apply_penalty == false %>
     <h3><%= SubmissionRule.model_name.human.capitalize %></h3>
-    <p><%= @assignment.submission_rule.class.human_attribute_name(:description) %></p>
+    <p><%= @assignment.submission_rule.class.human_attribute_name(:description) unless @assignment.is_timed %></p>
 
     <ul>
       <% if @grouping.nil? %>
         <li><%= t('grace_period_submission_rules.no_group_yet') %></li>
+      <% elsif @assignment.is_timed? %>
+        <li><%= t('assignments.timed.full_duration_warning_html') %></li>
```
david-yz-liu, line 46: This warning is no longer relevant (there's no more notion of a "final due date"). I think we *do* want to show the "due date with penalty" (in the else block), but only `unless @assignment.is_timed? && @grouping.start_time.nil?`.

app/views/assignments/_timed_due_info.html.erb
```diff
@@ -0,0 +1,10 @@
+<% due_date = @assignment.section_due_date(@current_user&.section) %>
+<h3><%= I18n.t('activerecord.models.timed_assignment.one') %></h3>
+<p>
+  <strong><%= I18n.t('assignments.timed.start_time_instructions') %></strong>
+  <%= I18n.l @assignment.section_start_time(@current_user&.section) %>
+<p>
+  <strong><%= I18n.t('assignments.timed.end_time_instructions') %></strong>
+  <%= I18n.l due_date %>
+</p>
+<%= render partial: 'timed_due_message', locals: { assignment: @assignment, grouping: @grouping, show_button: true } %>
```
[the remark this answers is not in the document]
david-yz-liu: (Of course, also showing any duration extension.)

config/locales/views/assignments/en.yml
```diff
@@ -28,6 +29,22 @@ en:
...
       overwrite: "Overwrite group files"
       title: "Starter Code"
+    timed:
+      start_button: "Start"
+      end_time: "End time"
+      start_time_instructions: "You may start any time after: "
+      end_time_instructions: "You must complete this timed assessment before the final due date: "
+      past_due_date: "You have completed this timed assessment."
+      past_end_time: "You cannot start this timed assessment at this time."
+      before_start_instructions: "Once this timed assessment has been started, you will have %{duration_string} to complete it without penalty."
```
david-yz-liu, line 39: Note: this message is currently inconsistent with the `duration_string` passed to it, which *includes* penalty periods. Given the other comments I've made, it might be better to rework/remove this string altogether, but I'm drawing your attention to it here in case we want to discuss further.

config/locales/models/assignments/en.yml
```diff
@@ -5,6 +5,8 @@ en:
         one: "Assignment"
       scanned_assignment:
         one: "Scanned Assignment"
+      timed_assignment:
+        one: "Timed Assignment"
```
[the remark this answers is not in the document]
mishaschwartz: I think I was just trying to be consistent with the other assignment types here. Do you think this deserves its own PR where we change them all? I'm not sure I like that "Assignment" and "Scanned Assignment" are still there if we're going to change the one.

spec/controllers/assignments_controller_spec.rb
mishaschwartz: I can do that but I prefer to use rspec matchers for rspec tests as much as possible because in general it gives better error messages. Also I'm pretty sure `have_http_status` calls `assert_response` so the mechanism should be the same.

david-yz-liu: Okay that sounds good, thanks for letting me know.

config/locales/models/assignments/en.yml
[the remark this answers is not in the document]
david-yz-liu: That's a good point. Instructors use "Assignment" to connote a specific type of "Assessment". (You'll notice that we chose "Assessment" as the superclass when merging Assignment and GradeEntryForm.)

For now I'd like you to change "Timed Assignment" to "Timed Assessment", and I would accept a future PR that changes "Scanned Assignment" to "Scanned Assessment", but not renaming "Assignment" to "Assessment".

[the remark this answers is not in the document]
mishaschwartz: makes sense

config/locales/views/assignments/en.yml
mishaschwartz: I'm still thinking that the "Once this assessment has been started" section is good information even if we're now repeating the duration numbers themselves later on so I'm inclined to leave this.

I've fixed the value though so that it doesn't include penalty periods. Thanks for catching that.

david-yz-liu: Hmm yes okay, I'm good with keeping this string.

PR #4680, opened 2020-06-17
Annotation category upload error handling

ignaspa:
## Motivation and Context

Previous error handling for uploading an annotation category had a bug, and this pull request minimally augments the error handling to alert the use of a missing criterion for deductive annotations.

## Your Changes

**Description**:
Changing the flash messages for upload issues in the annotation categories controller.
Adding a new key for a missing criterion in the locales.

## Testing

Testing to see if popups in UI were as expected.

## Checklist

### DIFF
--- a/app/controllers/annotation_categories_controller.rb
+++ b/app/controllers/annotation_categories_controller.rb
@@ -199,19 +199,23 @@ class AnnotationCategoriesController < ApplicationController
         flash_message(:success, result[:valid_lines]) unless result[:valid_lines].empty?
       elsif data[:type] == '.yml'
         successes = 0
-        annotation_line = 0
         data[:contents].each do |category, category_data|
           if category_data.is_a?(Array)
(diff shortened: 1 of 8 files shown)


ignaspa:
@david-yz-liu I just changed this into a draft again because of a thought I had. I was wondering, perhaps as a part of this PR it would make sense to change add_by_row such that if there is some issue in the execution of said method, whatever the method had been in the process of creating gets cleaned up/deleted. 

Currently, if there are any issues after the category is created but still within the add_by_row body, the category and any annotations created for the category before the issue was raised still exist after the raising. An example of this might be when there is an invalid deduction for one of the annotation texts that causes an error. The annotation category, and the annotation texts processed before the problematic one, still exist.

This could lead to partial uploads that may confuse the admin or increase the likelihood of annotations being uploaded more than once (as has happened to me when testing in the UI).

david-yz-liu:
@ignaspa yes that's a good idea, but I prefer waiting until this is all merged in and making the change on master.

PR #4695, opened 2020-06-24
Annotation Stats Panel

ignaspa:
## Motivation and Context

Instructors would ideally be able to see how many times an annotation has been used per assignment, to which submissions said annotation has been applied along with whom applied.

## Your Changes

**Description**:
Adding a new expandable react component for each annotation in the annotation settings page.
Adding a new route to get the data for the annotation.

## Checklist

Change the wiki page on [annotation settings](https://github.com/MarkUsProject/Markus/wiki/Instructor-Guide:-Annotations-Tab) to include a section on this feature.


app/assets/javascripts/Components/annotation_stat_panel.jsx
```diff
@@ -0,0 +1,99 @@
...
+          <div>
+            {row.original['users.first_name'] + ' ' + row.original['users.last_name']}
+          </div>
+        );
+      }
+    },
+    {
+      Header: I18n.t('activerecord.models.submission.one'),
+      accessor: 'groups.group_name',
+      aggregate: (vals, pivots) => {
```
david-yz-liu, line 43: For a single-expression lambda, you can do `(vals, pivots) => pivots.length`.

But also, this is a place where some extra text would be nice. E.g., `'Used {pivots.length} times'`

app/controllers/annotation_categories_controller.rb
```diff
@@ -204,6 +204,11 @@ def upload
     redirect_to assignment_annotation_categories_path(assignment_id: @assignment.id)
   end
 
+  def get_annotation_text_stats
```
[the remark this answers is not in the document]
david-yz-liu: Yeah, here you can just do `annotation_text_stats`

app/controllers/annotation_categories_controller.rb
```diff
@@ -204,6 +204,11 @@ def upload
...
+    annotation_stats = AnnotationText.find_by(id: params[:annotation_text_id]).get_stats
```
david-yz-liu, line 208: Use `find` instead of `find_by`. You can do this when you're searching by `id`, and `find` will raise an error if the record isn't found.

app/models/annotation_text.rb
```diff
@@ -14,4 +14,20 @@ def escape_content
     content.gsub('\\', '\\\\\\') # Replaces '\\' with '\\\\'
            .gsub(/\r?\n/, '\\n')
   end
+
+  def get_stats
+    applications = self.annotations.joins(result: { grouping: :group })
```
david-yz-liu, line 19: Add in a `TODO` to simplify the second `joins` after we make `creator` a non-polymorphic association.

app/views/annotation_categories/_annotation_text.html.erb
```diff
@@ -31,6 +31,13 @@
         <%= annotation_text.last_editor.user_name %>
       <% end %>
     </div>
+    <p>
```
david-yz-liu, line 34: Okay this is probably the biggest change. I want you to move the logic for showing/hiding the table into the React component itself. I.e., move the button into the component, and register some state for either showing or hiding the table. This means you can put the "Annotation count: ___" text above or alongside the button, which is nice. (Could rename the button to "Show Details" or "Show Submissions".)

The thing to watch out for is that the data for the table should only be fetched when the table is first shown, which is now different than `componentDidMount`. This means that if the table is toggled repeatedly, the data won't be reloaded.

app/assets/javascripts/Components/annotation_stat_panel.jsx
```diff
@@ -0,0 +1,119 @@
...
+    this.state = {
+      applications: null,
+      details: false,
+      num_times_used: props.num_used
+    }
+  }
+
+  toggle = () => {
+    if (this.state.applications === null) {
+      this.fetchData()
```
david-yz-liu, line 18: Add semicolons here (and elsewhere)

app/assets/javascripts/Components/annotation_stat_panel.jsx
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  columns = [
+    {
+      Header: I18n.t('annotations.used_by'),
+      accessor: row => '(' + row['user_name'] + ') ' + row['first_name'] + ' ' + row['last_name'],
+      id: 'user',
+      minWidth: 200,
+      maxWidth: 200,
+      resizeable: false,
```
david-yz-liu, line 31: I wouldn't make this column resizeable (possibly long names)

app/assets/javascripts/Components/annotation_stat_panel.jsx
```diff
@@ -0,0 +1,119 @@
...
+      PivotValue: ({value}) => value,
+      filterMethod: (filter, row) => {
+        return row[filter.id].toLowerCase().includes(filter.value.toLowerCase());
+      }
+    },
+    {
+      Header: I18n.t('activerecord.models.submission.one'),
+      accessor: row => row['group_name'],
```
david-yz-liu, line 39: You can now make the accessor `'group_name'` and omit the `id` field to have `'group_name'` be the default id.

config/locales/views/annotations/en.yml
```diff
@@ -1,8 +1,12 @@
 en:
   annotations:
+    count: 'Number of times used: '
     help_message: "Click on an annotation category to add annotations."
     last_edited_by: "Last Edited By:"
     text: "Text"
+    usage: 'Usage Details'
+    used_by: 'Used By'
+    used_times: "Used %{number} times"
```
david-yz-liu, line 9: Strings with counts are always a bit awkward (think "Used 1 times"). The standard way to resolve this is to have nested keys under `used_times`. Search "pluralization" at https://github.com/fnando/i18n-js.

app/assets/javascripts/Components/annotation_stat_panel.jsx
```diff
@@ -0,0 +1,119 @@
...
+      id: 'group',
+      aggregate: (vals, pivots) => I18n.t('annotations.used_times', {'number': pivots.length}),
+      sortable: false,
+      Aggregated: row => (
```
david-yz-liu, line 43: I don't think you need the `<span>` tag, so I'd just do `row => '(' + row.value ')'`

app/assets/javascripts/Components/annotation_stat_panel.jsx
```diff
@@ -0,0 +1,119 @@
...
+        console.log(row[filter.id].toLowerCase().includes(filter.value.toLowerCase()))
+        if (row._subRows === undefined){
+          return row[filter.id].toLowerCase().includes(filter.value.toLowerCase())
+        } else {
+          return true;
+        }
+      },
+      Cell: row => {
+        return (
+          <div>
```
david-yz-liu, line 59: You don't need this `<div>`

app/assets/javascripts/Components/annotation_stat_panel.jsx
```diff
@@ -0,0 +1,119 @@
...
+              row.original['submission_id'],
+              row.original['result_id']
+            )}
+            >
+              {row.original['group_name']}
+            </a>
+          </div>
+        );
+      },
+      maxWidth: 200,
```
david-yz-liu, line 71: You don't need a `maxWidth`, you can allow this column to fill the remaining space. (Maybe you need a `minWidth`, not sure)

app/assets/javascripts/Components/annotation_stat_panel.jsx
```diff
@@ -0,0 +1,119 @@
...
+      id: 'group',
+      aggregate: (vals, pivots) => I18n.t('annotations.used_times', {'number': pivots.length}),
+      sortable: false,
+      Aggregated: row => (
+        <span>
+          {'(' + row.value + ')'}
+        </span>
+      ),
+      filterMethod: (filter, row) => {
+        console.log(row[filter.id].toLowerCase())
```
david-yz-liu, line 49: Don't forget to remove these

app/assets/javascripts/Components/annotation_stat_panel.jsx
```diff
@@ -0,0 +1,119 @@
...
+        console.log(row[filter.id].toLowerCase().includes(filter.value.toLowerCase()))
+        if (row._subRows === undefined){
```
david-yz-liu, line 51: Include a space before the `{`

app/assets/javascripts/Components/annotation_stat_panel.jsx
```diff
@@ -0,0 +1,119 @@
...
+  fetchData = () => {
+    $.ajax({
+      url: Routes.annotation_text_stats_assignment_annotation_categories_path(this.props.assignment_id),
+      data: {
+        annotation_text_id: this.props.annotation_id
+      },
+      dataType: 'json'
+    }).then(res => {
+      console.log(res)
+      this.setState({applications: res['uses'], num_times_used: res['num_times_used'], details: true});
```
david-yz-liu, line 84: You don't need `num_times_used` any more, it's passed as a prop when the component is created. (You should remove this from the controller/model/tests too.)

app/assets/javascripts/Components/annotation_stat_panel.jsx
```diff
@@ -0,0 +1,119 @@
...
+
+  fetchData = () => {
+    $.ajax({
+      url: Routes.annotation_text_stats_assignment_annotation_categories_path(this.props.assignment_id),
+      data: {
+        annotation_text_id: this.props.annotation_id
+      },
+      dataType: 'json'
+    }).then(res => {
+      console.log(res)
```
david-yz-liu, line 83: Remove this line

app/assets/javascripts/Components/annotation_stat_panel.jsx
david-yz-liu: And since you're doing that, you can simply have the server respond with the "applications"; no need to have a `'uses'` key at all (since there would only be one key, it's an unnecessary level of indirection).

app/assets/javascripts/Components/annotation_stat_panel.jsx
```diff
@@ -0,0 +1,119 @@
...
+import { render } from 'react-dom';
+import ReactDOM from 'react-dom';
+
+class AnnotationStatPanel extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      applications: null,
+      details: false,
+      num_times_used: props.num_used
```
david-yz-liu, line 12: Don't make this state; simply use `this.props.num_used` throughout.

app/assets/javascripts/Components/annotation_stat_panel.jsx
```diff
@@ -0,0 +1,119 @@
...
+        <ReactTable
+          className='auto-overflow'
+          data={this.state.applications}
+          columns={this.columns}
+          filterable
+          pivotBy={['user']}
+        />);
+      return (<fieldset>
+        <p>{I18n.t('annotations.count') + this.state.num_times_used}</p>
+        <a onClick={() => this.toggle()} className='button usage-details'>
```
david-yz-liu, line 100: Ah, you should be able to do `onClick={this.toggle}` rather than `() => this.toggle()`.

david-yz-liu: Also, don't introduce a separate `usage-details` class for styling. Instead, wrap this element in a `<p>` tag.

app/assets/javascripts/Components/annotation_stat_panel.jsx
```diff
@@ -0,0 +1,119 @@
...
+      let annotation_table =  (
+        <ReactTable
+          className='auto-overflow'
+          data={this.state.applications}
+          columns={this.columns}
+          filterable
+          pivotBy={['user']}
+        />);
+      return (<fieldset>
+        <p>{I18n.t('annotations.count') + this.state.num_times_used}</p>
```
david-yz-liu, line 99: Pull out both the `p` tag here and the button below into separate variables; this allows you to use those variables in both cases of the if statement, rather than duplicating the code.

app/controllers/annotation_categories_controller.rb
```diff
@@ -204,6 +204,11 @@ def upload
     redirect_to assignment_annotation_categories_path(assignment_id: @assignment.id)
   end
 
+  def annotation_text_stats
+    annotation_stats = AnnotationText.find(params[:annotation_text_id]).stats
+    render json: annotation_stats
```
david-yz-liu, line 209: You don't need a temporary variable here, you can do `render json: AnnotationText....`

app/controllers/annotation_categories_controller.rb
```diff
@@ -204,6 +204,11 @@ def upload
     redirect_to assignment_annotation_categories_path(assignment_id: @assignment.id)
   end
 
+  def annotation_text_stats
```
david-yz-liu, line 207: Based on the data that's actually being returned, `annotation_text_uses` is probably a better name for the route. In fact, you should probably change "stats" -> "uses" throughout.

app/models/annotation_text.rb
```diff
@@ -14,4 +14,23 @@ def escape_content
...
+                       .joins('INNER JOIN users ON annotations.creator_id = users.id')
+                       .order('groups.group_name')
+                       .pluck_to_hash('results.id AS result_id',
+                                      'groupings.assessment_id AS assignment_id',
+                                      'results.submission_id AS submission_id',
+                                      'groups.group_name AS group_name',
+                                      'users.first_name AS first_name',
+                                      'users.last_name AS last_name',
+                                      'users.user_name AS user_name')
+    stats = {
```
david-yz-liu, line 30: There's no need for a local variable here. Simply return the hash.

app/views/annotation_categories/_annotation_text.html.erb
```diff
@@ -52,3 +53,15 @@
     <% end %>
   </div>
 </li>
+
+<script>
+  statsPanel = document.getElementById('annotation-text-' + <%= annotation_text.id %> + '-stat-panel');
```
david-yz-liu, line 58: Wrap this in an IIFE to prevent `statsPanel` from entering the global scope.

ignaspa:
@david-yz-liu should the changelog file be updated for this feature?

PR #4700, opened 2020-06-25
Annotation table sorting

ignaspa:
## Motivation and Context

After sourcing feedback from stakeholders, it seems the annotation table would preferably be sorted by the criterion of the deduction first, then the filename and annotation number, so that the relevant deductions are at the top of the table.

image

## Your Changes

**Description**:

- Changing the sort of the result's annotation table to first account for criteria with deductions.
- Ensuring the table does not show the deduction column if deductions are 0.

## Checklist

### DIFF
--- a/app/assets/javascripts/Components/Result/annotation_table.jsx
+++ b/app/assets/javascripts/Components/Result/annotation_table.jsx
@@ -150,7 +150,8 @@ export class AnnotationTable extends React.Component {
     if (this.props.detailed) {
       allColumns = allColumns.concat(this.detailedColumns);
     }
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/Components/Result/annotation_table.jsx
```diff
@@ -134,6 +134,21 @@ export class AnnotationTable extends React.Component {
         );
       }
     },
+    sortMethod: (a, b) => {
+      if (a === '') {
+        return 1;
+      }
+      else if (b === '') {
```
david-yz-liu, line 141: Move each `else if` onto the previous line

PR #4738, opened 2020-07-14
Student ui

ignaspa:
## Motivation and Context

The student experience on MarkUs has a variety of aspects that can be improved. This pull request will address some improvements that can be made, including the following:

1. Combining the display of marks spreadsheets and assignments in the landing page (since students need not appreciate the admin level differences in these evaluations.
2. Highlighting upcoming assessments before past assessments.

## Your Changes

**Description**:

- Changes to the list of assignments displayed on the home page for students.

## Checklist


ignaspa:
@david-yz-liu Waiting for Travis and Coveralls, but I am unsure about the tests for the `upcoming` method in the case the method is called from the assignment model. I covered the cases where the student is in a grouping and not in a grouping, but I did not write any tests that check for specific section due dates, because the factory setup was a little confusing. Let me know if I need to do this, then I will try tomorrow.

app/views/assignments/_list.html.erb
```diff
@@ -1,67 +1,24 @@
-<% assignments = Assignment.where(is_hidden: false).includes(:pr_assignment).order(:id) %>
-
-<% if assignments.empty? %>
+<% if assessments.empty? %>
   <p><%= t('assignments.none') %></p>
```
ignaspa, line 2: Perhaps this translation should be changed from  "There are currently no assignments." to something that is more inclusive of both grade entry forms and assignments, upcoming and past. Perhaps "No assessments to show"

app/views/grade_entry_forms/_list.html.erb
```diff
@@ -14,41 +14,7 @@
...
-                <% else %>
-                  <%= t('grade_entry_forms.students.detailed_marks_message') %>
-                <% end %>
-              <% elsif grade_entry_form.due_date.nil? ||
-                       grade_entry_form.due_date < Time.zone.now.to_date %>
-                <%= t('grade_entry_forms.students.no_results') %>
-              <% end %>
-            </td>
-          </tr>
+          <%= render partial: 'grade_entry_forms/row', locals: { grade_entry_form: grade_entry_form} %>
```
ignaspa, line 17: Add a space between form and }

app/views/assignments/_list.html.erb
david-yz-liu: Nice catch, let's go with "No assessments to display."

app/views/grade_entry_forms/_list.html.erb
```diff
@@ -14,41 +14,7 @@
       </thead>
       <tbody>
         <% grade_entry_forms.each do |grade_entry_form| %>
-          <% gef_text = "#{h(grade_entry_form.short_identifier)}: #{h(grade_entry_form.description)}" %>
```
david-yz-liu, line 17: Hmm, is this partial even used anywhere any more? If not, we could delete it altogether.

app/views/assignments/_list.html.erb
```diff
@@ -1,67 +1,24 @@
-<% assignments = Assignment.where(is_hidden: false).includes(:pr_assignment).order(:id) %>
```
david-yz-liu, line 1: This entire file should be moved to `assessments/_list.html.erb` I think.

app/views/assignments/_list.html.erb
ignaspa: Should I change the non english translations to `UPDATE ME!` like the other places? I'm worried about leaving the translations in a state where they could be missed being different than english.

david-yz-liu: Ah, this should actually be a new key (`assessments.none`) rather than modifying an existing key.

app/models/assignment.rb
```diff
@@ -144,6 +144,15 @@ def section_names_past_due_date
     sections_past
   end
 
+  def upcoming(current_user)
+    grouping = current_user.accepted_grouping_for(self.id)
+    due_date = grouping&.due_date
+    if due_date.nil?
+      due_date = assignment.section_due_date(current_user.section)
+    end
+    due_date > Time.current
```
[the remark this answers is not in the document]
ignaspa: The `collection_date` method is for groupings, so is this comment for the lines above where the user is in a grouping? Or am I supposed to use `collection_date_passed` for assignment policy to replace `assignment.section_due_date`?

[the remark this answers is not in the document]
david-yz-liu: Right, I meant this comment more generally. So yes, you should use `collection_date` for groupings, but *also* take into account submission rules when there is no grouping. I'm not sure about the best way to do this, but `collection_date_passed` is possible.

PR #4739, opened 2020-07-15
Marking annotation dropdown

ignaspa:
## Motivation and Context

Currently, the drop down menu for annotation texts can be hidden behind other panes/components when marking. This PR ensures the entire menu is visible regardless of the location of the category button near the right pane.

## Your Changes

**Description**:
Changes to the dropdown menu JS and the AnnotationHandler component.

## Testing

Visual testing in Firefox & Chrome.

## Checklist

### DIFF
--- a/app/assets/javascripts/Components/Result/annotation_manager.jsx
+++ b/app/assets/javascripts/Components/Result/annotation_manager.jsx
@@ -9,6 +9,20 @@ export class AnnotationManager extends React.Component {
     });
   }
 
+  componentDidUpdate(prevProps, prevState, snapshot) {
+    this.props.categories.forEach(cat => {
+      let list = $(`#annotation_text_list_${cat.id}`)[0].getBoundingClientRect();
+      let panel = $('.react-tabs-panel-action-bar')[0].getBoundingClientRect();
+      console.log(list)
+      console.log(panel)
+      console.log(list.width > panel.right - list.left)
+      if (list.width > panel.right - list.left) {
+        $(`#annotation_text_list_${cat.id}`)[0].style.position = relative;
+        $(`#annotation_text_list_${cat.id}`)[0].style.right = panel.right - list.left;
+      }
+    });
+  }
+
   render() {
     return [
       <button key='new_annotation_button'
--- a/app/assets/javascripts/Components/Result/annotation_manager.jsx
+++ b/app/assets/javascripts/Components/Result/annotation_manager.jsx
@@ -2,24 +2,10 @@ import React from 'react';
 
 
 export class AnnotationManager extends React.Component {
-  componentDidMount() {
+  componentDidUpdate(prevProps, prevState, snapshot) {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/Components/Result/annotation_manager.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 
 
 export class AnnotationManager extends React.Component {
-  componentDidMount() {
+  componentDidUpdate(prevProps, prevState, snapshot) {
```
david-yz-liu, line 5: In general, omit right-most parameters if they aren't used. Here we can omit all three, and just do `componentDidUpdate()`.

app/assets/javascripts/DropDownMenu/DropDownMenu.js
```diff
@@ -12,33 +12,25 @@ function DropDownMenu(trigger_node, menu_node) {
 
   // Set up the trigger_node click event
   var me = this;
-  $(this.trigger_node).click(function(event) {
+  $(this.trigger_node).hover(function(event) {
```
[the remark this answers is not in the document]
david-yz-liu: You can remove `event` here (see above comment).

app/assets/javascripts/DropDownMenu/DropDownMenu.js
```diff
@@ -12,33 +12,25 @@ function DropDownMenu(trigger_node, menu_node) {
...
+  let panel = $('.react-tabs-panel-action-bar')[0].getBoundingClientRect();
+
   // Position menu node so that it's directly under the trigger node,
   // and hide it
   $(this.getMenuNode()).css({
     "position" : "absolute",
-    "left" : "offset_left + 'px'",
-    "top" : "(offset_top + offset_height) + 'px'",
+    "left" : Math.min(0, panel.right - (offset_left + 340)),
+    "top" : (offset_height + 5) + 'px',
```
david-yz-liu, line 33: Hmm, since you were able to omit the "px" for `left`, I wonder if you can omit it here as well.

app/assets/javascripts/DropDownMenu/DropDownMenu.js
```diff
@@ -12,33 +12,25 @@ function DropDownMenu(trigger_node, menu_node) {
...
 
+  let panel = $('.react-tabs-panel-action-bar')[0].getBoundingClientRect();
+
   // Position menu node so that it's directly under the trigger node,
   // and hide it
   $(this.getMenuNode()).css({
     "position" : "absolute",
-    "left" : "offset_left + 'px'",
-    "top" : "(offset_top + offset_height) + 'px'",
+    "left" : Math.min(0, panel.right - (offset_left + 340)),
```
david-yz-liu, line 32: Leave a comment referring to the CSS rule that 340 comes from, so that we have a reference in case it changes later.

PR #4740, opened 2020-07-16
Student marks views

ignaspa:
## Motivation and Context

Students can benefit from a separation between the general interface of MarkUs and the information about their grades. This PR removes visible marks on login and brings about a new marks view for students, accessible from a tab in the top menu bar. This new view contains the information previously visible on the homepage, as well as a new histogram of marks for visual aid.

## Your Changes

**Description**:

- New marks tab with summary of assessment results for the duration of the course.

**Type of change** (select all that apply):

 
- [x] New feature (non-breaking change which adds functionality)

## Testing

Conducted visual testing of the new UI in both chrome and firefox.
Added tests for course summaries controller route `populate`.

## Checklist

- [x] I have performed a self-review of my own code.
- [x] I have fixed any Hound bot comments. 
- [x] I have verified that the TravisCI tests have passed. 
- [x] I have reviewed the test coverage changes reported on Coveralls. 
- [x] I have added tests for my changes.
- [x] I have updated the Changelog.md file. 
- [x] I have described any required documentation changes below.

PR #4747, opened 2020-07-24
Assignment marks chart

ignaspa:
## Motivation and Context

Building on PR #4740, the student experience can be improved by offering percentages for each mark in an assignment result (as a normalization of the achievement in each mark) as well as offering a visual display for instant analysis of marks (what went well and what did not).

## Your Changes

**Description**:
Add a modal in the results view that displays a bar chart with each mark as bar. (include old marks when remark given)

image

## Testing

Visual testing in FireFox.

How do extra marks and old marks fit into this chart?

## Checklist


spec/controllers/grade_entry_forms_controller_spec.rb
```diff
@@ -342,4 +342,23 @@
       end.to change { ActionMailer::Base.deliveries.count }.by(1)
     end
   end
+
+  describe '#student_interface' do
+    before :each do
+      allow(controller).to receive(:session_expired?).and_return(false)
```
david-yz-liu, line 348: You don't need this one or the next line (they just repeat the previous mocking). Only the `:current_user` mock needs to change.

app/assets/javascripts/Components/Result/summary_panel.jsx
```diff
@@ -11,13 +12,66 @@ export class SummaryPanel extends React.Component {
...
+  toggleMarksChart = () => {
+    let labels = [];
+    let markData = [];
+    let oldMarks = [];
+    let oldMarksExist = Object.keys(this.props.old_marks).length > 0;
+    this.props.marks.forEach(m => {
+      labels.push(m.name);
+      markData.push(Math.round(m.mark * 100) / m.max_mark);
+      if (oldMarksExist) {
+        if (this.props.old_marks[m.criterion_type + '-' + m.id]) {
```
david-yz-liu, line 57: This check is a bit strange. I think that if this is a remark result, then all of the relevant `old_marks` should be in this object, and we should rely on this. So we shouldn't need to have this check here, nor the else branch (which I don't think makes sense as the right behaviour anyway).

david-yz-liu:
@ignaspa please update the Changelog (and also finish up the checklist).

PR #4749, opened 2020-07-27
Use single table inheritance to unify criterion types

david-yz-liu:
## Motivation and Context

The code for criteria is overly complex because each criterion type is stored in a separate table. E.g., we needed a `Assignment#get_criteria` method instead of a simple association, and to pass around both criterion id and type to disambiguate between different criterion types.

## Your Changes

**Description**: Unify the criteria tables using single table inheritance.

**Type of change** (select all that apply):

 

- [x] Refactoring (internal change to codebase, without changing functionality)

## Testing

Ensured that existing tests all pass, and manual testing in the UI for places where criteria are used.

## Questions and Comments (if applicable)

This is the work of @ashih2018 (I'm just merging).

## Checklist

- [x] I have performed a self-review of my own code.
- [x] I have fixed any Hound bot comments. 
- [x] I have verified that the TravisCI tests have passed. 
- [x] I have reviewed the test coverage changes reported on Coveralls. 
- [x] I have added tests for my changes. 
- [x] I have updated the Changelog.md file.

PR #4752, opened 2020-07-27
Fix sti-criteria-2 comments

ashih2018:
Refactor code based on comments made in #4749 

## Motivation and Context
Fixes comments made in #4749 so it can be merged into master.

## Your Changes
- Refactored various parts of code.

- [ ] Other (please specify): 

## Testing
None

## Checklist


app/models/assignment.rb
```diff
@@ -259,11 +264,7 @@ def display_for_note
 
   # Returns the maximum possible mark for a particular assignment
   def max_mark(user_visibility = :ta_visible)
-    # TODO: sum method does not work with empty arrays. Consider updating/replacing gem:
```
ashih2018, line 262: This comment isn't necessary anymore.

app/models/assignment.rb
```diff
@@ -544,7 +545,7 @@ def summary_json(user)
       next if hide_unassigned && unassigned
 
       max_mark += crit.max_mark
-      accessor = "#{crit.class}-#{crit.id}"
```
ashih2018, line 547: This is as a result of updating `mark_hash` for `result.rb`

app/models/ta.rb
```diff
@@ -37,11 +37,14 @@ def percentage_grades_array(assignment)
     grades = []
 
     if assignment.assign_graders_to_criteria
-      criteria_data = self.criteria.where('criteria.assessment_id': assignment.id).sum(:max_mark)
-      return [] if criteria_data.zero?
+      criteria_ids = self.criterion_ta_associations.where(assessment_id: assignment.id).pluck(:criterion_id)
```
ashih2018, line 40: Made a comment on the other PR as well about the same thing, but we can't change this logic because we use criteria_data later when finding mark data, so we still want to keep track of the ids. Rather we should probably just change the name to `criteria_ids.

app/models/result.rb
```diff
@@ -85,8 +85,8 @@ def get_subtotal(assignment: nil)
       else
         user_visibility = :ta_visible
       end
-      criteria = assignment.criteria.where(user_visibility => true).map(&:id)
-      marks_array = (marks.to_a.select { |m| criteria.member? m.criterion_id }).map(&:mark)
+      criterion_ids = assignment.criteria.where(user_visibility => true).ids
+      marks_array = (marks.to_a.select { |m| criterion_ids.member? m.criterion_id }).map(&:mark)
```
david-yz-liu, line 89: So looking down a bit, it seems that this is similar to the "max mark" calculation we had before. You should be able to reduce this into a single query to return the subtotal now, something like `marks.where(criterion_id: criterion_ids).sum`

app/models/result.rb
```diff
@@ -204,7 +204,7 @@ def create_marks
   def mark_hash
     Hash[
       marks.map do |mark|
-        ["#{mark.criterion.type}-#{mark.criterion_id}",
+        [mark.criterion_id,
          mark.mark]
```
david-yz-liu, line 208: Move this onto the previous line now that it's shorter

PR #4756, opened 2020-07-29
UI for enable/disable variable permissions for graders

ANUSHIYAPRIYA:
## Motivation and Context
This request is for adding UI elements to enable/disable variable permissions for graders.

## Your Changes

**Description**:
Made changes to the user interface for the previous PR #4601 

## Testing
Manually tested this PR through the web interface

## Checklist


app/models/ta.rb
```diff
@@ -2,7 +2,9 @@ class Ta < User
 
   CSV_UPLOAD_ORDER = Rails.configuration.ta_csv_upload_order
   SESSION_TIMEOUT = Rails.configuration.ta_session_timeout
-
+  has_one :grader_permission, dependent: :destroy, foreign_key: :user_id
```
mishaschwartz, line 5: add an `inverse_of :ta` here so that creating the associated object with nested params will work as intended

app/models/ta.rb
```diff
@@ -2,7 +2,9 @@ class Ta < User
...
+  after_create :create_grader_permission
```
mishaschwartz, line 6: change this to a `before_validation` callback and add a validation for the existence of the associated object:

```
  validates_presence_of :grader_permission
  before_validation :create_grader_permission, on: :create 
```

app/models/ta.rb
```diff
@@ -87,4 +89,10 @@ def grade_distribution_array(assignment, intervals = 20)
 
     return distribution
   end
+
+  private
+
+  def create_grader_permission
+    self.grader_permission = GraderPermission.create!(user_id: self.id)
```
mishaschwartz, line 96: if you change this to a `before_validation` callback you should also update this to only build the GraderPermission object (and do it only if it doesn't exist so you don't overwrite a previously built one):

```
self.grader_permission ||= GraderPermission.new
```

app/models/grader_permission.rb
```diff
@@ -0,0 +1,15 @@
+# Contains the grader permissions for a particular grader
+class GraderPermission < ApplicationRecord
```
mishaschwartz, line 2: This method can be condensed to:

```
class GraderPermission < ApplicationRecord
  belongs_to :ta, class_name: 'Ta', foreign_key: :user_id
end
```

because:

- a `belongs_to` association requires the associated object to exist
- by setting `class_name: 'Ta'` it will check if the user is a Ta for you

config/locales/views/grader_permissions/en.yml
```diff
@@ -0,0 +1,8 @@
+en:
+  grader_permissions:
+    list: "Grader Permissions"
+    manage_assessments: "Manage assessments"
+    manage_submissions: "Manage submissions"
+    run_tests: "Run tests"
+    tooltip1: "Create and manage all the assessments which includes annotations, automated testing, exam templates."
+    tooltip2: "Collect and manage submissions, and release/unrelease the grades."
```
mishaschwartz, line 8: Good idea with the tooltips!
Please name them something more descriptive than "tooltip1" and "tooltip2". Also please add a 3rd tooltip for the "Run tests" option.

db/schema.rb
```diff
@@ -617,8 +625,12 @@
   add_foreign_key "extensions", "groupings"
   add_foreign_key "extra_marks", "results", name: "fk_extra_marks_results", on_delete: :cascade
   add_foreign_key "feedback_files", "submissions"
+<<<<<<< HEAD
```
mishaschwartz, line 628: Looks like you didn't finish resolving a merge conflict. Please fix this.

app/views/tas/new.html.erb
```diff
@@ -1,5 +1,11 @@
 <% content_for :title, t('tas.new') %>
 
-<%= render partial: 'layouts/userform',
-           locals: { controller: 'tas',
-                     action: 'create'} %>
+<div class='pane-wrapper'>
```
mishaschwartz, line 3: These 3 divs can be moved into the `grader_permission_pane` partial view.

mishaschwartz (changes requested):
Overall this looks quite good. I really like the layout of the form and your use of tooltips.
I notice that your tests are not here, please make sure to push your tests for this PR.

app/models/user.rb
```diff
@@ -208,6 +208,9 @@ def self.upload_user_list(user_class, user_list, encoding)
           conflict_target: [:user_name], columns: [:last_name, :first_name, :section_id, :email, :id_number]
         }
         User.where(id: imported.ids).each do |user|
+          if user_class == Ta
+            user.run_callbacks(:create) { false }
```
mishaschwartz, line 212: I'm pretty sure the block isn't necessary here.

app/views/grader_permissions/_grader_permission_pane.html.erb
```diff
@@ -0,0 +1,53 @@
+<div class='pane-wrapper'>
+  <div class='pane'>
+    <div id='grader_permission_pane'>
+      <%= render partial: 'shared/navigation_warning',
+                 formats: [:js],
+                 handlers: [:erb] %>
+      <%= form_for @user, as: :user do |f| %>
```
mishaschwartz, line 7: please use `form_with` (`form_for` has been deprecated since rails 5.1)

app/views/grader_permissions/_grader_permission_pane.html.erb
```diff
@@ -0,0 +1,53 @@
...
+          <%= f.text_field :last_name %>
+
+          <%= f.label :first_name, User.human_attribute_name(:first_name) %>
+          <%= f.text_field :first_name %>
+
+          <%= f.label :email, User.human_attribute_name(:email) %>
+          <%= f.text_field :email %>
+        </div>
+
+        <%= f.hidden_field :id %>
```
mishaschwartz, line 22: is this necessary?

app/views/grader_permissions/_grader_permission_pane.html.erb
```diff
@@ -0,0 +1,53 @@
...
+            <%= ff.check_box :manage_submissions %>
+            <%= ff.label :manage_submissions, t('grader_permissions.manage_submissions') %>
+          </div><br/>
+          <div class="tooltip">
+            <span class="tooltip tooltiptext"><%= t('grader_permissions.run_tests_tooltip') %></span>
+            <%= ff.check_box :run_tests %>
+            <%= ff.label :run_tests, t('grader_permissions.run_tests') %>
+          </div>
+        <% end %>
+        <p>
```
mishaschwartz, line 42: why a `<p>` tag here?

app/views/tas/edit.html.erb
```diff
@@ -2,6 +2,6 @@
   <%= t('tas.edit') %>
 <% end %>
 
-<%= render partial: 'layouts/userform',
+<%= render partial: 'grader_permissions/grader_permission_pane',
            locals: { controller: 'tas',
                      action: 'edit' } %>
```
mishaschwartz, line 7: please remove unused locals (action is not used)

app/views/tas/new.html.erb
```diff
@@ -1,5 +1,5 @@
 <% content_for :title, t('tas.new') %>
 
-<%= render partial: 'layouts/userform',
+<%= render partial: 'grader_permissions/grader_permission_pane',
            locals: { controller: 'tas',
                      action: 'create'} %>
```
mishaschwartz, line 5: please remove unused locals (action is not used)

mishaschwartz (changes requested):

app/views/grader_permissions/_grader_permission_pane.html.erb
mishaschwartz: That is only because the `update` method is looking for the id there: 

```ruby
  def update
    @user = Ta.find(params[:user][:id]) # <- this line
    @user.update(user_params)
    respond_with(@user)
  end
```
The id of the ta can be accessed with `params[:id]` which is passed to the controller either way. If you get the id through `params[:id]` the hidden field is not necessary.

spec/controllers/tas_controller_spec.rb
```diff
@@ -212,4 +212,47 @@
...
+            run_tests: true
+          },
+          user_name: 'c6conley',
+          first_name: 'Mike',
+          last_name: 'Conley'
+        }
+      }
+    end
+    before do
+      post :create, params: params
```
mishaschwartz, line 232: use a factory here please. These tests are meant to test the `update` method, not the `create` method.

app/models/user.rb
```diff
@@ -208,6 +208,10 @@ def self.upload_user_list(user_class, user_list, encoding)
           conflict_target: [:user_name], columns: [:last_name, :first_name, :section_id, :email, :id_number]
         }
         User.where(id: imported.ids).each do |user|
+          if user_class == Ta
```
david-yz-liu, line 211: I'm unsure why this block is necessary

ANUSHIYAPRIYA: When we are doing a bulk import TA list, which is not triggering the callback in the TA model. That callback is used for creating associated grader permissions. So I am triggering that callback explicitly here if it is a TA class.

Issue #4667, opened 2020-06-14
Total mark column: round grades


dijkstracula:
In the submissions view, the grades listed don't have a maximum number of digits, so numerical precision sometimes yields pretty lengthy grades:

![image](https://user-images.githubusercontent.com/8949748/84605914-77e6e180-ae5e-11ea-9c4f-76f83d1360ed.png)

It'd be nice if this was only rounded to one or maybe two decimal places.

PR #4762, opened 2020-07-30
views: round total marks for submissions

ignaspa:
## Motivation and Context

Resolves [issue #4667](https://github.com/MarkUsProject/Markus/issues/4667) 

## Your Changes

**Description**:
Add rounding to total grade value in the client.

- [x] Other (please specify): 
Minor UI alteration, same functionality.

## Testing

Checking rounding in UI with FireFox. (altering component prop values to introduce various decimal values)

Unclear if rounding should be done in UI where JS rounding may cause minor discrepancies. Currently, rounding in the UI is what is done for the max mark, so in keeping with pre existing code rounding for the total mark was implemented as such as well. However, could alternatively add rounding to assignment model method which creates the data.

## Checklist


david-yz-liu:
Yeah, I think in other places we normally round in the UI.

PR #4783, opened 2020-08-03
style: correct style of deductions listing in marks summary view

ignaspa:
## Motivation and Context

Students may be confused about the idea that 'deductions from annotations' or 'overridden deductions' seem clickable due to the font color and cursor pointer. This PR fixes those misleading style choices.

## Your Changes

**Description**:
Fix mark-deductions css.

## Testing

UI in FireFox

## Checklist

### DIFF
--- a/app/assets/stylesheets/grader.scss
+++ b/app/assets/stylesheets/grader.scss
@@ -184,9 +184,8 @@
 
 .mark-deductions {
   clear: both;
-  color: $dark-blue;
-  cursor: pointer;
-  font-size: 1.1em;
+  color: $dark-grey;
+  font-size: 1em;
   padding-bottom: 5px;
   padding-top: 2px;
   text-align: right;
--- a/app/assets/stylesheets/grader.scss
+++ b/app/assets/stylesheets/grader.scss
@@ -184,8 +184,6 @@
 
 .mark-deductions {
   clear: both;
-  color: $dark-grey;
-  font-size: 1em;
   padding-bottom: 5px;
   padding-top: 2px;
   text-align: right;


app/assets/stylesheets/grader.scss
```diff
@@ -184,9 +184,8 @@
 
 .mark-deductions {
   clear: both;
-  color: $dark-blue;
-  cursor: pointer;
-  font-size: 1.1em;
+  color: $dark-grey;
```
david-yz-liu, line 187: Hmmm, here can we just use the default value (which I'm assuming is `$black$`)?

Similarly, do we need to specify a `font_size` below, or can we use the default value there too?

app/assets/stylesheets/grader.scss
```diff
@@ -184,9 +184,8 @@
 
 .mark-deductions {
   clear: both;
-  color: $dark-blue;
-  cursor: pointer;
```
david-yz-liu, line 188: Just double-checking this: does the `cursor: pointer` still apply for the mark deductions themselves (e.g., "-1")? Because those are clickable.

app/assets/stylesheets/grader.scss
ignaspa: So I was basing the change for the default color on everything else on the page which is dark grey. When I remove the properties you describe, the default color is also $dark-grey, and the size is 0.8em/1.2em
image

app/assets/stylesheets/grader.scss
ignaspa: Yep!

PR #4794, opened 2020-08-07
views: fix error in result view

ignaspa:
## Motivation and Context

Fix the javascript errors when the result view components initially load.

## Your Changes

**Description**:
Update the submissions_selector component.

## Checklist

### DIFF
--- a/app/assets/javascripts/Components/Result/submission_selector.jsx
+++ b/app/assets/javascripts/Components/Result/submission_selector.jsx
@@ -92,6 +92,13 @@ export class SubmissionSelector extends React.Component {
       progressBarColour = '#FE2A2A';
     }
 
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/Components/Result/submission_selector.jsx
```diff
@@ -113,8 +113,8 @@ export class SubmissionSelector extends React.Component {
               value={this.props.num_marked}
               min={0}
               max={this.props.num_assigned}
-              low={this.props.num_assigned * 0.35}
```
david-yz-liu, line 116: I think the check rules out 0 as a possible value, but we should probably allow 0 (and that should also be the default, not 1)

Issue #3832, opened 2019-02-15
Bonus Mark in Assignment Marking Scheme


jessebett:
I believe #84 was closed without addressing the feature request.

While the feature you have included is useful, to be able to individually add/subtract bonus marks on top of the marking scheme, this is not very practical for the more common scenario of having a bonus question.

In my assignment there is a Bonus Question. Rather, one question has a bonus sub-question.
My TAs will be grading the bonus question along with the rest of the assignment.
If the question is /10, and there's a bonus sub-question /2, the students can achieve 12/10.
However, Markus provides no way to achieve this. 
It is not practical to have my TAs click summary, then click extra mark, then add an additional mark, for a question that is graded just like any other question.

Currently I am dealing with this very non-optimally. I just make the question /12, and then inform the students that they should manually recompute their grade. This is not great.

I propose one of two solutions:
1. Like the interface for Marks Spreadsheet, there could be an option to add a Bonus Mark category. In the case of subquestions, I will have to pull the bonus out of the parent question to have it individually recognized as bonus.
2. Allow me to optionally indicate that a question can have a mark greater than the criteria. This way I can have the TA just write 12/10 and it will be treated as bonus.

mishaschwartz:
I like this suggestion but I think that option 1 is the better way to go since: 
* it can be used for all criteria types (not just flexible)
* it reduces the chance that a grader will accidentally assign a bonus mark when one is not intended

PR #4804, opened 2020-08-11
criteria: add option to specify as bonus

ignaspa:
## Motivation and Context

Resolves [issue #3832](https://github.com/MarkUsProject/Markus/issues/3832) and pushes the criteria table closer to being merged with the grade entry item table.

## Your Changes

**Description**:
Migration to include bonus option for criteria.

## Checklist

### DIFF
--- a/db/migrate/20200811194846_add_bonus_to_criteria.rb
+++ b/db/migrate/20200811194846_add_bonus_to_criteria.rb
@@ -1,0 +1,5 @@
+class AddBonusToCriteria < ActiveRecord::Migration[6.0]
+  def change
+    add_column :criteria, :bonus, :boolean, default: false, null: false
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_07_26_012622) do
+ActiveRecord::Schema.define(version: 2020_08_11_194846) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -150,6 +150,7 @@ ActiveRecord::Schema.define(version: 2020_07_26_012622) do
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
     t.bigint "assessment_id", null: false
+    t.boolean "bonus", default: false, null: false
     t.index ["assessment_id"], name: "index_criteria_on_assessment_id"
   end
 


app/helpers/course_summaries_helper.rb
```diff
@@ -55,13 +55,10 @@ def get_table_json_data(current_user)
...
-    flexible_max = FlexibleCriterion.group(:assessment_id).sum(:max_mark)
-    checkbox_max = CheckboxCriterion.group(:assessment_id).sum(:max_mark)
-    @max_marks = Hash[Assignment.all.map do |a|
-      [a.id, rubric_max.fetch(a.id, 0) + flexible_max.fetch(a.id, 0) + checkbox_max.fetch(a.id, 0)]
-    end
-    ]
+    @max_marks = Hash[Assignment.all.joins(:criteria)
+                               .where('criteria.bonus': false)
+                               .group('assessments.id')
+                               .pluck('assessments.id', 'SUM(criteria.max_mark)')]
```
david-yz-liu, line 61: Hm, I think if you do a `.group` followed by a `.sum`, you get a hash already (mapping id to the value being summed). I think this is exactly what you want in this case.

app/models/assignment.rb
```diff
@@ -549,7 +549,7 @@ def summary_json(user)
       unassigned = !assigned_criteria.nil? && !assigned_criteria.include?(crit.id)
       next if hide_unassigned && unassigned
 
-      max_mark += crit.max_mark
+      max_mark += crit.bonus? ? 0 : crit.max_mark
```
david-yz-liu, line 552: Rather than doing a "+0", use `unless crit.bonus?` as an inline check.

app/models/criterion.rb
```diff
@@ -140,6 +143,7 @@ def self.update_assigned_groups_counts(assignment)
 
   # When max_mark of criterion is changed, all associated marks should have their mark value scaled to the change.
   def scale_marks
+    self.assignment.update_results_stats if previous_changes.key?('bonus')
```
david-yz-liu, line 146: Hmm interesting. I'm thinking that maybe `update_results_stats` should be called when the `max_mark` is changed as well, but called after the marks are updated (e.g. line 171).

ignaspa: In that case I would essentially change the guard clause to an `if` branch right? Because, I am essentially piggybacking off a callback that only serves the purpose of scaling marks and am now using the callback to update results based on a bonus designation.

david-yz-liu: Yes that's true, I'm a bit torn. Another approach would be to make a new callback function (e.g., "update result stats") that you have in an `after_update`, and then do

```ruby
def update_result_stats
  return unless max_mark changed or bonus changed
  scale marks if max_mark_changed
  self.assignment.update_results_stats
```

ignaspa: Hm ok, I have implemented the approach I mentioned, but I can easily change the approach to this one as well.

PR #4850, opened 2020-09-12
Autosaves the overall comments every 3 seconds if there are changes

### LINKED ISSUE #2786: The "Overall Comments" are not autosaved
In the assignment grading view, the marks are autosaved, implying that all changes should be autosaved. But the "Overall Comments" under the annotation tab is not autosaved unless "Save Changes" is clicked. I lost all my comments...

ihasan98:
## Motivation and Context

resolves #2786

## Your Changes

**Description**:

- [ ] Other (please specify): 

## Testing

Manual testing through web interface

## Checklist

### DIFF
--- a/app/assets/javascripts/Components/Result/annotation_panel.jsx
+++ b/app/assets/javascripts/Components/Result/annotation_panel.jsx
@@ -10,7 +10,6 @@ export class AnnotationPanel extends React.Component {
       overallComment: props.overallComment,
       unsavedChanges: false
     };
-    this.autoSaveOverallComments = this.autoSaveOverallComments.bind(this);
     this.submitOverallCommentButton = React.createRef();
   };
 
@@ -60,7 +59,7 @@ export class AnnotationPanel extends React.Component {
     event.preventDefault();
   };
 
-  autoSaveOverallComments() {
+  autoSaveOverallComments = () => {
     if (this.state.unsavedChanges) {
       this.persistOverallComments();
     }
--- a/app/assets/javascripts/Components/Result/annotation_panel.jsx
+++ b/app/assets/javascripts/Components/Result/annotation_panel.jsx
@@ -6,9 +6,7 @@ export class AnnotationPanel extends React.Component {
   persistChanges = (value) => {
     return $.post({
       url: Routes.update_overall_comment_assignment_submission_result_path(
-        this.props.assignment_id,
-        this.props.submission_id,
-        this.props.result_id
+        this.props.assignment_id, this.props.submission_id, this.props.result_id
       ),
       data: { result: { overall_comment: value } },
     });
@@ -17,7 +15,7 @@ export class AnnotationPanel extends React.Component {
   render() {
     let overallCommentElement;
     if (this.props.released_to_students || this.props.remarkSubmitted) {
-      overallCommentElement = <div id="overall_comment_text" key="overall_comment_text" />;
+      overallCommentElement = <div id='overall_comment_text' key='overall_comment_text' />;
     } else {
       overallCommentElement = (
         <TextForm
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/Components/Result/annotation_panel.jsx
```diff
@@ -10,20 +10,22 @@ export class AnnotationPanel extends React.Component {
       overallComment: props.overallComment,
       unsavedChanges: false
     };
+    this.autoSaveOverallComments = this.autoSaveOverallComments.bind(this);
```
mishaschwartz, line 13: Instead of using the "bind in constructor" technique, can we use the "class properties" syntax instead:

https://reactjs.org/docs/faq-functions.html#class-properties-stage-3-proposal

since we're using this syntax in a lot of other react components in MarkUs.

app/assets/javascripts/Components/Result/autosave_text_form.jsx
```diff
@@ -0,0 +1,111 @@
...
+    timeout = setTimeout(later, wait);
+
+    if (callNow) func.apply(context, args);
+  };
+}
+
+const SaveMessage = ({ unSaved }) => (
+  <div className="autosave-text">
+    {unSaved ? (
+      <p> There are unsaved changes </p>
```
david-yz-liu, line 29: Please internationalize these strings (see https://guides.rubyonrails.org/i18n.html). You can put the strings under the `views/results/` locale file. It's okay to use English only.

app/assets/javascripts/Components/Result/annotation_panel.jsx
```diff
@@ -1,91 +1,45 @@
...
+      data: { result: { overall_comment: value } },
     });
-    event.preventDefault();
   };
 
   render() {
     let overallCommentElement;
     if (this.props.released_to_students || this.props.remarkSubmitted) {
-      overallCommentElement = <div id='overall_comment_text' key='overall_comment_text' />;
+      overallCommentElement = (
```
david-yz-liu, line 20: Revert this change

app/assets/javascripts/Components/Result/annotation_panel.jsx
```diff
@@ -1,91 +1,45 @@
...
-          <div id="overall_comment_preview" className="preview"/>
-        </div>
+        <TextForm
+          initialValue={this.props.overallComment}
+          persistChanges={this.persistChanges}
+          previewId={"overall_comment_preview"}
+        />
       );
     }
-
```
david-yz-liu, line 83: Please revert the changes here and below

app/assets/javascripts/Components/Result/autosave_text_form.jsx
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }
+
+  handlePersist = debounce(
+    () => {
+      this.props.persistChanges(this.state.value).then(() => {
+        this.setState({ unsavedChanges: false });
+      });
+    },
+    1500,
```
david-yz-liu, line 65: I'm wondering what the relationship is between this number and the 3000. You should pull these out into constants and include a comment explaining their purpose.

david-yz-liu (commented):
Nice work, @ihasan98. In addition to the few comments I left, please also look into removing the flash message, and making the "success" message appear after a success response. Also the styling for the autosave text could be improved to distinguish between the two messages, maybe including the checkmark icon after "success" message?

app/assets/javascripts/Components/Result/autosave_text_form.jsx
ihasan98: This was my bad. I forgot to push the commit that removed the `setInterval`. I'll still make this a constant

david-yz-liu:
@ihasan98 I didn't notice this earlier, but currently the overall comment no longer appears when the result is released. The way you've implemented the `TextForm`'s `updatePreview` method isn't correctly rendering to the overall comment text div when the result is released (it's a different one than the preview).

Issue #4502, opened 2020-04-04
"Not collected" submission counts towards progress bar


Xyene:
If a submission was not collected, it still counts in the denominator of a marker's progress bar (this one):

![image](https://user-images.githubusercontent.com/1403503/78416961-397dac80-75fb-11ea-827c-9c100f148b68.png)

As far as I could find there was no way to mark that submission as "done", making (in my case) the highest possible completion 17/18.

PR #4854, opened 2020-09-17
Easy fix: Not collected submission counts towards progress bar 

ANUSHIYAPRIYA:
## Motivation and Context
This resolves issue #4502 

## Your Changes

**Description**:
Made changes so that the progress bar in the results page will display [ No.of submissions completed/No of. submissions collected ] ( Ex: 12/15 completed, where 15 - No of submissions collected )

## Testing

Manually tested in the UI
Added corresponding tests for the changes

## Checklist

### DIFF
--- a/app/assets/javascripts/Components/Result/result.jsx
+++ b/app/assets/javascripts/Components/Result/result.jsx
@@ -503,9 +503,8 @@ class Result extends React.Component {
           is_reviewer={this.state.is_reviewer}
           marks={this.state.marks || []}
           marking_state={this.state.marking_state}
-          num_assigned={this.state.num_assigned}
           num_marked={this.state.num_marked}
-          num_collectd={this.props.num_collected}
+          num_collected={this.state.num_collected}
           released_to_students={this.state.released_to_students}
           total={this.state.total}
           toggleFullscreen={this.toggleFullscreen}
--- a/app/assets/javascripts/Components/Result/result.jsx
+++ b/app/assets/javascripts/Components/Result/result.jsx
@@ -505,6 +505,7 @@ class Result extends React.Component {
           marking_state={this.state.marking_state}
           num_assigned={this.state.num_assigned}
           num_marked={this.state.num_marked}
+          num_collectd={this.props.num_collected}
           released_to_students={this.state.released_to_students}
           total={this.state.total}
           toggleFullscreen={this.toggleFullscreen}
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -153,7 +153,6 @@ class ResultsController < ApplicationController
           end
           data[:notes_count] = submission.grouping.notes.count
           data[:num_marked] = assignment.get_num_marked(current_user.admin? ? nil : current_user.id)
-          data[:num_assigned] = assignment.get_num_assigned(current_user.admin? ? nil : current_user.id)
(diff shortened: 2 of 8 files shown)


app/assets/javascripts/Components/Result/result.jsx
```diff
@@ -505,6 +505,7 @@ class Result extends React.Component {
           marking_state={this.state.marking_state}
           num_assigned={this.state.num_assigned}
           num_marked={this.state.num_marked}
+          num_collectd={this.props.num_collected}
```
mishaschwartz, line 508: - this prop is not used
- a prop named `num_collected` is already being passed to the `SubmissionSelector` on line 499
- `num_collected` should be accessed from the `this.state`, not `this.props` (see other comment below)

app/assets/javascripts/Components/Result/result.jsx
```diff
@@ -505,6 +505,7 @@ class Result extends React.Component {
           marking_state={this.state.marking_state}
           num_assigned={this.state.num_assigned}
```
mishaschwartz, line 506: - `num_assigned` should no longer be necessary, remove it here and everywhere else

app/views/results/edit.html.erb
```diff
@@ -4,6 +4,7 @@
   <%= javascript_include_tag 'mathjax' %>
   <%= javascript_pack_tag 'result' %>
   <%= javascript_include_tag 'Results/main' %>
+  <% ta_id = @current_user.admin? ? nil : @current_user.id %>
```
mishaschwartz, line 7: don't define unnecessary variables in views

app/views/results/edit.html.erb
```diff
@@ -15,6 +16,7 @@
           grouping_id: <%= @result.submission.grouping_id %>,
           assignment_id: <%= @result.submission.grouping.assessment_id %>,
           role: '<%= @current_user.class.name %>',
+          num_collected: <%= @assignment.get_num_collected(ta_id) %>,
```
mishaschwartz, line 19: this shouldn't be passed as a prop. Instead, get the information when setting the `state` of the Result component.

mishaschwartz (changes requested):

app/controllers/results_controller.rb
```diff
@@ -154,6 +154,7 @@ def show
           data[:notes_count] = submission.grouping.notes.count
           data[:num_marked] = assignment.get_num_marked(current_user.admin? ? nil : current_user.id)
           data[:num_assigned] = assignment.get_num_assigned(current_user.admin? ? nil : current_user.id)
```
mishaschwartz, line 156: If `num_assigned` is not being used please remove it here as well.
If `get_num_assigned` is not being used anywhere else please remove that function as well.

mishaschwartz (approved):

david-yz-liu:
@ANUSHIYAPRIYA I merged this in because it fixes this specific issue, but it seems like there are other places in the code that still use `get_num_assigned` where it might be more appropriate to use `get_num_collected` instead. So could you please review those places as well?

ANUSHIYAPRIYA:
> @ANUSHIYAPRIYA I merged this in because it fixes this specific issue, but it seems like there are other places in the code that still use `get_num_assigned` where it might be more appropriate to use `get_num_collected` instead. So could you please review those places as well?

Yes, David. I thought of creating a new PR for that.

PR #4861, opened 2020-09-22
Created new fake user type 'TestStudent'

ANUSHIYAPRIYA:
## Motivation and Context

This is the first step for the project 'Feature: Autotest Solution Table'.
The instructor can submit solutions for their tests when setting up automated testing and they can run tests on their solutions.
https://openproject.teach.cs.toronto.edu/projects/markus/work_packages/50/activity

## Your Changes

**Description**:
In order to treat the solution file as same as the submission file, I have created a fake user type 'Test Student'.
So that the instructor can test their solutions with the submitted test files.

## Testing

Added tests for the changes

## Checklist

### DIFF
--- a/app/models/grouping.rb
+++ b/app/models/grouping.rb
@@ -25,7 +25,6 @@ class Grouping < ApplicationRecord
   has_many :ta_memberships, class_name: 'TaMembership'
   has_many :tas, through: :ta_memberships, source: :user
   has_many :students, through: :student_memberships, source: :user
-  has_many :test_students, class_name: 'TestStudent', through: :student_memberships, source: :user
   has_many :pending_students,
            class_name: 'Student',
            through: :pending_student_memberships,
@@ -84,8 +83,7 @@ class Grouping < ApplicationRecord
 
   has_many :grouping_starter_file_entries, dependent: :destroy
   has_many :starter_file_entries, through: :grouping_starter_file_entries
-  validate :test_student_grouping_member, if: -> { self.test_students.exists? }
-  validates_associated :test_students
+  validate :test_student_grouping_member
   # Assigns a random TA from a list of TAs specified by +ta_ids+ to each
   # grouping in a list of groupings specified by +grouping_ids+. The groupings
   # must belong to the given assignment +assignment+.
@@ -782,8 +780,8 @@ class Grouping < ApplicationRecord
   end
 
   def test_student_grouping_member
-    members =  self.accepted_student_memberships.joins(:user).count
-    return if members == 1
+    members = self.accepted_student_memberships.joins(:user).where('users.type': 'TestStudent').count
+    return if members <= 1
 
     errors.add(:base, 'Grouping with test student should have no other members')
   end
--- a/app/models/grouping.rb
+++ b/app/models/grouping.rb
@@ -84,7 +84,8 @@ class Grouping < ApplicationRecord
 
   has_many :grouping_starter_file_entries, dependent: :destroy
   has_many :starter_file_entries, through: :grouping_starter_file_entries
-  validate :test_student_grouping_member, if: -> { !self.id.nil? && self.test_students.exists? }
+  validate :test_student_grouping_member, if: -> { self.test_students.exists? }
+  validates_associated :test_students
   # Assigns a random TA from a list of TAs specified by +ta_ids+ to each
   # grouping in a list of groupings specified by +grouping_ids+. The groupings
   # must belong to the given assignment +assignment+.
@@ -781,9 +782,9 @@ class Grouping < ApplicationRecord
   end
 
   def test_student_grouping_member
-    members = memberships.joins(:user).where(grouping_id: self.id).count
-    return true if members == 1
+    members =  self.accepted_student_memberships.joins(:user).count
+    return if members == 1
+
     errors.add(:base, 'Grouping with test student should have no other members')
-    false
   end
 end
--- a/app/models/grouping.rb
+++ b/app/models/grouping.rb
@@ -25,7 +25,7 @@ class Grouping < ApplicationRecord
   has_many :ta_memberships, class_name: 'TaMembership'
   has_many :tas, through: :ta_memberships, source: :user
   has_many :students, through: :student_memberships, source: :user
-  has_many :test_students, through: :student_memberships, source: :user
+  has_many :test_students, class_name: 'TestStudent', through: :student_memberships, source: :user
   has_many :pending_students,
            class_name: 'Student',
            through: :pending_student_memberships,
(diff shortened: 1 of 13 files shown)


mishaschwartz (changes requested):
Some other things that should be included in this PR:

- Make sure that a grouping with a TestStudent is a member is always a valid grouping
- Make sure that the TestStudent's groupings are never shown in the UI
- Make/change any policies that are relevant to the new user
- Add relevant associations to the TestStudent model
- Add rspec tests for the TestStudent model
- etc.

ANUSHIYAPRIYA:
> Some other things that should be included in this PR:
> 
> * Make sure that a grouping with a TestStudent is a member is always a valid grouping
> * Make sure that the TestStudent's groupings are never shown in the UI
> * Make/change any policies that are relevant to the new user
> * Add relevant associations to the TestStudent model
> * Add rspec tests for the TestStudent model
> * etc.

Points which are addressed:
1. Validating associated grouping. After adding the code for creating a grouping for test student, will add more validation
2. Made sure that the TestStudent's groupings are never shown in the UI
3. Added relevant associations to the TestStudent model
4. Added rspec tests for the TestStudent model

Points which are not necessary:
1. Make/change any policies that are relevant to the new user (I think for the upcoming PR's we may need some changes in policy.)

app/models/grouping.rb
```diff
@@ -25,6 +25,8 @@ class Grouping < ApplicationRecord
   has_many :ta_memberships, class_name: 'TaMembership'
   has_many :tas, through: :ta_memberships, source: :user
   has_many :students, through: :student_memberships, source: :user
+  has_many :test_students, class_name: 'TestStudent', through: :student_memberships, source: :user
```
mishaschwartz, line 28: I'm not sure if this is necessary. I don't think the grouping object needs to differentiate between students and test_students here.

app/models/test_student.rb
```diff
@@ -0,0 +1,14 @@
+# TestStudent - fake student, used for running tests on solution files
+class TestStudent < User
+  has_many :accepted_groupings, lambda {
```
mishaschwartz, line 3: use the arrow syntax `->` instead of `lambda`

mishaschwartz (changes requested):
Some additional validations and checks I would like to see added:
- a grouping with a test student as a member should have no other members
- a test student should not be allowed to be invited to join a grouping
- a test student can only be an 'inviter'

are there any others you think are appropriate to add?

app/models/test_student.rb
ANUSHIYAPRIYA: I used the -> first but Rubocop raised an offense and suggested using lambda. Which one is recommended?

ANUSHIYAPRIYA:
> Some additional validations and checks I would like to see added:
> 
> * a grouping with a test student as a member should have no other members
> * a test student should not be allowed to be invited to join a grouping
> * a test student can only be an 'inviter'
> 
> are there any others you think are appropriate to add?

Points addressed:
1. A grouping with a test student as a member should have no other members
2. A test student can only be an 'inviter'

Points not addressed:
1. A test student should not be allowed to be invited to join a grouping
Reason:
By default, the existing code does this behavior. It is only allowing the students of type 'Student' and the unhidden students. I checked all the corresponding method for inviting a member. Those are all already checking for the class as 'Student' class and hidden is false. So this will automatically exclude the test students.

app/models/test_student.rb
mishaschwartz: I'm going to make a decision here and say that we should go with `->`. I'll make a separate PR that disables that cop so houndbot won't complain in the future.

mishaschwartz (changes requested):

app/models/grouping.rb
```diff
@@ -778,4 +779,11 @@ def use_section_due_date?
       inviter.section.present? &&
       assignment.section_due_dates.present?
   end
+
+  def test_student_grouping_member
+    members = memberships.joins(:user).where(grouping_id: self.id).count
```
mishaschwartz, line 784: We care about `accepted_student_memberships` here not all `memberships`

Try starting with:

```
accepted_student_memberships.joins(:user).pluck('users.type')
```

and see if you can find a more efficient way of checking this

app/models/grouping.rb
```diff
@@ -83,7 +84,7 @@ class Grouping < ApplicationRecord
 
   has_many :grouping_starter_file_entries, dependent: :destroy
   has_many :starter_file_entries, through: :grouping_starter_file_entries
-
+  validate :test_student_grouping_member, if: -> { !self.id.nil? && self.test_students.exists? }
```
mishaschwartz, line 87: use `new_record?` instead of checking for nil.

also, can we do this validation with having a `test_students` association please.

also, this validation should be doable even if we have a new record

app/models/student_membership.rb
```diff
@@ -82,4 +84,14 @@ def one_accepted_per_assignment
     errors.add(:base, I18n.t('csv.memberships_not_unique'))
     throw(:abort)
   end
+
+  def test_student_membership_status
+    user.memberships.each do |membership|
```
mishaschwartz, line 89: this validation should only care about the membership referred to by `self`, not all memberships belonging to the user.

try something like:

```ruby
errors.add(...) if user.test_student? && membership_status != STATUSES[:inviter]
```

app/models/student_membership.rb
```diff
@@ -82,4 +84,14 @@ def one_accepted_per_assignment
...
+      unless membership.membership_status == 'inviter'
+        errors.add(:base, 'A test student can only be an inviter')
+        false
```
mishaschwartz, line 92: validations don't need to return true or false

spec/factories/test_students.rb
```diff
@@ -0,0 +1,5 @@
+FactoryBot.define do
+  factory :test_student, class: TestStudent, parent: :user do
+    hidden { true }
```
mishaschwartz, line 3: this is no longer necessary

app/models/student_membership.rb
```diff
@@ -82,4 +84,14 @@ def one_accepted_per_assignment
     errors.add(:base, I18n.t('csv.memberships_not_unique'))
     throw(:abort)
   end
+
+  def test_student_membership_status
+    user.memberships.each do |membership|
+      unless membership.membership_status == 'inviter'
```
mishaschwartz, line 90: always refer to the `STATUSES` hash, not the string value directly

mishaschwartz (changes requested):

app/models/grouping.rb
```diff
@@ -83,7 +84,8 @@ class Grouping < ApplicationRecord
 
   has_many :grouping_starter_file_entries, dependent: :destroy
   has_many :starter_file_entries, through: :grouping_starter_file_entries
-
+  validate :test_student_grouping_member, if: -> { self.test_students.exists? }
+  validates_associated :test_students
```
mishaschwartz, line 88: this can be removed

PR #4867, opened 2020-09-25
Don't cause double render when reporting error

mishaschwartz:
## Motivation and Context

If an error is raised, the render at the bottom of the file is still called causing a double render error.

## Your Changes

**Description**:

- return so a double render is not called

## Checklist


david-yz-liu:
@mishaschwartz nice catch. Maybe you could add a test that stubs `Student#create_group_for_working_alone_student` to raise an error and improve the code coverage here?

PR #4868, opened 2020-09-25
Replaced num_assigned with num_collected

ANUSHIYAPRIYA:
## Motivation and Context

This is an extension of PR #4854. 

## Your Changes

**Description**:
Replaced num_assigned with num_collected submissions in all the places except Grader distribution graph and Grader dashboard view

## Testing

Tested manually in the UI

## Checklist


app/models/assignment.rb
```diff
@@ -752,11 +752,11 @@ def get_num_valid
 
   def get_num_marked(ta_id = nil)
     if ta_id.nil?
-      results_join = groupings.left_outer_joins(:current_result)
+      results_join = groupings.includes(:current_result).where('groupings.is_collected': true)
```
david-yz-liu, line 755: It looks like we're only considering collected groupings now, which is good. This means this part can be simplified quite a bit, just doing a filter for results that have marking state 'complete', rather than the current indirect calculation. You still need to keep the `groupings.joins(:current_result)` part (the `:current_result` limits this to only one result per grouping).

david-yz-liu: (and similarly below)

david-yz-liu: @ANUSHIYAPRIYA You can use a `.joins` to simplify this query.

- Results only exist if submissions are collected, so doing a `joins(:current_result)` will make the `.where('groupings.is_collected': true)` unnecessary.
- Below, it will require that the associated result exists, and so the `where('results.id': nil)` can be removed (and you can remove the `.or` too, and just use a single `.where`)

app/models/assignment.rb
```diff
@@ -752,11 +752,11 @@ def get_num_valid
...
-      results_join = groupings.left_outer_joins(:current_result)
-      num_incomplete = results_join.where('results.id': nil)
-                                   .or(results_join.where('results.marking_state': 'incomplete'))
-                                   .count
-      get_num_assigned - num_incomplete
+      results_join = groupings.includes(:current_result).where('groupings.is_collected': true)
+      num_complete = results_join.where('results.id': nil)
+                                 .or(results_join.where('results.marking_state': 'complete'))
+                                 .count
+      num_complete
```
david-yz-liu, line 759: You don't need a separate variable here; this entire block of code can be a single query (which you just return the value of)

app/models/assignment.rb
```diff
@@ -752,11 +752,7 @@ def get_num_valid
 
   def get_num_marked(ta_id = nil)
     if ta_id.nil?
-      results_join = groupings.left_outer_joins(:current_result)
-      num_incomplete = results_join.where('results.id': nil)
-                                   .or(results_join.where('results.marking_state': 'incomplete'))
-                                   .count
-      get_num_assigned - num_incomplete
+      groupings.left_outer_joins(:current_result).where('results.marking_state': 'complete').count
```
david-yz-liu, line 755: You actually need to use `joins` here, not `left_outer_joins`. This was a change I noted earlier, since now we only want to count groupings that have a result.

You can read more about the difference at https://guides.rubyonrails.org/active_record_querying.html#left-outer-joins.

PR #4869, opened 2020-09-26
Fix for current time: Globally replace Time.now and Time.zone.now with Time.current

sujiayi317:
##  Motivation and Context

As sugested by Prof. Liu:
Problem: Currently MarkUs uses a few different methods to access the current time: Time.now, Time.zone.now, Time.current.
Solution: Do a global search-and-replace for the first two to replace them with Time.current.

## Your Changes

Globally replace Time.now and Time.zone.now with Time.current

**Description**:

## Testing

Run `docker-compose run test rspec`, 3562 examples, 0 failures, 23 pending

Markus/coverage/index.html is read-only, so no opration is performed for this file.
All other 42 files are updated with Time.current (as a replacement of Time.now or Time.zone.now).

Steps I have taken: Run `git diff` to compare the changes. No side effect.

## Checklist


app/lib/git_repository.rb
```diff
@@ -62,7 +62,7 @@ def self.do_commit(repo, author, message)
     index = repo.index
     commit_tree = index.write_tree(repo)
     index.write
-    commit_author = {email: 'markus@markus.com', name: author, time: Time.now}
+    commit_author = { email: 'markus@markus.com', name: author, time: Time.current }
```
david-yz-liu, line 65: Looks like this change might be causing an error in the git tests (see Travis report). Please change back for now, we can look into this later.

sujiayi317: Oh, yes! Sorry about that, I thought this change would be fine... Let me change back now.

david-yz-liu: Oh, no the issue isn't the spaces, it's the `Time.now` vs. `Time.current` (I think). You should change this back to `Time.now`.

sujiayi317: I've just changed this back to `Time.now`, hopefully this time it will work.

sujiayi317: All checks have passed according to Travis, yeah. I'm a little confused why`Time.now` is the only correct version in this case.

PR #4874, opened 2020-09-30
Change all option for annotation editing

robert-tan:
## Motivation and Context

Will help graders change singular cases of annotation use.

## Your Changes

**Description**: Added an 'Update This' and 'Update All' button to non-deductive categorized annotations. Allows graders to change a single instance of annotations. 

## Testing

Rspec tests run

## Checklist


spec/controllers/annotations_controller_spec.rb
```diff
@@ -273,6 +273,29 @@
...
       end
+
+      it 'successfully updates a singular annotation text' do
+        anno1 = create(
+          :text_annotation,
+          submission_file: submission_file,
+          creator: user,
+          result: result
+        )
+        anno2 = create(
```
david-yz-liu, line 284: I don't think this does what you want: the `:text_annotation` factory automatically creates a new `annotation_text` instance. So here you want to pass in an option for `annotation_text` to make it the same as the one for `anno1`.

app/views/annotations/_annotation_modal.html.erb
```diff
@@ -49,7 +49,10 @@
       <% end %>
 
       <section class='dialog-actions'>
-        <%= submit_tag title, id:'submit_annotation', data: submit_btn_data %>
+        <%= submit_tag (change_one_option ? title_all : title), id:'submit_annotation', data: submit_btn_data %>
+        <% if change_one_option %>
+          &nbsp;&nbsp;<%= submit_tag title_one, id: 'submit_this_annotation', name: 'change_one' %>
```
david-yz-liu, line 54: After looking at this more, I think having two buttons is a bit too cumbersome (especially since the labels are quite long). Instead, please keep the old submit button, and add a checkbox input to the form that says something like "Modify all occurrences of this annotation".

app/controllers/annotations_controller.rb
```diff
@@ -154,6 +154,14 @@ def update
         return
       end
     end
-    @annotation_text.update(content: params[:content])
+
+    if params[:change_one]
+      new_text = @annotation_text.dup
+      new_text.update(content: params[:content], annotation_category_id: nil)
+      new_text.save
```
david-yz-liu, line 161: `save` after `update` is redundant. Also let's wrap the two updates in a [transaction](https://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html).

app/assets/javascripts/Components/Result/result.jsx
```diff
@@ -297,7 +297,7 @@ class Result extends React.Component {
         this.setState({annotations: newAnnotations});
       }
     }
-    update_annotation_text(annotation.id, marked(annotation.content, {sanitize: true}));
+    update_annotation_text(annotation.annotation_text_id, annotation.content, annotation.id);
```
david-yz-liu, line 300: Leave the `marked` call (you might have removed it because it's being deprecated, but we should keep it for now).

app/controllers/annotations_controller.rb
```diff
@@ -154,6 +154,15 @@ def update
         return
       end
     end
-    @annotation_text.update(content: params[:content])
+
+    if params[:annotation_text] && params[:annotation_text][:change_one] && params[:annotation_text][:change_one] == '1'
+      new_text = @annotation_text.dup
```
david-yz-liu, line 159: Don't `dup` here, explicitly create the new annotation text inside the transaction with the relevant fields.

app/views/annotations/_annotation_modal.html.erb
```diff
@@ -47,9 +47,12 @@
           <%= t('annotations.default_deduction') %>
         </p>
       <% end %>
+      <% if change_one_option %>
+        <br><br><%= f.check_box :change_one, id: 'change_one_annotation' %> <%= title_one %>
```
david-yz-liu, line 51: A couple of things here:

1. Wrap the input and text label in a `p` tag rather than using `<br>`
2. Use an `f.label` to explicitly link the text to the checkbox input.
3. Don't pass `title_one` as a local, just call `t('annotations.update_one')` right here
4. Also, let's reverse the meaning of this checkbox. Call it something like "Change all existing uses of this annotation text." (I guess you can change the name from `change_one` to `change_all`.)

app/views/results/common/_annotations.js.erb
```diff
@@ -410,6 +413,13 @@
     });
     updatePreview('new_annotation_content', 'annotation_preview');
     modal.open();
+
+    $('#submit_annotation').click(() => {
```
david-yz-liu, line 417: Remove this---by making the checkbox more explicit (see above comment), we can drop the `confirm` altogether (and remove the existing string, or adapt it to the new checkbox label).

Also I think adding this caused the submit button to no longer close the modal? You should look into that more if removing this code still doesn't fix that issue.

david-yz-liu (commented):
@robert-tan good work, I left a few different comments.

Also in the UI, it looks like there's a bug where if you have two instances of the same annotation text, and then edit one and select "only change this one", both annotations get changed to refer to the new text. This goes away after refreshing the page (both annotations refer to the correct, distinct annotation texts), but you should look into and fix the problem of the initial update before reloading the page.

app/assets/javascripts/Components/Result/result.jsx
robert-tan: Actually the marked call causes the annotation text hover to be padded with HTML (like if my annotation.content said 'hi', calling marked on it would produce <p>hi</p> and this will also get displayed in the annotation hover exactly (with the HTML tags as text)

PR #4917, opened 2020-10-18
Converts the Ruby Annotation Modals into React Modals

ihasan98:
## Motivation and Context
This allows the creation and editing of annotations to be done client side. Previously the modals would be rendered server side.

## Your Changes
**Description**:
Implements the same functionality as the previous annotation forms, except now they are in React and rendered client side.

- [ ] Other (please specify): 

## Testing

Manual testing through the web interface

A couple of things that still need to be done:
* Deleting previous files that handles modals

## Checklist


app/assets/javascripts/Components/Modals/create_modify_annotation_panel_modal.jsx
```diff
@@ -0,0 +1,132 @@
...
+
+  checkCriterion = (event) => {
+    event.preventDefault();
+    this.handleChange(event);
+    let val = event.target.value;
+    let categories_with_criteria = this.props.categories
+      .filter((category) => category.flexible_criterion_id !== null)
+      .map((category) => category.id);
+
+    if (categories_with_criteria.includes(parseInt(val))) {
```
david-yz-liu, line 47: Use `parseInt(val, 10)` here to specify the base (to avoid problems described https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#Description).

app/assets/javascripts/Components/Modals/create_modify_annotation_panel_modal.jsx
```diff
@@ -0,0 +1,132 @@
...
+      $("#deduction_disclaimer").removeClass("hidden");
```
david-yz-liu, line 48: Since we're using React here, we can remove the jQuery and turn this into state for the modal.

app/assets/javascripts/Components/Result/result.jsx
```diff
@@ -271,15 +310,52 @@ class Result extends React.Component {
...
+              categoryId: "",
+            },
+          })
+          this.refreshAnnotations();
+        }
+      );
+    };
+
+    let annotation = this.state.annotations.find(
+      (annotation) => annotation.id === parseInt(annot_id)
```
david-yz-liu, line 342: Similar comment about `parseInt` as above

app/assets/javascripts/Components/Modals/create_modify_annotation_panel_modal.jsx
```diff
@@ -0,0 +1,132 @@
...
+        content: this.props.content,
+        category_id: this.props.category_id,
+      });
+    }
+  }
+
+  onSubmit = (event) => {
+    event.preventDefault();
+    this.props.onSubmit({ ...this.state });
+    this.setState({ content: "", category_id: "" });
```
david-yz-liu, line 30: This part is interesting. It seems like the parent will reset this component's `content` and `category_id` props after a submit request, making this `setState` unnecessary (I think?).

app/assets/javascripts/Components/Modals/create_modify_annotation_panel_modal.jsx
```diff
@@ -0,0 +1,132 @@
...
+      category_id: this.props.category_id,
+    };
+  }
+
+  componentDidMount() {
+    Modal.setAppElement("body");
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (prevProps !== this.props) {
```
david-yz-liu, line 19: Hmm, in general I prefer just comparing specific prop values (in this case, `content` and `category_id`.

Slightly related: one request we had was to preserve form data if the TA accidentally closed the modal (by pressing Esc or just clicking Cancel instead of Save), so that the next time the modal is opened in those cases, the previous annotation text is still there. I think if you make the change I described here, that should happen---but please test.

app/assets/javascripts/Components/Result/result.jsx
```diff
@@ -29,6 +30,13 @@ class Result extends React.Component {
       feedback_files: [],
       submission_files: {files: [], directories: {}, name: '', path: []},
       fullscreen,
+      annotationModal: {
```
david-yz-liu, line 33: This "initial annotationModal state" is used a couple of places in this file. SInce all the values are static, would be good to pull this out into a constant.

app/models/annotation.rb
```diff
@@ -37,6 +37,7 @@ def get_data(include_creator=false)
       content: annotation_text.content,
       annotation_category:
         annotation_text.annotation_category&.annotation_category_name,
+      category_id: annotation_text.annotation_category&.id,
```
david-yz-liu, line 40: Sorry I missed this one. The name is fine, but in general I'd prefer using the exact attribute names when possible: here that would be `annotation_category_id` instead of `category_id`. In fact, you can access `annotation_text.annotation_category_id` explicitly. :)

app/assets/javascripts/Components/Modals/create_modify_annotation_panel_modal.jsx
ihasan98: Yep, you were right. Though it took me a decent while to identify why this was the case. Really impressed that you identified this at a glance!

PR #4918, opened 2020-10-19
Added file manager for solution files

ANUSHIYAPRIYA:
## Motivation and Context

This is the next step after PR #4861.
The instructor can now upload solution files on the Automated Testing page. 

## Your Changes

**Description**:
Added file manager for the solution files.

## Testing

Manually tested in the UI and also added tests.

## Checklist


app/models/assignment.rb
```diff
@@ -1213,6 +1214,19 @@ def create_autotest_dirs
     FileUtils.mkdir_p self.autotest_files_dir
   end
 
+  def create_autotest_test_student
+    group_name = self.short_identifier + '_test_student_group'
```
mishaschwartz, line 1218: use the same group for all assignments

app/models/assignment.rb
```diff
@@ -1213,6 +1214,19 @@ def create_autotest_dirs
...
+    user_name = self.short_identifier + '_test_student'
```
mishaschwartz, line 1219: use the same TestStudent user for all assignments

app/views/automated_tests/_form.html.erb
```diff
@@ -7,7 +7,10 @@
   <script>
     document.addEventListener('DOMContentLoaded', () => {
       window.autotestManagerComponent = makeAutotestManager(document.getElementById('autotest-manager'),
-        { assignment_id: <%= @assignment.id %>});
+        { assignment_id: <%= @assignment.id %>,
+          grouping_id: <%= @assignment.groupings.joins(student_memberships: :user)
+                                      .where('users.type': 'TestStudent').first.id %>,
+          enableSubdirs: <%= allowed_to? :manage_subdirectories?, @assignment, with: SubmissionPolicy %> });
```
mishaschwartz, line 13: not needed. this permission only applies to actual student submissions. We should never restrict this for the solutions file manager

app/views/automated_tests/_form.html.erb
```diff
@@ -7,7 +7,10 @@
   <script>
     document.addEventListener('DOMContentLoaded', () => {
       window.autotestManagerComponent = makeAutotestManager(document.getElementById('autotest-manager'),
-        { assignment_id: <%= @assignment.id %>});
+        { assignment_id: <%= @assignment.id %>,
+          grouping_id: <%= @assignment.groupings.joins(student_memberships: :user)
```
mishaschwartz, line 11: don't do this here. Find or create the test student and its grouping and store it as an instance variable on the controller that renders this view. Or, create an association to the solution grouping on the assignment itself.

Also, change the prop name to something more descriptive. ex: `solution_grouping_id`

app/models/assignment.rb
```diff
@@ -79,6 +79,7 @@ class Assignment < Assessment
   has_many :starter_file_groups, dependent: :destroy, inverse_of: :assignment, foreign_key: :assessment_id
 
   after_create :create_autotest_dirs
+  after_create :create_autotest_test_student, if: -> { self.type == 'Assignment' }
```
mishaschwartz, line 82: only really needed if tests are enabled for this assignment. Think about a better place to create the relevant objects (maybe in a controller method).

app/assets/javascripts/Components/autotest_manager.jsx
```diff
@@ -316,6 +317,16 @@ class AutotestManager extends React.Component {
             disableActions={{rename: true}}
           />
         </fieldset>
+        <fieldset>
+          <legend><span>{I18n.t("submissions.files")}</span></legend>
+          <SubmissionFileManager
+            readOnly={!this.state.enable_test}
+            assignment_id={this.props.assignment_id}
+            grouping_id={this.props.grouping_id}
+            enableSubdirs={this.props.enableSubdirs}
```
mishaschwartz, line 326: should always be true here

app/assets/javascripts/Components/submission_file_manager.jsx
```diff
@@ -158,7 +158,7 @@ class SubmissionFileManager extends React.Component {
           />
         </div>
       );
-    } else {
+    } else if (!this.props.is_solution_file){
```
mishaschwartz, line 161: If you're trying to disable the file viewer, don't call the renderFileViewer function at all. Otherwise there is going to be the fieldset/legend/span etc. rendered without any content.

app/controllers/submissions_controller.rb
```diff
@@ -346,7 +346,8 @@ def update_files
         return
       end
     else
-      @grouping = @assignment.groupings.find(params[:grouping_id])
+      @grouping = @assignment.groupings.find(params[:grouping_id]) ||
+          @assignment.groupings.joins(student_memberships: :user).where('users.type': 'TestStudent').first
```
mishaschwartz, line 350: This should not be necessary, the `params[:grouping_id]` value should still be correct

config/locales/views/submissions/en.yml
```diff
@@ -55,6 +55,7 @@ en:
                                 Make sure you have worked on the most recent files before submitting."
 
     # Miscellaneous
+    files: "Solution Files"
```
mishaschwartz, line 58: put this in `locales/views/automated_tests` and name it something like `solution_files`, just `files` is not descriptive enough

mishaschwartz (changes requested):

app/assets/javascripts/Components/autotest_manager.jsx
```diff
@@ -316,6 +317,16 @@ class AutotestManager extends React.Component {
...
           />
         </fieldset>
+        <fieldset>
+          <legend><span>{I18n.t("automated_tests.solution_files")}</span></legend>
+          <SubmissionFileManager
+            readOnly={!this.state.enable_test}
+            assignment_id={this.props.assignment_id}
+            grouping_id={this.props.grouping_id}
+            enableSubdirs={true}
+            is_solution_file={true}
```
mishaschwartz, line 327: If the only purpose of this prop is to decide whether or not to render the file viewer, rename the prop to something that describes its function

app/controllers/automated_tests_controller.rb
```diff
@@ -30,6 +30,8 @@ def update
   def manage
     @assignment = Assignment.find(params[:assignment_id])
     @assignment.test_groups.build
+    find_or_create_test_student(params[:assignment_id])
```
mishaschwartz, line 33: why not have the `find_or_create_test_student` method return the grouping and then you don't need the query on the next line. If you do you might want to rename it to `find_or_create_test_grouping` so that its more obvious what the return value is.

app/controllers/automated_tests_controller.rb
```diff
@@ -220,4 +222,17 @@ def required_params
   def assignment_params
     params.require(:assignment).permit(*required_params)
   end
+
+  def find_or_create_test_student(assignment_id)
+    test_group = Group.find_or_create_by(group_name: 'test_student_group')
```
mishaschwartz, line 227: Since the Group.group_name attribute has to be unique, reserve this group name by assigning it to a constant in the Group class and add a validation to make sure that no other group can have this group_name. Otherwise you might end up with a real group that conflicts with this group_name.

Do the same thing below for the TestStudent.user_name.

app/controllers/automated_tests_controller.rb
```diff
@@ -220,4 +222,17 @@ def required_params
...
+    user = TestStudent.find_or_create_by(user_name: 'test_student',
+                                         first_name: 'Test', last_name: 'Student', hidden: true)
+    test_grouping = Grouping.find_by(group_id: test_group.id, assessment_id: assignment_id)
```
mishaschwartz, line 230: You can use `find_or_create_by` here as well and you can use it to find or create the student membership as well.

app/controllers/automated_tests_controller.rb
```diff
@@ -220,4 +222,17 @@ def required_params
   def assignment_params
     params.require(:assignment).permit(*required_params)
   end
+
+  def find_or_create_test_student(assignment_id)
+    test_group = Group.find_or_create_by(group_name: 'test_student_group')
+    user = TestStudent.find_or_create_by(user_name: 'test_student',
+                                         first_name: 'Test', last_name: 'Student', hidden: true)
```
mishaschwartz, line 229: I don't think the hidden is necessary here to create the TestStudent and in case the TestStudent has already been created but the hidden value has been switched, we'll end up accidentally creating a new one that we don't need.

mishaschwartz: Actually, the first_name and last_name could be accidentally changed later on as well. Instead, set the first and last name in the block passed to `find_or_create_by` instead (and do the same for the membership_status when you find or create the StudentMembership below).

app/controllers/automated_tests_controller.rb
```diff
@@ -220,4 +222,17 @@ def required_params
   def assignment_params
     params.require(:assignment).permit(*required_params)
   end
+
+  def find_or_create_test_student(assignment_id)
```
mishaschwartz, line 226: Think about what should happen if any of the objects don't get found or created in this method due to validation errors. What should be returned? What should be displayed to the user? Should successful changes be rolled back?

app/views/automated_tests/_form.html.erb
```diff
@@ -7,7 +7,9 @@
   <script>
     document.addEventListener('DOMContentLoaded', () => {
       window.autotestManagerComponent = makeAutotestManager(document.getElementById('autotest-manager'),
-        { assignment_id: <%= @assignment.id %>});
+        { assignment_id: <%= @assignment.id %>,
+          grouping_id: <%= @test_grouping.id %>
+        });
```
mishaschwartz, line 12: This is a really picky comment but... be consistent with where you place the brackets. See the default style (symmetrical) here: https://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Layout/MultilineHashBraceLayout

app/assets/javascripts/Components/autotest_manager.jsx
```diff
@@ -5,6 +5,7 @@ import Form from 'react-jsonschema-form';
 import Datepicker from './date_picker'
 import FileUploadModal from './Modals/file_upload_modal'
 import AutotestSpecsUploadModal from "./Modals/autotest_specs_upload_modal";
+import {SubmissionFileManager} from "./submission_file_manager";
```
mishaschwartz, line 8: Good re-use of an existing component!

app/assets/javascripts/Components/autotest_manager.jsx
```diff
@@ -316,6 +317,16 @@ class AutotestManager extends React.Component {
             disableActions={{rename: true}}
           />
         </fieldset>
+        <fieldset>
```
mishaschwartz, line 320: This is good! We might decide later that we want to move this to its own page but if we do, it will be easy to just copy and paste this code.

mishaschwartz (changes requested):

app/models/user.rb
```diff
@@ -27,13 +27,17 @@ class User < ApplicationRecord
   validates_uniqueness_of   :id_number, :allow_nil => true
 
   validates_format_of       :type, with: /\AStudent|Admin|Ta|TestServer|TestStudent\z/
+  validates :user_name_of_user_class
```
mishaschwartz, line 30: https://edgeguides.rubyonrails.org/active_record_validations.html#exclusion

app/controllers/automated_tests_controller.rb
```diff
@@ -220,4 +221,29 @@ def required_params
...
+      user.first_name = 'Test'
+      user.last_name = 'Student'
+    end
+    test_groupings = test_student.groupings
+    test_grouping = test_groupings.find_by(assessment_id: assignment_id)
+    return test_grouping unless test_grouping.nil?
+
+    if test_groupings.empty?
+      test_group = Group.new
+      test_group.group_name = test_group.get_autogenerated_group_name
```
mishaschwartz, line 236: Have you tested this? I'm pretty sure you're going to have a problem here because the group doesn't have an `id` value.

app/controllers/automated_tests_controller.rb
```diff
@@ -220,4 +221,29 @@ def required_params
...
+      test_group.save
+    else
+      test_group = test_groupings.first.group
+    end
+    test_grouping = test_group.groupings.create!(assessment_id: assignment_id)
+    StudentMembership.find_or_create_by(user_id: test_student.id, grouping_id: test_grouping.id) do |membership|
+      membership.membership_status = StudentMembership::STATUSES[:inviter]
+    end
+    test_grouping
+  rescue ActiveRecord::RecordInvalid, RecordNotSaved
```
mishaschwartz, line 246: This will only catch an error if the grouping isn't created. What if the TestStudent, Group, or StudentMembership can't be created?

config/locales/views/automated_tests/en.yml
```diff
@@ -19,6 +20,7 @@ en:
       problems: "error"
     title: "Automated Testing"
     display_output_title: "Display test output to students?"
+    no_test_grouping_for_solution_files: "Unexpected error happened while loading solution files."
```
mishaschwartz, line 23: This doesn't really describe the error. Why not display the actual validation error messages instead?

spec/controllers/automated_tests_controller_spec.rb
```diff
@@ -8,14 +8,43 @@
       # TODO: write tests
     end
     context 'GET manage' do
-      before :each do
```
mishaschwartz, line 11: I don't understand why this block is removed and then just repeated a bunch of times below. You can just leave this one in place.

spec/controllers/automated_tests_controller_spec.rb
```diff
@@ -8,14 +8,43 @@
...
+      context 'Getting the test student grouping' do
+        context 'When the assignment already has test grouping' do
+          let(:test_student) { create(:test_student, user_name: User::TEST_STUDENT_USER_NAME) }
+          let(:group) { create(:group, group_name: 'test_student_group') }
+          let(:grouping) { create(:grouping, group: group, assignment: assignment) }
+          let!(:membership) { create(:inviter_student_membership, user: test_student, grouping: grouping) }
+          it 'should return the existing test grouping' do
+            get_as user, :manage, params: params
+            expect(assigns(:test_grouping)).to eq(grouping)
+            expect(assigns(:test_grouping).memberships.first).to eq(membership)
```
mishaschwartz, line 31: this should be a separate test

spec/controllers/automated_tests_controller_spec.rb
```diff
@@ -8,14 +8,43 @@
...
+            get_as user, :manage, params: params
+          end
+          it 'should return a newly created test grouping' do
+            expect(assigns(:test_grouping)).to be_truthy
+          end
+          it 'should return a grouping which belongs to a test student' do
+            user_id = assigns(:test_grouping).memberships.first.user_id
+            user = User.find(user_id)
+            expect(user.is_a?(TestStudent)).to be true
+            expect(user.groupings.find_by(assessment_id: assignment.id)).to eq(assigns(:test_grouping))
```
mishaschwartz, line 45: this should be a separate test

app/controllers/automated_tests_controller.rb
```diff
@@ -220,4 +221,29 @@ def required_params
...
+    flash_now(:error, I18n.t('automated_tests.no_test_grouping_for_solution_files'))
+  end
```
mishaschwartz, line 248: What does this method return if an error occurs?

spec/controllers/automated_tests_controller_spec.rb
```diff
@@ -8,14 +8,43 @@
...
+            expect(assigns(:test_grouping)).to eq(grouping)
+            expect(assigns(:test_grouping).memberships.first).to eq(membership)
+          end
+        end
+        context 'When there is no test grouping exists for the assignment' do
+          before :each do
+            get_as user, :manage, params: params
+          end
+          it 'should return a newly created test grouping' do
+            expect(assigns(:test_grouping)).to be_truthy
```
mishaschwartz, line 39: `be_truthy` is too broad of a check and I'm pretty sure will always be true even if an error occurs.

spec/controllers/automated_tests_controller_spec.rb
```diff
@@ -8,14 +8,43 @@
...
+        it 'User should be able to view the Automated Testing manage page' do
+          expect(response.status).to eq(200)
+        end
+        it 'should render the assignment_content layout' do
+          expect(response).to render_template('layouts/assignment_content')
+        end
       end
-      it 'should render the assignment_content layout' do
-        expect(response).to render_template('layouts/assignment_content')
+      context 'Getting the test student grouping' do
```
mishaschwartz, line 22: Your tests need to cover the following cases:

- the test student exists
    - the grouping exists
        - the membership exists
        - the membership does not exist
    - the grouping does not exist
        - the group exists
        - the group does not exist
- the test student does not exist

In every case where a new object needs to be created, you also need a test for what happens if the creation fails because of a validation error.

app/models/user.rb
```diff
@@ -27,6 +30,8 @@ class User < ApplicationRecord
   validates_uniqueness_of   :id_number, :allow_nil => true
 
   validates_format_of       :type, with: /\AStudent|Admin|Ta|TestServer|TestStudent\z/
+  validates :user_name, exclusion: { in: [TEST_STUDENT_USER_NAME] }, if: -> { self.class != TestStudent }
```
mishaschwartz, line 33: use `validates_exclusion_of`

spec/models/user_spec.rb
```diff
@@ -46,6 +46,42 @@
     end
   end
 
+  context 'Creating user with user name test_student' do
```
mishaschwartz, line 49: you can replace most of this with calls to:

```ruby
it { is_expected.to validate_exclusion_of(:user_name).in_array([User::TEST_STUDENT_USER_NAME]) }
```

and just change the `User.type` of the [`subject`](https://github.com/thoughtbot/shoulda-matchers#on-the-subject-of-subject)

mishaschwartz: If you do this, also move it to the top of the file with the rest of the validation tests.

mishaschwartz (changes requested):

spec/controllers/automated_tests_controller_spec.rb
ANUSHIYAPRIYA: For the test case "When the test grouping, membership already exists" I had to create everything before calling the "get :manage" 
If I only one before block for all the test cases that will create the test_grouping, membership already. So I can't test the mentioned test case.

app/controllers/automated_tests_controller.rb
```diff
@@ -220,4 +221,32 @@ def required_params
...
+    else
+      test_group = test_groupings.first.group
+    end
+    test_grouping = test_group.groupings.create!(assessment_id: assignment_id)
+    StudentMembership.find_or_create_by!(user_id: test_student.id, grouping_id: test_grouping.id) do |membership|
+      membership.membership_status = StudentMembership::STATUSES[:inviter]
+    end
+    test_grouping
+  rescue ActiveRecord::RecordInvalid, RecordNotSaved => e
+    flash_now(:error, 'Something unexpected happened' + e.message)
```
mishaschwartz, line 249: make this an I18n string

spec/controllers/automated_tests_controller_spec.rb
mishaschwartz: in that case please use `let!` variables instead of repeating the before block

mishaschwartz: you know what, never mind, this is fine

mishaschwartz (approved):

Issue #4916, opened 2020-10-17
Remove preview for compressed archives


robert-tan:
Very minor feature but had to submit an assignment with a .tar.gz extension and the preview with random characters shows up, which isn't useful/necessary:

![image](https://user-images.githubusercontent.com/46432393/96346496-07620280-1094-11eb-9552-da35f2449fb9.png)

PR #4920, opened 2020-10-20
Easy Fix: Remove preview for compressed archives

ANUSHIYAPRIYA:
## Motivation and Context

This resolves issue #4916.

## Your Changes

**Description**:
Modified the views so that the file viewer in the repo browser won't display the preview for compressed archives.

## Testing

Manually tested in the UI

## Checklist

### DIFF
--- a/app/assets/javascripts/Components/Result/file_viewer.jsx
+++ b/app/assets/javascripts/Components/Result/file_viewer.jsx
@@ -147,15 +147,16 @@ export class FileViewer extends React.Component {
         annotationFocus={this.props.annotationFocus}
         {...commonProps}
       />;
-    } else if ((this.state.type !== '') && (this.state.type !== 'unknown')) {
+    } else if (!['binary', '', 'unknown'].includes(this.state.type)) {
       return <TextViewer
         type={this.state.type}
         content={this.state.content}
         focusLine={this.props.focusLine}
         {...commonProps}
       />;
-    } else {
+    } else if (this.state.type === 'binary'){
       return (I18n.t('submissions.cannot_display'));
-    }
+    } else
+      return ''
   }
 }
--- a/app/assets/javascripts/Components/Result/file_viewer.jsx
+++ b/app/assets/javascripts/Components/Result/file_viewer.jsx
@@ -147,7 +147,7 @@ export class FileViewer extends React.Component {
         annotationFocus={this.props.annotationFocus}
         {...commonProps}
       />;
-    } else if (!['binary', '', 'unknown'].includes(this.state.type)) {
+    } else if (!['binary', ''].includes(this.state.type)) {
       return <TextViewer
         type={this.state.type}
         content={this.state.content}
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/Components/Result/file_viewer.jsx
```diff
@@ -147,15 +147,15 @@ export class FileViewer extends React.Component {
         annotationFocus={this.props.annotationFocus}
         {...commonProps}
       />;
-    } else if (this.state.type !== '') {
+    } else if ((this.state.type !== '') && (this.state.type !== 'unknown')) {
```
mishaschwartz, line 150: As discussed in the meeting. Only if a file is binary (not the same as having an unknown MIME type) the viewer should display the `'submissions.cannot_display'` message. We would like the same behaviour as when the file is viewed on the results page.

Please see `SubmissionsController.get_file` and `SubmissionFile.is_binary` for reference.

mishaschwartz (changes requested):

app/assets/javascripts/Components/Result/file_viewer.jsx
ANUSHIYAPRIYA: Yes Misha, But we are setting the type to 'unknown' for binary files. 

submissions_controller.rb: line no: 444 to 446

I can add one more condition which checks for this type and display the 'submissions.cannot_display' message. And for the remaining type just return empty.

Why am I checking the unknown mime type in line no: 150 is, previously the Text Viewer including the binary files also.
So I eliminated that by checking the mime type should not be 'unknown' for text files.

mishaschwartz: The `get_file` method does check that and sets the `type` to `'unknown'` but that method isn't being called here.  The content that is displayed is from the `download` method, not the `get_file` method. This is why the zip content is being displayed in the first place.

app/controllers/submissions_controller.rb
```diff
@@ -718,6 +718,16 @@ def get_all_file_data(revision, grouping, path)
         data[:key] = file_name
         data[:modified] = data[:last_revised_date]
         data[:revision_by] = '' if anonymize
+        if data[:type] == 'unknown'
```
david-yz-liu, line 721: It's a bit awkward to check this here, modify `get_file_info` above instead.

app/assets/javascripts/Components/Result/file_viewer.jsx
```diff
@@ -147,15 +147,16 @@ export class FileViewer extends React.Component {
...
       return <TextViewer
         type={this.state.type}
         content={this.state.content}
         focusLine={this.props.focusLine}
         {...commonProps}
       />;
-    } else {
-      return '';
-    }
+    } else if (this.state.type === 'binary'){
```
david-yz-liu, line 157: Move this check after the previous one, that way you won't have to change any of the other code.

app/controllers/submissions_controller.rb
ANUSHIYAPRIYA: Misha suggested a simple solution for this instead of modifying the get_file_info method and making it more complex. I have updated the PR based on that.

PR #4934, opened 2020-10-27
Update starter file handling

mishaschwartz:
## Motivation and Context

After the first semester using the new starter files, the following improvements are suggested to the way we handle starter files:

1. starter files should be available from a single source so that instructors can be (more) sure that students have access to the most up to date version of the starter files.
2. when starter files are assigned randomly and then changed later on, students who already have starter files assigned should get the updated versions of the files they already have. The previous behaviour is that starter files were randomly assigned all over again. This can lead to students being given access to multiple (possibly outdated) versions of the starter files. 
3. building on 1. (above), students should be warned if they haven't downloaded the most up to date starter files when they try to upload submission files (this is made possible because they will have one source for those files). 

## Your Changes

**Description**:

- [x] assign starter files to groupings when the groupings are created but *don't* populate the repos with the starter files
- [x] if starter files change, try to give students the new version of files they already have been assigned (instead of randomly selecting entirely new files)
- [x] if a student tries to submit code when they have not yet downloaded the most recent version of the starter code: display an alert warning them that they might be uploading the wrong file version

## Checklist


app/models/grouping.rb
```diff
@@ -445,12 +446,17 @@ def select_starter_file_entries
       assignment.default_starter_file_group&.starter_file_entries || []
     when 'shuffle'
       assignment.starter_file_groups.includes(:starter_file_entries).map do |g|
-        StarterFileEntry.find_by(id: g.starter_file_entries.ids.sample)
+        # If this grouping has previous starter files, try to choose an entry with the same path as before
+        old_entry = g.starter_file_entries.where(path: self.starter_file_entries.map(&:path)).sample
```
david-yz-liu, line 450: The use of `sample` here is pretty non-obvious. I think I know what you're trying to do but it would be great to leave a more explanatory comment explaining the use. (Personally I would have just used a `find_by`, but you can leave the `sample` if you want.)

app/models/grouping.rb
```diff
@@ -756,14 +749,14 @@ def student_test_run_in_progress?
 
   def access_repo
     group.access_repo do |repo|
-      add_starter_files(repo)
+      add_assignment_folder(repo)
       yield repo
```
david-yz-liu, line 753: Please update this line to yield the repo if a block is given (this should allow you to call `grouping.access_repo` instead of `grouping.access_repo {}`).

PR #4936, opened 2020-10-28
Rename the parameter in get_file_info from id to assignment_id

ANUSHIYAPRIYA:
## Motivation and Context

The helper method get_file_info in submissions_helper uses the id to set up the URL. But it should use the assignment_id because the download route looking for the params[:assignment_id].

## Your Changes

**Description**:
Renamed 'id' into 'assignment_id'

## Testing

Added tests for the changes made.

## Checklist


spec/helpers/submission_helper_spec.rb
```diff
@@ -37,4 +37,30 @@
       end
     end
   end
+
+  describe '#get_file_info' do
+    let(:assignment) { create(:assignment) }
+    let(:grouping) { create(:grouping, assignment: assignment) }
+    let(:file_name) { 'test.zip' }
+    before do
+      grouping.group.access_repo do |repo|
```
mishaschwartz, line 46: use `grouping.access_repo` directly

mishaschwartz: Actually, you don't need to actually submit a file to the repo to test this method. You can create a RevisionFile object directly for use in your tests. That way, this test will fail only if there is an issue with the `get_file_info` method, not if there is an issue with repo access/setup.

spec/helpers/submission_helper_spec.rb
```diff
@@ -37,4 +37,30 @@
...
+      file_info = {}
+      grouping.group.access_repo do |repo|
+        revision = repo.get_latest_revision
+        full_path = File.join(grouping.assignment.repository_folder, '')
+        file_obj = revision.tree_at_path(full_path)[file_name]
+        dirname, basename = File.split(file_name)
+        dirname = '' if dirname == '.'
+        file_info = get_file_info(basename, file_obj, assignment.id, revision.revision_identifier, dirname, grouping.id)
+      end
+      expect(file_info[:url].include?(assignment.id.to_s)).to be true
```
mishaschwartz, line 63: Parse the url and make sure the assignment_id is actually correct. `include?` over the whole url string is too general.

mishaschwartz (changes requested):

mishaschwartz (dismissed):

PR #4938, opened 2020-10-29
Time zone setting

robert-tan:
## Motivation and Context

Adds the ability to configure time zone preferences for MarkUs. Useful for those not currently located in Toronto.

## Your Changes

**Description**:
* Added a time zone column to Users
* Added front end for time zone setting
* Added an `around_action` callback for every request to use client timezone

## Testing

TO DO!

## Checklist


app/controllers/application_controller.rb
```diff
@@ -18,6 +18,8 @@ class ApplicationController < ActionController::Base
 
   helper :all # include all helpers in the views, all the time
 
+  # set user time zone based on their settings
+  around_action :use_time_zone
```
david-yz-liu, line 22: Use `around_action :set_time_zone, if: :current_user` here (to avoid the extra `yield` down in the method).

config/locales/views/users/en.yml
```diff
@@ -45,3 +44,5 @@ en:
       reset: "Reset my API key"
       unavailable: "Unavailable"
     verify_settings_update: 'Your settings have been updated!'
+    time_zone: "Time Zone:"
```
david-yz-liu, line 47: This should be an attribute name (under `locales/models/users`). Look up how we use `human_attribute_name` elsewhere in the code.

db/migrate/20201029155930_add_timezone_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddTimezoneToUsers < ActiveRecord::Migration[6.0]
+  def change
+    add_column :users, :time_zone, :string, null: false, default: 'Eastern Time (US & Canada)'
```
david-yz-liu, line 3: Don't set the default in the schema. Instead, set a callback in the user model, and make the default based on `config.time_zone`.

spec/controllers/users_controller_spec.rb
```diff
@@ -30,28 +30,26 @@
     include ERB::Util
     RSpec.shared_examples 'changing particular mailer settings' do
       it 'can be enabled in settings' do
-        student.update!(setting => false)
+        student.update!(receives_invite_emails: false)
```
david-yz-liu, line 33: I'm not sure why you made these changes, but the point of this code is to allow us to use the same examples with multiple different settings. You should revert these changes (and the same for `settings` down below).

PR #4994, opened 2020-11-26
Add display name for admins

robert-tan:
## Motivation and Context

The current admins view under Users -> Admins doesn't work due to a bug in display name (thanks @ignaspa for catching it). This fix enables the display name settings for admins too so the view would work and admins can set their display names.

## Your Changes

**Description**:
- Make some `has_attribute?` checks in the user model
- Refactor settings to not just work for Students

## Testing

Unit tests and manually tested

## Checklist


config/locales/views/users/en.yml
```diff
@@ -38,6 +36,8 @@ en:
 
   # Miscellaneous
   users:
+    display_name: 'Display name'
```
david-yz-liu, line 39: @robert-tan I didn't notice this earlier, but this key should actually be under `models/users/en.yml` as an attribute.

Then you can also update the form to use `User.human_attribute_name`

PR #4996, opened 2020-11-26
Add localization setting

robert-tan:
## Motivation and Context

Move locale to the settings page and store it in the Users table in the backend

## Your Changes

**Description**:
- Moved locale selector to settings
- Added locale column in Users
- Set locale to use Users setting

## Testing

Manually tested

## Checklist

### DIFF
--- a/app/views/users/settings.html.erb
+++ b/app/views/users/settings.html.erb
@@ -18,8 +18,10 @@
       <%= f.check_box :receives_invite_emails %>
       <%= f.label :receives_invite_emails, t('students.toggle_invite_notifications') %>
     </p>
-    <%= f.submit t('students.update_settings') %>
   <% end %>
+  <p>
+    <%= f.submit t('users.update_settings') %>
+  </p>
 <% end %>
 
 <% if allowed_to?(:reset_api_key) %>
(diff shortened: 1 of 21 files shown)


app/views/users/settings.html.erb
```diff
@@ -25,41 +31,43 @@
 <% end %>
 
 <% if allowed_to?(:manage?, KeyPair) %>
-  <h2><%= t('key_pairs.key_pairs_title') %></h2>
+  <p>
```
david-yz-liu, line 34: Revert this change (starting from this line and extending below). There's more appropriate ways of adding some whitspace, if that's what you were trying to do, and either way it shouldn't be in this PR.

app/views/users/settings.html.erb
```diff
@@ -1,6 +1,10 @@
 <% content_for :title, t('students.account_settings') %>
-<% if @current_user.student? %>
-  <%= form_with url: 'update_settings', method: 'patch', model: @current_user, local: true do |f| %>
+<%= form_with url: 'update_settings', method: 'patch', model: User, local: true do |f| %>
+  <p>
+    <%= f.label :locale, t(:language) %>
```
david-yz-liu, line 4: You can check for other uses of the `:language` string, but now we should move it to `models/users/en.yml` under `attributes` as well.

PR #5013, opened 2020-12-02
Show most recent test run when released

mishaschwartz:
## Motivation and Context

Right now, the student can see all test group results when an assignment is released. Really only the most recent one for each test group is relevant since that is the one used to fill in mark criteria. Displaying past results is confusing to the student.

Instructors can avoid this by re-collecting in between each test run but that is time consuming and has the unfortunate side effect of wiping out any marks not associated with the test runs as well.

This PR changes the student's view of released test runs so that they only see the most recent one. This has the following advantages:

- Students don't see multiple test group results per test group (old ones are irrelevant to them)
- Instructors don't have to recollect to hide old results from students

## Your Changes

**Description**:

- filter the test runs served from the `test_runs_instructors_released` route so that only those from the most recent test group result per test group are returned 

## Testing

- added tests to ensure that only the most recent test run data is returned

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,8 +1,9 @@
 # Changelog
 ## [unreleased]
 - Fix bug where duplicate marks can get created because of concurrent requests (#5018)
-- Only display latest test run data to students viewing results from an released assignment (#5013)
+- Only display latest results for each test group to students viewing results from an released assignment (#5013)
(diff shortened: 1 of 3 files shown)


david-yz-liu:
@mishaschwartz this looks good to me! You should do a merge from master so that the Changelog is updated appropriately.

PR #5022, opened 2020-12-14
Set moment.locale properly

mishaschwartz:
## Motivation and Context

We don't set the locale in a dom element anymore. Instead it is set in a script on the I18n.locale attribute.

## Your Changes

**Description**:

Get the locale info properly instead of trying to get a DOM element that no longer exists

## Checklist

### DIFF
--- a/app/views/assignments/_boot.js.erb
+++ b/app/views/assignments/_boot.js.erb
@@ -117,9 +117,6 @@
   }
 
   $(document).ready(function() {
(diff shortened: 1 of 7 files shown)


david-yz-liu (commented):
This change is good, but let's actually take the opportunity to remove the code duplication and set `moment.locale` in `app/views/layouts/_i18n_config.html.erb`.

(Also there's now an unused localization string.)

PR #5032, opened 2020-12-21
Jupyter notebook render

mishaschwartz:
## Motivation and Context

 It would be nice to render jupyter notebooks... this PR does that.

## Your Changes

**Description**:

- added python dependency to render notebooks as html using nbconvert
- cache previously rendered notebooks on disk
- add helper to select range so that we can enable annotations for these files later on

## Testing

- added tests

## Checklist

- add description of jupyter notebook conversion

### DIFF
--- a/app/assets/javascripts/Components/Result/jupyter_notebook_viewer.jsx
+++ b/app/assets/javascripts/Components/Result/jupyter_notebook_viewer.jsx
@@ -1,10 +1,25 @@
 import React from "react";
+import {markupTextInRange} from "../Helpers/range_selector";
 
 export class JupyterNotebookViewer extends React.Component {
-  // TODO: create annotation layer and link annotations to html elements in the iframe
+
+  elemMarkup = (node) => {
+    const elem = document.createElement('span');
+    elem.style.backgroundColor = 'red';
+    return elem;
+  }
+
+  getSelection = () => {
+    const iframe = document.getElementById("jupyter-notebook");
+    const target = (iframe.contentWindow || iframe.contentDocument);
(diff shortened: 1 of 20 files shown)


app/models/submission_file.rb
```diff
@@ -88,6 +90,10 @@ def is_pdf?
     File.extname(filename).casecmp('.pdf') == 0
   end
 
+  def pynb?
```
david-yz-liu, line 93: I'd prefer `is_pynb?` here to be consistent with the other methods.

david-yz-liu (commented):
Looks great. I left a few very minor comments. I didn't review the annotation js code, as we'll enable that later.

Just a reminder that you're also adding some tests to cover some of the added lines.

app/models/submission_file.rb
mishaschwartz: It is consistent with other methods but not with the rubocop checks. We should actually swap the rest of them over to this style or if not, then disable the cop. I'm happy with either.

david-yz-liu: Ah interesting, you can disable the cop then.

app/assets/javascripts/Components/Result/jupyter_notebook_viewer.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+  constructor() {
+    super();
+    this.state = {
+      annotations: []
+    }
+  }
+
+  getSelection = () => {
+    const iframe = document.getElementById("jupyter-notebook");
```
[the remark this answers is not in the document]
mishaschwartz: sure, noted for later

PR #5033, opened 2020-12-22
Allow user to view binary file as plain text if they want

mishaschwartz:
## Motivation and Context

Sometimes student submissions are categorized by MarkUs as a binary files even if they are not. For example, students can submit the output of running gdb which often looks a lot like a binary file to MarkUs even if that is not the intention. 
While grading, it would be nice to have to option to say that a file is not actually binary and display it anyway.

Note: this was a requested feature from csc209

## Your Changes

**Description**:

- add new BinaryViewer which shows an additional link to download the view the binary file as text anyway
- add a new param to the `get_file` route which specifies that a binary file should be downloaded as text anyway

## Testing

- tested in the UI
- we currently have no tests for the `get_file` route so I added a TODO to add those later

## Checklist

- describe this additional option in the Grading View page

### DIFF
--- a/config/locales/views/submissions/en.yml
+++ b/config/locales/views/submissions/en.yml
@@ -56,8 +56,8 @@ en:
 
     # Miscellaneous
     cannot_display: "Cannot display this file. Please download it to view."
-    get_anyway: "Try to render this file anyway?"
     commit_date: "Submission Date"
+    get_anyway: Try to render this file anyway?
     grace_credits_used: "Grace Credits Used"
     grading_can_begin: "Grading can begin"
     grading_can_begin_after: "Grading can begin after %{time}"
--- a/config/locales/views/submissions/en.yml
+++ b/config/locales/views/submissions/en.yml
@@ -56,6 +56,7 @@ en:
 
     # Miscellaneous
     cannot_display: "Cannot display this file. Please download it to view."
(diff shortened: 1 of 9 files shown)


config/locales/views/submissions/en.yml
```diff
@@ -56,6 +56,7 @@ en:
 
     # Miscellaneous
     cannot_display: "Cannot display this file. Please download it to view."
+    get_anyway: "Try to render this file anyway?"
```
david-yz-liu, line 59: I'd like to actually keep the keys in alphabetical order if possible. If you could move this below `commit_date` that would be great!

david-yz-liu (commented):
I left one small comment. I also think you're now adding full tests for `get_file` in another PR, so no need to add tests here?

mishaschwartz:
I'd like to add tests here but I'm going to wait until #5032 is merged so I can use the structure I've set up there for the tests.

config/locales/views/submissions/en.yml
mishaschwartz: That's fine... but if we want to alphebetize the keys then lets just run `i18n-tasks normalize` in a different PR and be better about keeping it sorted from now on.

mishaschwartz: #5056

PR #5054, opened 2021-01-04
test-results: Update test results table UI.

david-yz-liu:
## Motivation and Context

When viewing test results, the "test output" column is often too small compared to the size of the output, making it hard to read.

## Your Changes

**Description**: Made the test output data a (`react-table`) subcomponent of each test result row. Updated the CSS so that the "expand" arrow only displays when there is some test output, and test output should be shown (this can be toggled for students).

I also condensed the "Mark" and "Total" columns into a single column, to make more room for the test name/description in each row.

## Testing

Tested the new view in both Firefox and Chrome.

## Checklist


app/assets/javascripts/Components/test_run_table.jsx
```diff
@@ -150,39 +150,35 @@ class TestGroupResultTable extends React.Component {
...
     },
     {
       id: 'marks_earned',
       Header: I18n.t('activerecord.attributes.test_result.marks_earned'),
       accessor: row => row['test_results.marks_earned'],
-      minWidth: 40,
+      Cell: row => {
+        const marksEarned = row.original['test_results.marks_earned'];
+        const marksTotal = row.original['test_results.marks_total'];
+        return `${marksEarned} / ${marksTotal}`;
```
mishaschwartz, line 169: marksEarned or marksTotal can be null if the result hasn't returned yet. I would recommend replacing this with:

```js
if (marksEarned !== null && marksTotal !== null) {
   return `${marksEarned} / ${marksTotal}`;
}
return '';
```

PR #5061, opened 2021-01-12
Move configurable settings to yml files served by rubyconfig

mishaschwartz:
## Motivation and Context

[Rubyconfig](https://github.com/rubyconfig/config) is a gem that parses yaml files for (environment) configuration options. It also has builtin environment variable support.

By using this gem to set configuration options we can improve/simplify deployment:

- local configuration options can be specified with a `settings.local.yml` file (instead of editing the environment specific config files)
- local configuration options can be specified with environment variables
- clear separation between what options should be configurable and which should not

## Your Changes

**Description**:

- install the `rubyconfig` gem
- create default `settings.yml` file and environment specific `production.yml`, `test.yml`, `development.yml` files
    - these environment specific files contain the same content as in `production.rb`, etc. but in a yml format that can be read by `rubyconfig`
- change environment variables in `docker-compose.yml` to support the new environment variable format. For example:

```
MARKUS__REDIS__URL=redis://localhost:6379/1
```  
is equivalent to setting this in a yml file:
```yaml
redis:
  url: 'redis://localhost:6379/1'
```
- update `application.rb` to parse the settings from `rubyconfig` and use them to set values in the `Rails.application.config` object.
- move configuration options that require execution of ruby code into initializer files:
    - `better_errors.rb`: in development only, add 0.0.0.0 as allowed IP (previously in `development.rb`)
    - `compressors.rb`: (new) in production only, set javascript and css compressors (previously in `production.rb`)
    - `i18n_js.rb`: set available/default locales (previously in `*.rb` files)
    - `responders.rb`: set `flash_keys` (previously in `application.rb`). (This is related to #3680 but sets the values properly on the responders object instead of on a configuration object which is ignored/set too late)
- change all calls to `Rails.configuration` and `Rails.configuration.x` to calls to the new `Settings` object
- empty `production.yml`, `test.yml`, `development.yml` files. These are still available to set environment specific configuration options that don't belong/can't be set in yml files. Note that any configuration options set here will NOT be settable by environment variables and/or settings.local  files.
- some minor changes to the code required to handle type conversions (usually string to symbol). For example, see `app/models/student.rb` and `app/models/ta.rb`
- added a default queue name that is used by background jobs by default.

- [x] Other (please specify): refactoring configuration options that does not change functionality but will break setup procedures

## Testing

Updated some rspec tests
Click-through in UI to make sure changes to views are respected

## Checklist

https://github.com/MarkUsProject/Markus/wiki/Pending-Changes#pr-5061


app/models/user.rb
```diff
@@ -81,13 +81,13 @@ def self.authenticate(login, password, ip: nil)
...
+      pipe = IO.popen("'#{Settings.validate_file}'", 'w+') # quotes to avoid choking on spaces
       to_stdin = [login, password, ip].reject(&:nil?).join("\n")
-      pipe.puts(to_stdin) # write to stdin of Rails.configuration.validate_file
+      pipe.puts(to_stdin) # write to stdin of Settings.validate_file
       pipe.close
       m_logger = MarkusLogger.instance
-      if !Rails.configuration.validate_custom_exit_status.nil? &&
-          $?.exitstatus == Rails.configuration.validate_custom_exit_status
+      if !Settings.validate_custom_exit_status.nil? &&
+          $?.exitstatus == Settings.validate_custom_exit_status
```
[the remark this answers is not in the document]
mishaschwartz: vote to ignore

mishaschwartz:
- job queue names can have "job" at the end
- should we put all possible settings in `settings.yml` (with null values if they should be set by environment yml files)?

config/application.rb
```diff
@@ -47,9 +38,80 @@ class Application < Rails::Application
...
+      key: Settings.rails.session_store.args.key,
+      path: Settings.rails.session_store.args.path,
+      expire_after: Settings.rails.session_store.args.expire_after.days,
+      secure: Settings.rails.session_store.args.secure,
+      same_site: Settings.rails.session_store.args.same_site.to_sym
+    )
+
+    # Email notifications
+    config.action_mailer.delivery_method = Settings.rails.action_mailer.delivery_method.to_sym
+    config.action_mailer.smtp_settings = Settings.rails.action_mailer.smtp_settings.to_h
```
david-yz-liu, line 71: This is interesting, why is the `to_h` necessary here? (What's the value of `Settings.rails.action_mailer.smtp_settings`?)

config/application.rb
```diff
@@ -47,9 +38,80 @@ class Application < Rails::Application
...
+    config.action_mailer.default_url_options = Settings.rails.action_mailer.default_url_options.to_h
+    config.action_mailer.asset_host = Settings.rails.action_mailer.asset_host
+    config.action_mailer.perform_deliveries = Settings.rails.action_mailer.perform_deliveries
+    config.action_mailer.deliver_later_queue_name = Settings.rails.action_mailer.deliver_later_queue_name
+
+    # Print deprecation notices to stderr.
```
david-yz-liu, line 77: I think it's good to move comments like this to the Wiki page and remove them from this file. (Many of these comments are assuming a particular setting value, which is misleading now.)

config/settings.yml
```diff
@@ -0,0 +1,91 @@
...
+      expire_after: 21 # days
+      secure: false
+      same_site: lax
+  action_mailer:
+    delivery_method: test
+    default_url_options:
+      host: 'localhost:3000'
+    asset_host: 'http://localhost:3000'
+    perform_deliveries: false
+    deliver_later_queue_name: <%= ENV.fetch('MARKUS__QUEUES__DEFAULT') { 'DEFAULT_QUEUE' } %> # Settings.queues.default
```
david-yz-liu, line 22: I see what you're doing here. I have a slight preference for making this setting optional, and have `application.rb` set the default queue name if this is `nil`. Feels more consistent with our treatment of the `job` queue settings.

david-yz-liu (commented):
Neat! Just left a few comments/questions, but overall I'm definitely happy with this.

db/data/autotest_files/jupyter/script_files/jupyter_test.ipynb
```diff
@@ -0,0 +1,126 @@
+{
```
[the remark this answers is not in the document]
mishaschwartz: oops

config/application.rb
mishaschwartz: `config.action_mailer.smtp_settings` requires a hash object. `Settings.rails.action_mailer.smtp_settings` is either a `Config` struct or `nil`. In either case we end up with a properly formatted hash (`nil.to_h  == {}`)

PR #5082, opened 2021-01-24
Partial load results

robert-tan:
## Motivation and Context

Make "Previous" and "Next" buttons on the submission results page do a partial reload rather than re-rendering the entire component. Aimed to improve load times when grading submissions.

## Your Changes

**Description**:
Modified `next_grouping` route to render a JSON object of relevant info instead of the next/previous submission page. Modified Result React component to use updated info on click to re-populate page.

## Testing

Manually tested - creating annotations, updating marks, releasing results.

## Checklist


app/assets/javascripts/Components/Result/feedback_file_panel.jsx
```diff
@@ -46,6 +46,10 @@ export class FeedbackFilePanel extends React.Component {
         {feedback_file_id: this.state.selectedFile}
       );
       file_obj = this.props.feedbackFiles.find(file => file.id === this.state.selectedFile);
+
+      if (!file_obj) {
+        file_obj = this.props.feedbackFiles[0];
```
david-yz-liu, line 51: Since this is in response to change results, I'm guessing `this.props.feedbackFiles` could be empty.

But also, I think a better approach would be to use a React lifecycle hook to update `this.state.selectedFile` when `this.props.feedbackFiles` changes (that keeps the props + state in sync, before the render).

app/assets/javascripts/Components/Result/result.jsx
```diff
@@ -27,6 +27,13 @@ class Result extends React.Component {
       submission_files: {files: [], directories: {}, name: '', path: []},
       fullscreen: false,
       annotationModal: INITIAL_ANNOTATION_MODAL_STATE,
+      props: {
```
david-yz-liu, line 30: It actually is a bit confusing to use "props" as a state key. Better to flatten: keep the "_id" parts as just top-level keys in the state, and actually `role` can remain a prop, it doesn't need to be a state since it'll never change.

app/assets/javascripts/Components/Result/result.jsx
```diff
@@ -43,12 +50,21 @@ class Result extends React.Component {
     });
   }
 
+  componentDidUpdate(prevProps, prevState) {
+    if (
+      this.state.props.submission_id !== prevState.props.submission_id ||
```
david-yz-liu, line 55: Let's actually only check for changing `result_id`. The way our data is set up, if the submission id changes then the result id will also change.

app/assets/javascripts/Components/Result/result.jsx
```diff
@@ -541,6 +557,33 @@ class Result extends React.Component {
     });
   };
 
+  nextSubmission = (direction) => () => {
+    const url = Routes.next_grouping_assignment_submission_result_path(
+      this.state.props.assignment_id, this.state.props.submission_id, this.state.props.result_id
+    );
+
+    $.ajax({
+      url: `${url}?direction=${direction}`
```
david-yz-liu, line 566: In general prefer passing in a `data` option rather than constructing the URL string manually.

app/assets/javascripts/Components/Result/result.jsx
```diff
@@ -541,6 +557,33 @@ class Result extends React.Component {
...
+    }).then((result) => {
+      if (!result.next_result || !result.next_grouping) {
+        alert('There are no more results in this direction!');
```
david-yz-liu, line 569: Make sure to internationalize this string.

app/assets/javascripts/Components/Result/result.jsx
```diff
@@ -541,6 +557,33 @@ class Result extends React.Component {
...
+        ...this.state.props,
+        result_id: result.next_result.id,
+        submission_id: result.next_result.submission_id,
+        grouping_id: result.next_grouping.id,
+      };
+      this.setState(prevState => ({ ...prevState, props: result_obj }));
+      let new_url = Routes.edit_assignment_submission_result_url(
+        this.state.props.assignment_id, this.state.props.submission_id, this.state.props.result_id
+      );
+      history.pushState({}, document.title, new_url)
```
david-yz-liu, line 583: Nice---does this update the browser URL bar as well?

app/assets/javascripts/Components/Result/result.jsx
```diff
@@ -541,6 +557,33 @@ class Result extends React.Component {
     });
   };
 
+  nextSubmission = (direction) => () => {
```
david-yz-liu, line 560: I see what you're doing here but it's probably better not to use the nested lambdas like this (since we don't do it elsewhere in the code), and instead wrap the callback in a thunk below.

app/assets/javascripts/Components/Result/submission_file_panel.jsx
```diff
@@ -72,6 +72,10 @@ export class SubmissionFilePanel extends React.Component {
       }
       this.setState({selectedFile});
     }
+
+    if (prevProps !== this.props) {
```
david-yz-liu, line 76: Hm I don't think it's good to do a full prop comparison. Perhaps just the `result_id`?

app/controllers/results_controller.rb
```diff
@@ -320,7 +320,7 @@ def next_grouping
         next_pr = assigned_prs.where('peer_reviews.id < ?', result.peer_review_id).last
       end
       next_result = Result.find(next_pr.result_id)
-      redirect_to action: 'edit', id: next_result.id
+      render json: { next_result: next_result, next_grouping: next_pr }
```
david-yz-liu, line 323: It would be great to not duplicate the code here. (Set `next_grouping` and `next_result` in both branches, and then keep the `render` call at the bottom of the file.)

david-yz-liu (commented):
Looks great! Left a few comments.

app/assets/javascripts/Components/Result/result.jsx
robert-tan: Yep, this updates the browser URL on modern browsers (up-to-date versions of Firefox/Chrome)

app/assets/javascripts/Components/Result/submission_file_panel.jsx
robert-tan: Comparing `props.annotations` because that is the actual prop change that should trigger a re-mount

app/assets/javascripts/Components/Result/result.jsx
```diff
@@ -615,7 +658,13 @@ class Result extends React.Component {
             <div id='drag'/>
             <div id='right-pane'>
               <RightPane
-                {...this.props}
+                {...{
```
david-yz-liu, line 661: Haha, well with this change you should just do

```
<RightPane
  assignment_id={this.state.assignment_id}
  ...
```

(No need for the spread operator any more.)

PR #5083, opened 2021-01-25
Add tags to summary table

chen-jonathan:
Added the Tags column in the Assignment Summary table for admin to view.
## Motivation and Context
Pull request is required in order to add tags info to the Assignment Summary table.

image

## Your Changes
**Description**:
First step was to fetch the tags data in summary_json function in assignment.rb, This involved fetching the data (tags_data, tag_info) and adding a tags attribute to extra_mark. After this, I added a tags column to be rendered in assignment_summary_table.jsx, displaying the tags data that I fetched

## Testing
Added some tags to various submissions and viewed that they were present in the summary table. Tried a few combinations (1 tag, multiple tags for one submission, etc.) to see if there were any issues.

Should TA"s also be able to see the tag info in the summary table?

## Checklist


app/assets/javascripts/Components/assignment_summary_table.jsx
```diff
@@ -98,6 +98,37 @@ class AssignmentSummaryTable extends React.Component {
...
+        if (filter.value) {
+
+          // Check tag names
+          const tag_matches = row._original.tags.some(
+            (tag) => tag.includes(filter.value)
+          );
+
+          if (tag_matches) {
+            return true;
+          }
```
david-yz-liu, line 126: This is quite nice, but can be simplified to just `return row._original.tags.some(...);`.

app/models/assignment.rb
```diff
@@ -576,12 +580,15 @@ def summary_json(user)
                                .map { |s| [s['users.user_name'], s['users.first_name'], s['users.last_name']] }
       end
 
+      tag_info = tag_data.fetch(g.id, {}
```
david-yz-liu, line 583: The logic in this part isn't right. You have a `{}.map ...` which I don't think you want.

The `fetch` isn't a bad idea, but you want the default value to be `[]`, and the `map` to go on the outside. After you do that you'll find that you can just do `tags: tag_info` below.

app/assets/javascripts/Components/assignment_summary_table.jsx
```diff
@@ -98,6 +98,37 @@ class AssignmentSummaryTable extends React.Component {
       filterable: false,
       defaultSortDesc: true,
     },
+    {
```
david-yz-liu, line 101: Move the column to before the final grade column.

Issue #4898, opened 2020-10-08
One-time annotations download


bogdansimion:
Currently the one-time annotations cannot be downloaded via the Download button, but it would be nice to save these as well somehow if possible, as a future feature.

david-yz-liu:
Thanks for suggesting this, added in #5088.

PR #5088, opened 2021-01-26
Adding UI element to allow an instructor to download one time annotations

YujinC:
## Motivation and Context

[This pull request solves a feature request here.](https://github.com/MarkUsProject/Markus/issues/4898)

## Your Changes
Screen Shot 2021-01-26 at 2 48 10 PM
In the csv below, you can see the group_name, last editor, creator, and content of the annotation.
Screen Shot 2021-01-26 at 3 16 12 PM

**Description**:
I modified the annotations controller to add a new route: onetime_annotation_download and I modified uncategorized_annotations to add a local assignment variable.  I also added a new display inline class in _markus.scss.

## Testing

I tested it manually by going to an assignment's annotations tab and clicking 'One Time Only Annotations' to see the one time annotations. Then I clicked the 'Download One Time Annotations' button I created, and checked that the output csv matched the parameters required.

Before I write any tests, I'd like some feedback on the way I implemented both the controller method and the UI.
Additionally, do I need to update the Changelog.md?
## Checklist


app/controllers/annotation_categories_controller.rb
```diff
@@ -285,7 +285,22 @@ def annotation_text_uses
   end
 
   def uncategorized_annotations
+    @assignment = Assignment.find(params[:assignment_id])
+    @texts = annotation_text_data(nil)
+  end
+
+  def onetime_annotation_download
```
david-yz-liu, line 292: Rather than create a separate route, you can use the same route as above (`uncategorized_annotations`) but have it respond to a different format (in this case, `csv`). You can do this using `respond_to` blocks. You can search for examples in our own code, and read the documentation [here](https://apidock.com/rails/ActionController/MimeResponds/respond_to).

It's a new idea, but basically you'll want a `respond_to.js` for the existing route, and a `respond_to.csv` with a "do block" for the new route.

app/views/annotation_categories/_uncategorized_annotations_list.html.erb
```diff
@@ -1,7 +1,17 @@
 <div id='annotation_list_holder'>
-  <h2>
-    <%= t('annotations.uncategorized') %>
-  </h2>
+  <div>
+    <h2 class='display-inline'>
+      <%= t('annotations.uncategorized') %>
+    </h2>
+    <div class='float-right'>
+      <%= link_to "Download One Time Annotations",
```
david-yz-liu, line 7: Always make sure to add new string to the locale files.

david-yz-liu (commented):
You should indeed update the Changelog for this PR.

PR #5090, opened 2021-01-28
user: Fix flaky locale test and change locale on setting update

david-yz-liu:
## Motivation and Context

There is currently a flaky test in `user_controllers_spec.rb:65` (`I18n.locale` wasn't being set properly). While fixing this, I noticed a small bug in the `UsersController#update_settings` method, which didn't change `I18n.locale` even when the user changed this setting.

## Your Changes

**Description**: Fixed the test setup and modified `UsersController#update_settings` to change `I18n.locale`.

## Testing

Ran the tests locally a few times, always passed. Checked in the UI that after changing the locale setting, the flash message that's rendered is in the updated locale, rather than the previous locale

## Checklist

### DIFF
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -1,6 +1,20 @@
 describe UsersController do
   let(:admin) { create :admin }
 
+  shared_examples 'change locale' do
+    describe 'change locale in settings' do
+      before { I18n.locale = :en }
+
+      it 'updates locale' do
+        locale = 'es'
+        patch_as user, 'update_settings', params: { 'user': { 'locale': locale } }
+
+        expect(user.reload.locale).to eq locale
+        expect(I18n.locale).to eq locale.to_sym
+      end
+    end
+  end
+
   describe 'GET settings' do
     before { get_as admin, :settings }
     it 'should respond with success' do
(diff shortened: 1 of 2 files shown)


spec/controllers/users_controller_spec.rb
```diff
@@ -150,21 +146,17 @@
...
+      before :each do
         I18n.locale = :en
       end
 
-      it 'updates locale for student' do
+      it 'updates locale' do
         locale = 'es'
-        expect(I18n.locale).to eq I18n.default_locale
-        patch_as student,
+        patch_as admin,
```
mishaschwartz, line 155: `patch_as student` ?

spec/controllers/users_controller_spec.rb
```diff
@@ -26,16 +26,16 @@
     end
 
     describe 'Change locale' do
-      it 'updates locale for admin' do
+      before :each do
```
mishaschwartz, line 29: If you want a more compact syntax:

```ruby
before { I18n.local = :en }
```

spec/controllers/users_controller_spec.rb
```diff
@@ -26,16 +26,16 @@
...
+        I18n.locale = :en
+      end
+
+      it 'updates locale' do
```
mishaschwartz, line 33: This looks like a good candidate for a shared test. Just swap out the `patch_as X` for `patch_as user` and then set the `user` let variable to admin, student, ta.

david-yz-liu:
Note: coverage decreased because the tests are no longer accessing the `check_timeout` route, but that's separate from this PR. We'll add tests for that route separately.

PR #5160, opened 2021-03-05
starter-files: always reset starter files for groupings with no starter files

david-yz-liu:
## Motivation and Context

Groupings that were created before uploading starter files had their `starter_file_changed` attribute set to `false`, and could not actually download starter files.

## Your Changes

**Description**: Changed the check in `GroupingsController#download_starter_file` so that if the grouping has no starter file entries, it always calls `reset_starter_file_entries`. 

Note: nothing will happen if there are no starter file groups at all.

## Testing

Tested this in the UI and added a test case for this.

I also added a `reload` call on the `grouping.starter_file_entries`. I'm not sure if this is necessary (further discussion would be helpful).

## Checklist


mishaschwartz (commented):
When starter files are changed, all groupings for that assignment should have their `starter_file_changed` value set to `true`:

https://github.com/MarkUsProject/Markus/blob/master/app/controllers/assignments_controller.rb#L436

If that is not happening, then I think there is likely a bug in the `AssignmentsController#update_starter_file` method and we should look to fix that instead of checking if entries do or do not exist. 

I can investigate further.

david-yz-liu:
Ah, interesting. I don't have time before our meeting today to look into this further, please go ahead and we can discuss when we meet.

PR #5164, opened 2021-03-09
Fix starter file change warning

mishaschwartz:
## Motivation and Context

The rules governing when students get warned if their starter files have changed were incorrect. In some cases, students would get warned even if their starter files had not changed, in other cases, the students were not warned if the starter files had changed. This PR makes sure that students are warned that starter files changed under the following conditions:

When a file is added/removed/updated in a starter file group:
- `starter_file_type == 'simple'`: all groupings for the given assignment get warned
- `starter_file_type == 'sections'`: all groupings that have students in a section that is associated with the changed starter file group get warned
- `starter_file_type == 'shuffle'`: all groupings that are assigned a starter file entry that is affected by the file change are warned. Additionally, all groupings without any associated starter file entries get warned (in the case that they were created before the starter files were assigned).
- `starter_file_type == 'group'`: all groupings that are assigned a starter file entry from the updated starter file group are warned. Additionally, all groupings without any associated starter file entries get warned (in the case that they were created before the starter files were assigned).

When a starter file group is destroyed:
- all groupings that are assigned a starter file entry from the updated starter file group are warned

When the starter file type is changed or the default starter file group has changed:
- if the `starter_file_type` was changed to/from 'shuffle': warn all groupings for the given assignment
- otherwise: warn all groupings without any associated starter file entries (all other groupings will keep the same starter files)

When `entry_rename` or `use_rename` is changed:
- all groupings that are assigned a starter file entry that is affected by the name change are warned. Additionally, all groupings without any associated starter file entries get warned (in the case that they were created before the starter files were assigned).

Additional changes:

- do not allow starter file groups with no files to be assigned to groupings

## Your Changes

**Description**:

see above

## Checklist

### DIFF
--- a/app/models/assignment_properties.rb
+++ b/app/models/assignment_properties.rb
@@ -140,11 +140,14 @@ class AssignmentProperties < ApplicationRecord
     errors.add(:base, msg) if default_sfgroup && default_sfgroup.starter_file_entries.empty?
   end
 
+  # Warn groupings that are affected by a change in starter_file_type. If the type changed to/from 'shuffle' then this
+  # should be all groupings. Otherwise, it only affects groupings that have not been assigned any starter files yet.
+  # Similarly, if the default starter file group changes, only warn those that don't have any assigned starter files.
   def warn_if_starter_file_change
     if saved_change_to_default_starter_file_group_id?
-      assignment.starter_file_groups
-                .find_by(id: attribute_before_last_save(:default_starter_file_group_id))
-                .warn_affected_groupings
+      assignment.groupings.left_outer_joins(:starter_file_entries)
+                .where('starter_file_entries.path': nil)
+                .update_all(starter_file_changed: true)
     end
     if saved_change_to_starter_file_type?
       if saved_change_to_starter_file_type.include?('shuffle')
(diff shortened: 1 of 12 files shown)


app/models/assignment_properties.rb
```diff
@@ -131,4 +133,29 @@ def not_timed_and_scanned
...
   end
+
+  def default_starter_group_not_empty
+    default_sfgroup = StarterFileGroup.find_by(id: default_starter_file_group_id)
+    msg = I18n.t('activerecord.errors.models.assignment_properties.attributes.default_starter_file_group_id.is_empty')
+    errors.add(:base, msg) if default_sfgroup&.starter_file_entries&.empty?
+  end
+
+  # Warn groupings that are affected by a change in starter_file_type. If the type changed to/from 'shuffle' then this
+  # should be all groupings. Otherwise, it only affects groupings that have not been assigned any starter files yet.
```
david-yz-liu, line 144: I'm not sure this is correct. For example, if we go from 'simple' to 'section', won't we want to reassign students who are in a section that needs a different starter file group?

david-yz-liu (commented):
This looks really great! I left one comment for discussion, and it does look like the tests need to be updated, and it probably makes sense to add new tests that illustrate the more fine-grained checking you've now implemented.

app/models/assignment_properties.rb
mishaschwartz: No, because we prioritize making sure that students who have already received starter code keep the same version of that starter code if at all possible

mishaschwartz:
I'm tempted to just ditch these changes. They're going to be a real pain to maintain and add a ton of complexity to the code for minimal benefit.
If we don't include these changes then we just need to make it clear that changes to starter files after groups are created are not recommended and that additional actions by the students are required (we have a warning message to the instructors that does this).

PR #5170, opened 2021-03-11
Do not delete or modify an annotation used in a remark request

mishaschwartz:
## Motivation and Context

An issue occurred where deductive annotations were deleted that were associated with a result that was pending a remark request. This meant that a mark associated with the result had a `null` value. When viewing the remark request result, MarkUs tries to display the old marks (one of which is now null) and raises an error.

We currently have a rule that annotation texts that are associated with released result cannot be modified or deleted. This rule should be extended to include annotation texts that are no longer released because a remark request has been made. In this case, the initial result should still count as "released" while a remark request is pending.

## Your Changes

**Description**:

- add a condition to a callback that is triggered when updating or deleting annotation texts that aborts if a remark request exists for an associated result.

## Checklist


app/models/annotation_text.rb
```diff
@@ -36,7 +36,12 @@ def escape_content
 
   def check_if_released
     # Cannot update if any results have been released with the annotation and the deduction is non nil
```
david-yz-liu, line 38: Nice fix! Could you update this comment as well?

PR #5193, opened 2021-03-22
Set autocomplete to off

mishaschwartz:
## Motivation and Context

This change is a recommended minor security improvement. Previously we did not want to set this because it interfered with password managers. However, most modern browsers now take this as a suggestion (not a rule) and allow users to enable password managers even if this is set (https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#the_autocomplete_attribute_and_login_fields)

## Your Changes

**Description**:

- added attribute to password input

- [x] Other (please specify): security update

## Checklist


david-yz-liu:
Please add to role switch dialog as well.

PR #5207, opened 2021-04-08
Better repository permission file updates

mishaschwartz:
## Motivation and Context

There is a possibility that actions that required updating the repository permissions file were taking a really long time if too many updates were happening simultaneously. This is because we were setting an exclusive file lock on the permissions file when updating and files. This could potentially cause threads to stall as they wait to lock the resource, especially since we did not implement a priority queue (and it's not clear if ruby's file locking mechanism does either) for the lock so a thread could theoretically block for a while if it keeps on losing races.

This PR moves the work of writing to the permissions file to a background job. This means that any file locking will not block the main thread since jobs are run asynchronously. This is similar to how we are currently writing the authorized_keys files.  

In order to make sure that we are not writing multiple times unnecessarily and that the data in the file reflects the most recent state of the database, the job uses the following mechanism:

1. before enqueuing the job, check if a redis key is set. 
    - If it is set, this indicates that there is already an enqueued job, so we abort the enqueueing. 
    - If it is not set, set the key to some value (the job id is used) and enqueue the job.
    - Note that checking for the key and setting it if it doesn't exist can be done atomically.
2. when the job is run, unset the key just before getting the relevant data from the database and writing the file
    - we unset the key _before_ getting data from the database because this ensures that the data in the file reflects the most recent state of the database. If we did it afterwards, there is a chance that another process could update the db but then not enqueue the job because it sees the key is still set.

I also added the same functionality to the `UpdateKeysJob`.

## Your Changes

**Description**:

- added the `UpdateRepoPermissionsJob` and updated the `UpdateKeysJob`
- removed thread synchronization logic from the `repository.rb` file because it is handled in the jobs now
- updated tests

## Checklist


app/jobs/update_keys_job.rb
```diff
@@ -4,20 +4,36 @@ class UpdateKeysJob < ApplicationJob
...
-          f.write(KeyPair.full_key_string(name, key) + "\n")
-        end
+        data.each { |name, key| f.write(KeyPair.full_key_string(name, key) + "\n") }
       ensure
         f.flock(File::LOCK_UN)
       end
     end
+  ensure
+    redis = Redis::Namespace.new(Rails.root.to_s)
+    if redis.get('repo_permissions') == self.job_id
```
david-yz-liu, line 35: Unless I'm missing something, should this be `authorized_keys`?

PR #5211, opened 2021-04-14
marked: sanitize html using dompurify

david-yz-liu:
## Motivation and Context

The `sanitize` option for the marked library has been deprecated since version 0.7.0, and is [not recommended](https://marked.js.org/#usage) for sanitizing output.

## Your Changes

**Description**: This PR introduces the [DOMPurify](https://github.com/cure53/DOMPurify) library for doing HTML sanitization.

## Testing

Tested in the UI the places where we use marked to render markdown content.

## Checklist

### DIFF
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -18,9 +18,9 @@ window.ajax_events = ajax_events;
 // vendor libraries
 import 'javascripts/jquery.easyModal';
 
-// marked (markdown support)
-import marked from 'marked';
-window.marked = marked;
+// Markdown support (using marked and DOMpurify)
+import safe_marked from 'javascripts/safe_marked';
+window.safe_marked = safe_marked;
 
 // moment (date/times manipulation)
 import moment from 'moment';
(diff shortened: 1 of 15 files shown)


app/javascript/packs/application.js
```diff
@@ -18,9 +18,9 @@ window.ajax_events = ajax_events;
 // vendor libraries
 import 'javascripts/jquery.easyModal';
 
-// marked (markdown support)
-import marked from 'marked';
-window.marked = marked;
+// Markdown support (using marked and DOMpurify)
+import safe_marked from 'javascripts/safe_marked';
```
mishaschwartz, line 22: Do we want to move towards a model where we import js libraries only in the files we need them instead of setting things globally on `window`?

If so, this might be a good opportunity to start doing that for this function at least.

I'm happy to make those changes if you want.

david-yz-liu: Yeah sure, please go ahead and make those changes!

mishaschwartz: This is actually going to require a bit more configuration that is out of scope for this PR. I'll do it separately once this is pulled in.

PR #5222, opened 2021-04-22
Remove counter cache

mishaschwartz:
## Motivation and Context

Counter caches, and specifically the `groupings_count` cache on the assessments table cause all sorts of grief when running in a high concurrency environment.  If multiple groupings get created simultaneously (beginning of a timed assessment), postgres locks the same row to update counter caches and brings the app to a halt.

Removing the counter caches adds a tiny bit of overhead in some queries but it is worth the trade off.

## Your Changes

**Description**:

- removed counter caches and corresponding columns in the tables

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,6 @@
 # Changelog
 ## [unreleased]
+- remove counter caches (#5222)
 
 ## [v1.12.0]
 - Remove annotations context menu from peer assignments view (#5116)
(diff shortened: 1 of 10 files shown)


david-yz-liu (commented):
Good stuff, I think you just need to update the `schema.rb` with re-running the migration to remove the `annotation_texts_count` column.

PR #5226, opened 2021-04-27
Option to hide starter files after collection date

mishaschwartz:
## Motivation and Context

Provides the option to hide starter files (can't download) from students once the collection date has passed for their grouping.

## Your Changes

**Description**:

- added a new boolean attribute to assignment_properties indicating whether starter files are visible after the collection date
- added a checkbox on the starter files page to toggle this attribute
- added condition in grouping policy to check for this attribute

## Checklist

https://github.com/MarkUsProject/Markus/wiki/Pending-Changes#pr-5226

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,7 @@
 - Remove counter caches (#5222)
 - Add copy to clipboard button for plaintext submission files in Results view (#5223)
 - Delay grouping creation for working-alone timed assessments to when the student starts the assessment (#5224)
+- Add option to hide starter files after collection date (#5226)
 
 ## [v1.12.0]
 - Remove annotations context menu from peer assignments view (#5116)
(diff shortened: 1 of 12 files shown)


david-yz-liu (commented):
Thanks @mishaschwartz! Two comments:

1. The file `config/autotest.api_key` should be removed.
2. I think I wasn't clear before, but I actually want to prevent downloading the starter files after the _last time to start_, rather than the collection time. E.g. if the start window is 8am-4pm, and the duration is 2 hours, students who haven't started shouldn't be able to download the files after 4pm, not 6pm. But the collection date is good for non-timed-assessments, so you might need  a conditional to handle both cases.

    - Related: please change the checkbox label to "... to students after the due date (or start window for a timed assessment) has passed". _Or_ if you don't like that, we could have two separate labels, one for regular assignments and one for timed assessments.

mishaschwartz:
- The collection date for a student that hasn't started a timed assessment is the same as the due date (4pm in your example). 
- I don't think the label needs to be changed, it really is the collection date in both cases (but we don't use "collection date" as a term in the views so we say "due date"). Saying "start window" for a timed assessment implies that a student who has started the timed assessment near the end of start window will not be able to access the starter files even if they still have time left on the duration. If the start window ends at 4pm and the student starts at 3:30 with a 2 hour duration, they have access to the starter files until 5:30, not until 4.

Issue #5280, opened 2021-05-08
Remove activerecord-import dependency


david-yz-liu:
Now that Rails 6 supports [bulk inserts/upserts](https://github.com/rails/rails/pull/35077), we no longer need to use the `activerecord-import` gem. We can replace calls to `<Model Class>.import` with [`.insert_all`](https://edgeapi.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-insert_all) or [`.upsert_all`](https://edgeapi.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-upsert_all).

The following files have occurrences of `.import` that should be replaced:

- [ ] `app/models/criterion.rb` (1 occurrence)
- [ ] `app/models/criterion_ta_association.rb` (1 occurrence)
- [ ] `app/models/grade_entry_form.rb` (3 occurrences)
- [ ] `app/models/grade_entry_student_ta.rb` (1 occurrence)
- [ ] `app/models/grouping.rb` (1 occurrence)
- [ ] `app/models/ta_membership.rb` (1 occurrence)
- [ ] `app/models/tag.rb` (2 occurrences)
- [ ] `app/models/user.rb` (1 occurrence)
- [ ] `app/controllers/api/groups_controller.rb` (2 occurrences)

PR #5291, opened 2021-05-15
Issue update imports grade entry form

devan-srinivasan:
## Motivation and Context

Rails 6 recently introduced bulk inserts/upserts so I updated this code to accommodate that.

#5280 

## Your Changes

**Description**
I replaced the .import call with a .insert_all instead in the method `create_all_grade_entry_students`

## Testing

I reset the docker container database which uses the method I changed. I ran the spec test suite and the specific file that grades this file separately as well. Lastly, I triggered the UI element that calls the method I changed. All tests passed.

## Checklist

### DIFF
--- a/app/models/grade_entry_form.rb
+++ b/app/models/grade_entry_form.rb
@@ -89,11 +89,8 @@ class GradeEntryForm < Assessment
     # columns = [:user_id, :assessment_id, :released_to_student]
     new_data = []
     Student.all.each do |student|
-      # grade_entry_students.build(user_id: student.id, released_to_student: false)
       new_data << {:user_id => student.id, :assessment_id => id, :released_to_student => false}
     end
-    # GradeEntryStudent.import columns, values, validate: false, on_duplicate_key_ignore: true
-    # array_of_hashes = values.collect{ |record| Hash[columns.zip record] }
     GradeEntryStudent.insert_all(new_data, returning: false) unless new_data.length == 0
   end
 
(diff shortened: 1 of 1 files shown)


david-yz-liu (commented):
@devan-srinivasan good start, but you can simplify the code further. Since `insert_all` takes a list of hashes, it's easier to modify the original loop (lines 91-94) to accumulate this list directly. Something like

```ruby
new_data = {}

Student.all.each do |student|
  new_data << { ... }
end
```

Note that each hash should have three keys (the ones specified in `columns`), and if you create the hashes manually, you can delete the `columns` variable as well.

Also when you're done you should remove the old code rather than commenting it out.

david-yz-liu (commented):
@devan-srinivasan good progress! Two more comments:

1. You still have a line of code commented out, and you should remove it instead.
2. `hound` is an automatic style checker. You should fix all Hound comments. In particular, there needs to be a space inside curly braces `{ ... }`, and instead of `:user_id => ...` you should right `user_id: ...`. The last comment tells you exactly what to change.

devan-srinivasan:
Hey @david-yz-liu how does it look now? I committed this yesterday but just realized maybe I should reply here so you get notified so sorry for the late reply.

david-yz-liu:
@devan-srinivasan don't worry, I get email notifications when pull requests get updated. :) Your PR is definitely looking better, but you missed my first point (about the commented out line).

devan-srinivasan:
Oh my gosh, sorry. Hopefully, it's all good now. I removed that last commented line :)

PR #5294, opened 2021-05-17
Refactored import to be .upsert_all issue #5280

emilychangg:
## Motivation and Context

#5280 

## Your Changes

**Description**:
Changed .import to .upsert_all in `app/models/user.rb`

## Testing

Went through the UI to ensure it still worked like it did before and also used rspec

## Checklist


app/models/user.rb
```diff
@@ -208,44 +208,57 @@ def self.upload_user_list(user_class, user_list, encoding)
...
       end
-      unless parsed[:invalid_records].empty?
-        raise ActiveRecord::Rollback
+
+      user_hash = users.collect { |record| Hash[user_columns.zip record] }
+      imported = user_class.upsert_all(user_hash, unique_by: :user_name) unless user_columns.empty? ||
+        users.empty? || :user_name.nil? || user_hash.empty?
+
+      unless imported.nil?
+        User.where(id: imported.rows.flatten).each do |user|
```
david-yz-liu, line 224: You use `imported.rows.flatten` a few different times, so this would be good to save in a variable.

app/models/user.rb
```diff
@@ -208,44 +208,57 @@ def self.upload_user_list(user_class, user_list, encoding)
...
-        parsed[:invalid_records] += "#{user.user_name}: #{error_message}"
       end
-      unless parsed[:invalid_records].empty?
-        raise ActiveRecord::Rollback
+
+      user_hash = users.collect { |record| Hash[user_columns.zip record] }
+      imported = user_class.upsert_all(user_hash, unique_by: :user_name) unless user_columns.empty? ||
+        users.empty? || :user_name.nil? || user_hash.empty?
+
+      unless imported.nil?
```
david-yz-liu, line 223: I'm not sure whether this `nil` check is necessary. The Rails documentation doesn't mention the possibility of `upsert_all` returning `nil`, also that could be an oversight. For now, remove this check (both here and below) to simplify your code.

app/models/user.rb
```diff
@@ -208,44 +208,57 @@ def self.upload_user_list(user_class, user_list, encoding)
...
+          parsed[:invalid_records] += "#{user.user_name}: #{error_message}"
+        end
+        unless parsed[:invalid_records].empty?
+          raise ActiveRecord::Rollback
+        end
       end
     end
-    unless imported.failed_instances.empty?
+    unless imported.nil?
+      # unless imported.failed_instances.empty?
```
david-yz-liu, line 242: The error reporting is a bit tricky, since it looks like `imported` doesn't return the rows that failed.

My recommendation:

1. Obtain a set of the user names from the original `users` hash.
2. Obtain a set of the user names that were successfully inserted (from imported, see above comment).
3. Do the set difference to obtain the set of user names that *weren't* successfully imported.
4. Use this set in the error reporting code here.

app/models/user.rb
```diff
@@ -208,44 +208,57 @@ def self.upload_user_list(user_class, user_list, encoding)
...
-        error_message = user.errors
-                            .messages
-                            .map { |k, v| "#{k} #{v.flatten.join ','}" }.flatten.join MarkusCsv::INVALID_LINE_SEP
-        parsed[:invalid_records] += "#{user.user_name}: #{error_message}"
       end
-      unless parsed[:invalid_records].empty?
-        raise ActiveRecord::Rollback
+
+      user_hash = users.collect { |record| Hash[user_columns.zip record] }
+      imported = user_class.upsert_all(user_hash, unique_by: :user_name) unless user_columns.empty? ||
```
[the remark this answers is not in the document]
david-yz-liu: Another comment: pass in `returning: %w[id user_name]` so that the returned `imported.rows` contains both id and user name. This will make it slightly harder to extract just the ids (use `map`) but will help with the error reporting---see my comment below.

david-yz-liu (commented):
@emilychangg a good start, this one is definitely a bit tricky! I left some comments to help you simplify your code, please work on addressing those and then fixing any remaining failing tests.

app/models/user.rb
```diff
@@ -208,44 +208,57 @@ def self.upload_user_list(user_class, user_list, encoding)
...
     User.transaction do
-      imported = user_class.import user_columns, users, on_duplicate_key_update: {
-        conflict_target: [:user_name],
-        columns: [:last_name, :first_name, :section_id, :email, :id_number, :display_name, :time_zone]
-      }
-      User.where(id: imported.ids).each do |user|
-        if user_class == Ta
-          # This will only trigger before_create callback in ta model, not after_create callback
-          user.run_callbacks(:create) { false }
+      users.each do |user|
```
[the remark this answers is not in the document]
emilychangg: Hi David, thanks so much for all the help! In my newest commit, I tried to implement this, however, it would cause imported to always be nil. Unsure why this is happening, but I will look into it later.

app/models/user.rb
```diff
@@ -220,12 +220,12 @@ def self.upload_user_list(user_class, user_list, encoding)
       end
       imported = user_class.upsert_all(user_hash, unique_by: :user_name, returning: %w[id user_name]) \
 unless user_hash.empty?
-      unless imported.nil?
-        successful_imports = imported.rows.map { |x| [x[1]] }.flatten
-        imported_ids = imported.rows.map { |x| [x[0]] }.flatten
-      else
+      if imported.nil?
```
emilychangg, line 223: Hey David, I know you didn't want me to add the nil check but I know exactly why this happens and it is because the user_hash is empty, thus the .upsert_all returns nil. As for the `successful_imports = []` and `imported_ids = []`, I added those because I used byebug with the master branch and found out that those are what the values would be with .import.

david-yz-liu: You're right that the above line will never execute if `user_hash` is empty. However, this is a place where the code can be simplified. Notice that `user_hash` is computed from `users`, so we can actually return much earlier. Look for where `users` is computed, and after that point you can do a `return parsed if users.empty?`, which is actually an improvement since it saves all of the other work.

Looking at the code, `parsed` has its `:invalid_records` key set on line 214, so you'll want to move that earlier so that even with the early return, `parsed` has this key.

david-yz-liu: (And then you can remove the `unless user_hash.empty?` guard from the previous line.)

app/models/user.rb
```diff
@@ -203,16 +206,23 @@ def self.upload_user_list(user_class, user_list, encoding)
...
-        columns: [:last_name, :first_name, :section_id, :email, :id_number, :display_name, :time_zone]
-      }
-      User.where(id: imported.ids).each do |user|
+      user_hash = users.collect { |record| Hash[user_columns.zip record] }
+      user_hash.each do |user|
+        user[:type] = user_class.name
+        all_user_names.push(user[:user_name])
+      end
+      imported = user_class.upsert_all(user_hash, unique_by: :user_name, returning: %w[id user_name])
+      successful_imports = imported.rows.map { |x| [x[1]] }.flatten
```
david-yz-liu, line 223: If you just use `x[1]` instead of wrapping it in a list, you can remove the `flatten`. Same as the line below as well.

david-yz-liu (commented):
@emilychangg okay looking good, I just left one final comment, and after you're done with that this will be ready to be merged in!

PR #5305, opened 2021-05-21
replace .import with .insert_all

JunruL:
## Motivation and Context

Replace calls to <Model Class>.import with .insert_all or .upsert_all.
#5280

## Your Changes

**Description**:
Created an array of hash and replace the import method with the insert_all method.

## Testing

I ran "docker-compose run rails rspec" and all tests passed. I also test on the MarkUs application: "assign TAs" and "randomly assign TAs" under "Assignment->Graders".

I deleted "columns" variable in line 82 and line 97, which seemed not needed. However, this led to some errors and I could not figure out why. Keeping this variable does not lead to errors, so I don't remove it for this submission.

## Checklist


david-yz-liu (commented):
@JunruL your code looks good, but this change actually enables further simplification of this method:

1. You can remove the `columns` variable entirely.
2. You can also remove the call to `values.map` on line 98.

Also, in the code that you added it's better to use `map` rather than `each`; the former returns a list, so you can do `mappings = values.map ...`.

app/models/criterion.rb
```diff
@@ -94,12 +93,9 @@ def self.assign_tas(criterion_ids, ta_ids, assignment)
 
     # Add assessment_id column common to all rows. It is not included above so
```
david-yz-liu, line 94: You can now remove this comment as well!

PR #5308, opened 2021-05-25
updated the import statement in the method update_grade_entry_items 

devan-srinivasan:
## Motivation and Context

It upgraded the code to use a more recent means of bulk upserting and removed the need for the active-record import gem.

#5280 

## Your Changes

We adapted the method 'grade_entry_form.rb#update_grade_entry_items' to use 'upsert_all' instead of the 'import' gem. This is because of the introduction of bulk inserts/upserts with Rails 6. This change was done in **collaboration** with @JunruL 

## Testing

We first tested the UI by creating a dummy assignment, then navigating to the grades section, and uploading a sample grades report (that we downloaded from a pre-loaded assignment). This action called the method 'from_csv' which called the helped 'update_grade_entry_items', which was the method we updated.

## Checklist


devan-srinivasan:
Hey David, I think I closed in on the root cause. I noticed that (as you said) filtering of the `Total` column should be automatic. After some debugging and inspection I noticed that the filtering case was not executing as it would require `show_total => true`. However, as per default instantiation `show_total` was `false` which makes sense as we don't want the total to be included. So I changed the conditional being evaluated to be negated and that solved the problem much cleaner.

app/models/grade_entry_form.rb
```diff
@@ -154,7 +154,7 @@ def from_csv(grades_data, overwrite)
         next
       elsif totals.empty?
         totals = row.drop(1)
-        if self.show_total && names.last == GradeEntryForm.human_attribute_name(:total)
+        if !self.show_total && names.last == GradeEntryForm.human_attribute_name(:total)
```
david-yz-liu, line 157: @devan-srinivasan You're getting closer, but the logic in this function is correct. You should be fixing the test itself, I suspect it's setting up the data incorrectly.

devan-srinivasan: Aw man, okay. Will do.

spec/controllers/grade_entry_forms_controller_spec.rb
```diff
@@ -149,12 +149,13 @@
     end
 
     it 'ignores the total column if given in the csv file' do
+      # show_total needs to be on when uploading a csv file with the "Total" column included
```
david-yz-liu, line 152: Rather than including this in a comment, why don't you change the description on the previous line to specify a "when ...".

spec/controllers/grade_entry_forms_controller_spec.rb
```diff
@@ -149,12 +149,13 @@
...
+      grade_entry_form_with_data.update(show_total: true)
       post :upload, params: { id: grade_entry_form_with_data.id, upload_file: @file_total_included, overwrite: true }
       expect(response.status).to eq(302)
       expect(flash[:error]).to be_nil
       expect(response).to redirect_to(grades_grade_entry_form_path(grade_entry_form_with_data))
-
-      # Check that the total column is the actual total and not the incorrect total given in the file
+      # Check that the total column is the actual total and not the incorrect  v
```
david-yz-liu, line 158: Please revert these changes: include the blank line, and restore the previous comment, which I think you accidentally modified.

devan-srinivasan: yup sorry, I meant to fix that before pushing silly mistake.

PR #5315, opened 2021-05-27
add section association and delete the section method

JunruL:
## Motivation and Context

1. Replace the section method in the Grouping model with the section association, which is through inviter.
2. Check where the section method is used and edit the code, since it will now return a section object rather than a string. And I need to pay attention to a nil section, which should return '-'.

## Your Changes

**Description**: I added the section association in grouping.rb and removed the section method.

- [x] Other (please specify): Add a new association

## Testing

I ran "docker-compose run rails rspec" and all tests passed.

I could not find grouping.section or groupings.section in other files, and thus don't know where the original method is used. Since all the tests pass after I add the new association, I am not sure whether other change is needed.

## Checklist


david-yz-liu:
@JunruL wow, perhaps this method was no longer used! In any case, one additional check I'd like you to make is to look for code of the form `<grouping object>.inviter.section` and replace it with `<grouping object>.section`. Note that in `grouping.rb`, you might see either `self.inviter.section` or just `inviter.section` in instance methods.

app/models/assignment.rb
```diff
@@ -769,7 +769,7 @@ def section_groupings(section)
     groupings.select do |grouping|
       grouping.inviter.present? &&
       grouping.inviter.has_section? &&
-      grouping.inviter.section.id == section.id
+      grouping.section.id == section.id
```
david-yz-liu, line 772: Good find---actually, with the new association, this entire block can be simplified. We no longer need to have the first two checks, and can instead use the [safe navigation operator `?.`](https://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/). Please read about that and then use this to simplify this code.

app/models/assignment.rb
```diff
@@ -767,9 +767,7 @@ def average_annotations(ta_id = nil)
   # Returns the groupings of this assignment associated with the given section
   def section_groupings(section)
     groupings.select do |grouping|
-      grouping.inviter.present? &&
-      grouping.inviter.has_section? &&
-      grouping.inviter.section.id == section.id
+      grouping.inviter&.section && grouping.section.id == section.id
```
david-yz-liu, line 770: This is better, but can be simplified further. Notice that `nil == x` is `false` whenever `x` is a number.

PR #5319, opened 2021-05-29
Errors groups controller remove group

devan-srinivasan:
## Motivation and Context

This PR repurposed the existing '@errors' list to create 'flash_message' errors for the user, improving the overall user experience.

## Your Changes

**Description**
Firstly, I added a _descriptive but brief_ `flash_message` for errors in deleting a grouping. Additionally, I changed the status broadcast to only be `:ok` when there were no errors at all, while otherwise, it would be `:unprocessable_entity`

## Testing

I ran the `rspec` full suite on the program and all tests passed. I also triggered this change in the UI by testing that an attempt to remove an **unremovable** grouping flashed the appropriate error. I did so by making dummy groupings on an assignment, then setting their submissions to be collected, making them unremovable. Then by attempting to remove each unremovable group, the appropriate error was flashed, signifying a successful test.

## Checklist


app/controllers/groups_controller.rb
```diff
@@ -47,7 +46,15 @@ def remove_group
...
       end
     end
-    head :ok
+    unless @errors.empty?
+      err_msg = 'The following groupings could not be deleted since they have submissions: ' +
+        @errors.join(', ')
+      flash_message(:error, err_msg)
+      head :unprocessable_entity
+    end
+    if @errors.empty?
```
david-yz-liu, line 55: Instead of having two consecutive conditionals, use an if-else construct.

app/controllers/groups_controller.rb
```diff
@@ -47,7 +46,15 @@ def remove_group
         @removed_groupings.push(grouping)
       end
     end
-    head :ok
+    unless @errors.empty?
+      err_msg = 'The following groupings could not be deleted since they have submissions: ' +
```
david-yz-liu, line 50: This is a very good error message. You should move this to the localization files, by adding a new key (maybe under `views/groups/`). You can add the corresponding keys for each language, but use "UPDATE ME" for the other languages.

david-yz-liu (commented):
@devan-srinivasan Good work. I left a few comments, and here are two two general comments as well:

1. You can make `@errors` just `errors`, it doesn't need to be an instance variable.
2. Please add some new tests for this method under the `groups_controller_spec.rb` file. This is a good experience for adding tests using RSpec! Your tests should check both error and non-error situations, and the status code using the `have_http_status` method.

devan-srinivasan:
Hey @david-yz-liu doesn't `errors` need to be an instance attribute in order for the rspec cases to access it? when I make it a local variable rspec fails tests that involve the '@errors' array and I think it is due to inaccessibility. The full error is  
image

devan-srinivasan:
Hey @david-yz-liu when you remove `@errors` as an instance attribute, it fails a few rspec cases in `group_controllers_spec.rb` specifically the ones that test the contents of the `errors` array generated in the method `group_controller#remove_group`. The errors are all of the following type: 
image
After investigating I think this is because by making `errors` a local variable, the rspec tests cannot access the variable. Because of this, when it tests the array using `:errors.to_a` or something of that nature, it is testing the symbol which has no meaning and hence the test fails. Let me know what you think.

david-yz-liu:
Yes, now that this test is using `flash_message`, you should update the test to check that the flash is set properly instead of relying on `@errors`. You can look at how we test the contents of the flash hash in other tests.

devan-srinivasan:
Oh! Okay sorry didn't realize I should be making such large changes. Then yea I will update the tests there.

devan-srinivasan:
In the spec file that I was instructed to change, I changed the test to check the flash queue instead of the `errors` array, as well as added two new tests (one for the `:ok` status code and the other for `:unprocessable_entity`) for a total of **4 new tests**. Hope that it is enough. When I used `have_https_status(422)` or `have_https_status(:unprocessable_entity)` I got an error that reads 
image
I inspected this using byebug and saw that `response.responds_to?(has_https_status) => false`. Backtracking through the source code for this matcher, however (https://github.com/rspec/rspec-rails/blob/main/lib/rspec/rails/matchers/have_http_status.rb) I don't understand why this call is even made. When the matcher checks for a match, it first calls the class method `as_test_response`. This is where `response.responds_to?` is called and where the error is raised I believe. But this call shouldn't be made as `::ActionDispatch::TestResponse === obj => true` so I am stuck. I found a substitute for the rspec test which is by just manually checking the status code stored in the object, although I am aware this isn't what you intended for. If this is insufficient or you want more details on my issue let me know, don't know what to do about this so I set the PR to be draft. I did my best to research but couldn't find much documentation on this matcher and found nothing on the error at all. Sorry for troubling you with this.

spec/controllers/groups_controller_spec.rb
```diff
@@ -109,6 +114,11 @@
           delete :remove_group, params: { grouping_id: [grouping.id], assignment_id: assignment }
         end
 
+        it 'should return the :unprocessable_entity status code' do
+          # expect(response).to have_https_status(:unprocessable_entity)
```
david-yz-liu, line 118: @devan-srinivasan compare this line *very carefully* with Line 93 (which you added in an above test). It might help you to copy and paste so that the lines are side by side.

devan-srinivasan: oh my god, haha such a simple mistake. digging through source code was unnecessary!

PR #5321, opened 2021-05-31
refactored sanitize filename and it's tests

TheXanMan321:
## Motivation and Context

Moves the method `sanitize_file_name` into a more general module (in app/lib) since many models, views and controllers use it.

## Your Changes

**Description**:
I created app/lib/file_helper.rb and moved `sanitize_file_name` into it. I moved the test for  `sanitize_file_name`  from spec/helpers/submission_helper_spec.rb into a new file; spec/helpers/file_helper.rb.

## Testing

I tested places were `sanitize_file_name` was called in the UI. I also ran rspec and all tests (including the new ones) passed.
I tested places were get_file_type was called in the UI. (except for where it is called in automated testing) I also ran rspec.

There are calls to `sanitize_file_name` in coverage/index.html. From my understanding, this file is generated automatically and doesn't need to be changed, but I'm not entirely sure.
## Checklist


app/lib/file_helper.rb
```diff
@@ -0,0 +1,11 @@
+# Helper for replacing unwanted characters in filenames.
+module FileHelper
+  def sanitize_file_name(file_name)
+    # If file_name is blank, return the empty string
+    return '' if file_name.nil?
+    File.basename(file_name).gsub(
+      SubmissionFile::FILENAME_SANITIZATION_REGEXP,
```
david-yz-liu, line 7: Move both of these constants from `SubmissionFile` into this file.

spec/helpers/file_helper_spec.rb
```diff
@@ -0,0 +1,38 @@
+describe FileHelper do
+  # Put some confidence in our submission filename sanitization
```
david-yz-liu, line 2: You can delete this comment

spec/helpers/file_helper_spec.rb
```diff
@@ -0,0 +1,38 @@
+describe FileHelper do
+  # Put some confidence in our submission filename sanitization
+  context 'A new file when submitted' do
```
david-yz-liu, line 3: Try rephrasing this now so that it's more generic, and doesn't refer to "submitted", since the helper can be applied to all filenames.

app/helpers/repository_helper.rb
```diff
@@ -1,5 +1,5 @@
 module RepositoryHelper
-  include SubmissionsHelper
+  include FileHelper
```
david-yz-liu, line 2: I believe that all files under `app/lib` should be automatically loaded, so I think you should be able to remove this include (and the one below).

spec/helpers/file_helper_spec.rb
```diff
@@ -0,0 +1,38 @@
+describe FileHelper do
```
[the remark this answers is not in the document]
david-yz-liu: Hmm, so this is a bit of a red flag. Line endings can be a bit tricky with git and Windows. First, you can open this file in RubyMine and go to the bottom-right corner of the window. Look for where it says "CLRF" and switch it to "LF".

david-yz-liu:
@TheXanMan321 okay, I looked into the importing thing a bit more. I think the best fix is to make the `sanitize_file_name` method a class method (use `def self.sanitize_file_name`), and then call it by doing `FileHelper.sanitize_file_helper`. I like this more than doing the "include" :)

PR #5338, opened 2021-06-03
improve the group_sections calculation

JunruL:
## Motivation and Context

Since the association to section was added to groupings model, we can simplify the calculation of grouping_sections.

## Your Changes

**Description**:
I simplified the calculation of grouping_sections.

## Testing

I ran all tests.

## Checklist


app/models/assignment.rb
```diff
@@ -957,12 +957,9 @@ def current_grader_data
       groups[[group_id, group_name, count]]
       groups[[group_id, group_name, count]] << ta unless ta.nil?
     end
-    # TODO: improve the group_sections calculation.
-    # In particular, this should be unified with Grouping#section.
     group_sections = {}
-    self.groupings.includes(:accepted_students).find_each do |g|
-      s = g.accepted_students.first
-      group_sections[g.id] = s&.section_id
+    self.groupings.each do |g|
```
david-yz-liu, line 961: Use `.groupings.includes(:section)` to prevent an N+1 query (read about these here: https://guides.rubyonrails.org/active_record_querying.html#eager-loading-associations).

app/models/assignment.rb
```diff
@@ -957,12 +957,9 @@ def current_grader_data
...
+      group_sections[g.id] = g.section.id
```
david-yz-liu, line 962: Use the safe navigation operator here :)

app/models/assignment.rb
```diff
@@ -957,12 +957,9 @@ def current_grader_data
       groups[[group_id, group_name, count]]
       groups[[group_id, group_name, count]] << ta unless ta.nil?
     end
-    # TODO: improve the group_sections calculation.
-    # In particular, this should be unified with Grouping#section.
     group_sections = {}
-    self.groupings.includes(:accepted_students).find_each do |g|
-      s = g.accepted_students.first
-      group_sections[g.id] = s&.section_id
+    self.groupings.include(:section).each do |g|
```
david-yz-liu, line 961: @JunruL This should be `include`, not `includes`. This indicates that this method is not actually tested, since otherwise an error would have been raised.

So, in this PR please also add some tests for this model method. Note that this method returns a hash; you only need to write some tests for the `groups` key, and the `section` value inside each group. Writing new tests will take a bit of time, but is an excellent experience. My advice is to have some tests which create groupings with no members, groupings with inviters that don't belong to a section, and groupings that have inviters that do belong to sections.

JunruL: It seems that what I used was `include`. So do I need to use `includes` instead?

david-yz-liu: Yes, that's correct. You can see the error when you write tests for this method, or you can find where this method is called from the UI.

spec/models/assignment_spec.rb
```diff
@@ -2486,4 +2486,35 @@ def grouping_count(groupings)
...
     end
   end
+
+  describe '#current_grader_data' do
+    let!(:assignment) { create :assignment }
+
+    context 'groupings with no members' do
+      let!(:grouping) { create :grouping, assignment: assignment }
+      it 'should not have sections' do
+        expect(assignment.current_grader_data[:groups][0][:_id]).to eq(grouping.id)
```
david-yz-liu, line 2496: @JunruL overall the tests are excellent. Just make sure to call `assignment.current_grader_data` once and save it in a variable. This is more efficient in general, and though it doesn't matter for these tests (since the data is so small), it's a good habit to adopt.

PR #5341, opened 2021-06-08
Adding automated test file modified times

TheXanMan321:
## Motivation and Context

Showing the time an assignment's automated test file was submitted in the test manager, so that users know when  test files were modified (similar to files submitted to an assignment)

## Your Changes

**Description**:
I added a submitted_date key to the file_data hash rendered by populate_autotest_manager in automated_tests_controller.rb. I added a check to make sure the files being uploaded existed, since the tests do not create real files.

## Testing

I tested my changes in the UI, using multiple languages and time zones. I also ran rspec on automated_tests_controller_spec. 

I still need to update this so that the time uploaded is in the correct time zone.

## Checklist


app/controllers/automated_tests_controller.rb
```diff
@@ -91,8 +91,17 @@ def populate_autotest_manager
...
       else
         file_keys << file
-        { key: file, size: 1,
-          url: download_file_assignment_automated_tests_url(assignment_id: assignment.id, file_name: file) }
+
+        if files_dir.join(file).exist?
+          date = File.mtime(files_dir.join(file))
+          local_date = date.in_time_zone(current_user.time_zone)
+          ""
+          { key: file, size: 1, submitted_date: local_date.strftime("%A, %B, %d, %Y, %I:%M:%S %p %Z"),
```
david-yz-liu, line 99: Good research here, but you should use the `I18n.l` (that's a lower case L) localization method to render dates, as different locales render dates differently. You can read more about this here: https://guides.rubyonrails.org/i18n.html#adding-date-time-formats

app/controllers/automated_tests_controller.rb
```diff
@@ -91,8 +91,17 @@ def populate_autotest_manager
         { key: "#{file}/" }
       else
         file_keys << file
-        { key: file, size: 1,
-          url: download_file_assignment_automated_tests_url(assignment_id: assignment.id, file_name: file) }
+
+        if files_dir.join(file).exist?
+          date = File.mtime(files_dir.join(file))
+          local_date = date.in_time_zone(current_user.time_zone)
```
david-yz-liu, line 97: Hmmm, I wonder if calling `in_time_zone` is still necessary after using `I18n.l` (see comment below). Could you look into this? If you used similar code from elsewhere, it might be that we can simplify that too, in a separate PR.

david-yz-liu (commented):
@TheXanMan321 looking great! I left a few comments, and it looks like there are some Hound comments as well (I'm not sure if you meant to just remove the extra `""`).

david-yz-liu:
@TheXanMan321 good work! I reviewed the coveralls check in this case, it's okay (I think it looks lower because this branch isn't up to date with master).

PR #5343, opened 2021-06-12
fixed the submitted time of starter files

JunruL:
## Motivation and Context

The column "Submitted At" under the "Starter File" tab of an assignment is empty.

## Your Changes

**Description**: I added the time in `starter_file_group_file_data` method in `assignment_controller.rb`.

## Testing

I uploaded files on the web, and the time was correct, with the same format as submit time of files under submission tab.

## Checklist


app/controllers/assignments_controller.rb
```diff
@@ -570,7 +570,7 @@ def starter_file_group_file_data(starter_file_group)
       if (starter_file_group.path + file).directory?
         { key: "#{file}/" }
       else
-        { key: file, size: 1,
+        { key: file, size: 1, submitted_date: l(Time.current),
```
david-yz-liu, line 573: @JunruL It seems that you've found the right place where file data is being sent, but you're sending the wrong value. The "Submitted at" time should be based on the property of the file that's been actually saved on disk, not just the current time.

app/controllers/assignments_controller.rb
```diff
@@ -570,7 +570,7 @@ def starter_file_group_file_data(starter_file_group)
       if (starter_file_group.path + file).directory?
         { key: "#{file}/" }
       else
-        { key: file, size: 1,
+        { key: file, size: 1, submitted_date: l(File.mtime(starter_file_group.path + file)),
```
david-yz-liu, line 573: @JunruL Good change! But it looks like you need the `in_time_zone` call in order to make sure the time appears according to the current user's time zone. (See #5341 or talk to @TheXanMan321 about this)

BTW I tested this by going to the "Settings" tab and changing the current user's timezone.

PR #5347, opened 2021-06-15
Updated automated_test_controller_spec to check for file modified times

TheXanMan321:
## Motivation and Context

The tests in automated tests do not check that the correct times have been saved.

## Your Changes

**Description**:
I use `allow` in rspec so that automated_test_file tests methods could get a modified time without actually needing a file. 

- [ ] Other (please specify): 

## Testing

I ran my tests and they passed. I also changed with the timezone of my computer to make sure my tests still passed in different time zones.

## Checklist


spec/controllers/automated_tests_controller_spec.rb
```diff
@@ -89,11 +89,18 @@
         end
       end
       context 'files data' do
+        before(:each) do
```
david-yz-liu, line 92: @TheXanMan321 this is better but still not ideal. It's a bit tricky, but check out line 22---that's where you can see what object is being used as the user making the request. You'll want to take this object and set its timezone attribute (make sure to use `.update` so that the change gets saved to the database) before making the request.

TheXanMan321: Hi David, I am having trouble updating the time_zone of the user in line 22, (`subject { get_as user, :populate_autotest_manager, params: params `}) without modifying the subject used in other tests. Should I be modifying the subject?  At the moment I am creating a new user with a fixed time zone in my test, and using that instead of the user in line 22.

david-yz-liu: @TheXanMan321 oh, okay I think you misunderstood my last message. Line 22 uses the `user` variable to create the request, but you can access this variable from within your tests as well. (The "scope" of `user` is the entire test suite, since this is a `shared_examples` test, which allows you to define tests that have a parameter (like `user`) that can be set separately in different test blocks. See Line 317 and below.)

So what you actually need to do is modify the `user` variable *in the test code*, not on Line 22. And you should do it before calling `subject`, so that the change is saved before making the request.

david-yz-liu:
Manually ran tests and checked coverage (91.89%).

PR #5353, opened 2021-06-19
fix blank page issue in submission

JunruL:
## Motivation and Context

The submission page of an assignment is blank. Submitted files do not show up in this page.

## Your Changes

**Description**: I remove .where(test_group_result_id: nil) in submission.feedback_files.where(test_group_result_id: nil).map since test_group_result_id does not exist in feedback files. show function stops at this place because of this error, and thus the page could not load.

## Testing

I review the submission of several groups' submission, and all page can load. Besides, I also find some submissions with feedback files, and the data of these files are stored in result.feedback_files correctly (using the React Developer Tools to check this).

I think the condition of where(test_group_result_id: nil) is useless so I delete it. Should I use other parameters instead of avoiding it?

## Checklist

### DIFF
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -65,7 +65,7 @@ class ResultsController < ApplicationController
         if is_reviewer
           data[:feedback_files] = []
         else
-          data[:feedback_files] = submission.feedback_files.where(test_group_result_id: nil).map do |f|
+          data[:feedback_files] = submission.feedback_files.map do |f|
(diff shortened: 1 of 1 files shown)


david-yz-liu:
@JunruL ah, good investigation. One thing to check for when you get missing attribute (or similar) errors is whether your database schema is up to date. In this particular case, the `test_group_result_id` column was added in a previous commit that changed the database schema in a migration. So you can fix th problem you were encountering by running `rails db:migrate` in your Docker container.

Issue #5172, opened 2021-03-13
feature request: switch between marking view and repo view


dianehorton:
When handling special cases, I often switch back and forth between repo view and the marking view.  Currently, I go back to Submissions (or keep that open in another browser window) to do so. It would be great to be able to go back and forth automatically.

PR #5355, opened 2021-06-21
Issue-5172: Adding a link between grading view and repository

TheXanMan321:
## Motivation and Context

This link was added to make navigating between the two views more convenient. (resolve Issue #5172 )

## Your Changes

**Description**:
I added a header and a link in the tag/notes tab of the grading view that links to the group repository. (link and header names use I18n) I also passed assignment_id and submission_id as props from right_pane to tags_panel, so that I could create the link.
image

## Testing

I tested my link on different assignments and groups.

## Checklist

### DIFF
--- a/app/assets/javascripts/Components/Result/right_pane.jsx
+++ b/app/assets/javascripts/Components/Result/right_pane.jsx
@@ -65,6 +65,9 @@ export class RightPane extends React.Component {
              removeTag={this.props.removeTag}
              newNote={this.props.newNote}
              role={this.props.role}
+             assignment_id={this.props.assignment_id}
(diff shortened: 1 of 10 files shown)


david-yz-liu (commented):
Overall this is great, but I want you to add new strings to make this change clearer. Note: you just need to add real strings for the `en.yml` files; for the other languages, add the key and use "UPDATE ME" as a placeholder.

1. The link to the repository should probably be something more like "View group repository". You can add this under the "results" view locale files.
2. The *tab name* should change. To be honest I'm not sure what the best name is, but for now we can go with "Other Info" unless you think of something better. You can add this as a string under the `common/` locale files.

PR #5356, opened 2021-06-22
Fixed TODO in remove_extra_mark regarding AJAX data formatting

devan-srinivasan:
## Motivation and Context
This PR fixed the TODO on line 474 in `app/assets/javascripts/Components/Result/result.jsx`

## Your Changes

**Description**:
 I fixed the ajax request's data to use `result.id` as opposed to the `extra_mark.id`, and instead passed that in as the `data` parameter. I also updated the controller in accordance with these changes. Secondly, this PR also changed the HTTPS request type from a `POST` to `DELETE`. Lastly, I updated all the `rspec` cases to accommodate these changes. 

## Testing

I tested this change through the UI by making an extra mark on a user's submission for a dummy assignment and then deleted it **successfully**. Secondly, I ran the `rpsec` test suite, and all tests **passed**.

## Checklist

### DIFF
--- a/app/assets/javascripts/Components/Result/result.jsx
+++ b/app/assets/javascripts/Components/Result/result.jsx
@@ -470,11 +470,11 @@ class Result extends React.Component {
 
     $.ajax({
       url: Routes.remove_extra_mark_assignment_submission_result_path(
-        this.state.assignment_id, this.state.submission_id,
+        this.state.assignment_id, this.state.submission_id, this.state.result_id
         // TODO: Fix this route so that the id refers to a Result rather than ExtraMark.
-        id
       ),
-      method: 'POST',
+      method: 'DELETE',
+      data: {extra_mark_id: id}
     }).then(this.fetchData)
   };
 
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/Components/Result/result.jsx
```diff
@@ -470,11 +470,11 @@ class Result extends React.Component {
 
     $.ajax({
       url: Routes.remove_extra_mark_assignment_submission_result_path(
-        this.state.assignment_id, this.state.submission_id,
+        this.state.assignment_id, this.state.submission_id, this.state.result_id
         // TODO: Fix this route so that the id refers to a Result rather than ExtraMark.
```
david-yz-liu, line 474: You can delete the TODO, since you've addressed it.

app/controllers/results_controller.rb
```diff
@@ -653,8 +653,8 @@ def add_extra_mark
   end
 
   def remove_extra_mark
-    extra_mark = ExtraMark.find(params[:id])
-    result = extra_mark.result
+    result = Result.find(params[:id])
+    extra_mark = ExtraMark.find(params[:extra_mark_id])
```
david-yz-liu, line 657: We can be a bit better with using associations, something like `result.extra_marks.find(...)`, so that only the extra marks associated with that result are searched.

(This prevents a user putting in a URL with an unrelated result id and extra mark id.)

david-yz-liu (commented):
@devan-srinivasan Good work again, I just left a few minor comments.

PR #5373, opened 2021-07-01
Converting the second grade entry form chart to use react-chartjs-2

devan-srinivasan:
## Motivation and Context
This PR updated the current chart system to use the latest version of React charts, that is chartjs version 2.

## Your Changes

**Description**:
My changes consisted of the following:
- adding a new route
- adding a corresponding controller method
- adding a new attribute in the react component's state to represent the chart data and options
- making an AJAX request to retrieve data for this chart
- creating new tests for the route and controller

## Testing

I tested my change in the UI confirming that the chart loads the correct data. I also ensured the rspec test suite passed all cases including the new ones I wrote.

## Checklist


devan-srinivasan:
Hey David, I recall before you said to disregard coverage/coveralls reports. Should I disregard it again in this case?

app/assets/javascripts/Components/dashboard.jsx
```diff
@@ -30,16 +34,66 @@ class Dashboard extends React.Component {
             backgroundColor: 'rgb(75, 192, 192)',
           },
         ],
-      }
+      },
+      gradeEntryFormChart2: {}
```
david-yz-liu, line 38: You don't need to create a new state for this: instead, set the `data` state above with the correct data from the server response.

You should, however, create a new state for the Chart *options* (default can be an empty object), and pass that as props to the chart component below.

app/assets/javascripts/Components/dashboard.jsx
```diff
@@ -30,16 +34,66 @@ class Dashboard extends React.Component {
...
     };
   }
 
+  GradeEntryFormColumnBreakdownUpdateData = (data) => {
+    // Helper function to update the chart state
+    let label_list = Array();
+    for(let i = 0; i < 105; i+=5) {
```
david-yz-liu, line 45: Most of this processing should be done in the controller method. That is, that method's JSON response should already be in the form of a list of objects with `label` and `data` keys. The only thing you should add in this method is the `backgroundColor` key.

app/assets/javascripts/Components/dashboard.jsx
```diff
@@ -30,16 +34,66 @@ class Dashboard extends React.Component {
...
+    })
+  }
+
   componentDidUpdate(prevProps, prevState) {
     if (prevState.assessment_id !== this.state.assessment_id) {
       if (this.state.display_course_summary) {
         // TODO
       } else if (this.state.assessment_type === 'GradeEntryForm') {
-        // TODO
+        // Make an AJAX request to retrieve chart data
```
david-yz-liu, line 88: In terms of structuring the code, you should really move this AJAX call into the helper method. That way, the helper method can be something like `getGradeEntryFormColumnBreakdown` and handle both the AJAX request and updating the component state.

app/controllers/grade_entry_forms_controller.rb
```diff
@@ -47,6 +47,22 @@ def view_summary
     @date = params[:date]
   end
 
+  # returns the column distribution/breakdown for each grade_entry_item in a grade_entry_form
+  #   (see the second chart in the summary of a mark's spreadsheet, accessed from the dashboard)
+  def column_breakdown
+    @grade_entry_form = GradeEntryForm.find(params[:id])
```
david-yz-liu, line 53: Because no views are being rendered, you don't need to worry about making this an instance attribute. So you can drop the `@` from the variable name.

app/controllers/grade_entry_forms_controller.rb
```diff
@@ -47,6 +47,22 @@ def view_summary
...
+    @grade_entry_items = @grade_entry_form.grade_entry_items unless @grade_entry_form.nil?
```
david-yz-liu, line 54: The `nil` check isn't necessary; `.find` raises an error if the grade entry form isn't found.

Also see my comment about the `@`, and about how you should be formatting the JSON in the response.

app/controllers/grade_entry_forms_controller.rb
```diff
@@ -47,6 +47,22 @@ def view_summary
     @date = params[:date]
   end
 
+  # returns the column distribution/breakdown for each grade_entry_item in a grade_entry_form
+  #   (see the second chart in the summary of a mark's spreadsheet, accessed from the dashboard)
```
david-yz-liu, line 51: mark's -> marks

david-yz-liu (commented):
@devan-srinivasan a great start. I left a few comments, and also you should add a test case for when the grade entry form actually has columns and some marks. You can look at the existing model tests to see how we've set up such data before.

PR #5377, opened 2021-07-04
Use react-chartjs-2 for assignment grade distribution chart

JunruL:
## Motivation and Context

Use react-chartjs-2 to render chart data on the dashboard page using React.

## Your Changes

**Description**: 
- created method in assignment controller
- added routes and corresponding tests
- created `options` parameter in the component's state (the four types of charts should have different options)
- added ajax request to get assignment grade distribution data 
- will add tests for the new method created in assignment controller (but now I'm not sure whether I will change the method in the future. I will add tests when the method won't be changed any more).

## Testing

The new graph is the same as the old one except:
- The background area color (I find that react-charts-2 does not support [this](https://stackoverflow.com/questions/38493564/chart-area-background-color-chartjs))
- The labels on the x-axis (also, probably react-charts-2 does not support [this](https://github.com/reactjs/react-chartjs/issues/182))

## Checklist

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -317,29 +317,29 @@ class AssignmentsController < ApplicationController
     datasets = [
       {
         label: '2',
-        data: assignment.grade_distribution_array(),
+        data: assignment.grade_distribution_array,
         backgroundColor: [
-          'rgba(36, 81, 133, 1)',
+          'rgba(36, 81, 133, 1)'
         ],
         borderColor: [
-          'rgba(36, 81, 133, 1)',
+          'rgba(36, 81, 133, 1)'
         ],
-        borderWidth: 1,
-      },
+        borderWidth: 1
+      }
     ]
-    data = {labels: labels, datasets: datasets}
+    data = { labels: labels, datasets: datasets }
     options = {
       scales: {
         xAxes: [{
-              ticks: {
-                min: 0,
-                max: 100,
-                stepSize: 10,
-              }
-            }]
+                  ticks: {
+                    min: 0,
+                    max: 100,
+                    stepSize: 10
+                  }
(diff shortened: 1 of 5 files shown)


app/controllers/assignments_controller.rb
```diff
@@ -309,6 +309,38 @@ def batch_runs
     end
   end
 
+  # Return the chart data of assignment grade distributions
+  def grade_distribution_graph_data
+    assignment = Assignment.find(params[:id])
+    labels = []
+    (0..19).each { |i| labels.push((5 * i).to_s + '-' + (5 * i + 5).to_s) }
```
david-yz-liu, line 316: Use `.map` instead of `.each`, since the former returns a list, so you don't have to `push` manually.

app/controllers/assignments_controller.rb
```diff
@@ -309,6 +309,38 @@ def batch_runs
...
+          'rgba(36, 81, 133, 1)'
+        ],
+        borderColor: [
+          'rgba(36, 81, 133, 1)'
+        ],
+        borderWidth: 1
+      }
+    ]
+    data = { labels: labels, datasets: datasets }
+    options = {
```
david-yz-liu, line 330: This is a valid approach, but I'd rather keep these "display" options in the front-end. So what you should do is define these options in the React code, rather than in this method.

david-yz-liu (commented):
@JunruL good work, this is excellent progress. I left a few inline comments, but also you should be adding tests for your new route in the `assignments_controller_spec.rb` file. You'll need to generate some real grade data to test the returned JSON of the new route, and I recommend looking in the model test file and the `assignments.rb` factory file to see what you can use.

david-yz-liu:
Thanks, @JunruL!

PR #5378, opened 2021-07-04
React-chartjs-2 for course summary grade distribution graph

emilychangg:
## Motivation and Context

Use react-chartjs-2 to create the course summary grade distribution chart

## Your Changes

**Description**:
created a new route and request
created method in course summary controller

## Checklist


app/assets/javascripts/Components/dashboard.jsx
```diff
@@ -3,6 +3,7 @@ import { render } from 'react-dom';
 
 import { Bar } from 'react-chartjs-2';
 
+import {grade_distribution_course_summaries_path} from "../../../javascript/routes";
```
david-yz-liu, line 6: You don't need to import this because `Routes` is already set as a global object (in `application.js`). So below, you can just do `Routes.grade...`.

app/controllers/course_summaries_controller.rb
```diff
@@ -46,6 +46,33 @@ def populate
...
 
+  def grade_distribution
+    table_data = []
+    marking_schemes = current_user.student? ? MarkingScheme.none : MarkingScheme
+
+    marking_schemes.order(id: :asc).each do |m|
+      table_data << m.students_weighted_grade_distribution_array(current_user).data
+    end
+    labels = []
+    (0..19).each do |i|
```
david-yz-liu, line 57: Same comment as above

app/controllers/course_summaries_controller.rb
```diff
@@ -46,6 +46,33 @@ def populate
     }
   end
 
+  def grade_distribution
+    table_data = []
+    marking_schemes = current_user.student? ? MarkingScheme.none : MarkingScheme
+
+    marking_schemes.order(id: :asc).each do |m|
```
david-yz-liu, line 53: Use map instead of each, since the former returns an array, so you don't need to `table_data <<` manually in the block.

david-yz-liu (commented):
@emilychangg a great start. I left a few comments, and also you should add to your tests to check for the actual grades that are returned. To do this, you'll need to generate test data for assessments and marking schemes; you can look in the controller and model test files for `CourseSummary` to see how existing tests generate fake data.

app/controllers/course_summaries_controller.rb
```diff
@@ -46,6 +46,33 @@ def populate
...
+      table_data << m.students_weighted_grade_distribution_array(current_user).data
+    end
+    labels = []
+    (0..19).each do |i|
+      labels << i.to_s + '-' + (i + 1).to_s
+    end
+    datasets = [
+      {
+        label: 'Weighted Total Grades',
```
david-yz-liu, line 62: One other comment I just noticed: This should be an internationalized string (using `I18n`). Check how the string is rendered in the existing dashboard page. And it it was hard-coded there too, please add this string to the locale files (under something like `views/course_summary`).

PR #5412, opened 2021-07-17
course summary graphs in new component + accounted for no marking schemes and multiple schemes

emilychangg:
## Motivation and Context

Need to account for when there are no marking schemes or when there are multiple marking schemes
Need to add statistics summary and move course summary graphs into their own component.

## Your Changes

**Description**:
- moved where the colours are being set so that it isn't hardcoded and will work for multiple marking schemes
- fixed route so it won't cause errors when there are no marking schemes
- wrote tests 
- added translations

- [x] refactoring 

## Testing

Tested through UI and also using rspec

## Checklist

### DIFF
--- a/app/assets/javascripts/Components/course_summary_chart.jsx
+++ b/app/assets/javascripts/Components/course_summary_chart.jsx
@@ -17,6 +17,7 @@ export class CourseSummaryChart extends React.Component {
         labels: [],
         datasets: [],
       },
+      marking_scheme_ids: []
     };
   }
   componentDidMount() {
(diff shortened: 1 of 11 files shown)


app/assets/javascripts/Components/dashboard.jsx
```diff
@@ -114,7 +106,17 @@ class Dashboard extends React.Component {
 
   render() {
     if (this.state.display_course_summary) {
-      return <Bar data={this.state.data} />;
+      if (this.state.data.datasets === []) {
+        return (
+          <div>
+            {/*DOESN'T SHOW UP*/}
+            <p>{I18n.t('Create a Marking Scheme to display course summary graph.')}</p>
```
emilychangg, line 113: I'm not quite sure why this doesn't show up and I'm currently doing research on this

david-yz-liu: @emilychangg In Javascript, `===` compares objects and arrays by identity (i.e., memory address), like Python's `is`. If you want to check if an array is empty, you can check its `.length` attribute and see whether it's 0.

app/assets/javascripts/Components/course_summary_chart.jsx
```diff
@@ -0,0 +1,77 @@
...
+      for (const [index, element] of res["datasets"].entries()){
+        element["label"] = I18n.t("main.weighted_total_grades") + " " + res["marking_schemes_id"][index]
+        element["backgroundColor"] = colours[index]
+      }
+      var data = {
+        labels: res['labels'],
+        datasets: res['datasets']
+      }
+      this.setState({data: data})
+      this.setState({summary: res['average']})
```
david-yz-liu, line 43: Don't do two consecutive `setState` calls. Instead, create just one combined object (with both `data` and `summary`) and pass that into a single call to `setState`.

app/assets/javascripts/Components/course_summary_chart.jsx
```diff
@@ -0,0 +1,77 @@
...
+    this.fetchData();
+  }
+
+  fetchData = () => {
+    $.ajax({
+      url: Routes.grade_distribution_course_summaries_path(),
+      type: 'GET',
+      dataType: 'json'
+    }).then(res => {
+      var average;
```
david-yz-liu, line 32: In general, prefer the newer `let` rather than `var`.

david-yz-liu (commented):
@emilychangg I know you're still working on this, but I left just a few comments to help you along.

One other thing to point out is to add tests for when there are multiple marking schemes.

PR #5420, opened 2021-07-30
Remove starter file timestamp

mishaschwartz:
## Motivation and Context

Starter file timestamps were used to help determine if files had been added after the starter files were sent to the repo. BUT, we are not sending starter files to the repo anymore, any changes made to the repo must be made by the student (or other user). Because of this, we can just check if there are any files in the repo instead of comparing their commit time to a timestamp.

This is also the simplest way to resolve a bug that was caused by setting the submissions.is_empty column value incorrectly when an assignment did not have any starter files. 

## Your Changes

**Description**:

- remove groupings.starter_file_timestamp column and all uses

## Testing

- removed unused tests

## Checklist

### DIFF
--- a/app/models/grouping.rb
+++ b/app/models/grouping.rb
@@ -464,12 +464,6 @@ class Grouping < ApplicationRecord
     self.update!(starter_file_changed: false)
   end
 
-  def changed_starter_file_at?(revision)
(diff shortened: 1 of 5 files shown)


mishaschwartz:
tests failing but fixed in #5421

PR #5445, opened 2021-08-03
Add ability to hide assessments on a per section basis 

pretendWhale:
## Motivation and Context

This pull request adds a feature allowing users to hide assignments on a per section basis, if section assignment properties (formerly section due dates) are enabled. Hiding (or making visible) an assignment for a section will override the assignment's hidden setting.

## Your Changes

**Description**:

- renames model from SectionDueDate to AssessmentSectionProperties
- Moves relation on SectionDueDates from Assignment to Assessment for potential future functionality of hiding GradeEntry Forms for sections
- adds new nullable boolean is_hidden on AssessmentSectionProperties. a null value indicates that the assignment's hidden setting should be used. 
- adds UI for hiding on a section by section basis
- - because there are three possible values for section hiding (default, hidden, visible), three radio buttons are used. Assignment hidden has been changed to radio buttons as well for consistency.
- Adds a new assignment policy, see_hidden, as well as a method on the user model visible_assessments
modifies assignments_controller and grade_entry_form controller
- any time we check "is-hidden" we also check "is hidden for section"
View changes TBA

## Testing

This has been tested via the web interface by creating new assignments and setting them to various states of hidden, both on the assignment and the section level, and then logging in as students from different sections (including no section) to confirm that the behavior works as intended. Tests for the visible_assessments function have also been added to the user spec.

## Checklist

Requires adding documentation of section hiding on the [Assignment Set Up](https://github.com/MarkUsProject/Markus/wiki/Instructor-Guide--Assignments--Setting-Up) wiki page


app/models/assignment.rb
```diff
@@ -153,15 +152,15 @@ def section_due_date(section)
       return due_date
     end
 
-    SectionDueDate.due_date_for(section, self)
+    SectionDueDate.due_date_for(section, self.assessment_id)
```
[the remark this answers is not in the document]
mishaschwartz: Also, you can still just pass `self` here. Assignments don't have their own id since they are a type of assessment. In other words, for all assignments `self.id == self.assessment_id`. 

The only reason why there is an `assessment_id` attribute available for this object is that assignments have extra information stored in the `assignment_properties` table and `assignment_properties.assessment_id` is the foreign key that describes the association between rows in that table and the assessments table. The `Assignment` model has the following line:

```
delegate_missing_to :assignment_properties
```

which means that if rails is looking up an attribute on the Assignment model and you can't find it then rails will look for it in the AssignmentProperties model. This means that `assignment.assessment_id` is really `assignment.assignment_properties.assessment_id`.

app/models/assignment.rb
```diff
@@ -153,15 +152,15 @@ def section_due_date(section)
...
   end
 
   # Return the start_time for +section+ if it is not nil, otherwise return this
   # assignments start_time instead.
   def section_start_time(section)
     return start_time unless section_due_dates_type
 
-    section&.section_due_dates&.find_by(assignment: self)&.start_time || start_time
+    section&.section_due_dates&.find_by(assessment: self.assessment_id)&.start_time || start_time
```
mishaschwartz, line 163: See comment above about `self.assessment_id` vs. `self`

app/models/section_due_date.rb
```diff
@@ -1,6 +1,6 @@
 class SectionDueDate < ApplicationRecord
   belongs_to :section
-  belongs_to :assignment, inverse_of: :section_due_dates, foreign_key: :assessment_id
+  belongs_to :assessment, inverse_of: :section_due_dates, foreign_key: :assessment_id
```
mishaschwartz, line 3: see comment above about the need to specify `foreign_key`

app/views/shared/_assignments_dropdown_menu.html.erb
```diff
@@ -29,8 +29,12 @@
 <li class='section'><%= Assignment.model_name.human.pluralize %></li>
 
 <% if @current_user.student? %>
-  <% assignments = Assignment.where(is_hidden: false) %>
-  <% grade_entry_forms = GradeEntryForm.where(is_hidden: false) %>
+  <% assignments = Assignment.joins(:section_due_dates).where(is_hidden: false,
+                                                          'section_due_dates.section_id': @current_user.section,
```
mishaschwartz, line 33: I'm surprised the linter didn't complain about this one...

You should align the arguments to the where method (like you did on lines 35-37)

mishaschwartz: What will happen here when `@current_user.section` is `nil` or there isn't a `section_due_date` associated to the given section/assignment? (Hint: `joins` does an inner join by default)

db/migrate/20210804184031_add_is_hidden_to_section_due_date.rb
```diff
@@ -0,0 +1,5 @@
+class AddIsHiddenToSectionDueDate < ActiveRecord::Migration[6.1]
+  def change
+    add_column :section_due_dates, :is_hidden, :boolean
```
mishaschwartz, line 3: I would recommend making this not nullable and false by default. This will allow you to simplify your code in `app/views/shared/_assignments_dropdown_menu.html.erb` (for example) because you can just check for boolean values and won't have to deal with `nil`.

app/models/user.rb
```diff
@@ -274,6 +274,24 @@ def reset_api_key
     self.update(api_key: Base64.encode64(md5.to_s).strip)
   end
 
+
+  def visible_assessments(type: nil, assessment_id: nil)
```
mishaschwartz, line 278: Because you're using the `@current_user.visible_assessments(assessment_id: some_id).exists?` pattern a lot. It might simplify things if you add a policy function in the assignments and grade_entry_form policy:

```ruby
def visible?
  user.visible_assessments(assessment_id: record.id).exists?
end
```

and then you can call as a policy with `allowed_to? :view?`

spec/models/user_spec.rb
```diff
@@ -160,4 +162,98 @@
...
+    context "when there are no assessments" do
+      let(:new_user) {create :user, user_name: 'testusername', first_name: 'testfname',
+                               last_name: "testlname", type: "Student"}
+      it "should return an empty list" do
+        expect(new_user.visible_assessments()).to match_array([])
+      end
+    end
+    context "when there are assessments" do
+      before :each do
+        @new_section = create(:section)
```
mishaschwartz, line 177: use `let` variables instead of creating instance variables in a before block

mishaschwartz: for example: 

```ruby
let(:new_section) { create :section }
let(:assignment_visible) { create :assignment, due_date: 2.days.from_now, assignment_properties_attributes: { section_due_dates_type: true }  }

# etc. etc.
```

If you want to force the creation of each object before each test (ie. they're not lazily loaded) you should use the `let!` syntax (https://relishapp.com/rspec/rspec-core/v/2-5/docs/helper-methods/let-and-let)

spec/models/user_spec.rb
```diff
@@ -160,4 +162,98 @@
...
+        @section_due_date_hidden = create(:section_due_date, assessment: @assignment_hidden,
+                                           section: @new_section,
+                                           due_date: 2.days.from_now,
+                                           is_hidden: true)
+      end
+      context "when section_due_dates disabled" do
+        let(:new_user_2) {create :user, user_name: 'testusername', first_name: 'testfname',
+                                 last_name: "testlname", type: "Student"}
+        it 'does return all unhiddden assignments' do
+          expect(new_user_2.visible_assessments()).to match_array([@assignment_visible, @assignment_hidden])
```
mishaschwartz, line 197: The `match_array` matcher is good if you care about order. In this case I would look at the `contains_exactly` matcher instead (https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/contain-exactly-matcher)

That way your tests won't accidentally fail if the query returns things in a different order.

spec/models/user_spec.rb
```diff
@@ -160,4 +162,98 @@
       end
     end
   end
+
+  describe '#visible_assessments' do
+
+    context "when there are no assessments" do
+      let(:new_user) {create :user, user_name: 'testusername', first_name: 'testfname',
```
mishaschwartz, line 169: If you don't care about username, last name, etc. for your tests. You can just do: 

```ruby
create :student
```

The factory will fill in values as needed.

spec/models/user_spec.rb
```diff
@@ -160,4 +162,98 @@
...
+                                last_name: "2testlname", type: 'Student'}
+       it "does not return the hidden assignment" do
+          expect(new_user.visible_assessments()).to match_array([@assignment_visible, @assignment_hidden])
+       end
+     end
+     context "when a visible assignment is given" do
+       let (:new_user) {create :student, user_name: '2testusername', first_name: '2testfname',
+                               last_name: "2testlname", section_id: @new_section.id}
+       it "does return an array with the assignment" do
+         #byebug
```
mishaschwartz, line 239: don't commit debugging stuff

app/models/user.rb
```diff
@@ -274,6 +274,24 @@ def reset_api_key
...
+    type ||= Assessment.type
+    if self.section_id
```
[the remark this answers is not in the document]
mishaschwartz: Just to clarify the comment here. I don't think that you need to change your code, it looks great as is! I just wanted to demo how you can take advantage of the ruby syntax to move things around so you are aware of the options.

db/migrate/20210804184031_add_is_hidden_to_section_due_date.rb
pretendWhale: note to self: decided to make this nullable where null represents "default" - i.e. the is_hidden property of the associated assessment

app/models/user.rb
```diff
@@ -286,7 +286,13 @@ def visible_assessments(assessment_type: nil, assessment_id: nil)
     if self.section_id
       assessments = Assessment.left_outer_joins(:section_due_dates)
```
mishaschwartz, line 287: I think this can be simplified a bit to just make it a bit clearer what the conditions we're looking for are. 

From what I remember, we want to return all assessments where:

```
section_due_date.is_hidden = false  # The specific case
OR 
section_due_date.is_hidden = null AND assessment.is_hidden = false  # The default case
```

So unless I'm missing something, I think checking for both `false` and `nil` together on line 289 is doing too much.

Does something like this work?:

```ruby
assessments = Assessment.left_outer_joins(:section_due_dates)
assessments = assessments.where('section_due_dates.section_id': self.section_id, 'section_due_dates.is_hidden': false)
                         .or('section_due_dates.is_hidden': nil, 'assessments.is_hidden': false))
```

mishaschwartz, line 287: I think this can be simplified a bit to just make it a bit clearer what the conditions we're looking for are. 

From what I remember, we want to return all assessments where:

```
section_due_date.is_hidden = false  # The specific case
OR 
section_due_date.is_hidden = null AND assessment.is_hidden = false  # The default case
```

So unless I'm missing something, I think checking for both `false` and `nil` together on line 289 is doing too much.

Does something like this work?:

```ruby
assessments = Assessment.left_outer_joins(:section_due_dates)
                        .where('section_due_dates.section_id': self.section_id)
assessments = assessments.where('section_due_dates.is_hidden': false)
                         .or('section_due_dates.is_hidden': nil, 'assessments.is_hidden': false))
```

mishaschwartz: Note to self: we changed up the logic of what is shown so the comment above no longer applies

app/models/user.rb
```diff
@@ -276,6 +281,19 @@ def reset_api_key
     self.update(api_key: Base64.encode64(md5.to_s).strip)
   end
 
+  def visible_assessments(assessment_type: nil, assessment_id: nil)
```
mishaschwartz, line 284: This look great! Can you add a docstring to this method that describes generally what it does, what the optional arguments are and what the defaults are if the arguments aren't specified. 

We use a style of ruby docstrings that looks like:

```ruby
# Here is a description of the method below and what it does... 
# It takes three arguments +a+ +b+ and +c+ and does something with them ...
def my_function(a, b: nil, c: 5)
   .....
end

```

app/policies/grouping_policy.rb
```diff
@@ -75,7 +75,7 @@ def download_starter_file?
     return false if record.assignment.is_hidden?
     return false if !record.assignment.starter_files_after_due && record.past_collection_date?
     return true unless record.assignment.is_timed?
-
+    record.assignment ?  !check?(:see_hidden?, record.assignment) : !check?(:see_hidden?, record.grade_entry_form)
```
mishaschwartz, line 78: groupings are only ever associated to assignments, not to grade entry forms so you can simplify this to:

```ruby
!check?(:see_hidden?, record.assignment)
```

Also, I'm assuming that this is meant to replace the check on line 75 (`return false if record.assignment.is_hidden?`) so you should replace that line instead of adding it to the bottom. Something like:

```ruby
return false unless check?(:see_hidden?, record.assignment)
```

app/views/assignments/student_assignment_list.html.erb
```diff
@@ -1,8 +1,9 @@
 <% content_for :title, t('menu.home') %>
 
-<% upcoming, past = Assessment.where(is_hidden: false)
+<% upcoming, past = @current_user.visible_assessments()
```
mishaschwartz, line 3: You don't need the `()` after visible assessments here. Then align the next few lines so the `.` lines up

app/views/layouts/menus/_student_sub_menu.html.erb
```diff
@@ -65,7 +65,7 @@
         </li>
       <% end %>
 
-      <% if @assignment.has_peer_review_assignment? && !@assignment.pr_assignment.is_hidden %>
+      <% if @assignment.has_peer_review_assignment? && allowed_to?(:see_hidden?, @assignment.pr_assignment.id) %>
```
mishaschwartz, line 68: Remove the `.id` at the end. Otherwise the `record` will be the assignment's id (an integer) not the assignment object itself.

db/migrate/20210804184031_add_is_hidden_to_section_due_date.rb
```diff
@@ -0,0 +1,5 @@
+class AddIsHiddenToSectionDueDate < ActiveRecord::Migration[6.1]
+  def change
```
mishaschwartz, line 2: Another thing we should change in the database is the name of the table itself. From section_due_dates to assessment_section_properties (or whatever it was we decided).

This will require updates to the model name too but that should be easy to fix with some find/replace

spec/models/user_spec.rb
```diff
@@ -161,4 +163,249 @@
       end
     end
   end
+
+  describe '#visible_assessments' do
```
mishaschwartz, line 167: These look great. Can you add one more test in for when the user is a student and the assignment is a peer review assignment.

mishaschwartz (changes requested):

app/controllers/assignments_controller.rb
```diff
@@ -62,7 +62,7 @@ def show
   def peer_review
     assignment = Assignment.find(params[:id])
     @assignment = assignment.is_peer_review? ? assignment : assignment.pr_assignment
-    if @assignment.nil? || @assignment.is_hidden
+    unless @assignment.nil? || !allowed_to?(:see_hidden?)
```
mishaschwartz, line 65: Please see my previous comment, the logic is still backwards here. It should be either:

```ruby
if @assignment.nil? || !allowed_to(:see_hidden?)

# OR 

unless !@assignment.nil? || allowed_to(:see_hidden?)
```

app/controllers/assignments_controller.rb
```diff
@@ -62,7 +62,7 @@ def show
   def peer_review
     assignment = Assignment.find(params[:id])
     @assignment = assignment.is_peer_review? ? assignment : assignment.pr_assignment
-    if @assignment.nil? || @assignment.is_hidden
+    unless @assignment.nil? || allowed_to?(:see_hidden?)
```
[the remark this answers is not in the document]
mishaschwartz: The logic is still not right here, please see my previous comment

app/models/user.rb
mishaschwartz: I'm going to be really picky here, but I think it's worth it to be precise when writing documentation:

- change `"Defaults to Assessment"` to a description of what passing the parameter actually does. Something like: 
`If passed one of these options, only assessments of this type will be returned. Otherwise returns all assessment types`
- change `if passed and assessment...` to something like `if passed an assessment id, only the assessment with the given id will be returned if it is visible to the current user`. As it stands, it sounds like you are expecting an assessment object (not an id) and the function will return a boolean, not a collection of assessments.

config/locales/models/assignments/en.yml
```diff
@@ -25,7 +25,7 @@ en:
         remark_due_date: Remark Due Date
         remark_message: Remark Request Instructions
         repository_folder: Repository Folder
-        section_due_dates_type: Enable section due dates
+        section_due_dates_type: Enable section properties
```
mishaschwartz, line 28: I'm thinking about what wording makes the most sense to a typical user here. What do you think about:

- Enable section specific settings
- Set section specific settings
- Set settings for each section
- ... 

Something like that? I don't know if calling it section properties makes the most sense to a user (who doesn't have a view of the database and knows that's what the table name/ model name is.

pretendWhale: Agreed, "section properties" doesn't make a ton of sense for end users. I like "Enable section specific settings", because it seems to me to be the most precise description of what checking the box does. If I saw "set", I would probably assume that it will automatically generate some sort of default. It kind of does this for the hidden property, but doesn't (at least not visibly) for due date. If checking the section-specific settings box prefilled the section's due date with the assignment's due date, for example, I'd feel a bit better about using set. Either way, I've gone with section specific preferences, but none of these are strong preferences.

app/controllers/assignments_controller.rb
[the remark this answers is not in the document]
pretendWhale: Somehow I got both of these confused in my head and ended up actually replicating the problematic behavior in a different way. sorry, fixed now

app/models/user.rb
pretendWhale: Fair enough, I agree! I've added the description of the `assessment_type` parameter and changed the description of `assessment_id` to indicate it will return a collection

app/controllers/assignments_controller.rb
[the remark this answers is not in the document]
mishaschwartz: no worries :)

config/locales/models/assignments/en.yml
mishaschwartz: That makes sense, why don't we go with "Enable section specific settings" then

mishaschwartz (approved):

david-yz-liu (commented):
@pretendWhale nice work! I just left a few inline comments. Also, I noticed there were several whitespace changes (additions or removals of blank lines). I see that Misha said you don't need to fix them for this PR, but please don't do this in future work.

Changelog.md
```diff
@@ -1,7 +1,7 @@
 # Changelog
 
 ## [unreleased]
-
+- Rename SectionDueDates to AssessmentSectionProperties, and add the ability to hide assignments from individual sections. Note this setting overrides the assignment's default hidden status.
```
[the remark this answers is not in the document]
mishaschwartz: I'll also add that this should be shortened... the change descriptions in the changelog file should be very concise, usually only a single clause. I'd shorten this to just:

> Add the ability to hide assignments from individual sections

app/models/section.rb
```diff
@@ -18,7 +18,7 @@ def count_students
   end
 
   def section_due_date_for(aid)
```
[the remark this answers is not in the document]
mishaschwartz: It's being called by Student.due_date_for_assignment but that one isn't being used anywhere so I would remove that one as well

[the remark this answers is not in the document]
pretendWhale: I've removed both of these.

app/views/assignments/_form.html.erb
```diff
@@ -98,22 +98,29 @@
...
         <% unless @sections.empty? %>
           <%= ap_f.label :section_due_dates_type, Assignment.human_attribute_name(:section_due_dates_type) %>
           <span>
             <%= ap_f.check_box :section_due_dates_type %>
             &nbsp; (<%= t('assignments.form.section_due_dates_option') %>)
           </span>
         <% end %>
+        <%= f.label :is_hidden, Assignment.human_attribute_name(:is_hidden) %>
+        <span>
+          <%= f.label :is_hidden, I18n.t('assignments.section_hidden.hidden') %>
```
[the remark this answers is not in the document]
pretendWhale: Thanks for the heads up! Done, and will keep in mind for the future

app/views/assignments/_peer_review_form.html.erb
```diff
@@ -69,35 +69,52 @@
             &nbsp; (<%= t('assignments.form.section_due_dates_option') %>)
           </span>
         <% end %>
+        <%= f.label :is_hidden, Assignment.human_attribute_name(:is_hidden) %>
```
david-yz-liu, line 72: You should make the same changes in this file as the assignment form.

app/models/user.rb
```diff
@@ -276,6 +281,25 @@ def reset_api_key
     self.update(api_key: Base64.encode64(md5.to_s).strip)
   end
 
+  # Determine what assessments are visible to the user.
+  # By default, returns all assessments visible to the user.
+  # Optional parameter assessment_type takes values "Assignment" or "GradeEntryForm". If passed one of these options,
+  # only returns assessments of that type. Otherwise returns all assessment types.
+  # Optional parameter assessment_id: if passed an assessment id, returns a collection containing
+  # only the assessment with the given id, if it is visible to the current user. If it is not visible, returns an empty collection.
```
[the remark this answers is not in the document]
david-yz-liu: Split this line

app/controllers/assignments_controller.rb
```diff
@@ -160,14 +160,16 @@ def edit
       end
     end
 
-    # build section_due_dates for each section that doesn't already have a due date
+    # build assessment_section_properties for each section that doesn't already have a due date
```
david-yz-liu, line 163: You should change the "due date" at the end of this comment as well

david-yz-liu (approved):
Looks good @pretendWhale!

PR #5469, opened 2021-08-24
Make sure downloadable test specs file is portable between assignments and instances

mishaschwartz:
## Motivation and Context

Test spec files could be downloaded but contained unique ids for criteria and test groups that only made sense in the context of the assignment that they were downloaded from. This meant that these files could not be used to (easily) move test settings between assignments.

This PR removes/replaces those ids so that these spec files can be used to move settings in this way. 

## Your Changes

**Description**:

- spec file now identifies criteria based on type and name (instead of id)
- test group ids are not downloaded as part of the spec file (new test groups will be created when re-uploaded)

## Checklist

### DIFF
--- a/app/controllers/automated_tests_controller.rb
+++ b/app/controllers/automated_tests_controller.rb
@@ -182,9 +182,15 @@ class AutomatedTestsController < ApplicationController
     assignment = Assignment.find(params[:assignment_id])
     file_path = assignment.autotest_settings_file
     if File.exist?(file_path)
-      send_file file_path, filename: params[:file_name]
+      specs = JSON.parse File.read(file_path)
+      specs['testers']&.each do |tester_info|
+        tester_info['test_data']&.each do |test_info|
+          test_info['extra_info']&.delete('test_group_id')
+        end
+      end
+      send_data specs.to_json, filename: TestRun::SPECS_FILE
     else
-      send_data '{}', filename: file_path
+      send_data '{}', filename: TestRun::SPECS_FILE
     end
   end
 
@@ -193,12 +199,13 @@ class AutomatedTestsController < ApplicationController
     if params[:specs_file].respond_to? :read
       file_content = params[:specs_file].read
       begin
-        JSON.parse file_content
+        test_specs = JSON.parse file_content
       rescue JSON::ParserError
         flash_now(:error, I18n.t('automated_tests.invalid_specs_file'))
         head :unprocessable_entity
       else
         File.write(assignment.autotest_settings_file, file_content, mode: 'wb')
+        update_test_groups_from_specs(assignment, test_specs)
         @current_job = AutotestSpecsJob.perform_later(request.protocol + request.host_with_port, assignment)
         session[:job_id] = @current_job.job_id
         render 'shared/_poll_job.js.erb'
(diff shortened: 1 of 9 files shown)


app/helpers/automated_tests_helper.rb
```diff
@@ -38,18 +38,21 @@ def update_test_groups_from_specs(assignment, test_specs)
...
+      tester_specs['test_data']&.each do |test_group_specs|
         test_group_specs['extra_info'] ||= {}
         extra_data_specs = test_group_specs['extra_info']
         test_group_id = extra_data_specs['test_group_id']
         display_output = extra_data_specs['display_output'] || TestGroup.display_outputs.keys.first
         test_group_name = extra_data_specs['name'] || TestGroup.model_name.human
         criterion_id = nil
         unless extra_data_specs['criterion'].nil?
-          criterion_id = extra_data_specs['criterion']
+          criterion_id = criteria_map[extra_data_specs['criterion']]
```
mishaschwartz, line 55: If we can't find the criterion id here then show a warning message that the criterion can't be found

pretendWhale (approved):
This looks good to me!

app/controllers/automated_tests_controller.rb
```diff
@@ -193,12 +199,13 @@ def upload_specs
...
       file_content = params[:specs_file].read
       begin
-        JSON.parse file_content
+        test_specs = JSON.parse file_content
       rescue JSON::ParserError
         flash_now(:error, I18n.t('automated_tests.invalid_specs_file'))
         head :unprocessable_entity
       else
         File.write(assignment.autotest_settings_file, file_content, mode: 'wb')
+        update_test_groups_from_specs(assignment, test_specs)
```
david-yz-liu, line 208: I like reusing this method, but given this call there's now more opportunity to have a runtime error (I just encountered one when uploading an old specs file that has an int instead of string for the criterion field).

How about we use a begin-rescue block around this call and flash an error message if an exception gets raised. Additionally, I think the file should only be saved and sent to the autotesting server if there's no exception?

app/helpers/automated_tests_helper.rb
```diff
@@ -38,18 +38,25 @@ def update_test_groups_from_specs(assignment, test_specs)
...
     test_group_ids = []
-    test_specs['testers'].each do |tester_specs|
+    criteria_map = assignment.ta_criteria.pluck(:type, :name, :id).map do |type, name, id_|
+      ["#{type}:#{name}", id_]
+    end.to_h
+    test_specs['testers']&.each do |tester_specs|
       next if tester_specs['test_data'].nil?
 
-      tester_specs['test_data'].each do |test_group_specs|
+      tester_specs['test_data']&.each do |test_group_specs|
```
david-yz-liu, line 47: In this case, I think the check on the preceding line ensures that `tester_specs['test_data']` is not nil. I'm hesitant to use `&.` when it's not required.

david-yz-liu (commented):
Thanks @mishaschwartz! I left two minor comments, but also I'd love to see two additional `upload_specs` tests for the `criterion` field: one where the criterion doesn't exist and one where it does.

david-yz-liu (approved):
Thanks!

Issue #5385, opened 2021-07-07
Migrate assignments between instances


mishaschwartz:
Allow instructors to download/upload assignment content to make it easier to transfer assignments from one instance to another.

This will allow instructors to more easily migrate assignments between terms (for example).

mishaschwartz:
The assignment information that should be transferred includes:

- assignment properties
- tags
- criteria
- annotations
- starter files and settings
- automated test files and settings


No information about students or groups should be transferred, this includes things like:

- group membership information
- marks
- associations between annotations and submissions
- etc.

Note that the primary database keys may not be consistent when copying assignments. They should not be relied on to copy associations between objects in the database.

mishaschwartz:
First PR:

Create a new route that downloads an empty file (to be filled in later)

- create a download route
- create a download controller method
- create a policy for the controller method
- create tests

PR #5498, opened 2021-09-15
Modified UI and added buttons for new assignment copy feature

anoverga17:
## Motivation and Context

Second part of creating a new feature to download and upload assignment settings. This pull request makes some changes to the UI in order for the feature to be easily used by an end user (and also to aid in testing the feature). More specifically, the layout of the assignment settings and creation page has been changed to accommodate header buttons that allow the user to download or upload the necessary zip files. 

## Your Changes

**Description**:
- Changed the edit and create pages to use the assignment_content layout
- Changed the edit and create pages to use headers provided by assignment_content
- Added a download button to the edit assignment page
- Added an upload button to the create assignment page
- Created new download/upload modals for each button
- Added a new translation for use with the upload modal

These changes should not affect other areas of the application. The changes made ensured that only layout of the edit and create assignment pages were modified. In addition, the edit and create pages themselves were tested to ensure the only noticeable change was the addition of new buttons. 

## Testing

These changes have primarily been tested by manual testing through the web interface. The create and edit pages have been tested to ensure the change in layout did not create significant UI changes. This was done by comparing the layout of the new create and edit pages with their previous layout. 

In addition, the new buttons have been tested by manually clicking the buttons to open the new modals and browsing for/downloading files. 

## Checklist


app/views/assignments/_cp_assignment_download_modal.html.erb
```diff
@@ -0,0 +1,19 @@
+<%= content_for :modal_id, 'download_dialog' %>
```
mishaschwartz, line 1: We're not going to be offering any options here for how to download the file (in other places we offer different file formats but here it's just going to be a zip file). So there is no need to have a modal for this. When you click on the download link it should just download the file without any intermediate steps.

config/locales/views/assignments/en.yml
```diff
@@ -73,4 +73,5 @@ en:
       started_on: 'You started on:'
       starter_file_prompt: Click below to download the starter files for this assessment.
       time_until_due_warning: This assessment is due before %{due_date} without penalty.
+    upload_config_help_html: ' This zip file will configure and create this assignment.<br> For more information please visit <a href="https://github.com/MarkUsProject/Markus/wiki/Instructor-Guide--Importing-and-Exporting-Data">this page</a>.'
```
mishaschwartz, line 76: Please remove the whitespace at the beginning of the string.

mishaschwartz: This is good but we should make sure to actually include information about this in the wiki. You can add proposed changes to the wiki by editing this file https://github.com/MarkUsProject/Markus/wiki/Pending-Changes.

You can edit that file to describe the changes that you would make to the wiki including:

- which page will to changes go (in this case it's the "Instructor-Guide--Importing-and-Exporting-Data" page)
- where on the page will you add the new information
- what information will you add/change/delete etc.
- which pull request this relates to

When you've done that, update the link here to reference the precise section on the page that you have added so when users click on it, it will take them to the precise location on the page with the relevant information. 

When we merge these changes into production we will then add the content of the Pending Changes page to actual wiki.

mishaschwartz (changes requested):
Using the assignment_content layout is a good idea but if you're going to do that you should remove the `<%= render 'shared/flash_message' %>` line from _form.html.erb. Otherwise flash messages will be displayed twice on these pages (assignment_content.html.erb also includes that line and we don't need it twice).

app/views/assignments/_cp_assignment_upload_modal.html.erb
```diff
@@ -0,0 +1,30 @@
+<%= content_for :modal_id, 'upload_dialog_config' %>
+<%= content_for :modal_title, t('upload_the', item: 'Configuration Zip') %>
+<%= content_for :modal_open_link, '#uploadModalConfig' %>
+
+<%= content_for :modal_content do %>
+
+  <!-- Insert assignment-configuration section_id once documentation has been created -->
+  <%= t('assignments.upload_config_help_html', section_id: '') %>
```
mishaschwartz, line 8: I would just add the section_id argument here based on your proposed wiki changes ('assignment-configuration'). That way you don't have to remember to come back to this and fix it later on.

As a side note... usually it's good practice to add `TODO: ` at the beginning of these sort of comments. That makes it easier to find them later when you're doing a final cleanup to make sure you didn't forget anything.

anoverga17: I was concerned this may lead to a broken link but I see your point. I will add it in. I will also keep the TODO in mind

mishaschwartz: It's a good concern but the nice thing is in this case if the section_id is wrong it will still take you to the top of the file. Anything after the '#' in a url will take you to the element with that id but if it can't find it, it will just stay at the top of the file. So something like this just takes you to the top of the page: https://github.com/MarkUsProject/Markus/wiki/Instructor-Guide--Importing-and-Exporting-Data#this-is-not-an-actual-secion

mishaschwartz (approved):

PR #5510, opened 2021-09-22
Prevent cookie overflow error if the nbconvert fails

mishaschwartz:
## Motivation and Context

If the nbconvert operation fails, the reported error is really really big. This can result in a cookie overflow error since the error is reported as a flash message and the flash cookie can only take up to 4 Kb. 

## Your Changes

**Description**:

- only report the last line of the error message. Since this is a python error, this should be the most informative bit of the traceback anyway.

## Checklist


app/controllers/submissions_controller.rb
```diff
@@ -707,7 +707,8 @@ def notebook_to_html(file_contents, unique_path, type)
       end
       _stdout, stderr, status = Open3.capture3(*args, stdin_data: file_contents)
       unless status.exitstatus.zero?
-        flash_message(:error, stderr)
+        # Only display last line to avoid a cookie overflow error
+        flash_message(:error, stderr.lines.last)
```
david-yz-liu, line 711: Please move this message into the returned string on the next line

david-yz-liu (approved):

PR #5516, opened 2021-09-23
Added upload and download assignment properties

anoverga17:
## Motivation and Context

Third part of creating a new feature to download and upload assignment settings. This pull request adds the ability to download and upload all the properties that relate to an assignment.

## Your Changes

**Description**:
- Changed upload route to be a collection (since uploading adds to a collection)
- Added a get_assignment_properties method to the assignment model that compiles all the properties of an assignment to a yaml format
- Modified download_config_files method to create an assignment_properties yaml file and write it to the zip file

- [ ] Other (please specify):

## Checklist

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -561,7 +561,9 @@ class AssignmentsController < ApplicationController
     FileUtils.rm_f(zip_path)
 
     Zip::File.open(zip_path, create: true) do |zipfile|
-      zipfile.get_output_stream('sample file') { |f| f.write 'Just a file' }
+      zipfile.get_output_stream("#{assignment.short_identifier}-properties.yml") { |f| 
+        f.write assignment.get_assignment_properties
+      }
     end
     send_file zip_path, filename: zip_name
   end
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -1156,6 +1156,24 @@ class Assignment < Assessment
     self.assignment_properties.attributes.merge(self.attributes).symbolize_keys.to_json(options)
   end
 
+  # Returns an assignment's relevant properties for uploading/downloading an assignment's configuration
+  def get_assignment_properties
+    properties = {}
+    
+    # Insert DEFAULT_FIELDS first to make file more readable
+    DEFAULT_FIELDS.each do |f|
+      properties[f] = self.assignment.send(f)
+    end
+    all_properties = self.assignment_properties.attributes.merge(self.attributes).symbolize_keys
+    # Delete irrelevant/included keys
+    all_properties.delete(:id)
+    all_properties.delete(:assessment_id)
+    DEFAULT_FIELDS.each do |f|
+      all_properties.delete(f)
+    end
+    properties.merge(all_properties).to_yaml
+  end
+
   # zip all files in the folder at +self.autotest_files_dir+ and return the
   # path to the zip file
   def zip_automated_test_files(user)
--- a/app/views/assignments/_cp_assignment_upload_modal.html.erb
+++ b/app/views/assignments/_cp_assignment_upload_modal.html.erb
@@ -13,7 +13,7 @@
                              button_id: 'submit_upload_for_config',
                              nonce: true %>
 
-  <%= form_tag upload_config_files_assignment_path(:id),
+  <%= form_tag upload_config_files_assignments_path,
                { multipart: true, size: 1 } do %>
 
     <%= file_field_tag 'upload_file', required: true, accept: '.zip' %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -69,13 +69,13 @@ Rails.application.routes.draw do
       get 'download'
       post 'upload'
       get 'batch_runs'
+      post 'upload_config_files'
     end
 
     member do
       get 'download_sample_starter_files'
       get 'download_starter_file_mappings'
       get 'download_config_files'
-      post 'upload_config_files'
       get 'view_summary'
       post 'update_starter_file'
       get 'peer_review'


app/models/assignment.rb
```diff
@@ -1156,6 +1156,31 @@ def to_json(options = {})
...
 
+  # Returns an assignment's relevant properties for uploading/downloading an assignment's configuration as a hash
+  def assignment_properties_config
+    properties = {}
+    # Insert DEFAULT_FIELDS first to make file more familiar
+    DEFAULT_FIELDS.each do |f|
+      properties[f] = self.assignment.send(f)
+    end
+    # Get all other assignment properties
+    other_properties = self.assignment_properties.attributes.merge(self.attributes).symbolize_keys
```
mishaschwartz, line 1167: I don't love the idea of just grabbing all remaining keys from assignment_properties. At the very least I would prefer to have which keys you want listed somewhere. Either update `DEFAULT_FIELDS` to include the other keys you want to grab or create another global variable similar to `DEFAULT_FIELDS` that includes the rest of the keys you want to grab and use that.

app/models/assignment.rb
```diff
@@ -1156,6 +1156,31 @@ def to_json(options = {})
...
+    section_dates = SectionDueDate.where(assessment_id: self.id).pluck_to_hash(:due_date, :section_id, :start_time)
```
mishaschwartz, line 1168: use `self.section_due_dates` instead of `SectionDueDate.where(assessment_id: self.id)`

app/models/assignment.rb
```diff
@@ -1156,6 +1156,31 @@ def to_json(options = {})
...
+    required_files = AssignmentFile.where(assessment_id: self.id).pluck_to_hash(:filename)
```
mishaschwartz, line 1169: use `self.assignment_files`

app/models/assignment.rb
```diff
@@ -1156,6 +1156,31 @@ def to_json(options = {})
...
+    late_policy = SubmissionRule.where(assessment_id: self.id)
```
mishaschwartz, line 1170: use `self.submission_rule.periods`

app/models/assignment.rb
```diff
@@ -1156,6 +1156,31 @@ def to_json(options = {})
...
+                                .joins('submission_rules LEFT JOIN periods
+                                        ON periods.submission_rule_id = submission_rules.id')
+                                .pluck_to_hash(:type, :submission_rule_id, 'periods.id',
+                                               :deduction, :hours, :interval, :submission_rule_type)
+    # Remove unneeded attributes and merge all properties together
+    other_properties.delete(:id)
```
mishaschwartz, line 1176: This is related to the comment above, but I would rather just grab the attributes that we need initially instead of getting them all at once and then deleting them afterwards

app/models/assignment.rb
```diff
@@ -1156,6 +1156,31 @@ def to_json(options = {})
...
+    other_properties.delete(:assessment_id)
+    properties.merge(other_properties)
+              .merge('section_due_dates' => section_dates)
```
mishaschwartz, line 1179: This is a style comment:

use this syntax for hashes `section_due_dates: section_dates` instead of the (old style) syntax with `=>`

app/controllers/assignments_controller.rb
```diff
@@ -561,14 +561,28 @@ def download_config_files
...
+        f.write assignment.assignment_properties_config.to_yaml }
     end
     send_file zip_path, filename: zip_name
   end
 
   # Uploads a zip file containing all the files specified in download_config_files
   # and modifies the assignment settings according to those files.
   def upload_config_files
+    begin
+      data = process_file_upload
```
mishaschwartz, line 574: `process_file_upload` handles uploading csv or yml files. You are uploading a zip file so you have to handle this upload differently

app/controllers/assignments_controller.rb
```diff
@@ -561,14 +561,28 @@ def download_config_files
...
+      flash_message(:error, t('upload_errors.syntax_error', error: e.to_s))
+    rescue StandardError => e
+      flash_message(:error, e.message)
+    else
+      if data[:type] == '.zip'
+        flash_message(:error, result[:invalid_lines]) unless result[:invalid_lines].empty?
+        flash_message(:success, result[:valid_lines]) unless result[:valid_lines].empty?
+      end
+    end
+    redirect_to action: 'create'
```
mishaschwartz, line 585: This method will create the new assignment so afterwards it would make more sense to redirect to the edit route for the newly created assignment

app/controllers/assignments_controller.rb
mishaschwartz: For example, look at the `UploadHelper#upload_files_helper` for an example of how to unzip an uploaded zip file

mishaschwartz (changes requested):
Overall I think you have the right idea here, there are just a few things you can do to improve the efficiency and style of the code. Let me know if you have questions

app/models/assignment.rb
```diff
@@ -1156,6 +1156,31 @@ def to_json(options = {})
...
+      properties[f] = self.assignment.send(f)
+    end
+    # Get all other assignment properties
+    other_properties = self.assignment_properties.attributes.merge(self.attributes).symbolize_keys
+    section_dates = SectionDueDate.where(assessment_id: self.id).pluck_to_hash(:due_date, :section_id, :start_time)
+    required_files = AssignmentFile.where(assessment_id: self.id).pluck_to_hash(:filename)
+    late_policy = SubmissionRule.where(assessment_id: self.id)
+                                .joins('submission_rules LEFT JOIN periods
+                                        ON periods.submission_rule_id = submission_rules.id')
+                                .pluck_to_hash(:type, :submission_rule_id, 'periods.id',
```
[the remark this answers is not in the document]
anoverga17: I considered that but I am concerned that some information may be lost/duplicated. Namely, the schema seems to suggest that an assignment can have multiple submission types. If I were to remove the ids, I'm concerned that I would not know which period belongs to which submission_rule. 
I tried to make the hash hierarchical to fix this but the code I ended up writing was pretty inefficient compared to just joining the two tables.  
Any suggestions on how to approach this would be appreciated.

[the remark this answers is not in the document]
anoverga17: Never mind, I was looking at the schema.rb file for a UNIQUE keyword but after taking a look at the top of the assignment.rb file, it seems there is only one submission_rule per assignment. That makes a lot more sense. I will definitely make this change then

app/models/assignment.rb
```diff
@@ -1156,6 +1156,30 @@ def to_json(options = {})
...
+    fields.each do |f|
+      properties[f] = self.assignment.send(f)
+    end
+    # Get all other assignment properties
+    properties[:submission_rule] = self.submission_rule.type
+    late_periods = self.submission_rule.periods.pluck_to_hash(:deduction, :hours, :interval)
+    section_dates = self.section_due_dates.pluck_to_hash(:due_date, :start_time)
+    required_files = self.assignment_files.pluck_to_hash(:filename)
+    # Merge all properties together
+    properties.merge('submission_rule_periods': late_periods)
```
mishaschwartz, line 1178: Don't chain merge like this. You can just pass multiple keys/value pairs as the arguments to a single merge call

app/views/assignments/_cp_assignment_upload_modal.html.erb
```diff
@@ -8,21 +8,21 @@
...
   <%= javascript_include_tag 'upload_button_control.js',
-                             id: 'upload_script_for_config',
-                             upload_id: 'upload_config_files',
+                             id: 'upload_script',
+                             upload_id: 'upload_files_for_config',
                              button_id: 'submit_upload_for_config',
                              nonce: true %>
 
-  <%= form_tag upload_config_files_assignment_path(:id),
-               { multipart: true, size: 1 } do %>
```
mishaschwartz, line 17: why remove the size?

app/views/assignments/_boot.js.erb
```diff
@@ -36,7 +36,7 @@
   function check_required_fields() {
     var invalid = false;
     $('input[required]:not([disabled])').each(function() {
-      if (this.value === '') {
+      if (this.value === '' && this.id !== 'upload_files_for_config') {
```
mishaschwartz, line 39: Can you explain the reason for these changes? Maybe I'm missing something?

mishaschwartz (changes requested):

app/views/assignments/_boot.js.erb
anoverga17: The function here disables all submit functionality unless all required fields are filled. Since the upload modal is on the same page as the normal form and because the zip file is a required field, one cannot submit nor upload an assignment without both filling out the required form fields and providing a zip file to submit. 
This was a quick and dirty way to ignore the upload zip file as a required field for the form as a whole while still making it a required field for the upload modal.

app/views/assignments/_cp_assignment_upload_modal.html.erb
anoverga17: I did not know what it did (I had thought it limited the upload capacity to one file but that was wrong) and when I passed the equivalent html into an html validator (as I was debugging the disabled buttons issue), it raised an error so I removed it to avoid any unnecessary code

app/controllers/assignments_controller.rb
```diff
@@ -561,18 +561,106 @@ def download_config_files
...
+          flash_message(:error, 'Assignment could not be created')
+          redirect_to action: 'new'
+        end
+      end
+    end
   end
 
   private
 
+  def configure_properties(zipfile)
```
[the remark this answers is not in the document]
anoverga17: I had thought the tempfile was necessary in order to safely extract the contents from the zip file but now that you point it out, that makes more sense. I will change this.

app/views/assignments/_cp_assignment_upload_modal.html.erb
mishaschwartz: Got it. If you want to limit it to one file, then don't set the `:multiple` argument on the `file_field_tag` (https://apidock.com/rails/ActionView/Helpers/FormTagHelper/file_field_tag)

app/controllers/assignments_controller.rb
```diff
@@ -554,25 +554,135 @@ def download_sample_starter_files
...
+                 :due_date, :is_hidden, :show_total, :type]
+    YAML.safe_load(
+      file.get_input_stream.read.encode(Encoding::UTF_8, 'UTF-8'),
+      [Date, Time, Symbol, ActiveSupport::TimeWithZone, ActiveSupport::TimeZone,
+       ActiveSupport::HashWithIndifferentAccess, ActiveSupport::Duration],
+      [],
+      true
+    ).deep_symbolize_keys.each do |attribute_pair|
+      attribute, value = attribute_pair
+      case attribute
```
mishaschwartz, line 668: Since you're also deciding the format of the yaml file that gets downloaded, why not just download it in the same format that you want when you upload it instead of reconfiguring the hash here

mishaschwartz (changes requested):

app/controllers/assignments_controller.rb
```diff
@@ -554,25 +554,135 @@ def download_sample_starter_files
...
+            redirect_to new_assignment_path
+            return nil
+          end
+          assignment.repository_folder = assignment.short_identifier
+          Assignment.transaction do
+            child_assignment = create_assignment_from_zip(zipfile, 'pr-config-files/properties.yml')
+            unless child_assignment.nil?
+              child_assignment.parent_assignment = assignment
+              child_assignment.repository_folder = assignment.repository_folder
+              raise ActiveRecord::Rollback unless child_assignment.save
```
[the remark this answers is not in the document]
anoverga17: It did not give me an error and the child assignment seemed to be created fine. I can save the assignment first though

app/controllers/assignments_controller.rb
```diff
@@ -554,25 +554,135 @@ def download_sample_starter_files
...
   end
 
   # Uploads a zip file containing all the files specified in download_config_files
   # and modifies the assignment settings according to those files.
   def upload_config_files
+    upload_file = params.require(:upload_files_for_config)
+    if upload_file.size == 0
+      flash_message(:error, I18n.t('upload_errors.blank'))
+    end
+    if File.extname(upload_file.path).casecmp?('.zip')
```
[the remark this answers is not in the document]
anoverga17: I just checked this and since I am using `casecmp?` here this should return a boolean regarding whether two strings are equal.

app/controllers/assignments_controller.rb
```diff
@@ -554,25 +554,144 @@ def download_sample_starter_files
...
+    end
+    if File.extname(upload_file.path).casecmp?('.zip') && !error
+      Zip::File.open(upload_file.path) do |zipfile|
+        Assignment.transaction do
+          # Build assignment from properties
+          prop_file = zipfile.find_entry('properties.yml')
+          assignment = build_uploaded_assignment(prop_file)
+          if assignment.nil?
+            error = true
+            break
```
mishaschwartz, line 594: a break statement will not rollback a transaction as of rails 6.1 (what we are currently using) but will for rails 6.2 (https://www.bigbinary.com/blog/rails-6-1-deprecates-the-use-of-return-break-or-throw-to-exit-a-transaction-block)

To avoid confusion moving forward, raise a rollback exception instead of breaking, returning, throwing from inside a transaction block : `raise ActiveRecord::Rollback`

app/models/assignment.rb
```diff
@@ -1156,6 +1156,36 @@ def to_json(options = {})
...
+                         :student_form_groups, :remark_due_date, :remark_message,
+                         :assign_graders_to_criteria, :enable_test, :enable_student_tests,
+                         :allow_remarks, :display_grader_names_to_students, :display_median_to_students,
+                         :group_name_autogenerated, :vcs_submit, :has_peer_review, :group_name_displayed,
+                         :invalid_override, :section_groups_only, :section_due_dates_type, :unlimited_tokens,
+                         :only_required_files, :token_start_date, :token_period, :non_regenerating_tokens,
+                         :scanned_exam, :anonymize_groups, :hide_unassigned_criteria, :duration,
+                         :start_time, :is_timed, :starter_file_type, :starter_files_after_due]
+    fields.each do |f|
+      properties[f] = self.assignment.send(f)
```
mishaschwartz, line 1174: what does `self.assignment` refer to and is it different from `self` ?

app/controllers/assignments_controller.rb
```diff
@@ -603,7 +603,7 @@ def upload_config_files
             end
             assignment.save!
             redirect_to edit_assignment_path(assignment.id)
-          rescue Exception => e
+          rescue StandardError, SyntaxError => e
```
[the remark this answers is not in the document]
mishaschwartz: Oh I see.. don't raise SyntaxError manually. SyntaxErrors are raised when *ruby* code cannot be parsed by the interpreter.

If you're going to raise a custom error either subclass StandardError (if a new custom error class is necessary) or just raise a generic (runtime) error (`raise "my error message goes here"`)

app/controllers/assignments_controller.rb
```diff
@@ -554,25 +554,164 @@ def download_sample_starter_files
...
+                                   form_type: form_type,
+                                   upload_type: upload_type))
+      return
+    end
+    assignment.repository_folder = assignment.short_identifier
+    assignment
+  end
+
+  def read_yaml_file(file)
+    begin
```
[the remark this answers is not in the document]
mishaschwartz: please fix this one

PR #5518, opened 2021-09-23
Add counts to marking state filters

### LINKED ISSUE #5389: Add counts to "marking status" dropdown
On the submissions page: show how many groupings fall into each "marking status" category in the dropdown menu

pretendWhale:
## Motivation and Context

Resolves #5389  by adding counts to the marking state filter on the submission and assignment summary tables.

## Your Changes

**Description**:
Updates submission and assignment summary table components, as well as table helper functions

## Testing

Manual testing of both tables in the web interface for multiple assignments with different marking states. Tested with multiple filters (text filter, selection filter), as well as more than one filter at once.

## Checklist

### DIFF
--- a/app/assets/javascripts/Components/assignment_summary_table.jsx
+++ b/app/assets/javascripts/Components/assignment_summary_table.jsx
@@ -7,12 +7,21 @@ import ReactTable from "react-table";
 class AssignmentSummaryTable extends React.Component {
   constructor() {
     super();
+    const markingStates = {
+      not_collected: 0,
+      incomplete: 0,
+      complete: 0,
+      released: 0,
+      remark: 0,
+      before_due_date: 0,
+    };
     this.state = {
       data: [],
       criteriaColumns: [],
       loading: true,
       num_assigned: 0,
       num_marked: 0,
+      marking_states: markingStates,
     };
   }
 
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/Components/assignment_summary_table.jsx
```diff
@@ -29,96 +38,127 @@ class AssignmentSummaryTable extends React.Component {
...
+      remark: 0,
+      before_due_date: 0,
+    };
+    data.forEach(row => {
+      markingStates[row["marking_state"]] += 1;
+    });
+    return markingStates;
+  }
+
+  getWrappedInstance() {
```
mishaschwartz, line 74: Instead of redefining this here, you can just change [line 54](https://github.com/MarkUsProject/Markus/pull/5518/files#diff-e4273cac2d46f2e394362d102b79e399bb43095a55e25ea151767b064ceb15c2R54) to:

```
    const summaryTable = this.wrappedInstance;
```

app/assets/javascripts/Components/assignment_summary_table.jsx
```diff
@@ -29,96 +38,127 @@ class AssignmentSummaryTable extends React.Component {
...
-          // Check group name
-          if (row._original.group_name.includes(filter.value)) {
-            return true;
+  onFilteredChange = () => {
+    const summaryTable = this.getWrappedInstance();
+    const markingStates = this.getMarkingStates(summaryTable.state.sortedData);
+    this.setState({marking_states: markingStates});
+  };
+
+  getMarkingStates(data) {
```
mishaschwartz, line 59: Instead of defining this twice (here and in the submissions table), move this to the helper file and import it in both files

config/locales/views/submissions/en.yml
```diff
@@ -61,6 +61,13 @@ en:
       not_collected: Not Collected
       released: Released
       remark_requested: Remark Requested
+    state_filter:
```
mishaschwartz, line 64: Instead of defining new keys here that are essentially the same text as the `state:` keys (above), just use those ones and just append the count to the end of the string. That way we don't have to maintain two sets of translations that are essentially the same.

So instead of calling:

```
I18n.t("submissions.state_filter.in_progress", {count: marking_states["incomplete"]})
```

do something like:

```
I18n.t("submissions.state.in_progress") + ` (${marking_states["incomplete"]})`
```

mishaschwartz (changes requested):

config/locales/views/submissions/en.yml
pretendWhale: Makes sense! Done

mishaschwartz (approved):

app/assets/javascripts/Components/assignment_summary_table.jsx
```diff
@@ -1,18 +1,27 @@
...
 import {render} from "react-dom";
-import {markingStateColumn} from "./Helpers/table_helpers";
+import {markingStateColumn, getMarkingStates} from "./Helpers/table_helpers";
 
 import ReactTable from "react-table";
 
 class AssignmentSummaryTable extends React.Component {
   constructor() {
     super();
+    const markingStates = {
```
david-yz-liu, line 10: Eh, instead of duplicating the keys here, why don't you call `getMarkingStates` on an empty list? (Similar for the other constructor below.

david-yz-liu (commented):
@pretendWhale thanks, looks nice! In terms of the UI issue we discussed on Monday, you can go ahead and make the change that Sophia proposed (hiding the counts for the other marking states when filtering by a marking state).

app/assets/javascripts/Components/assignment_summary_table.jsx
```diff
@@ -29,96 +31,109 @@ class AssignmentSummaryTable extends React.Component {
...
-        } else {
-          return <span>{row.original.group_name}</span>;
-        }
-      },
-      filterMethod: (filter, row) => {
-        if (filter.value) {
-          // Check group name
-          if (row._original.group_name.includes(filter.value)) {
-            return true;
+  onFilteredChange = () => {
```
david-yz-liu, line 46: This callback function can actually take 2 arguments (https://github.com/tannerlinsley/react-table/tree/v6#fully-controlled-component). This should allow you to set `markingStateFilter` without searching through `summaryTable.state.filtered`.

Also, if the marking state column is what's changed, we don't need to recompute `markingStates`. So I think it's possible for this function to either `setState` for `marking_states` or `markingStateFilter` but not both.

app/assets/javascripts/Components/assignment_summary_table.jsx
```diff
@@ -1,18 +1,20 @@
...
   constructor() {
     super();
+    const markingStates = getMarkingStates([]);
     this.state = {
       data: [],
       criteriaColumns: [],
       loading: true,
       num_assigned: 0,
       num_marked: 0,
+      marking_states: markingStates,
```
david-yz-liu, line 17: Initialize the state for `markingStateFilter` here as well (to `"all"`). This, plus the `onFilteredChange` comment I made below, should allow you to ensure we never pass `undefined` to `markingStateColumn`, so you can remove that case from the code.

app/assets/javascripts/Components/assignment_summary_table.jsx
pretendWhale: Thanks for pointing that out!

And you are correct about splitting this to only ever `setState` for one of the two options

david-yz-liu (approved):

PR #5525, opened 2021-09-29
Add course table

pretendWhale:
## Motivation and Context

Adds Courses table to database

## Your Changes

**Description**:
Adds Courses table. Also adds course controller and policy. Primarily Boilerplate currently

## Testing

Created new tables via the newly added views.

Note that much of this functionality will need to change as we add the additional features required to support multiple courses.

## Checklist


app/models/course.rb
```diff
@@ -0,0 +1,4 @@
+# Class describing a course
+class Course < ApplicationRecord
+  validates_presence_of :name, presence: true, uniqueness: true
```
mishaschwartz, line 3: It would be better to have two distinct validations here instead, `validates_presence_of` and `validates_uniqueness_of`. 

You also don't need to specify `presence: true` for the validates_presence_of validator

app/views/courses/_form.html.erb
```diff
@@ -0,0 +1,27 @@
+<%= form_with(model: course) do |form| %>
+  <% if course.errors.any? %>
```
mishaschwartz, line 2: I know this is boilerplate but we are 100% not going to be keeping this so I would just remove this entire file.

app/views/courses/edit.html.erb
```diff
@@ -0,0 +1,6 @@
+<h1>Editing Course</h1>
```
mishaschwartz, line 1: I know this is boilerplate but we are 100% not going to be keeping this so I would just remove this entire file.

app/views/courses/index.html.erb
```diff
@@ -0,0 +1,25 @@
+<p id="notice"><%= notice %></p>
```
mishaschwartz, line 1: Don't delete this file but just remove the content for now. I don't want to keep code that we're definitely not going to use.

app/views/courses/new.html.erb
```diff
@@ -0,0 +1,5 @@
+<h1>New Course</h1>
```
mishaschwartz, line 1: I know this is boilerplate but we are 100% not going to be keeping this so I would just remove this entire file.

app/controllers/courses_controller.rb
```diff
@@ -0,0 +1,50 @@
+# Manages actions relating to editing and modifying
+# courses.
+class CoursesController < ApplicationController
```
mishaschwartz, line 3: We only actually need the index and show methods here and we can delete the rest.

app/policies/course_policy.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def show?
+    true
+  end
+
+  def index?
+    true
+  end
+
+  def manage?
```
mishaschwartz, line 13: This policy inherits from ApplicationPolicy which already has the manage? method defined (with the same body) so we can remove this method here.

app/views/courses/show.html.erb
```diff
@@ -0,0 +1,14 @@
+<p id="notice"><%= notice %></p>
```
mishaschwartz, line 1: The show route is going to end up rendering the current landing page (MainController.index) probably so we won't actually need a separate show view. I would remove this file as well.

config/routes.rb
```diff
@@ -449,6 +449,8 @@
     end
   end
 
+  resources :courses
```
mishaschwartz, line 452: To make sure that Rails handles the fact that we're not using all the builtin routes we should add 

```
only: [:show, :index]
```

app/models/course.rb
mishaschwartz: Let's also add a validation to restrict the name of the course. We can use the same validation that we use on the assessment short_identifier: 

https://github.com/MarkUsProject/Markus/blob/0519dec2aa024b880db86ebbdf690cae0f324d65/app/models/assessment.rb#L23-L25

spec/models/course_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
```
mishaschwartz, line 1: You can remove this line

spec/models/course_spec.rb
```diff
@@ -0,0 +1,5 @@
...
+
+RSpec.describe Course, type: :model do
```
mishaschwartz, line 3: Change this to:

```
describe Course do
```

like the other model specs

spec/models/course_spec.rb
```diff
@@ -0,0 +1,5 @@
...
+  pending "add some examples to (or delete) #{__FILE__}"
```
mishaschwartz, line 4: You can add the validation tests here

mishaschwartz (changes requested):
Add specs for the following files:
- courses_controller.rb
- course_policy.rb

Add tests in the routes_spec.rb file for the new routes

spec/controllers/courses_controller_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    allow(controller).to receive(:current_user).and_return(admin)
+  end
+
+  context 'accessing course pages' do
+    it 'responds with success on index' do
+      get :index
+      expect(response.status).to eq(200)
+    end
+    # TODO: enable this once we have a show action.
+    xit 'responds with success on show' do
```
mishaschwartz, line 18: can we enable this now?

spec/controllers/courses_controller_spec.rb
```diff
@@ -0,0 +1,23 @@
+describe CoursesController do
+  let(:admin) { create :admin }
+  let(:course) { create :course }
+
+  before :each do
```
mishaschwartz, line 5: This isn't necessary, just use `get_as admin` instead of `get` to make requests

app/models/course.rb
```diff
@@ -0,0 +1,13 @@
...
+class Course < ApplicationRecord
+  attribute :is_hidden, :boolean, default: true
+  validates_presence_of :name
+  validates_uniqueness_of :name
+  validates :name, format: { with: /\A[a-zA-Z0-9\-_]+\z/,
+                             message: 'name must only contain alphanumeric, hyphen, or '\
+                                      'underscore' }
+  validates_presence_of :is_hidden
+  validates :display_name, format: { with: /\A[a-zA-Z0-9\-_ ]+\z/,
+                                     message: 'name must only contain alphanumeric, hyphen, '\
```
mishaschwartz, line 11: should be `'display_name must only ...'`

app/models/course.rb
```diff
@@ -0,0 +1,13 @@
+# Class describing a course
+class Course < ApplicationRecord
+  attribute :is_hidden, :boolean, default: true
```
mishaschwartz, line 3: This isn't necessary. This just duplicates the column definition in the courses table and will just cause problems if we ever update one without changing the other.

spec/policies/course_policy_spec.rb
```diff
@@ -0,0 +1,13 @@
+describe CoursePolicy do
+  let(:context) { { user: user } }
+  let(:user) { create :student }
+  describe_rule :manage? do
```
mishaschwartz, line 4: `manage?` is no longer defined for this policy (it's inherited) so you can remove this test here

spec/routing/routes_spec.rb
```diff
@@ -1267,3 +1267,32 @@
...
 end
 # end course summary route tests
+
+# Start course route tests
+describe 'courses' do
+  let(:path) { '/courses' }
+  let(:ctrl) { 'courses' }
+  let(:course) { create :course }
+  it 'routes GET course properly' do
+    expect(get: path << '/' << course.id.to_s)
```
mishaschwartz, line 1277: Use string interpolation fo these strings. ex: `"#{path}/#{course.id}"`

mishaschwartz (changes requested):
Some super small changes and then this is good to go

PR #5527, opened 2021-09-30
Disable run test button

chen-jonathan:
image

## Motivation and Context

This PR resolves the issues with disabling the "Run Tests" button. Now when the marks are released, the "Run Tests" button in the grading view is disabled.

## Your Changes

**Description**:
Added a new method `disableRunTestButton() `in the `LeftPane` React Component. This method contains the logic for disabling/enabling the "Run Tests" button. 

## Testing

I tested the code changes through the UI. 
When the marks are not released yet, the "Run tests" button is enabled and can be clicked. 
image
However, when marks are released, for the same submission, the "Run tests" button is now disabled
image
I did similar steps for some of the other autotesting assignments as well.
## Checklist


david-yz-liu (commented):
@chen-jonathan your approach works, but there is actually a deeper problem. The prop `can_run_tests` should already be set to `true`/`false` depending on whether the result is released. So what you should do is:

1. Use the dev tools to check the data being sent from the server, and what value for `can_run_tests` is being used.
2. On the server side, check how the (presumably erroneous) value of `can_run_tests` is being set.
3. Consult the documentation for various methods of the `action-policy` gem here: https://actionpolicy.evilmartians.io/#/behaviour.

david-yz-liu (approved):

PR #5547, opened 2021-10-04
repo: Restrict permissions for assignment is_hidden and anonymize_groups attributes

david-yz-liu:
## Motivation and Context

1. Follow up on #5499: when groups should be anonymized to graders, graders should not have access to the group repositories, since the commit history may reveal information about the group members. (Grader still have the ability to download the collected submission files.)
2. When reviewing the assignment attributes, I found that `is_hidden` should also restrict repo permissions: when an assignment is hidden, students shouldn't be able to access their group's repository, even if they are assigned a group. (This is very unlikely to come up in practice, since students won't be able to see their group's repository URL until the assignment is unhidden. However, I still think it's worth making this change to the permission calculation.)

## Your Changes

**Description**: Modified the `Assignment` repository permission callback function to handle cases for when the `anonymize_groups` and `is_hidden` attributes change. Also made the logical changes in `Repository.get_all_permissions` to restrict students from accessing repos of hidden assignments, and graders from access repos of assignments with anonymized grading.

## Testing

Added a new test case, and verified that the permission file reflects the above changes through manual updates in the web UI.

## Checklist

Add a note to the graders Wiki page to say that enabling the "anonymize groups" option prevents TAs from accessing student repositories.


app/lib/repository.rb
```diff
@@ -254,7 +254,9 @@ def self.get_all_permissions
       end
       # NOTE: this will allow graders to access the files in the entire repository
```
mishaschwartz, line 255: Unrelated to this PR but we should deal with this note eventually. Especially in this context where one assignment could have anonymization enabled and the other has it disabled.

app/models/assignment.rb
```diff
@@ -904,7 +906,7 @@ def each_group_repo
   def self.get_repo_auth_records
     records = Assignment.joins(:assignment_properties)
                         .includes(groupings: [:group, { accepted_student_memberships: :user }])
-                        .where(assignment_properties: { vcs_submit: true })
+                        .where(is_hidden: false, assignment_properties: { vcs_submit: true })
```
mishaschwartz, line 909: Great catch! The only problem is that we now hide assessments based on section membership as well. Implementing that is going to be tricky and I think it should actually be handled elsewhere. I can think about it and commit a change

app/models/assignment.rb
```diff
@@ -1285,4 +1287,17 @@ def create_associations
...
   end
+
+  # Update the repository permissions file if one of the following attributes was changed after a save:
+  # - vcs_submit
+  # - is_hidden
+  # - anonymize_groups
+  def update_repo_permissions
+    return unless
+      saved_change_to_vcs_submit? ||
+        saved_change_to_is_hidden? ||
```
mishaschwartz, line 1298: We should also handle a change to the section based is_hidden value

mishaschwartz (changes requested):

app/models/assignment.rb
```diff
@@ -1310,15 +1314,37 @@ def create_associations
...
+  # - only_required_files
+  # - is_hidden or section-specific is_hidden
+  # - any assignment files
+  def update_repo_required_files
+    return unless Settings.repository.type == 'git'
+    return unless
+      saved_change_to_only_required_files? ||
+        assignment_files.any?(&:saved_changes?) ||
+        visibility_changed? ||
+        @prev_assessment_section_property_ids != self.assessment_section_properties.ids
```
mishaschwartz, line 1339: Shouldn't this be:

```
@prev_assignment_file_ids != self.reload.assignment_files.ids
```

david-yz-liu: Nice catch, thank you!

PR #5553, opened 2021-10-13
Prevented page from refreshing when running automated tests

MShwed:
## Motivation and Context

Currently when running automated tests the page refreshes. The desired behaviour in this case would be that a flash message is still displayed but the page does not refresh.

This can be seen with the following video demonstrating the behaviour: https://share.getcloudapp.com/OAuGw447

This occurs for both admins running autotests and students.

## Your Changes

**Description**:

The related fixes for this bug could be seen on the backend and front-end. 

The front-end fix involved adding the form attribute to the html element `remote-data: true`. However on the backend we also need to return data from this ajax request which has also been added here. This will update the js so the flash message will be able to be rendered on the front-end.

The changes can be seen in this video: https://share.getcloudapp.com/Wnuk1qDl

Further this includes the same set of changes for student run autotests. 

## Testing

**Administrator Perspective**
Manual testing was done through the interface as can be seen by the before and after videos of the change

Before:  https://share.getcloudapp.com/OAuGw447
After: https://share.getcloudapp.com/Wnuk1qDl

**Student Perspective**
Before: https://share.getcloudapp.com/geuplW61
After: https://share.getcloudapp.com/xQujpWOD

## Checklist

cc @david-yz-liu

### DIFF
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -273,9 +273,6 @@ class ResultsController < ApplicationController
                                                 [submission.grouping.group_id])
     session[:job_id] = @current_job.job_id
     flash_message(:success, I18n.t('automated_tests.tests_running'))
-    respond_to do |format|
-      format.js
-    end
   end
 
   def stop_test
--- a/app/controllers/automated_tests_controller.rb
+++ b/app/controllers/automated_tests_controller.rb
@@ -67,12 +67,10 @@ class AutomatedTestsController < ApplicationController
                                                   [grouping.group_id],
                                                   collected: false)
       session[:job_id] = @current_job.job_id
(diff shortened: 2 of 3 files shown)


app/controllers/results_controller.rb
```diff
@@ -272,8 +272,10 @@ def run_tests
                                                 submission.assignment.id,
                                                 [submission.grouping.group_id])
     session[:job_id] = @current_job.job_id
-    flash_message(:notice, I18n.t('automated_tests.tests_running'))
-    redirect_back(fallback_location: root_path)
+    flash_message(:success, I18n.t('automated_tests.tests_running'))
+    respond_to do |format|
```
david-yz-liu, line 276: You can remove this block---it's not wrong, but our code base doesn't generally restrict the request types unless we need to distinguish between multiple format types.

PR #5560, opened 2021-10-15
Added Upload/Download Criteria for Assignment Copy

anoverga17:
## Motivation and Context

Part of creating a new feature to download and upload assignment settings. This pull request adds the criteria file needed to upload/download tags for the copy assignment feature. In order to accommodate this, upload functionality from the `upload` function in the criteria controller was moved to the parent application controller so that the criteria upload could be accessed by the assignment controller. 

## Your Changes

**Description**:
- Added assignment/peer review assignment criteria file when creating zip in `download_config_files`
- `upload_config_files` can now process a `criteria.yml` and `peer-review-config-files/criteria.yml` file in order to copy over assignment and peer review assignment tags
- Moved main upload functionality of `upload` and `reset_results_total_mark` in the criteria controller to the parent application controller for so that assignment controller has access to them. 

## Testing

These changes have primarily been tested by doing manual testing through the web interface. Various sample sssignments with various criteria were downloaded (with their contents examined to see if downloads were working properly). These assignments were then later reuploaded and the database was checked to ensure the criteria attributes remained unchanged. 

rspec tests will be added in a later pull request. 

## Checklist

### DIFF
--- a/app/models/tag.rb
+++ b/app/models/tag.rb
@@ -36,7 +36,7 @@ class Tag < ApplicationRecord
     admins = Hash[Admin.pluck(:user_name, :id)]
     begin
       tag_data = data.map do |row|
-        row.symbolize_keys!
+        row = row.symbolize_keys
         name, description, user_id = row[:name], row[:description], admins[row[:user]]
         if name.nil? || name.strip.blank? || user_id.nil?
           raise ArgumentError, I18n.t('invalid_tag_data', item: row)


app/controllers/application_controller.rb
```diff
@@ -133,4 +133,60 @@ def implicit_authorization_target
   def identification_params
     params.permit(:id)
   end
+
+  def upload_criteria(assignment, data)
```
mishaschwartz, line 137: I like that you've factored this out of the criteria controller so that we can use it in multiple places. However I don't think that this belongs in the ApplicationController. The ApplicationController is for methods that affect all application routes, this one is specific to criteria. 

I would move this function to a new helper (app/helpers/criteria_helper.rb)

mishaschwartz: The other thing I'd like you to think about is the fact that you now have a transaction in a transaction. If an error is raised during the criteria creation do you want everything to roll back or just the criteria creation step?

Also, this method displays flash messages which is fine but does not follow the pattern we established earlier of raising errors and reporting the errors as flash messages _at the end_.

Instead of just moving this function, think about how you can tweak it a bit so that way you're handling transactions and error messages stays consistent.

PR #5561, opened 2021-10-15
Test Results Endpoint and Summary Download Modal

MShwed:
## Motivation and Context

This PR adds support for downloading the most recent test results that have been run associated with assignment submissions in JSON format. 

The goal will be to also add support for downloading the test results in CSV format.

## Your Changes

**Description**:

Currently this PR adds an endpoint to download a file representing the most recently run test results in JSON format. 
This endpoint currently is restricted to administrators. 

This change adds a new modal which has been added to the `assignment_summary_table.jsx`.

This change can be previewed here: https://share.getcloudapp.com/bLudD97r

Screen Shot 2021-10-15 at 4 41 38 PM

## Testing

Currently testing has been done manually through the web interface

## Checklist

cc @david-yz-liu


app/models/assignment.rb
```diff
@@ -600,6 +600,20 @@ def summary_json(user)
       numMarked: self.get_num_marked(user.admin? ? nil : user.id) }
   end
 
+  # Generate a JSON summary of the test results associated with an assignment
+  def summary_test_result_json(user)
+    return {} unless user.admin? || user.ta?
```
david-yz-liu, line 605: You can remove this, and the `user` parameter. We do authorization at the controller level, and the default rule, `AssignmentPolicy#manage?` will apply and block requests from students (and, in fact, from graders that don't have permissions to manage assignments).

app/models/assignment.rb
```diff
@@ -600,6 +600,20 @@ def summary_json(user)
...
+    {
```
david-yz-liu, line 606: I don't think the `submission_test_results` key part is necessary: I'd like to return just the list (produced by `map`) instead.

app/models/assignment.rb
```diff
@@ -600,6 +600,20 @@ def summary_json(user)
...
+      submission_test_results: self.submissions.map do |submission|
```
david-yz-liu, line 607: Use `self.current_submissions_used` instead of `self.submissions`. A group can have more than one collected submission, but generally we only want to most recent one, which is what `current_submissions_used` gives us.

app/models/assignment.rb
```diff
@@ -600,6 +600,20 @@ def summary_json(user)
...
+        {
+          group_name: submission.grouping.group.group_name,
+          submission_date: submission.created_at,
```
david-yz-liu, line 610: You can remove this, I don't think it's necessary

app/models/assignment.rb
```diff
@@ -600,6 +600,20 @@ def summary_json(user)
...
+          test_results: [*submission.test_group_results.order('created_at').last]
```
david-yz-liu, line 611: This is a good start, but needs some refinement.

1. I want to see the `TestResult` objects here, in addition to the fields from `TestGroupResult`.
2. An assignment can have more than one `TestGroup`, so what you really want is the last `TestGroupResult` *for each `TestGroup`*. That's a bit trickier to do. One possibility is using fancier SQL to get all of the "most recent test group results" for an assignment (across all submissions), and then divide those up by submission id. We do something similar here: https://github.com/MarkUsProject/Markus/blob/438b7c36d69c9cdd293aaea77c048fe601e3962b/app/models/assignment.rb#L793-L802.
3. Right now this code has an [N+1 query problem](https://guides.rubyonrails.org/active_record_querying.html#eager-loading-associations), since the test group results are queried once per submission. If you do what I suggested in point 2 above, that'll fix the issue for `TestGroupResult`, but not `TestResult`. For the latter, one option is to load all of the relevant `TestResult` objects and then group them by `test_group_result_id`.
4. We don't need every field in these objects. You can use the `pluck_to_hash` query method to grab only certain fields from the database. 

    - For `TestGroupResult`, we need the `name` of the associated `TestGroup`, `marks_earned`, `marks_total`, `extra_info`, and `error_type`.
    - For `TestResult`, we need `name`, `status`, `marks_earned`, `marks_total`, and `output`.

app/controllers/assignments_controller.rb
```diff
@@ -269,6 +269,20 @@ def summary
     end
   end
 
+  def download_test_results
+    @assignment = Assignment.find(params[:id])
+    respond_to do |format|
+      format.json do
+        data = @assignment.summary_test_result_json(@current_user).to_json
```
david-yz-liu, line 276: Move the `.to_json` call into the assignment method, since the method name suggests it returns JSON.

app/models/assignment.rb
```diff
@@ -600,6 +600,59 @@ def summary_json(user)
...
+                      'test_results.output',
+                      'test_results.marks_earned as test_result_marks_earned',
+                      'test_results.marks_total as test_result_marks_total',
+                      'test_results.id as id',
+                      :status, :marks_earned, :marks_total, :extra_info, :error_type)
+  end
+
+  # Generate a JSON summary of the most recent test results associated with an assignment.
+  def summary_test_result_json
+    self.summary_test_results.to_json
```
david-yz-liu, line 629: Okay, so right now the JSON structure is a flat list, but looking at this more carefully, I'd like to see some grouping of the results, first by grouping id, and then by test group name. So the overall JSON should look like:

```
{
  <grouping_id>:
    {
      <test group name>: [<test result 1>, <test result 2>, ...],
      <test group name>: [<test result 1>, <test result 2>, ...],
      ...
    },
  ...
}
```
Note that you can use Ruby's [`Enumerable#group_by`](https://ruby-doc.org/core-3.0.2/Enumerable.html#method-i-group_by) to do a lot of the grouping for you, just be careful when handling the nested structure.

app/models/assignment.rb
```diff
@@ -600,6 +600,59 @@ def summary_json(user)
...
+  end
+
+  # Generate a CSV summary of the most recent test results associated with an assignment.
+  def summary_test_result_csv
+    summary_test_result = self.summary_test_results
+    CSV.generate do |csv|
+      submissions = {}
+      summary_test_result.map do |test_result|
```
david-yz-liu, line 637: This approach is pretty good, but has two issues:

1. It assumes that the test results are well-ordered, so that results for each grouping will be added to `submissions` in the same order.
2. It doesn't handle the case when a grouping is missing some results---logically this should create a "gap" in a row, but instead the row's other elements will be shifted over.

So I'll suggest a more explicit creation of a table, by first querying all distinct test result names from `summary_test_result`, then grouping the results by grouping id (similar to what I suggested in the JSON method), and for each grouping constructing the correct row by cross-referencing the results against the collection of test result names. I'll leave the details of that part for you to work out.

app/assets/javascripts/Components/Modals/download_test_results_modal.jsx
```diff
@@ -0,0 +1,68 @@
+import React from "react";
+import Modal from "react-modal";
+
+class DownloadTestResultsModal extends React.Component {
+  constructor(props) {
```
david-yz-liu, line 5: You can remove the constructor if you aren't doing anything other than calling the super class one.

spec/factories/assignments.rb
```diff
@@ -42,6 +42,24 @@
...
+      3.times { create(:flexible_criterion, assignment: a) }
+      3.times { create(:grouping_with_inviter_and_submission, assignment: a) }
+      3.times { create(:test_group, assignment: a) }
+
+      a.groupings.each do |grouping|
+        a.test_groups.each do |test_group|
+          5.times do
+            test_run = create(:test_run, grouping: grouping)
+            test_group_result = create(:test_group_result, test_run: test_run, test_group: test_group)
+            create(:test_result, id: test_run.id, test_group_result: test_group_result)
```
david-yz-liu, line 56: Similar to one of my above comments, you shouldn't set `id: test_run.id`.

config/locales/defaults/download_upload/en.yml
```diff
@@ -5,6 +5,8 @@ en:
   download_csv: Download in CSV Format
   download_errors:
     unrecognized_format: Could not recognize %{format} format to download with
+  download_json: Download in JSON Format
+  download_tests: Download Test Results
```
david-yz-liu, line 9: Instead of creating a new key, re-use the `download_the` key (see below), together with the model name for test results (see `config/locales/models/test_results`).

spec/controllers/assignments_controller_spec.rb
```diff
@@ -167,6 +167,75 @@
     end
   end
 
+  context 'download the most recent test results as JSON' do
+    let(:user) { create(:admin) }
+    let(:assignment) { create(:assignment_with_criteria_and_test_results) }
+
+    it 'responds with the appropriate status' do
+      get_as user, :download_test_results, params: { id: assignment.id }, format: 'json'
+      expect(response.status).to eq(200)
```
david-yz-liu, line 176: Use `expect(response).to have_http_status :success`

spec/controllers/assignments_controller_spec.rb
```diff
@@ -167,6 +167,75 @@
...
+      filename = response.header['Content-Disposition'].split[1].split('"').second
+      expect(filename).to eq("#{assignment.short_identifier}_test_results.json")
+    end
+
+    it 'returns application/json type' do
+      get_as user, :download_test_results, params: { id: assignment.id }, format: 'json'
+      expect(response.media_type).to eq 'application/json'
+    end
+
+    it 'returns the most recent test results' do
```
david-yz-liu, line 201: For this test, you should still make the request (`get_as ...`) and then access `response.parsed_body` to check the data in the body. Then, you should add tests in the Assignment model's test file for the `summary_test_result_json` method.

Similar comment for the "CSV" version below.

app/models/assignment.rb
```diff
@@ -600,6 +600,76 @@ def summary_json(user)
...
+    self.summary_test_results.group_by(&:group_name).transform_values do |grouping|
+      grouping.group_by(&:name)
+    end.to_json
+  end
+
+  # Generate a CSV summary of the most recent test results associated with an assignment.
+  def summary_test_result_csv
+    summary_test_results = self.summary_test_results.as_json
+
+    headers = Set.new
```
david-yz-liu, line 640: Hmm this is a bit tricky. Let's use `SortedSet` for `headers`---this will make the test names be sorted in alphabetical order. (In particular, this will guarantee that the test results appear grouped by their test group.)

app/models/assignment.rb
```diff
@@ -600,6 +600,76 @@ def summary_json(user)
...
+    groups = Set.new
+    summary_test_results.each do |test_result|
+      groups << test_result['group_name']
+      headers << "#{test_result['group_name']}:#{test_result['test_result_name']}"
```
david-yz-liu, line 644: The keys may need to change based on my comment about renaming above.

app/models/assignment.rb
```diff
@@ -600,6 +600,76 @@ def summary_json(user)
...
+    end
+
+    CSV.generate do |csv|
+      csv << headers
```
david-yz-liu, line 648: Hmmm okay, so there's one last data format change I need you to make. The *first column should be the group name* (and the remaining columns should be the actual test results). So for the header row, the first entry should be `nil`, and then followed by the `headers`. For all subsequent rows, the first entry should be the group name, and then the test results for that specific group.

app/models/assignment.rb
```diff
@@ -600,6 +600,76 @@ def summary_json(user)
...
+      groups.each do |group|
```
david-yz-liu, line 649: The overall approach in this loop is correct but inefficient (looping over all groups, and then for each of those, all test results). I'd recommend a different approach: you can pre-process the test results by grouping by group name (similar to the JSON summary), and then for each group create a hash mapping test name to the status, similar to what you're doing in L654-658, except on a per-group basis.

app/models/assignment.rb
```diff
@@ -605,6 +605,65 @@ def summary_json(user)
...
+    self.summary_test_results.group_by(&:group_name).transform_values do |grouping|
+      grouping.group_by(&:name)
+    end.to_json
+  end
+
+  # Generate a CSV summary of the most recent test results associated with an assignment.
+  def summary_test_result_csv
+    summary_test_results = self.summary_test_results
+    results = summary_test_results.map do |result|
+      ["#{result['group_name']}:#{result['test_result_name']}", result['status']]
```
david-yz-liu, line 643: You're still confusing `group_name` with `test_groups.name`.

You need `test_groups.name` here, since that's the actual name of the test group. 

`group_name` is different, it identifies the student group. Overall you need a *nested hash* so that the outer hash keys are the group names, and the inner hash keys are what you had before (the test group name and then the test name).

This way, you can take then do something like `results[some_group_name][header]` to actually get the status.

app/models/assignment.rb
```diff
@@ -605,6 +605,65 @@ def summary_json(user)
...
+    end.to_h
+
+    group_names = summary_test_results.group_by(&:group_name).keys
+
+    CSV.generate do |csv|
+      headers = SortedSet.new(results.keys)
```
david-yz-liu, line 649: If you do what I said above, you won't be able to directly extract the keys from `results`. Use a separate loop over `summary_test_results` to get the headers, using the approach you had before.

app/models/assignment.rb
```diff
@@ -605,6 +605,65 @@ def summary_json(user)
...
+      csv << [nil, *headers]
+      group_names.each do |name|
+        row = [name]
+
+        headers.each do |header|
+          _, test_result_name = header.split(':')
```
david-yz-liu, line 655: Don't split on colons, since there might actually be colons in the test name (this is configurable for some testers). But you won't need to do this if you change what I said above, since the `headers` here should exactly match the keys in the "inner" hashes in `results`.

david-yz-liu (approved):

PR #5570, opened 2021-10-28
Required files should be per-course

mishaschwartz:
## Motivation and Context

Required files should be added to each repo for the course that they belong to only.

NOTE: `require 'csv'` was added to git_repository.rb because if an Assignment isn't accessed before the `update_permissions_file` then the csv module isn't loaded and that method will fail (for example, when the git_repository_spec.rb spec is run in isolation). Since we shouldn't assume that Assignment will always be loaded first we should explicitly require csv here as well.

## Your Changes

**Description**:

- add course argument to repository creation so that course specific required files can be accessed

## Checklist

### DIFF
--- a/app/lib/git_repository.rb
+++ b/app/lib/git_repository.rb
@@ -1,3 +1,5 @@
+require 'csv'
+
 # Implements AbstractRepository for Git repositories
 # It implements the following paradigm:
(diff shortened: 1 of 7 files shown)


pretendWhale (approved):
Looks good.

PR #5574, opened 2021-10-31
scanned-exam: Enable non-letter paper sizes in exam templates

david-yz-liu:
## Motivation and Context

Previously, exam templates with later page sizes than letter paper (e.g., legal size) were being cropped MarkUs used them to generated PDFs with QR codes.

## Your Changes

**Description**: Changed the PDF generation stage to replicate the page size of the original exam template, on a page-by-page basis. This change supports test papers of varying sizes, and templates where the pages have different sizes.

## Testing

Manual testing by creating an exam template that uses legal paper size, and running the "Generate" and "Split" actions with this template.

## Checklist


app/jobs/generate_job.rb
```diff
@@ -31,15 +31,15 @@ def perform(exam_template, num_copies, start)
     generated_pdf = CombinePDF.new
     (start..start + num_copies - 1).each do |exam_num|
       m_logger.log("Now generating: #{exam_num}")
-      pdf = Prawn::Document.new(margin: 15) do
-        exam_template.num_pages.times do |page_num|
+      pdf = Prawn::Document.new(margin: 15, skip_page_creation: true) do
```
mishaschwartz, line 34: This change makes sense but is there a reason why we're not persisting some of the other layout settings while we're at it? I feel like if we don't then we'll run into other issues in the future where the template files are not copied correctly:

```ruby
pdf = Prawn::Document.new(skip_page_creation: true) do
    template_pdf.pages.each_with_index do |page, page_num|
        start_new_page(size: page.page_size[2..4], layout: page.layout, margins: page.margins.dup)
```

david-yz-liu: Nice catch! This is complicated a little by the fact that we're converting between two PDF formats (Prawn and CombinePDF). I couldn't easily find "margin" data from the CombinePDF pages, but we can investigate that later.

mishaschwartz: I see... that makes sense with the two different libraries. Do you think we should make an issue so we can look into this later. I'll approve this PR for now though.

mishaschwartz (approved):

PR #5592, opened 2021-11-02
Exam template automatic parsing

david-yz-liu:
## Motivation and Context

An instructor asked about the UI *automatic parsing* feature of scanned exams: why was there only one crop box but multiple fields that could be selected?

Prompted by this question, I reviewed this feature and found a few different bugs:

1. The code in `read_chars.py` needed a bit of updating, and contained one step (`straighten`) that failed to work correctly on examples I tried out.
2. The development docker image was missing a shared library required by `python-opencv`, `libGL.so.1`.
3. The automatic parsing form doesn't work correctly when there is more than one exam template, because it assumes unique `id`s for its form elements.

## Your Changes

**Description**:

1. Made a few updates to `read_chars.py` and left a TODO to investigate the problematic step, which I will later make an issue for.
2. Added the Ubuntu `libgl1` package to the Dockerfile.
3. Fixed the automatic parsing form so that it correctly handles multiple instances being rendered on the same page.
    - Also made a change to ensure that the exam templates appear in the order they were created.
4. To address the original issue, I simplified the form to only allow one field, which is either a user's `user_name` or `id_number`, and made the corresponding changes in the controller and `SplitPdfJob`, which use this field.

Finally, I added some tests for the automatic parsing workflow.

## Testing

Tested the automatic parsing process in the UI (processing scans that contain handwritten digits) and added a few additional test cases.

In the future we'd like to move `read_chars.py` and its dependencies into a separate Python module and review it more thoroughly, including its dependencies. I'll make an issue for this.

## Checklist

I'll add a new section for the "Scanned Exams" wiki page describing the automatic parsing functionality.


app/jobs/split_pdf_job.rb
```diff
@@ -298,9 +270,21 @@ def save_pages(exam_template, partial_exams, filename=nil, split_pdf_log=nil)
...
+  # the first is the result of attempting to parse alphabetic characters, and the second is the result of
+  # attempting to parse numeric digits.
+  def match_student(parsed, exam_template)
+    return nil if parsed.size < 2
+
+    case exam_template.cover_fields
+    when 'id_number'
+      Student.find_by(id_number: parsed[1])
+    when 'user_name'
+      Student.find_by('lower(user_name) = ?', parsed[0].downcase)
```
mishaschwartz, line 284: If we want to not rely on sql syntax and use activerecord as much as possible it would be better to do:

```ruby
Student.where(Student.arel_table[:user_name].matches(parsed[0])).first # matches is case insensitive
```

OR

```ruby
Student.where(Student.arel_table[:user_name].lower.eq(parsed[0].downcase)).first
```

mishaschwartz (approved):
Just one small query thing but otherwise looks good

PR #5601, opened 2021-11-05
Added Starter File Upload/Download to Copy Assignment Feature

anoverga17:
## Motivation and Context

This pull request adds the ability to download/upload information about the starter files and their groups for every assignment that is downloaded and uploaded as part of the copy assignment feature. 

## Your Changes

**Description**:
- Added ability to download assignment starter file group information (stored in a starter file config file)
- Moved starter file information for the copy assignment feature from assignment properties file to the starter file config file to group related data. 
- Added information about what is the default starter file group to starter file config file
- Added the ability to download starter files for each group
- Added the ability to upload starter file groups
- Added the ability to upload starter files
- Added download/upload starter files for peer review assignments
- Added tests for starter files upload/download

## Testing

These changes have primarily been tested by doing manual testing through the web interface. Various sample assignments with various starter files and starter file groups were downloaded (with their contents examined to see if downloads were working properly). These assignments were then later reuploaded and the database was checked to ensure the starter file settings and their contents remained unchanged.

In addition, rspec tests were added whereby the content of the starter files was checked after both downloading and uploading to ensure the correct data was being sent and received. 

## Checklist


app/controllers/assignments_controller.rb
```diff
@@ -582,6 +586,10 @@ def download_config_files
         yml_criteria = assignment.criteria.reduce({}) { |a, b| a.merge b.to_yml }
         f.write yml_criteria.to_yaml
       end
+      zipfile.get_output_stream(CONFIG_FILES[:annotations]) do |f|
+        f.write convert_to_yml(assignment.annotation_categories)
+      end
+      assignment.starter_file_config_to_zip(zipfile, 'starter-file-config-files', CONFIG_FILES[:starter_files])
```
mishaschwartz, line 592: `'starter-file-config-files'` should also be a string in CONFIG_FILES (avoid using [magic strings](https://en.wikipedia.org/wiki/Magic_string))

app/controllers/assignments_controller.rb
```diff
@@ -582,6 +586,10 @@ def download_config_files
         yml_criteria = assignment.criteria.reduce({}) { |a, b| a.merge b.to_yml }
         f.write yml_criteria.to_yaml
       end
+      zipfile.get_output_stream(CONFIG_FILES[:annotations]) do |f|
+        f.write convert_to_yml(assignment.annotation_categories)
```
mishaschwartz, line 590: I should have commented this in a previous PR. But if you're going to use the `convert_to_yml` function from the annotation_categories_helper.rb file, please change the name to make it clear that this isn't a generic function to convert some thing to yml. Or even better, put it in the AnnotationCategory model so you can call it like: `AnnotationCategory.convert_to_yml(assignment.annotation_categories)`

app/models/assignment.rb
```diff
@@ -1199,6 +1199,35 @@ def assignment_properties_config
...
+        directory_name: group_name,
+        name: starter_file_group.name,
+        use_rename: starter_file_group.use_rename,
+        entry_rename: starter_file_group.entry_rename
+      }
+      download_count += 1
+    end
+    starter_file_settings = {
+      starter_file_type: self.starter_file_type,
+      allow_starter_files_after_due: self.starter_files_after_due,
```
mishaschwartz, line 1224: Shouldn't these two settings already be set in the assignment properties file?

app/models/starter_file_group.rb
```diff
@@ -28,17 +28,25 @@ def zip_starter_file_files(user)
     zip_path = File.join('tmp', zip_name + '.zip')
     FileUtils.rm_rf zip_path
     Zip::File.open(zip_path, create: true) do |zip_file|
-      self.files_and_dirs.map do |file|
+      write_starter_files_to_zip(zip_file)
+    end
+    zip_path
+  end
+
+  def write_starter_files_to_zip(zip_file, dir = '')
```
mishaschwartz, line 36: use `nil` as a default here since you're checking explicitly for whether this variable is set or not. If you wanted to leave it as the empty string you can just use `File.join(dir, file)` because it will be correct whether `dir` is empty or not (`this//path/` is the same as `/this/path/`)

mishaschwartz (changes requested):

app/models/assignment.rb
anoverga17: I have decided to move those here just so that related settings are group together

app/controllers/assignments_controller.rb
```diff
@@ -658,7 +669,49 @@ def upload_config_files
 
   private
 
-  # Build the tag/criteria file specified by +hash_to_build+ found in +zip_file+
+  # Configures the starter files for an +assignment+ provided in the +zip_file+
+  def config_starter_files(assignment, zip_file)
+    assignment.is_peer_review? ? starter_config_file = :peer_review_starter_files : starter_config_file = :starter_files
```
mishaschwartz, line 674: slightly clearer syntax:

```ruby
starter_config_file = assignment.is_peer_review? ? :peer_review_starter_files : :starter_files
```

app/controllers/assignments_controller.rb
```diff
@@ -658,7 +669,49 @@ def upload_config_files
...
+    starter_file_settings = build_hash_from_zip(zip_file, starter_config_file).symbolize_keys
+    assignment.starter_file_type = starter_file_settings[:starter_file_type]
+    assignment.starter_files_after_due = starter_file_settings[:allow_starter_files_after_due]
```
mishaschwartz, line 677: Just to clarify, you're *not* setting this as part of the assignment properties and you *are* assigning this here or you're doing it in both places?

app/models/assignment.rb
```diff
@@ -1199,6 +1197,37 @@ def assignment_properties_config
...
   end
 
+  # Writes this assignment's starter file settings to the file located at +settings_filepath+ located in the
+  # +zip_file+. Also writes the starter files for this assignment in the same directory as +settings_filepath+.
+  def starter_file_config_to_zip(zip_file, settings_filepath)
+    default_starter_group = nil
+    download_count = 1
+    group_data = []
+    directory_path = File.dirname(settings_filepath)
+    self.starter_file_groups.find_each do |starter_file_group|
```
mishaschwartz, line 1207: There's an `each_with_index` method too if you don't want to have to keep track of the counter manually.

mishaschwartz (changes requested):
Looks good overall. You probably want to merge/rebase on the copy-assignments branch as well just to clean up the changes a bit

app/controllers/assignments_controller.rb
anoverga17: Yes that is correct. Stater file settings will be set from information in this file and will no longer be provided in the properties file

PR #5616, opened 2021-11-30
Copy assignments

### LINKED ISSUE #5385: Migrate assignments between instances

mishaschwartz:
## Motivation and Context

PR to merge the copy-assignments branch into master. This is the culmination of the [Download/Upload entire assignments project](https://github.com/orgs/MarkUsProject/projects/4). 

Resolves #5385 

## Your Changes

**Description**:

See all PRs in [the project](https://github.com/orgs/MarkUsProject/projects/4) for details.

## Testing

- [x] unittests added
- [ ] final UI review and tests

## Checklist

https://github.com/MarkUsProject/Markus/wiki/Pending-Changes#assignment-configuration


app/models/assignment.rb
```diff
@@ -1247,29 +1247,122 @@ def to_json(options = {})
     self.assignment_properties.attributes.merge(self.attributes).symbolize_keys.to_json(options)
   end
 
+  # Returns an assignment's relevant properties for uploading/downloading an assignment's configuration as a hash
+  def assignment_properties_config
```
mishaschwartz, line 1251: I think this method is much more complex than it needs to be and is going to be a huge pain to maintain. Every time we change an attribute for an assignment object, we are going to have to change this method as well and I really would like to avoid that.

Here are a few suggestions:

- use `self.attributes` instead of repeating `self.send`
- just get all the attributes for each type of assignment. If they're not relevant for that assignment type then they'll have a nil/default value anyway.
- exclude only the attributes that are specific to this course (id, created_at, updated_at, all foreign keys, etc.). By excluding instead of including, we run less of a risk that changes will break this (if we add something new it will be included by default, if we remove something it won't be included anyway)

For example:

```ruby
def assignment_properties_config
    exclude = %w[id created_at updated_at outstanding_remark_request_count]
    properties = self.attributes.except(*exclude).reject { |attr| attr.end_with?('_id') }
    properties['assignment_properties_attributes'] = self.assignment_properties.attributes.except(*exclude).reject { |attr| attr.end_with?('_id') }
    properties['assignment_files_attributes'] = self.assignment_files.pluck_to_hash(*(AssignmentFile.column_names - exclude).reject { |attr| attr.end_with?('_id') })
    properties['submission_rule_attributes'] = self.submission_rule.attributes.except(*exclude).reject { |attr| attr.end_with?('_id') }
    properties['submission_rule_attributes']['periods_attributes'] = self.submission_rule.periods.pluck_to_hash(*(Period.column_names - exclude).reject { |attr| attr.end_with?('_id') })
end
```

app/models/assignment.rb
```diff
@@ -1247,29 +1247,122 @@ def to_json(options = {})
...
+        default_starter_group = group_name
+      end
+      group_data << {
+        directory_name: group_name,
+        name: starter_file_group.name,
+        use_rename: starter_file_group.use_rename,
+        entry_rename: starter_file_group.entry_rename
+      }
+    end
+    starter_file_settings = {
```
mishaschwartz, line 1309: For the same reason as in the comment for `assignment_properties_config` I would like to avoid explicitly including attribute names whenever possible. I know that you want to keep all the information about starter files together in one zip file entry but in this case, the `starter_file_type` and the `starter_files_after_due` attributes should really be with the assignment properties, not here.

If we don't set them initially when we create the assignment then we run the risk of the new assignment not being valid (for example, if one of these attributes becomes a required attributes later on)

app/models/assignment.rb
```diff
@@ -1247,29 +1247,122 @@ def to_json(options = {})
...
+    self.add_test_files_to_zip(zip_file, zip_dir)
+    test_specs_path = self.autotest_settings_file
+    test_specs = File.exist?(test_specs_path) ? JSON.parse(File.open(test_specs_path, &:read)) : {}
+    test_specs['testers']&.each do |tester_info|
+      tester_info['test_data']&.each do |test_info|
+        test_info['extra_info']&.delete('test_group_id')
+      end
+    end
+    misc_settings = {}
+    misc_settings_attr = [:enable_test, :enable_student_tests, :tokens_per_period, :unlimited_tokens,
```
mishaschwartz, line 1343: For the same reason as in the comment for assignment_properties_config I would like to avoid explicitly including attribute names whenever possible. I know that you want to keep all the information about automated tests together in one zip file entry but in this case, these attributes should really be with the assignment properties, not here.

If we don't set them initially when we create the assignment then we run the risk of the new assignment not being valid (for example, if one of these attributes becomes a required attributes later on)

config/locales/defaults/download_upload/es.yml
```diff
@@ -12,6 +12,7 @@ es:
   upload: Subir
   upload_errors:
     invalid_csv_row_format: El formato de la fila es inválido
+    invalid_file_type: UPDATE ME
```
mishaschwartz, line 15: If a field should have a variable, include it even when we're just doing 'UPDATE ME'. So this should be:

```
invalid_file_type: UPDATE ME %{type}
```

And the same for the other translations below

app/models/assignment.rb
```diff
@@ -1247,29 +1247,97 @@ def to_json(options = {})
     self.assignment_properties.attributes.merge(self.attributes).symbolize_keys.to_json(options)
   end
 
+  # Returns an assignment's relevant properties for uploading/downloading an assignment's configuration as a hash
+  def assignment_properties_config
+    # Data to avoid including
+    exclude = %w[id created_at updated_at outstanding_remark_request_count repository_folder]
+    should_reject = ->(attr) { attr.end_with?('_id', '_created_at', '_updated_at') }
```
david-yz-liu, line 1254: I'm not sure what the `_created_at` and `_updated_at` suffixes are checking for?

app/controllers/assignments_controller.rb
```diff
@@ -562,8 +583,240 @@ def download_sample_starter_files
...
+    yaml_file = zip_file.get_entry(CONFIG_FILES[hash_to_build])
+    yaml_content = yaml_file.get_input_stream.read.encode(Encoding::UTF_8, 'UTF-8')
+    properties = parse_yaml_content(yaml_content)
+    if [:tags, :peer_review_tags].include?(hash_to_build)
+      properties.each { |row| row[:user] = @current_user.user_name }
+    end
+    properties
+  end
+
+  # Ensure that the +assignment+ type (scanned, timed, neither) matches the params
```
david-yz-liu, line 773: This is the only functional part which I'm not a huge fan of. Instead of having the upload link on the assignment form (and passing params for `is_scanned` and `is_timed`), let's move that upload link to the `assignments/index` page's heading links. You can use the name "Upload Configuration Zip File" (from the modal title) to distinguish it from the "Upload" link that already appears. Then, you can remove this method entirely, allowing the instructor to upload any time of assignment by using that modal.

(You can keep the layout changes to the assignment forms that you made, though.)

david-yz-liu (commented):
@anoverga17 excellent work, thank you! I left a few inline comments that you should be able to resolve pretty quickly.

app/models/assignment.rb
anoverga17: The `starter_file_updated_at attribute` in the assignment_properties model has to be rejected (adds unnecessary information that will be overwritten after uploading). 
The `_created_at` was added for future-proofing. If an attribute similar to `created_at` is made, it likely won't be necessary for this feature and can therefore be ignored.

mishaschwartz: I requested this change https://github.com/MarkUsProject/Markus/pull/5616#discussion_r761197563

app/models/assignment.rb
```diff
@@ -1247,29 +1247,97 @@ def to_json(options = {})
...
+  end
+
+  # Writes this assignment's starter file settings to the file located at +settings_filepath+ located in the
+  # +zip_file+. Also writes the starter files for this assignment in the same directory as +settings_filepath+.
+  def starter_file_config_to_zip(zip_file, settings_filepath)
+    default_starter_group = nil
+    group_data = []
+    directory_path = File.dirname(settings_filepath)
+    self.starter_file_groups.each.with_index(1) do |starter_file_group, download_count|
+      group_name = "#{download_count}: #{starter_file_group.name}"
```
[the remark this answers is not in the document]
mishaschwartz: starter file group names are not unique so we need something to differentiate

david-yz-liu (approved):

PR #5635, opened 2021-12-03
assessments: Fix bug in User#visible_assessments for students in a section

david-yz-liu:
## Motivation and Context

Currently, `User#visible_assessments` has a bug where if a student is in a section, then the `assessment_type` parameter is ignored. This causes assessments to be duplicated in the student assessment selector dropdown.

## Your Changes

**Description**: Fixed the bug in `User#visible_assessments`.

## Testing

Added some additional test cases, and manually verified that the student assessment dropdown only shows assessments once (under either assignments or grade entry forms).

## Checklist

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -289,15 +289,19 @@ class User < ApplicationRecord
   # only the assessment with the given id, if it is visible to the current user.
   # If it is not visible, returns an empty collection.
   def visible_assessments(assessment_type: nil, assessment_id: nil)
-    assessments = Assessment.where(is_hidden: false, type: assessment_type || Assessment.type)
+    assessments = Assessment.where(type: assessment_type || Assessment.type)
+    assessments = assessments.where(id: assessment_id) if assessment_id
     if self.section_id
-      assessments = Assessment.left_outer_joins(:assessment_section_properties)
-                              .where('assessment_section_properties.section_id': [self.section_id, nil])
+      assessments = assessments.left_outer_joins(:assessment_section_properties)
+                               .where(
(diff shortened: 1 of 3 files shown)


mishaschwartz (commented):
This looks fine but I don't understand why this would make a difference? To me this seems like it solves the problem where an assignment that is un-hidden for a specific section not being shown to that section. But I don't understand how this fixes the assessment_type issue.

app/models/user.rb
```diff
@@ -289,15 +289,19 @@ def reset_api_key
   # only the assessment with the given id, if it is visible to the current user.
   # If it is not visible, returns an empty collection.
   def visible_assessments(assessment_type: nil, assessment_id: nil)
-    assessments = Assessment.where(is_hidden: false, type: assessment_type || Assessment.type)
+    assessments = Assessment.where(type: assessment_type || Assessment.type)
+    assessments = assessments.where(id: assessment_id) if assessment_id
     if self.section_id
-      assessments = Assessment.left_outer_joins(:assessment_section_properties)
```
david-yz-liu, line 294: @mishaschwartz the root of the problem was this line. The previous version used `Assessment.left_outer_joins` and overwrite the query from the previous line, and so `assessment_type` was never used in this block of the code.

So the main change I made was to switch `Assessment.left_outer_joins` to `assessments.left_outer_joins`, so that in the new version both branches share the queries for `assessment_type` and `assessment_id`. (I could have left the `assessment_id` query below, but I thought it made a bit more sense to group the shared queries together. I don't care too strongly either way though.)

mishaschwartz (approved):

PR #5647, opened 2021-12-07
tags: update views for tags

pretendWhale:
## Motivation and Context

- ensures that all exam template views work as intended and all routing is up to date
- Note: layouts (specifically `sub_sub_menu`) have not been updated because the changes have been made in other PRs

## Your Changes

**Description**:

- [x] Other (please specify): multi-course app rewrite

## Checklist

### DIFF
--- a/app/assets/javascripts/Components/tag_table.jsx
+++ b/app/assets/javascripts/Components/tag_table.jsx
@@ -19,6 +19,7 @@ class TagTable extends React.Component {
   fetchData = () => {
     $.get({
       url: Routes.course_tags_path(this.props.course_id),
+      data: {assignment_id: this.props.assignment_id},
       dataType: "json",
     }).then(res => {
       this.setState({
--- a/app/assets/javascripts/Components/tag_table.jsx
+++ b/app/assets/javascripts/Components/tag_table.jsx
@@ -18,7 +18,7 @@ class TagTable extends React.Component {
 
   fetchData = () => {
     $.get({
-      url: Routes.assignment_tags_path(this.props.assignment_id),
+      url: Routes.course_tags_path(this.props.course_id),
       dataType: "json",
     }).then(res => {
       this.setState({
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/Components/tag_table.jsx
```diff
@@ -18,7 +18,7 @@ class TagTable extends React.Component {
 
   fetchData = () => {
     $.get({
-      url: Routes.assignment_tags_path(this.props.assignment_id),
+      url: Routes.course_tags_path(this.props.course_id),
```
mishaschwartz, line 21: We still need to pass the assignment_id to the tags controller when this is called from this component:

```ruby
Routes.course_tags_path(this.props.course_id, assignment_id: this.props.assignment_id)
```

We don't have a view for all tags yet

app/controllers/tags_controller.rb
```diff
@@ -5,7 +5,9 @@ class TagsController < ApplicationController
   layout 'assignment_content'
 
   def index
-    @assignment = Assignment.find_by(id: params[:assignment_id])
+    if params[:assignment_id]
```
mishaschwartz, line 8: This should not be necessary... am I missing something?

app/views/tags/index.html.erb
```diff
@@ -4,7 +4,7 @@
                handlers: [:erb] %>
   <%= javascript_tag nonce: true do %>
     document.addEventListener('DOMContentLoaded', () => {
-      makeTagTable(document.getElementById('tags_table'), {assignment_id: <%= @assignment.id %>});
```
mishaschwartz, line 7: We still need the assignment_id prop. Please pass both instead of replacing

app/views/tags/index.html.erb
```diff
@@ -15,11 +15,7 @@
   <% end %>
 <% end %>
 
-<% content_for :title,
```
mishaschwartz, line 18: We still want this. Please add it back

mishaschwartz (changes requested):

mishaschwartz (approved):

PR #5652, opened 2021-12-08
Tag edit modal should close only when cancel button is clicked

mishaschwartz:
## Motivation and Context

- tag edit modal should close when the cancel button is clicked, not when any part of the modal is clicked

## Your Changes

**Description**:

- update jquery selector

## Checklist


anoverga17 (approved):
Works great. 
It feels a little weird to me to be targeting all elements in the dialog with the reset type (as opposed to marking the cancel button with an id and targeting that). 
In this case though, it works out to be the same so it should be good either way.

mishaschwartz:
Yeah you're probably right. I'll change it up

david-yz-liu (approved):

PR #5665, opened 2021-12-13
chart-js: Fix registering scales for DataChart component

david-yz-liu:
## Motivation and Context

After doing the chart.js (#5626) and react-chartjs-2 (#5624) updates, we need to register `Chart` elements everywhere we import `Chart`. (See [here](https://www.chartjs.org/docs/latest/getting-started/integration.html#bundlers-webpack-rollup-etc).)

Without this registering, there was a runtime error when the `DataChart` component was used (e.g., Results view -> Summary tab -> Marks Chart modal).

## Your Changes

**Description**: Added the registering code to where we import `Chart` from `chart.js` in the `DataChart` component file.

## Testing

Ensure the `DataChart` component loads properly in the UI.

## Checklist

### DIFF
--- a/app/assets/javascripts/Components/Helpers/data_chart.jsx
+++ b/app/assets/javascripts/Components/Helpers/data_chart.jsx
@@ -2,6 +2,9 @@ import React from "react";
 import {render} from "react-dom";
 import {Bar} from "react-chartjs-2";
 
+import {Chart, registerables} from "chart.js";
+Chart.register(...registerables);
+
 export class DataChart extends React.Component {
   static defaultProps = {
     legend: true,


mishaschwartz (approved):
Looks good! Is there any downside to registering everything as opposed to just the components we need?

david-yz-liu:
@mishaschwartz great question, I should have left a comment about this. The disadvantage of registering everything is an increased bundle size. Since we use Chart.js in a limited way, we *really* don't need to register everything, but we're actually already doing so in https://github.com/MarkUsProject/Markus/blob/e4749a9439c624945408ed3f00fb42d0462f5906/app/javascript/packs/application.js#L46-L48.

So in the interest of expediency I've merged in this PR, but I'm going to make an issue now to review this code and likely simplify the registering (or rely on `react-chartjs-2` to [do the bundling for us](https://react-chartjs-2.netlify.app/docs/migration-to-v4#tree-shaking).

PR #5675, opened 2021-12-15
Created a view for content uploaded via URL

anoverga17:
## Motivation and Context

This pull request is part of a series of pull requests that aim to allow users to upload/grade larger files. As part of this, support for uploading URL submissions is necessary. This pull request creates a frontend for those submitted URL links. 

## Your Changes

**Description**:
- Created a new URLViewer React class that loads iframes
- Modified result and submission controller CSP header to allow iframes to load content from certain sources
- Created a new submission option to submit a URL and created a modal for the option

Note: currently only URLs from Youtube, Google Drive, and MyMedia Videos are accepted. 

## Testing

As this is a UI change, only manual testing of the feature was done. All the new views were loaded and accessed as various users to ensure the correct views was working and were being accessed accordingly. 

## Checklist

### DIFF
--- a/app/assets/javascripts/Components/Modals/submission_url_submit_modal.jsx
+++ b/app/assets/javascripts/Components/Modals/submission_url_submit_modal.jsx
@@ -1,0 +1,91 @@
+import React from "react";
+import Modal from "react-modal";
+
+class SubmitUrlUploadModal extends React.Component {
+  constructor() {
+    super();
+    this.state = {
+      newUrl: "",
+      newUrlText: "",
+    };
+  }
+
+  componentDidMount() {
+    Modal.setAppElement("body");
+  }
+
+  onSubmit = event => {
+    event.preventDefault();
+    this.props.onSubmit(this.state.newUrl, this.state.newUrlAlias);
+  };
+
+  handleUrlChange = event => {
+    const urlInput = event.target.value;
+    try {
+      const validatedURL = new URL(urlInput);
+      if (this.state.newUrlText === "") {
+        const suggestedText = validatedURL.hostname;
+        this.setState({newUrlText: suggestedText});
+      }
+    } catch (e) {
+      // If here, URL object creation failed and URL is invalid. Thus, no recommended text
+    } finally {
+      this.setState({newUrl: urlInput});
+    }
+  };
+
+  handleUrlAliasChange = event => {
+    this.setState({newUrlText: event.target.value});
+  };
+
+  handleModalClose = () => {
+    this.setState({
+      newUrl: "",
+      newUrlText: "",
+    });
+    this.props.onRequestClose();
+  };
+
+  render() {
+    return (
+      <Modal
+        className="react-modal"
+        isOpen={this.props.isOpen}
+        onRequestClose={this.handleModalClose}
+      >
+        <h2>{I18n.t("submit_the", {item: I18n.t("submissions.student.link")})}</h2>
+        <form onSubmit={this.onSubmit}>
+          <div className={"modal-container-vertical"}>
+            <div className={"modal-container"}>
+              <label className={"modal-inline-label"}>{I18n.t("submissions.student.url")}</label>
+              <input
+                type={"url"}
+                name={"new_url"}
+                value={this.state.newUrl}
+                onChange={this.handleUrlChange}
+                required={true}
+              />
+            </div>
+            <div className={"modal-container"}>
+              <label className={"modal-inline-label"}>
+                {I18n.t("submissions.student.url_text")}
+              </label>
+              <input
+                type={"text"}
+                name={"new_url_text"}
+                value={this.state.newUrlText}
+                onChange={this.handleUrlAliasChange}
+                required={true}
+              />
+            </div>
+          </div>
+          <div className={"modal-container"}>
+            <input type="submit" value={I18n.t("save")} />
+          </div>
+        </form>
+      </Modal>
+    );
+  }
+}
+
+export default SubmitUrlUploadModal;
--- a/app/assets/javascripts/Components/Result/url_viewer.jsx
+++ b/app/assets/javascripts/Components/Result/url_viewer.jsx
@@ -1,0 +1,21 @@
+import React from "react";
+
+export class URLViewer extends React.Component {
+  constructor(props) {
+    super(props);
+    //TODO: Remove state url and replace its usage with url props when backend is working
+    this.state = {
+      url: "https://www.youtube.com/embed/dQw4w9WgXcQ", // Sample Youtube embed link
+      //url: "https://drive.google.com/file/d/1yQh8bzo4IzmtSt-qZ0Vb-BzzEXsp45v_/preview" // sample google drive embed link
+      //url: "https://play.library.utoronto.ca/watch/4493e554354a7e51bdf90f3197185b5f" // sample mymedia link
+    };
+  }
+
+  render() {
+    return (
+      <div className="url-container" key={"url_container"}>
+        <iframe className="url-display" src={this.state.url} allowFullScreen />
+      </div>
+    );
+  }
+}
--- a/app/assets/javascripts/Components/markus_file_manager.jsx
+++ b/app/assets/javascripts/Components/markus_file_manager.jsx
@@ -21,6 +21,11 @@ class RawFileManager extends RawFileBrowser {
     this.handleActionBarAddFolderClick(event);
   };
 
+  onActionBarSubmitURLClick = event => {
+    event.preventDefault();
+    this.props.onActionBarSubmitURLClick();
+  };
+
   folderTarget = selectedItem => {
     // treat multiple selections as not targeting a folder
     const selectionIsFolder = !!selectedItem && selectedItem.relativeKey.endsWith("/");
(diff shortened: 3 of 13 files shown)


app/assets/stylesheets/common/core.scss
```diff
@@ -469,6 +469,11 @@ th.required::after {
   vertical-align: middle;
 }
 
+.fa-submit-link-o::before {
+  content: asset-url('icons/link_add.png');
```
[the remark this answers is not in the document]
anoverga17: Ignoring since all other similar asset links use single quotes

app/assets/javascripts/Components/markus_file_manager.jsx
```diff
@@ -173,6 +178,17 @@ class RawFileManager extends RawFileBrowser {
         </li>
       );
 
+      if (this.props.enableUrlSubmit === true) {
```
mishaschwartz, line 181: Don't check a boolean by comparing it to true (or false either)

mishaschwartz: This option disappears if a file or folder is selected

app/assets/javascripts/Components/Modals/submission_url_submit_modal.jsx
```diff
@@ -0,0 +1,69 @@
+import React from "react";
+import Modal from "react-modal";
+
+class SubmitUrlUploadModal extends React.Component {
+  state = {
```
mishaschwartz, line 5: State should be set in the `constructor` not as a static attribute on the class.

app/assets/javascripts/Components/Modals/submission_url_submit_modal.jsx
```diff
@@ -0,0 +1,69 @@
...
+    newUrlAlias: "",
+  };
+
+  componentDidMount() {
+    Modal.setAppElement("body");
+  }
+
+  onSubmit = event => {
+    event.preventDefault();
+    this.props.onSubmit("https://youtu.be/dQw4w9WgXcQ", "");
```
mishaschwartz, line 16: Make this actually send what is in the state. Even though we are not handling this in the backend just yet you can still make the front end work as intended.

app/assets/javascripts/Components/Modals/submission_url_submit_modal.jsx
```diff
@@ -0,0 +1,69 @@
...
+  };
+
+  handleUrlChange = event => {
+    this.setState({newUrl: event.target.value});
+  };
+
+  handleUrlAliasChange = event => {
```
mishaschwartz, line 23: What is the goal of including this field as well? I'm not sure I understand what its purpose is.

app/assets/javascripts/Components/Modals/submission_url_submit_modal.jsx
```diff
@@ -0,0 +1,69 @@
...
+    const {isOpen, onRequestClose} = this.props;
+    return (
+      <Modal className="react-modal" isOpen={isOpen} onRequestClose={onRequestClose}>
+        <h2>{I18n.t("submissions.student.create_link")}</h2>
+        <form onSubmit={this.onSubmit}>
+          <div className={"modal-container-vertical"}>
+            <div className={"modal-container"}>
+              <label>
+                {I18n.t("submissions.student.url")}
+                <br />
```
mishaschwartz, line 37: In all other modals of this type we put the label to the left of the input. Please do the same here (instead of the label on top of the input)

app/assets/javascripts/Components/Result/url_viewer.jsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import "../../../stylesheets/common/_url_viewer.scss";
```
mishaschwartz, line 2: Don't import css files in jsx. Import it in the same we do do for the `_file_viewer.scss` file (in `_markus.scss)

app/assets/javascripts/Components/Result/url_viewer.jsx
```diff
@@ -0,0 +1,25 @@
...
+
+export class URLViewer extends React.Component {
+  state = {
```
mishaschwartz, line 5: State should be set in the `constructor` not as a static attribute on the class.

app/assets/javascripts/Components/Result/url_viewer.jsx
```diff
@@ -0,0 +1,25 @@
...
+    url: "https://www.youtube.com/embed/dQw4w9WgXcQ", // Sample Youtube embed link
+    //url: "https://drive.google.com/file/d/1yQh8bzo4IzmtSt-qZ0Vb-BzzEXsp45v_/preview" // sample google drive embed link
+    //url: "https://play.library.utoronto.ca/watch/4493e554354a7e51bdf90f3197185b5f" // sample mymedia link
+  };
+
+  componentDidMount() {
```
mishaschwartz, line 11: You can just use the prop as the default value for state if you're setting state in the constructor:

```js
  constructor(props) {
    super(props);
    this.state = {
      url: this.props.url,
    };
  }
```

mishaschwartz: However, I don't think you even need to set the state at all here. Just use the prop directly:

```js
        <iframe className="url-display" src={this.props.url} allowFullScreen />
```

You don't need to update this value inside this component.

app/assets/javascripts/Components/markus_file_manager.jsx
```diff
@@ -21,6 +21,11 @@ class RawFileManager extends RawFileBrowser {
     this.handleActionBarAddFolderClick(event);
   };
 
+  onActionBarSubmitURLClick = (event, selectedItem) => {
+    event.preventDefault();
+    this.props.onActionBarSubmitURLClick(selectedItem);
```
mishaschwartz, line 26: I don't think we need to pass the selected item do we? It doesn't/shouldn't matter if an item is selected when we upload a URL.

app/assets/javascripts/Components/submission_file_manager.jsx
```diff
@@ -161,7 +167,11 @@ class SubmissionFileManager extends React.Component {
   };
 
   openUploadModal = uploadTarget => {
-    this.setState({showModal: true, uploadTarget: uploadTarget});
+    this.setState({showUploadModal: true, uploadTarget: uploadTarget});
+  };
+
+  openSubmitURLModal = uploadTarget => {
+    this.setState({showURLModal: true, uploadTarget: uploadTarget});
```
mishaschwartz, line 174: I don't think the uploadTarget matters here

app/controllers/submissions_controller.rb
```diff
@@ -122,6 +123,7 @@ def populate_file_manager
       end
       entries = get_all_file_data(revision, grouping, '')
     end
+    puts entries
```
mishaschwartz, line 126: remove debugging code

app/controllers/results_controller.rb
```diff
@@ -13,6 +13,7 @@ class ResultsController < ApplicationController
     # required because MathJax dynamically changes
     # style. # TODO: remove this when possible
     p.style_src :self, "'unsafe-inline'"
+    p.frame_src 'https://www.youtube.com', 'https://drive.google.com', 'https://play.library.utoronto.ca'
```
mishaschwartz, line 16: Make these a constant on the class and refer to that constant here (and in the submissions_controller.rb)

config/locales/views/submissions/en.yml
```diff
@@ -63,6 +63,7 @@ en:
       remark_requested: Remark Requested
     status: Status
     student:
+      create_link: Create Link
```
mishaschwartz, line 66: I don't love "Create Link".  I don't think that it really conveys the intended use of "submit a file or video hosted at a different url". But let's discuss this at a meeting to figure out the best way to go with this

mishaschwartz (changes requested):

config/locales/views/submissions/en.yml
mishaschwartz: From the meeting:

- change "Create Link" to "Submit Link"
- change "Upload File" to "Submit File"

app/assets/javascripts/Components/Modals/submission_url_submit_modal.jsx
anoverga17: url_alias is the name of the url (if the user wants to give the link a proper name as opposed to just having the name be the link itself). Right now, it is optional and just helps the filename for links be more readable.

anoverga17:
Thanks for the feedback Misha. A lot of these are old habits from when I last programmed with React so I'll be sure to keep these in mind moving forward.

app/assets/javascripts/Components/Modals/submission_url_submit_modal.jsx
mishaschwartz: Ok that makes sense. Let's think about a different label for this one as well, "url alias" to me sounds like they can provide a secondary URL. I had a look at a few ways other software labels this:

- slack uses "Text"
- google docs uses "Text"
- word uses "Text to display"

PR #5721, opened 2022-01-04
starter-files: Validate uniqueness of starter file group name for assignments

david-yz-liu:
## Motivation and Context

I'd like to ensure starter file group names are unique to help instructors disambiguate them (most importantly, when downloading starter file mappings) and exporting starter files (I requested this in #5616).

## Your Changes

**Description**: Added a uniqueness validation for the starter file group name. Updated logic to automatically generate starter file group names for an assignment, inspired by handling of duplicated file names.

## Testing

Added additional tests and ran test suite; checked starter file functionality in UI.

I kept the `:name` param in the `StarterFileGroupsController#create` method, even though our current code no longer uses it, in case we want to use it in the future. I could also remove it for now.

## Checklist


app/controllers/starter_file_groups_controller.rb
```diff
@@ -6,7 +6,11 @@ class StarterFileGroupsController < ApplicationController
 
   def create
     assignment = Assignment.find_by(id: params[:assignment_id])
-    assignment.starter_file_groups.create(update_params)
+
+    new_params = update_params
```
mishaschwartz, line 10: Do you think it makes more sense to put this logic in a `before_validate` callback in the StarterFileGroup model?

Something like:

```ruby
# app/models/starter_file_group.rb

before_validate :set_name, on: :create

private

def set_name
    return if self.name || self.assignment.nil?

    current_names = self.assignment.starter_file_groups.pluck(:name).to_set
    new_name = I18n.t('assignments.starter_file.new_starter_file_group')
    i = 1
    while current_names.include? new_name
      new_name = "#{I18n.t('assignments.starter_file.new_starter_file_group')} (#{i})"
      i += 1
    end
    self.name = new_name
end
```

That way all the logic is contained in the model (so that it will be enforced no matter where it is created) and you don't have to worry about modifying the params in the controller(s).

david-yz-liu: Yeah that's a good idea, thanks. I'll make the change soon.

mishaschwartz (approved):

PR #5738, opened 2022-01-10
Only check for timeout if not logged in remotely

mishaschwartz:
## Motivation and Context

If logged in with remote authentication, the remote authenticator handles its own timeouts so we don't need to artificially handle timeouts in MarkUs. Otherwise it is just annoying for the user:

- if MarkUs times out before the remote authenticator, the user has to re-log in to MarkUs even though their login credentials haven't expired
- if the remote authenticator times out before MarkUs, there is no warning and pages will just stop loading all of a sudden

This change also ensures that if the remote authenticator does time out, MarkUs picks up on that the next time a request is made and logs the user out so that they can re-log in with the remote authenticator.

## Your Changes

**Description**:

- check for the remote user name in the real_user method so that it gets checked on every request
- only check for timeouts when the user is not authenticated remotely
- periodically check to see if the user has timed out yet if the user is not authenticated remotely

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,7 @@
 ## [unreleased]
 - Remove unmaintained locales (#5727)
 - Fix bug where "Create Note" button was displaying when a course had no noteables (#5745)
+- Do not timeout users who are logged in using remote authentication (#5738)
 
 ## [v2.0.3]
 - Fix bug where repository access files were not taking multiple courses into account (#5734)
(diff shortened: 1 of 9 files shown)


mishaschwartz:
TODO: don't check for imminent timeout if logged in remotely

mishaschwartz:
failing tests are unrelated to this PR

david-yz-liu (approved):

PR #5745, opened 2022-01-11
notes: Fix Note.noteables_exist? to handle multiple courses

david-yz-liu:
## Motivation and Context

The current implementation of `Note.noteables_exist?` looks for the existence of assignments/groupings/students, but now that we support multiple courses, we need to take into account the current course as well. 

## Your Changes

**Description**: Modify the method to take into account the course to check.

## Testing

Added some test cases, and tested in the UI (on the "Notes" page).

## Checklist

### DIFF
--- a/app/models/note.rb
+++ b/app/models/note.rb
@@ -15,9 +15,9 @@ class Note < ApplicationRecord
     I18n.l(created_at)
   end
 
-  def self.noteables_exist?
+  def self.noteables_exist?(course_id)
     NOTEABLES.each do |classname|
-      unless Kernel.const_get(classname).all.empty?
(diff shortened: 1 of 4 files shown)


mishaschwartz (approved):

app/models/note.rb
```diff
@@ -15,9 +15,9 @@ def format_date
     I18n.l(created_at)
   end
 
-  def self.noteables_exist?
+  def self.noteables_exist?(course_id)
     NOTEABLES.each do |classname|
-      unless Kernel.const_get(classname).all.empty?
+      unless Kernel.const_get(classname).joins(:course).where('courses.id': course_id).empty?
```
mishaschwartz, line 20: The newer/rails syntax is `classname.constantize` but this works too

PR #5752, opened 2022-01-12
Redesign login page

mishaschwartz:
## Motivation and Context

Users report that displaying the local authentication login form as well as the remote authentication button is confusing because some users will try to put their remote authentication credentials into the form and will not be able to log in.

This PR attempts to reduce the confusion by changing the login form:

- if only remote authentication is enabled: show remote authentication button

markus-remote-only

- if only local authentication is enabled: show local authentication form

markus-local-only

- if both are enabled: show local authentication and remote authentication buttons
    - when the local authentication button is clicked, show local authentication form.

markus-local-remote-hidden
markus-local-remote-visible

## Your Changes

**Description**:

 ## Testing

## Checklist

### DIFF
--- a/app/views/main/login.html.erb
+++ b/app/views/main/login.html.erb
@@ -1,5 +1,17 @@
 <% content_for :head do %>
   <%= javascript_include_tag 'cookies_eu', nonce: true %>
+  <%= javascript_tag nonce: true do %>
+    document.addEventListener('DOMContentLoaded', () => {
+      document.getElementById('local-login-selector').addEventListener('click', () => {
+        const formElem = document.getElementById('local-login-form');
+        if (getComputedStyle(formElem).display === 'none') {
+          formElem.style.display = 'block';
+        } else {
+          formElem.style.display = 'none';
+        }
+      })
+    })
+  <% end %>
 <% end %>
 
 <main class='login'>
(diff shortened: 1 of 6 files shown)


app/views/main/login.html.erb
```diff
@@ -1,5 +1,20 @@
 <% content_for :head do %>
   <%= javascript_include_tag 'cookies_eu', nonce: true %>
+  <%= javascript_tag nonce: true do %>
+    document.addEventListener('DOMContentLoaded', () => {
+      document.getElementById('local-login-selector').addEventListener('click', () => {
+        const formElem = document.getElementById('local-login-form');
+        const orElem = document.getElementById('login-or-container');
```
david-yz-liu, line 7: Maybe this is picky but I'd like to see the "OR" div always visible when both forms of authentication is supported, so that the "expanded MarkUs login" view looks like 

image

I guess this makes the HTML structure below slightly more complex since you need an extra conditional for the OR, but I think the UI looks a bit better that way. :)

config/locales/common/en.yml
```diff
@@ -24,3 +24,4 @@ en:
   search: Search
   skip: Skip
   working: Loading
+  or: Or
```
david-yz-liu, line 27: Please run the `i18n-tasks normalize` task

david-yz-liu (approved):

Issue #5670, opened 2021-12-14
Cannot delete folders


hykilpikonna:
The delete button is hidden when a folder is selected.

![image](https://user-images.githubusercontent.com/22280294/145908156-d28e50e3-8b25-4981-bbe5-ddc313326a87.png)

mishaschwartz:
Looking at this now. I think we have a policy that says that students cannot create or delete folders but that is not being strictly adhered to because we allow students to upload a zip file that we unzip, and create folders in the process.

We should revisit this policy to see if it still makes sense. If it does then we should also prevent students from uploading zip files in this way. If it does not, then we should remove this policy altogether instead of half applying it like we do now.

hykilpikonna:
For me, the upload zip functionality is very important, especially when I have a lot of project files. Finding the changed files is a hassle, and updating every single file manually takes a lot of time. So I created a script that will automatically pack my project files into a zip file so that it can make sure every submitted file is up to date without unnecessary effort.

The folder was created because I forgot to test the packing script before I submitted the first zip file, and I can't delete the folder afterwards.

mishaschwartz:
That makes sense. We are going to go with the following changes:

- Update the policy so that all roles (including students) can create/destroy directories in the file viewer
- Ensure that parent directories can still be created when strict required files are enforced. For example, if a student must submit a file `subfolder/example.py` then the student will be able to create `subfolder/` before creating/uploading `example.py` into that directory.

PR #5763, opened 2022-01-18
allow students to create folders for submissions

pretendWhale:
## Motivation and Context

Allow students to create folders when creating submissions via the web interface. Addresses #5670 

## Your Changes

**Description**:
- Remove manage_subdirectories policy and checks on this policy
- Repo manager and Submission Manager: add default enableSubdirs prop (true by default)

## Testing

tested by navigating to the web interface and adding/creating folders

- This allows students to create arbitrary folders in web submissions at all times, including when there are strict required files enforced. We may want to disallow this behavior in that circumstance.
- This is probably a different pr, but when there are strict required files that include directories, it might be nice to allow uploading a zip with an arbitrary name that unzips into the required file structure. Currently, a zip file will be rejected as not one of the required files.

## Checklist

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -317,10 +317,6 @@ class SubmissionsController < ApplicationController
     # The folders that will be deleted
     delete_folders = params[:delete_folders] || []
 
-    unless delete_folders.empty? && new_folders.empty?
-      authorize! to: :manage_subdirectories?
-    end
-
     if delete_files.empty? && new_files.empty? && new_folders.empty? && delete_folders.empty?
       flash_message(:warning, I18n.t('student.submission.no_action_detected'))
     else
--- a/app/policies/submission_policy.rb
+++ b/app/policies/submission_policy.rb
@@ -18,10 +18,6 @@ class SubmissionPolicy < ApplicationPolicy
     role.instructor? || role.ta?
   end
 
-  def manage_subdirectories?
-    role.instructor? || role.ta?
-  end
-
   def view_files?
     true
   end
--- a/app/assets/javascripts/Components/repo_browser.jsx
+++ b/app/assets/javascripts/Components/repo_browser.jsx
@@ -11,6 +11,10 @@ class RepoBrowser extends React.Component {
     };
   }
 
+  static defaultProps = {
+    enableSubdirs: true,
+  };
+
   componentDidMount() {
     this.fetchRevisions();
   }
--- a/app/assets/javascripts/Components/submission_file_manager.jsx
+++ b/app/assets/javascripts/Components/submission_file_manager.jsx
@@ -23,6 +23,7 @@ class SubmissionFileManager extends React.Component {
     readOnly: false,
     revision_identifier: undefined,
     starterFileChanged: false,
+    enableSubdirs: true,
   };
 
   componentDidMount() {
--- a/app/views/submissions/file_manager.html.erb
+++ b/app/views/submissions/file_manager.html.erb
@@ -14,7 +14,6 @@
           assignment_id: <%= @assignment.id %>,
           grouping_id: <%= @grouping.id %>,
           readOnly: <%= !@assignment.allow_web_submits %>,
-          enableSubdirs: <%= allowed_to? :manage_subdirectories? %>,
           starterFileChanged: <%= @grouping.starter_file_changed %>
         });
     });
--- a/app/views/submissions/repo_browser.html.erb
+++ b/app/views/submissions/repo_browser.html.erb
@@ -15,7 +15,6 @@
           late_penalty: <%= !@grouping.assignment.submission_rule.is_a? NoLateSubmissionRule %>,
           grouping_id: <%= @grouping.id %>,
           collected_revision_id: '<%= @collected_revision&.revision_identifier %>',
-          enableSubdirs: <%= allowed_to? :manage_subdirectories? %>,
           collection_date: '<%= l(@grouping.collection_date) %>',
           due_date: '<%= l(@grouping.due_date) %>',
           is_timed: <%= @grouping.assignment.is_timed %>,


spec/controllers/submissions_controller_spec.rb
```diff
@@ -184,6 +184,59 @@
           expect(files['TestShapes.java']).to be_nil
         end
       end
+
+      context 'when creating a folder with required files' do
+        before :each do
+          @assignment.update(
```
mishaschwartz, line 190: make this `update!` so that it doesn't fail silently for some reason

spec/controllers/submissions_controller_spec.rb
```diff
@@ -184,6 +184,59 @@
...
+        it 'should upload a folder named test_zip' do
+          post_as @student, :update_files,
+                  params: { course_id: course.id, assignment_id: @assignment.id,
+                            new_folders: ['test_zip'] }
+          expect(response).to have_http_status :ok
+        end
+        it 'should upload a subdirectory' do
+          post_as @student, :update_files,
+                  params: { course_id: course.id, assignment_id: @assignment.id,
+                            new_folders: ['test_zip/zip_subdir'] }
```
mishaschwartz, line 204: shouldn't the params be `path: 'test_zip', new_folders: ['zip_subdir']`?

and the same for the test at line 213 below

spec/controllers/submissions_controller_spec.rb
```diff
@@ -184,6 +184,59 @@
           expect(files['TestShapes.java']).to be_nil
         end
       end
+
+      context 'when creating a folder with required files' do
```
mishaschwartz, line 188: Add some tests to make sure the folder actually get created or not (don't just check the http status)

mishaschwartz (changes requested):
Just some testing changes. Otherwise looks great!

spec/controllers/submissions_controller_spec.rb
pretendWhale: I don't think so.  `update_files` does take a path parameter. it's used when creating new files, but it is not used in creating new folders (see `handleCreateFolder` in `submission_file_manager.jsx`). 

Instead just the grouping_id and the list of new folders (with full relative paths) are sent to update_files. I'd actually like to discuss that briefly with you - I think I understand why, but I want to confirm

spec/controllers/submissions_controller_spec.rb
pretendWhale: Right, of course. This did lead me to catch an edge case (not sure if it is known): 

When attempting to upload a subdirectory without initially creating its parent directory, it fails silently (that is, it returns a 200, but no folders are actually added to the repo).

I haven't fully diagnosed why this is happening, but in practice I'm fairly certain it never will. Using the web interface, only one directory can be created at a time. using zip files, all of the directories are created in order. Someone could make a post with those parameters, but it isn't supported. Either way, do we think this is at all worth investigating and either creating the parent directory if necessary or (more likely) catching this and returning an error?

spec/controllers/submissions_controller_spec.rb
mishaschwartz: I see... you're right. It probably doesn't matter but I think I expect it to be the other way just so that it's consistent with the file upload bit. I guess the function handles it either way so let's leave it like this for now since that mimics what is being sent by the front end.

spec/controllers/submissions_controller_spec.rb
mishaschwartz: Can you add a test for that case where you check if it's displaying flash messages? I don't think we're necessarily changing the return code if a transaction fails (we probably should but we don't) but we should at least be writing the error messages to the flash hash.

mishaschwartz (approved):

app/controllers/submissions_controller.rb
```diff
@@ -337,10 +337,9 @@ def update_files
         else
           required_files = nil
         end
-
```
david-yz-liu, line 340: Please revert this change

app/helpers/repository_helper.rb
```diff
@@ -132,6 +132,14 @@ def add_folder(folder_path, user, repo, path: '/', txn: nil)
 
     folder_path = current_path.join(folder_path)
     folder_path = folder_path.to_s
+
+    # check if only required files are allowed for a submission
+    # allowed folders = paths in required files
+    if required_files.present? && required_files.none? { |file| file.starts_with?(folder_path) }
+      messages << [:extra_files, new_folder_path]
```
david-yz-liu, line 139: `new_folder_path` doesn't exist (a bit disappointing that Rubocop didn't pick this up, I might look into that).

Also, please create a new error message for this case rather than reusing the one for `:extra_files`, as the action that triggers this is different.

app/policies/submission_policy.rb
```diff
@@ -19,7 +19,7 @@ def manage_files?
   end
 
   def manage_subdirectories?
-    role.instructor? || role.ta?
+    user.end_user?
```
david-yz-liu, line 22: Eh, it seems like this policy is no longer necessary. You can clean up the code a bit by removing its uses, and the corresponding `enableSubdirs` prop from the file browser React component(s).

mishaschwartz: We talked about this in the developer meeting last Wednesday... we would like to keep this in place because I would like to make this an assignment option in the future (allow subdirectories or no). 

@pretendWhale This does remind me though that we should be checking this policy in the `update_files` method to ensure that the current user is allowed to create directories

app/helpers/repository_helper.rb
mishaschwartz: @pretendWhale I think you wrote a test that should fail here? ... can you double check at least one test is hitting this line as well.

app/policies/submission_policy.rb
david-yz-liu: Good point, sounds good to me @mishaschwartz!

@pretendWhale could you please also make a new issue for adding an assignment option to allow subdirectory creation/deletion, and link it to this PR?

david-yz-liu (approved):

PR #5768, opened 2022-01-21
groups: Ensure student 'Delete Group' button has correct path

david-yz-liu:
## Motivation and Context

The current code for the student interface Delete Group button calls a path helper as follows: `course_assignment_group_path(@current_course.id, @assignment.id)`. This omits an `:id` argument, which is automatically filled in as the `@assignment.id`. For example, `DELETE courses/1/assignments/10/groups/10`.

The `GroupsController` ignores the `:id` parameter so this previously wasn't an issue, but now that we're doing stricter URL checking and model associations, this `:id` is used to look up a grouping, which doesn't necessarily belong to the assignment, or even part of the same course, resulting in a 404 error.

## Your Changes

**Description**: Fix the path helper.

## Testing

Tested manually in the UI (the bug is in the view file).

Even though I fixed the `:id` parameter, it's still not being used in the controller method. We can change this if we move the method into a `Groupings` controller in the future.

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,7 @@
 
 ## [unreleased]
 - Remove unmaintained locales (#5727)
+- Fix bug for "Delete Group" button generating an invalid path (#5768)
 
 ## [v2.0.5]
 - Add ability to annotate notebook (jupyter and Rmd) submissions (#5749)
--- a/app/views/assignments/show.html.erb
+++ b/app/views/assignments/show.html.erb
@@ -205,7 +205,7 @@
             @grouping.accepted_students.size == 1 &&
             @grouping.extension.nil? %>
             <%= button_to t('helpers.submit.delete', model: Group.model_name.human),
-                          course_assignment_group_path(@current_course.id, @assignment.id),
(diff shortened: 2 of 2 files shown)


mishaschwartz (changes requested):
Thanks for catching that. I think the way to solve this though is not to pass an additional parameter that is ignored in the controller method.

I think instead we should make the destroy method a collection route instead (like we did with remove_group) since the way it does not require a grouping id passed as a parameter.

mishaschwartz:
We really should have a different grouping controller but that's a different conversation

david-yz-liu:
@mishaschwartz I see what you're saying and did consider this, but I disagree here. I prefer sticking with the Rails default, which is to treat `destroy` as a member route. My preference is to modify the controller method itself, but I didn't do so in this pull request to keep the change small.

With the `remove_group` method, it would also make sense to move this to a `GroupingsController` (which I mentioned in the PR description), and turn it into a member route. It already takes in a `grouping_id` parameter.

mishaschwartz (approved):

PR #5776, opened 2022-01-24
add test for admin user model

TheTallJerry:
## Motivation and Context

Added tests for uncovered method in the AdminUser model. Used constant in AdminUser model. 
## Your Changes

**Description**: Added a new test file under spec/models for the AdminUser model, testing both the internal behavior and the return value of the tested method. In the same model, for the call to `find_or_create_by`, changed the passed in string to a predefined constant. This change doesn't affect other areas of the application. 

## Testing

Ran RSpec with this newly added test file. Got 100% coverage for the method covered and all the exisiting tests passed. 

## Checklist

### DIFF
--- a/spec/models/admin_user_spec.rb
+++ b/spec/models/admin_user_spec.rb
@@ -1,10 +1,10 @@
-describe AdminUser do 
-    let(:testuser) { AdminUser.find_or_create }
-    it "return type should be User" do
-        expect(testuser).to be_a_kind_of(User)
-    end
-    it "first_name and last_name should equal to the specified values" do
-        expect(testuser.first_name).to eq('admin')
-        expect(testuser.last_name).to eq('admin')
-    end
+describe AdminUser do
+  let(:testuser) { AdminUser.find_or_create }
+  it 'return type should be User' do
+    expect(testuser).to be_a_kind_of(User)
+  end
+  it 'first_name and last_name should equal to the specified values' do
+    expect(testuser.first_name).to eq('admin')
+    expect(testuser.last_name).to eq('admin')
+  end
 end
--- a/spec/models/admin_user_spec.rb
+++ b/spec/models/admin_user_spec.rb
@@ -1,22 +1,36 @@
 describe AdminUser do
-  it 'return type should be User' do
-    testuser = AdminUser.find_or_create
-    expect(testuser).to be_a_kind_of(User)
-  end
-  context 'no existing admin' do
-    it 'first_name and last_name should equal to the specified values' do
-      testuser = AdminUser.find_or_create
-      expect(testuser.first_name).to eq('admin')
-      expect(testuser.last_name).to eq('admin')
+  describe '.find_or_create' do
+    let(:admin_user) { AdminUser.find_or_create }
+    it 'returns an instance of AdminUser' do
+      expect(admin_user).to be_an_instance_of(AdminUser)
     end
-  end
-  context 'existing admin' do
-    it 'first_name and last_name should ot be changed for existing admin' do
-      adminuser = AdminUser.create(user_name: AdminUser::ADMIN_USERNAME, first_name: 'jerry', last_name: 'foo')
-      testuser = AdminUser.find_or_create
-      expect(testuser.first_name).to eq('jerry')
-      expect(testuser.last_name).to eq('foo')
-      expect(testuser).to eq(adminuser)
+    it 'ensures the returned AdminUser api_key is not null' do
+      expect(admin_user.api_key).not_to be(nil)
+    end
+    context 'when there is no existing AdminUser' do
+      it 'will create a new AdminUser' do
+        expect { AdminUser.find_or_create }.to change { AdminUser.count }.by(1)
+      end
+      it 'will set first_name and last_name to the default values' do
+        expect(admin_user.first_name).to eq('admin')
+        expect(admin_user.last_name).to eq('admin')
+      end
+    end
+    context 'when there is existing AdminUser' do
+      let!(:existing_admin_user) do
+        AdminUser.create(user_name: AdminUser::ADMIN_USERNAME, first_name: 'jerry', last_name: 'foo')
+      end
+      let(:returned_user) { AdminUser.find_or_create }
+      it 'will not create a new AdminUser' do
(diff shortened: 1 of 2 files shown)


david-yz-liu (commented):
@TheTallJerry great start! I think there are some other cases of arguments that you should be testing as well.

TheTallJerry:
Looks like CI failed - looking into it

spec/models/admin_user_spec.rb
```diff
@@ -0,0 +1,22 @@
+describe AdminUser do
+  it 'return type should be User' do
```
david-yz-liu, line 2: Tests for a method are grouped together in a separate `describe` block (see our other test files). In this case it would be `describe '.find_or_create'`.

spec/models/admin_user_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    testuser = AdminUser.find_or_create
```
david-yz-liu, line 3: Instead of calling this in each method, pull it out into a [`let` statement](https://relishapp.com/rspec/rspec-core/docs/helper-methods/let-and-let) inside the describe block for this method.

spec/models/admin_user_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    expect(testuser).to be_a_kind_of(User)
```
david-yz-liu, line 4: In general we don't test for types, but you can keep this test case but change it to `AdminUser` to make it more specific.

spec/models/admin_user_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+  context 'no existing admin' do
```
david-yz-liu, line 6: For context blocks, the text should introduce a condition with "when"/"where" and use complete phrases. For example, `'when there is no existing admin'`.

spec/models/admin_user_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    it 'first_name and last_name should equal to the specified values' do
```
david-yz-liu, line 7: Again, the labels are meant to form human phrases, with "it" referring to the method. So for example, `it 'returns a user...'`

Also: "specified value" is vague. You can instead use "default value".

Finally, split the two expect statements into separate tests.

spec/models/admin_user_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+      testuser = AdminUser.find_or_create
+      expect(testuser.first_name).to eq('admin')
+      expect(testuser.last_name).to eq('admin')
+    end
+  end
+  context 'existing admin' do
+    it 'first_name and last_name should ot be changed for existing admin' do
```
david-yz-liu, line 14: In addition to my comments above, there's a typo with "ot". And once again, make sure the sentence makes grammatical sense.

david-yz-liu: Also make sure to split these up into separate tests as well.

You can use a `let!(:admin_user)` in this context block to ensure an admin user is already created at each test (you can check out the documentation for the difference between `let` and `let!`).

david-yz-liu (commented):
@TheTallJerry okay great, I left a more detailed review this time. In addition to the inline comments, please also add a test to check that the `api_key` attribute is not `nil` (this is the other line of code in the `find_or_create` method).

spec/models/admin_user_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  describe '.find_or_create' do
+    let(:admin_user) { AdminUser.find_or_create }
+    it 'returns an instance of AdminUser' do
+      expect(admin_user).to be_an_instance_of(AdminUser)
+    end
+    it 'ensures the returned AdminUser api_key is not null' do
+      expect(admin_user.api_key).not_to be(nil)
+    end
+    context 'when there is no existing AdminUser' do
+      it 'will create a new AdminUser' do
```
david-yz-liu, line 11: In general, use present tense for the descriptions ("creates" instead of "will create").

spec/models/admin_user_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+        expect { AdminUser.find_or_create }.to change { AdminUser.count }.by(1)
+      end
+      it 'will set first_name and last_name to the default values' do
```
david-yz-liu, line 14: split this up into two separate tests

spec/models/admin_user_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+        expect(admin_user.first_name).to eq('admin')
+        expect(admin_user.last_name).to eq('admin')
+      end
+    end
+    context 'when there is existing AdminUser' do
+      let!(:existing_admin_user) do
+        AdminUser.create(user_name: AdminUser::ADMIN_USERNAME, first_name: 'jerry', last_name: 'foo')
+      end
+      let(:returned_user) { AdminUser.find_or_create }
```
david-yz-liu, line 23: It's preferable to reuse the same object as before (`admin_user`). I don't think you need to worry about the variable name being less explicit, since the tests are short and will rely on the English descriptions to convey what they're doing.

spec/models/admin_user_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      it 'will not create a new AdminUser' do
+        expect { AdminUser.find_or_create }.not_to(change { AdminUser.count })
+      end
+      it 'will return the existing AdminUser' do
+        expect(returned_user).to eq(existing_admin_user)
+      end
+      it 'will leave first_name and last_name unchanged' do
```
david-yz-liu, line 30: same comment as above about splitting

spec/models/admin_user_spec.rb
```diff
@@ -0,0 +1,36 @@
+describe AdminUser do
+  describe '.find_or_create' do
+    let(:admin_user) { AdminUser.find_or_create }
+    it 'returns an instance of AdminUser' do
+      expect(admin_user).to be_an_instance_of(AdminUser)
+    end
+    it 'ensures the returned AdminUser api_key is not null' do
```
david-yz-liu, line 7: Oops I missed this, but please also change "null" to "nil".

david-yz-liu (approved):

PR #5778, opened 2022-01-25
v1.14.1: move wiki documentation to this repo

mishaschwartz:
This PR moves the wiki documentation from this repositories wiki into the `doc/wiki` folder. 

We are doing this so that we can better handle version specific wiki docs.

This also bumps the version to v1.14.1


mishaschwartz:
Forget this... I don't want to add all this to the repo. It will make cloning the repo problematic because it will grab all the docs (and associated images) as well.

PR #5784, opened 2022-01-27
Add test for Assignment#get_valid_num (on feature branch)

Sophieeeeeeeee:
## Motivation and Context
Added tests for uncovered method get_valid_num of model Assignment in file /spec/models/assignment_spec.rb

## Your Changes

**Description**:
Added codes that test if the method returns the correct number of valid groups of the assignment in different scenarios such as different status of student_membership in a group. 

- [ ] Other (please specify):

## Testing

Ran test file assignment_spec.rb for class Assignment which is where the newly added codes are. All tests passed with no failure and a light increase (0.06%) in test coverage for the model.

## Checklist

### DIFF
--- a/spec/models/assignment_spec.rb
+++ b/spec/models/assignment_spec.rb
@@ -2228,36 +2228,34 @@ describe Assignment do
   end
 
   describe '#get_num_valid' do
-    before :each do
-      @assignment = create(:assignment, assignment_properties_attributes: { group_min: 2, group_max: 3 })
-      @groupings = Array.new(4) { create(:grouping, assignment: @assignment) }
-    end
+    let(:assignment) { create(:assignment, assignment_properties_attributes: { group_min: 2, group_max: 3 }) }
+    let!(:groupings) { create_list :grouping, 4, assignment: assignment }
 
-    context 'When two groups meet min size requirement' do
-      before :each do
-        create(:accepted_student_membership, grouping: @groupings.first)
-        create(:inviter_student_membership, grouping: @groupings.first)
-        create(:student_membership, grouping: @groupings.second)
-        create(:student_membership, grouping: @groupings.second)
+    context 'when two groups meet min size requirement' do
+      before do
+        create(:accepted_student_membership, grouping: groupings.first)
+        create(:inviter_student_membership, grouping: groupings.first)
+        create(:student_membership, grouping: groupings.second)
+        create(:student_membership, grouping: groupings.second)
       end
-      it 'should return # of valid groups that meet size requirement' do
-        expect(@assignment.get_num_valid).to eq(2)
+      it 'returns # of valid groups that meet size requirement' do
+        expect(assignment.get_num_valid).to eq(2)
       end
     end
-    context 'When one group meets min size requirement and another is instructor approved' do
-      before :each do
-        create(:accepted_student_membership, grouping: @groupings.first)
-        create(:inviter_student_membership, grouping: @groupings.first)
-        @groupings.second.update_attribute(:instructor_approved, true)
+    context 'when one group meets min size requirement and another is instructor approved' do
+      before do
+        create(:accepted_student_membership, grouping: groupings.first)
+        create(:inviter_student_membership, grouping: groupings.first)
+        groupings.second.update(instructor_approved: true)
       end
-      it 'should return # of valid groups that meet size requirement or instructor approved' do
-        expect(@assignment.get_num_valid).to eq(2)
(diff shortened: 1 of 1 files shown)


spec/models/assignment_spec.rb
```diff
@@ -197,7 +197,7 @@
 
           it 'shows the criteria visible to tas only' do
             expect(@assignment.ta_criteria.ids).to match_array(@ta_criteria.map(&:id) +
-                                                               @ta_and_peer_criteria.map(&:id))
+                                                                 @ta_and_peer_criteria.map(&:id))
```
david-yz-liu, line 200: please revert this change

spec/models/assignment_spec.rb
```diff
@@ -2227,6 +2227,42 @@ def grouping_count(groupings)
     end
   end
 
+  describe '#get_num_valid' do
+    before :each do
+      @assignment = create(:assignment, assignment_properties_attributes: { group_min: 2, group_max: 3 })
```
david-yz-liu, line 2232: So for new tests, we prefer using `let`/`let!` instead of instance variables. You can change this to something like `let(:assignment) { create(:assignment, ...) }`, and then the variable `assignment` will be available for use in the tests.

You also do *not* need the `before :each` block when using `let`/`let!`.

spec/models/assignment_spec.rb
```diff
@@ -2227,6 +2227,42 @@ def grouping_count(groupings)
...
+      @groupings = Array.new(4) { create(:grouping, assignment: @assignment) }
```
david-yz-liu, line 2233: Use `let!` for this one---see above comment, and also `let!` will ensure all these groupings get created before each test.

Also, instead of manually creating an array, use the [create_list](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#building-or-creating-multiple-records) FactoryBot helper.

spec/models/assignment_spec.rb
```diff
@@ -2227,6 +2227,42 @@ def grouping_count(groupings)
...
+    end
+
+    context 'When two groups meet min size requirement' do
+      before :each do
```
david-yz-liu, line 2237: In this case it's okay to use `before :each`, but remove the `:each` since that's the default. (fix in cases below as well)

spec/models/assignment_spec.rb
```diff
@@ -2227,6 +2227,42 @@ def grouping_count(groupings)
...
+        create(:accepted_student_membership, grouping: @groupings.first)
+        create(:inviter_student_membership, grouping: @groupings.first)
+        create(:student_membership, grouping: @groupings.second)
+        create(:student_membership, grouping: @groupings.second)
+      end
+      it 'should return # of valid groups that meet size requirement' do
```
david-yz-liu, line 2243: replace "should return" with "returns" (fix in cases below as well)

spec/models/assignment_spec.rb
```diff
@@ -2227,6 +2227,42 @@ def grouping_count(groupings)
     end
   end
 
+  describe '#get_num_valid' do
+    before :each do
+      @assignment = create(:assignment, assignment_properties_attributes: { group_min: 2, group_max: 3 })
+      @groupings = Array.new(4) { create(:grouping, assignment: @assignment) }
+    end
+
+    context 'When two groups meet min size requirement' do
```
david-yz-liu, line 2236: start these labels with lowercase letters rather than uppercase (this is just a style convention, and is useful when nesting `context` blocks)

(fix in cases below as well)

spec/models/assignment_spec.rb
```diff
@@ -2227,6 +2227,42 @@ def grouping_count(groupings)
...
+      it 'should return # of valid groups that meet size requirement' do
+        expect(@assignment.get_num_valid).to eq(2)
+      end
+    end
+
+    context 'When one group meets min size requirement and another is instructor approved' do
+      before :each do
+        create(:accepted_student_membership, grouping: @groupings.first)
+        create(:inviter_student_membership, grouping: @groupings.first)
+        @groupings.second.update_attribute(:instructor_approved, true)
```
david-yz-liu, line 2252: Use [`update`](https://guides.rubyonrails.org/active_record_basics.html#update) instead of `update_attribute`.

david-yz-liu (commented):
Thanks, @Sophieeeeeeeee! I left a couple of comments.

spec/models/assignment_spec.rb
david-yz-liu: @Sophieeeeeeeee I think you might have missed this comment

PR #5788, opened 2022-01-28
Added Hyperlink Button to URL preview

anoverga17:
## Motivation and Context

This pull request adds an important add on to the url uploader, that is the ability to see links uploaded from a markusurl file and be able to visit the link in a new tab. To do this, minor changes were also made to the layout of the `url_viewer`

## Your Changes

**Description**:
- Added a bar that shows a clickable text of the entire URL. Clicking the link opens it in a new page.   
- Reworked layout of `url_viewer` to be adjustable according to the length of the url.
- Fixed minor bug that did not reset url modal content after submitting
- Added error messages for unsupported and/or broken URLs

## Testing

Manual testing of the feature was done to ensure no additional features were broken by the `url_viewer` rework. Several types of links were uploaded and viewed to ensure that the the proper messages were being displayed and the appropriate link was shown. 

## Checklist

### DIFF
--- a/app/assets/javascripts/Components/Modals/submission_url_submit_modal.jsx
+++ b/app/assets/javascripts/Components/Modals/submission_url_submit_modal.jsx
@@ -16,11 +16,11 @@ class SubmitUrlUploadModal extends React.Component {
 
   onSubmit = event => {
     event.preventDefault();
+    this.props.onSubmit(this.state.newUrl, this.state.newUrlText);
     this.setState({
       newUrl: "",
       newUrlText: "",
     });
-    this.props.onSubmit(this.state.newUrl, this.state.newUrlText);
   };
 
   handleUrlChange = event => {
--- a/app/assets/javascripts/Components/Modals/submission_url_submit_modal.jsx
+++ b/app/assets/javascripts/Components/Modals/submission_url_submit_modal.jsx
@@ -16,6 +16,10 @@ class SubmitUrlUploadModal extends React.Component {
 
   onSubmit = event => {
     event.preventDefault();
+    this.setState({
+      newUrl: "",
+      newUrlText: "",
+    });
     this.props.onSubmit(this.state.newUrl, this.state.newUrlText);
   };
 
--- a/app/assets/javascripts/Components/Result/url_viewer.jsx
+++ b/app/assets/javascripts/Components/Result/url_viewer.jsx
@@ -5,6 +5,7 @@ export class URLViewer extends React.Component {
     super(props);
     this.state = {
       url: "",
+      isInvalidUrl: false
     };
   }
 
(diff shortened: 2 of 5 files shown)


app/assets/javascripts/Components/Modals/submission_url_submit_modal.jsx
```diff
@@ -16,6 +16,10 @@ class SubmitUrlUploadModal extends React.Component {
 
   onSubmit = event => {
     event.preventDefault();
+    this.setState({
```
mishaschwartz, line 19: `setState` is asynchronous meaning it _might_ update the state after the `this.props.onSubmit` is triggered but then again it might update it before, and then you'll end up passing empty strings as arguments to that function.

app/assets/javascripts/Components/Result/url_viewer.jsx
```diff
@@ -5,15 +5,16 @@ export class URLViewer extends React.Component {
...
   }
 
   componentDidMount() {
     this.configDisplay();
   }
 
-  componentDidUpdate(prevProps) {
-    if (prevProps.externalUrl !== this.props.externalUrl) {
+  componentDidUpdate(prevProps, prevState) {
+    if (prevProps !== this.props) {
```
mishaschwartz, line 17: Javascript objects are not equal as long as they don't refer to the same object in memory. This means that this if statement may be true even if the contents of `prevProps` and `this.props` are identical. If you're doing this comparison, it is safer to compare the content of the props objects, not the objects themselves.

app/assets/javascripts/Components/Result/url_viewer.jsx
```diff
@@ -5,15 +5,16 @@ export class URLViewer extends React.Component {
...
     };
   }
 
   componentDidMount() {
     this.configDisplay();
   }
 
-  componentDidUpdate(prevProps) {
-    if (prevProps.externalUrl !== this.props.externalUrl) {
+  componentDidUpdate(prevProps, prevState) {
```
mishaschwartz, line 16: don't include prevState as an argument if the function doesn't use it.

app/assets/javascripts/Components/Result/url_viewer.jsx
```diff
@@ -5,15 +5,16 @@ export class URLViewer extends React.Component {
     super(props);
     this.state = {
       url: "",
```
mishaschwartz, line 7: Do we need to store the url in state anymore? Can't we just use the `this.props.externalUrl` value instead?

config/locales/views/results/en.yml
```diff
@@ -5,7 +5,7 @@ en:
       across_all_submission_files: Click on the filename to jump to the annotation.
       common: Common Annotations
       include_in_download: Include Annotations
-      include_in_download_warning: 'Warning: Annotations on PDFs are not currently downloaded.'
+      include_in_download_warning: 'Warning: Annotations on PDFs and URLs are not currently downloaded.'
```
mishaschwartz, line 8: This isn't necessary here. We haven't enabled any sort of annotations on URLs yet so this warning is misleading. Please revert this change.

mishaschwartz (changes requested):

app/assets/javascripts/Components/Result/url_viewer.jsx
anoverga17: We can. This was a design decision though since `this.props.externalUrl` can be a string that says the file cannot be loaded. I thought therefore it was cleaner to store urls specifically in the state since other ways I thought of of flagging an unreadable URL wasn't as readable.

mishaschwartz (approved):

PR #5791, opened 2022-01-31
Add Option to Enable/Disable URL Uploading

anoverga17:
This pull request gives instructors the option to toggle whether students can submit URLs for an assignment or not. 

## Motivation and Context

This gives instructors the freedom to choose whether URLs are allowed to be submitted or not. URLs require additional work on behalf of the instructor to verify if the content was handed in on time and thus, is not something instructors would always want. 
This makes it so that the new URL upload feature is entirely optional. 

## Your Changes

**Description**:
- Added a `url_submit` property to an assignment that indicates if the assignment should accept URLs or not
- Disallowed backend creation of URL files depending on if the assignment disallows URL uploading
- Added toggle button to enable/disable an assignment's ability to upload a URL
- Changed backend so that the file viewer only receives and loads a `markusurl` file iff URL uploading is enabled
- Passed in url submit enabled prop to frontend to enable/disable URL submit button
- Added rspec tests for the changes

## Testing

Manual testing of the feature was done. Assignments were created with and without the `url_submit` property enabled. Various `.markusurl` files were then uploaded to each type of assignment to see if the views changed appropriately

## Checklist

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -432,10 +432,8 @@ class SubmissionsController < ApplicationController
           file_contents = repo.download_as_string(raw_file)
           file_contents.encode!('UTF-8', invalid: :replace, undef: :replace, replace: '�')
 
-          if file_type == 'markusurl' && assignment.url_submit
-            file_contents = extract_url(file_contents)
-          elsif file_type == 'markusurl' && !assignment.url_submit
-            file_type = 'unknown'
+          if file_type == 'markusurl'
+            assignment.url_submit ? file_contents = extract_url(file_contents) : file_type = 'unknown'
           end
 
           if params[:force_text] != 'true' && SubmissionFile.is_binary?(file_contents)
@@ -803,7 +801,7 @@ class SubmissionsController < ApplicationController
     return [] unless revision.path_exists?(full_path)
 
     anonymize = current_role.ta? && grouping.assignment.anonymize_groups
-    url_allowed = grouping.assignment.url_submit
+    url_submit = grouping.assignment.url_submit
 
     entries = revision.tree_at_path(full_path).sort do |a, b|
       a[0].count(File::SEPARATOR) <=> b[0].count(File::SEPARATOR) # less nested first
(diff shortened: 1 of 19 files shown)


app/views/assignments/_form.html.erb
```diff
@@ -202,6 +202,11 @@
                          Assignment.human_attribute_name(:vcs_submit) %>
           <%= ap_f.check_box :vcs_submit %>
         <% end %>
+        <% unless @assignment.scanned_exam || @assignment.is_timed %>
```
mishaschwartz, line 205: This should be allowed for timed assessments

db/migrate/20220128182721_add_url_submit_to_assignment_properties.rb
```diff
@@ -0,0 +1,5 @@
+class AddUrlSubmitToAssignmentProperties < ActiveRecord::Migration[6.1]
+  def change
+    add_column :assignment_properties, :url_submit, :boolean, default: false
```
mishaschwartz, line 3: I always forget what the default is here but we don't want to allow nil values so add a `null: false` if the default is that it allows null values (or just add it to be safe if you're not sure)

app/controllers/submissions_controller.rb
```diff
@@ -431,8 +432,10 @@ def get_file
           file_contents = repo.download_as_string(raw_file)
           file_contents.encode!('UTF-8', invalid: :replace, undef: :replace, replace: '�')
 
-          if file_type == 'markusurl'
+          if file_type == 'markusurl' && assignment.url_submit
```
mishaschwartz, line 435: This can be simplified:

```ruby
if file_type == 'markusurl'
  file_contents = assignment.url_submit ? extract_url(file_contents) : 'unknown'
end
```

app/controllers/submissions_controller.rb
```diff
@@ -800,6 +803,7 @@ def get_all_file_data(revision, grouping, path)
     return [] unless revision.path_exists?(full_path)
 
     anonymize = current_role.ta? && grouping.assignment.anonymize_groups
+    url_allowed = grouping.assignment.url_submit
```
mishaschwartz, line 806: This is nitpicky but....

You have 3 variables that are named differently but all have the same semantics: `url_allowed`, `url_enabled`, `url_submit`. Pick one and stick with it please

spec/controllers/submissions_controller_spec.rb
```diff
@@ -123,6 +123,9 @@
 
     context 'submitting a url' do
       describe 'should add url files' do
+        before :each do
+          @assignment.update(url_submit: true)
```
mishaschwartz, line 127: make all of these `update!` so that it doesn't fail silently if there is an issue setting the value

mishaschwartz (changes requested):

db/migrate/20220128182721_add_url_submit_to_assignment_properties.rb
anoverga17: the default here is false (i.e. url submissions are disabled). I will still add the null: false just to be safe though

mishaschwartz (approved):

Issue #5607, opened 2021-11-17
Navigation buttons not working on mobile


DemetreJou:
On mobile or when desktop window is small the top left corner nav menu button doesn't work, neither does the MarkUs logo.
Mouse does change when hovering the nav burger button even if clicking doesn't do anything
No network requests are initiated as far as I can tell.
Clicking on other things such as assignments, results, submission files, annotations, marks, summary all work as expected
Have checked on both desktop and mobile firefox and without extensions/addons.
Let me know if I can provide more details

PR #5818, opened 2022-02-03
Issue 5607

TheTallJerry:
## Motivation and Context

Resolves https://github.com/MarkUsProject/Markus/issues/5607 issue of mobile menu/navigation bar not working. 

## Your Changes

**Description**:
The two files concerned by this commit both had an assignment to window.onload. This caused one of them to override the other. The fix is to add to window.onload instead of assigning. 

## Testing

I have performed manual testing using the latest version of google chrome on icognito mode (i.e. with no site data/cache). I used google developer tools to switch to several mobile devices, as well as resizing the desktop browser window to hit the size limit of the appearance of the navgation bar. In both case I was able to open/collapse the menu and use its menu items. There were no visible UI discrepancies. 

## Checklist

### DIFF
--- a/app/assets/javascripts/menu.js
+++ b/app/assets/javascripts/menu.js
@@ -32,6 +32,6 @@ function initMenu() {
     }
   };
 }
-window.addEventListener("onload", () => {
+window.addEventListener("DOMContentLoaded", () => {
   initMenu();
 });
--- a/app/assets/javascripts/menu.js
+++ b/app/assets/javascripts/menu.js
@@ -32,6 +32,8 @@ function initMenu() {
     }
   };
 }
+// using addEventListener as opposed to direct assignment so that event listeners added elsewhere
+// don't get overridden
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/menu.js
```diff
@@ -32,7 +32,6 @@ function initMenu() {
     }
   };
 }
-
-window.onload = () => {
+window.addEventListener("onload", () => {
```
david-yz-liu, line 35: Let's use the `"DOMContentLoaded"` event here and below instead, since we just need the DOM to be loaded. (https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event)

david-yz-liu (commented):
@TheTallJerry thank you! In addition to the inline comment, please also add to the Changelog to record this change.

david-yz-liu (approved):

PR #5833, opened 2022-02-06
Switch from webpacker to rails-jsbundling and upgrade to webpack v5

david-yz-liu:
## Motivation and Context

Rails 7 (the latest version) has a [new default way](https://world.hey.com/dhh/rails-7-will-have-three-great-answers-to-javascript-in-2021-8d68191b) of bundling Javascript files using jsbundling-rails (but also still using webpack). Currently, we use webpacker, but the latest version does not support webpack 5+, which has delayed our migration from webpack v4 to v5 for a while now. 

So by switching to jsbundling-rails, we can stick with Rails defaults (generally a good thing), and upgrade webpack to v5. (Another option would have been to switch to a friendly fork of webpacker called [shakalacker](https://github.com/shakacode/shakapacker), which would also allow the upgrade.)

## Your Changes

**Description**:

Essentially, followed two migration guides: (1) from [webpacker to rails-jsbundling](https://github.com/rails/jsbundling-rails/blob/main/docs/switch_from_webpacker.md) and (2) from [webpack v4 to v5](https://webpack.js.org/migrate/5/).
Major changes:

- `webpacker` gem has been replaced by `jsbundling-rails`
  - webpack outputs are now placed in `app/assets/builds` and then included as part of the Rails asset pipeline
  - Added two new yarn commands to `package.json`, `yarn build` and `yarn build-dev`, which respectively build assets with webpacker for production and development. `yarn build` is run by the Rails task `javascript:build`, and included in the `assets:precompile` task used in production.
  - Removed `webpack-dev-server`: `jsbundling-rails` recommends running `webpack` with `--watch` to automatically recompile when source files change
  - Because `webpacker` was removed, we now need explicit requires for `open3` and `net/ssh` (these were previously required by `webpacker` and made available in other parts of our code).
- Upgraded `webpack` to v5, and upgraded plugins as well (some were locked on previous versions that were required for webpack v4)
  - Also moved webpack and babel packages from `dependencies` to `devDependencies` in `package.json`.
  - Webpack configuration is now found in `webpack.[common|production|development].yml`.
  - Removed some unnecessary configuration in `babel.config.yml` (which had been automatically populated by webpacker on a first install)

## Testing

Ensured the assets build properly with both `yarn build` and `yarn build-dev`; also tested that the assets are correctly served in the local development environment using Docker.

I've left the `webpacker` docker-compose entry for now, even though I don't know that we need to keep it.

## Checklist


pretendWhale:
[pr review webpack](https://trello.com/c/cym8k9Bt/2-pr-review-webpack)

pretendWhale (approved):
@david-yz-liu Looks good! The webpacker docker container is still needed because webpack doesn't run anywhere else. If we wanted to just have it run in the background of the rails container we could add `yarn build-dev &` to `.dockerfiles/entrypoint-dev-rails.sh`. I see you updated a line in `bin/markus` to do this, but as far as I can tell that file is legacy (or at least isn't actively used in the docker setup). At that point the webpacker container would be redundant. Since this works as is and only concerns development, I think it could be a separate PR if we decide we want to do that.

david-yz-liu:
@pretendWhale you're totally right, I'm hoping to remove the webpacker docker container in a subsequent PR!

PR #5834, opened 2022-02-06
Remove assignments properties table unused column

therealfake:
## Motivation and Context

Removed the column group_name_displayed in the assignments properties table because this attribute isn't actually used anywhere in the code.

## Your Changes

**Description**:
- Created and ran database migration to remove the column group_name_displayed in the assignments properties table. 
- Removed (all but 1, see Questions) uses of group_name_displayed in the code.

- I think the deletion of [this](https://github.com/MarkUsProject/Markus/blob/master/app/views/assignments/show.html.erb#L130-L136) usage of group_name_displayed should just result in [line 132](https://github.com/MarkUsProject/Markus/blob/master/app/views/assignments/show.html.erb#L132) being inside the h3 tags. However, I am not confident in this solution.
- When should a change be added to the Changelog.md file?

## Checklist


app/views/assignments/show.html.erb
```diff
@@ -128,11 +128,7 @@
         <%# Since the student has a group already, we display information about the group only %>
         <%# Check if groupname should be displayed %>
```
david-yz-liu, line 129: You can delete this comment as well, since it no longer applies

db/migrate/20220206065135_remove_group_name_displayed_from_assignment_properties.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveGroupNameDisplayedFromAssignmentProperties < ActiveRecord::Migration[6.1]
+  def change
+    remove_column :assignment_properties, :group_name_displayed, :boolean
```
david-yz-liu, line 3: Please add `default: false` and `null: false` as arguments here, so that if the migration is rolled back, the column has the same properties as before.

david-yz-liu (commented):
@therealfake good work! I left just a few small comments.

Regarding your question about the changelog: we typically only update it for user-facing changes. This case is a bit of a grey area because the setting didn't show up in the UI, but does show up when exporting the settings for an assignment. So you can add an entry to the changelog in this case.

app/views/assignments/show.html.erb
therealfake: Done!

db/migrate/20220206065135_remove_group_name_displayed_from_assignment_properties.rb
therealfake: Done!

therealfake:
> @therealfake good work! I left just a few small comments.
> 
> Regarding your question about the changelog: we typically only update it for user-facing changes. This case is a bit of a grey area because the setting didn't show up in the UI, but does show up when exporting the settings for an assignment. So you can add an entry to the changelog in this case.

david-yz-liu (approved):

PR #5839, opened 2022-02-11
Redesigned Workflow for Scanned Exams

### LINKED ISSUE #5613: UI change for scanned assessment divisions
This is a feature request (and to some extent, a bug report) for the UI used to set up divisions for scanned assessments.

By default, MarkUs assigns page 1 to an automatic division named COVER, except that currently, it's possible for users to include page 1 in one of their own division — with the result that the automatic COVER is empty, and the interface for assigning copies is rendered useless.

Suggestion: show explicitly the default COVER division, set to page 1 and greyed out so that it cannot be changed. And if possible, make it impossible for users to include page 1 in any of their own divisions. This would make it clear that page 1 is **already** handled by the process, and remove the possibility of users mistakenly including it in their own custom divisions.

### LINKED ISSUE #5829: Simplify/improve scanned exam workflow
Right now the "Exam Template" tab is not that intuitive to work with. This problem doesn't have an immediate/obvious fix so solving this issue will require some planning/discussion with users.

Some things to keep in mind:

- there are multiple forms on this page and it is not entirely obvious which order to complete them in and what each of them do
  - we should really be only showing one form per view
- commands like generate and split should be explained better or put somewhere else so that their purpose is clearer
- interactions between multiple exam templates should be obvious from the layout
- it should be more obvious how to assign scans to groups once they've been uploaded (right now you have to go to the graders tab)
- other things! Make it prettier I guess

due date: to be completed by February 28th 2022

anoverga17:
This pull request makes the required UI and controller changes to make the scanned exam workflow more intuitive. 

## Motivation and Context

Currently, the exam template feature is not very intuitive and is confusing to use without the aid of documentation. This pull request aims to make the exam template view more akin to criteria/annotation views and more noticeably show how features relate to each other. This pull request also adds a simple help page for simpler of documentation. 

This pull request aims to resolve #5829
As part of this pull request, UI and minor backend changes were also made to prevent any errors with creating template divisions and thus also resolves #5613

## Your Changes

**Description**:
- Redesigned scanned exam view to look more like criteria/annotation views
- Moved all scanned exam related views (except those related to error fixing) to main scanned exam page
- Reworked split scanned exam and the pdf log to be grouped together and better relate to all scanned exams as opposed to any particular one. 
- Added extra step for create modal to specify template divisions
- Moved generate exam into it's own modal
- Added a help home page that explains the more confusing parts of the scanned exam feature.
- Changed Template Division Model and UI to only accept start/end pages greater than or equal to 2.

- [x] View Update

## Testing

Manual testing of the pull request was done to ensure all the appropriate views work and are rendered properly. 

## Checklist

https://github.com/MarkUsProject/Wiki/pull/95


anoverga17:
Let me know if tests need updating since what we have is pretty bare bones.

app/views/exam_templates/_create_exam_template_info.html.erb
```diff
@@ -0,0 +1,12 @@
+<div id="create-template-info-modal-content">
+  <div class="inline-labels">
+    <%= f.label :name, ExamTemplate.human_attribute_name('name'), class: "required" %>
+    <%= f.text_field :name, required: true %>
+    <%= f.label :file_io, t('exam_templates.create.upload'), class: "required" %>
+    <%= f.file_field :file_io, required: true %>
+  </div>
+  <section class='dialog-actions'>
+    <input id="create-template-info-next" class='create-template-view-change' type='button' value='Next'/>
```
david-yz-liu, line 9: After playing around with this, I don't think we need the template divisions in the modal. (I didn't realize this until using it myself, and seeing that there's not much of a delay between saving the exam template and then seeing the "edit" pane appear, where we can configure template divisions and other settings).

Let's stick with the modal just having the exam template file, and then after the user saves that, they do other configuration in the pane, not in the modal. This will also simplify the code by removing a bunch of the Javascript/HTML you added (sorry I want you to now remove it).

app/controllers/exam_templates_controller.rb
```diff
@@ -31,14 +31,29 @@ def create
                                                     filename: filename,
                                                     name: name)
       if exam_template&.valid?
+        exam_template.update(exam_template_params) unless params[:exam_template].nil?
```
david-yz-liu, line 34: You can probably revert this change (see my comments about the modal below).

app/views/exam_templates/index.html.erb
```diff
@@ -16,38 +16,29 @@
...
-      <h3><%= t('exam_templates.generate.title') %></h3>
-      <%= render partial: 'generate_form',
-                 locals: { exam_template: exam_template } %>
-    </div>
-    <hr>
-    <div class="split-section">
-      <h3><%= t('exam_templates.split.title') %></h3>
-      <%= render partial: 'split_form',
-                 locals: { exam_template: exam_template } %>
+<div class='wrapper'>
```
david-yz-liu, line 25: Delete this

app/views/exam_templates/_create_new_template.html.erb
```diff
@@ -4,18 +4,9 @@
   <%= form_for :create_template,
                url: course_assignment_exam_templates_path(@current_course, @assignment),
                html: { multipart: true, size: 1 } do |f| %>
```
david-yz-liu, line 6: You can add a `remote: true` here to make this an AJAX request (which allows you to render JS instead of HTML from the controller route).

app/controllers/exam_templates_controller.rb
```diff
@@ -31,14 +31,29 @@ def create
...
+        redirect_to edit_course_exam_template_path(current_course, exam_template)
+        return
       else
         flash_message(:error, t('exam_templates.create.failure'))
       end
     end
     redirect_to course_assignment_exam_templates_path(current_course, assignment)
   end
 
+  def edit
```
david-yz-liu, line 45: You can add a simple test for this method just to check for the correct response status (for both format types).

app/views/exam_templates/_create_new_template.html.erb
anoverga17: I don't think that is required here. I need to refresh the page to update the list of templates on the side panel. This doesn't happen when I set `remote: true`

david-yz-liu (commented):
@anoverga17 okay looking good! I left another round of feedback. In addition, there are four small additional UI improvements to the exam template form I'd like you to make while you are at it:

1. Remove the horizontal line/border between "Number of pages" and "Template divisions". These two sections are part of the same form, with the same `Save` button, so it's misleading to separate them so dramatically.
2. Remove the number of pages as a field, and instead put it beside the filename, e.g. `midterm1-v2-test.pdf (6 pages)`. Keep the link only on the filename.
3. In the template division table, make the table headers "Label", "Start", and "End" all be required (with the asterisk). You might need to look into the CSS styles to see if we have something to apply to a non-label field, or you might need to make your own style rule.
4. (issue 5613) Make the `min` attribute of every "Start" and "End" field `2`. Add a note to the documentation explaining that the minimum number for the template division is 2, because the first page is reserved as a cover page. Please also modify the PR description to specifically mention this change, and use the GitHub magic text to indicate that this PR fixes issue 5613.

app/views/exam_templates/_create_new_template.html.erb
david-yz-liu: @anoverga17 Yeah, so it is possible to refresh both lists (we do this for criteria/annotations), but it doesn't need to be in this PR. We can go with what you have!

app/views/exam_templates/_template_division.html.erb
```diff
@@ -5,11 +5,11 @@
                         'aria-label': TemplateDivision.human_attribute_name('label') %>
   </td>
   <td>
-    <%= form.number_field :start, min: 1, disabled: disabled,
+    <%= form.number_field :start, min: 2, disabled: disabled,
```
david-yz-liu, line 8: This is good, but you should also modify the `add_template_division` function in `_boot.js` to add the `min` attribute (looks like it wasn't specified at all before).

app/views/exam_templates/_template_division_pane.html.erb
```diff
@@ -3,15 +3,17 @@
...
+        <th class='required'><%= TemplateDivision.human_attribute_name('label') %></th>
+        <th class='required'><%= TemplateDivision.human_attribute_name('start') %></th>
+        <th class='required'><%= TemplateDivision.human_attribute_name('end') %></th>
         <th><%= t(:delete) %></th>
       </tr>
     </thead>
     <tbody>
-    <%= render partial: 'exam_templates/template_division',
-               locals: { f: f, disabled: disabled } %>
+    <% unless exam_template.nil? %>
```
david-yz-liu, line 13: Similar to what I noted previous, I don't think `exam_template` should ever be nil when this partial is rendered?

config/locales/models/exam_templates/en.yml
```diff
@@ -6,7 +6,7 @@ en:
         cover_fields: Field to match by
         filename: Filename
         name: Name
-        num_pages: Number of pages
+        num_pages: Pages
```
david-yz-liu, line 9: use lowercase here

david-yz-liu (approved):

PR #5844, opened 2022-02-13
Replace uses of first name/last name with display name

therealfake:
## Motivation and Context

In the instructor's dashboard, for each assignment there is a grade distribution for each grader. The legend for the chart show's each grader's name and currently uses the grader's ```first_name``` and ```last_name``` attributes. This can be an issue for international students, if they have a preferred English, or if someone that identifies as transgender has a preferred name that is not their legal first and last name in the system. It is important that we respect the name they wish to be called by. Thus this pull request not only changes the name used in the chart to the ```display_name``` attribute, but also cases which this change can apply.

## Your Changes

**Description**:
Change uses of first name/last name with display name when applicable. Such as changing name used in grade distribution chart for graders from using ```first_name``` and ```last_name``` attribute to ```display_name``` attribute.

How can I test a change to the front-end like this?

## Checklist


therealfake (commented):
The previous order in the [student model](https://github.com/MarkUsProject/Markus/pull/5844/commits/dff054f42186ef3a690b39063785dc2e7a7923ed#diff-84bdcad805a596abdda1c90bce6e08ce4341e2b0da260a5cb0240691c88ebfd6L89-R89) last name, first name. Would I need to split display name, have last name first, and then add a comma in between?

david-yz-liu:
@therealfake great question. Please switch the text for that method to the display name. (This also improves the consistency of the name displays, which is great!)

app/controllers/groups_controller.rb
```diff
@@ -136,7 +136,7 @@ def assign_scans
       return
     end
     names = next_grouping.non_rejected_student_memberships.map do |u|
-      "#{u.user.first_name} #{u.user.last_name}"
+      u.user.display_name.to_s
```
david-yz-liu, line 139: You don't need the `to_s` here, since the display name is already a string. (This applies for all changes below.)

david-yz-liu (commented):
@therealfake great work! I just left one comment here.

Changelog.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [unreleased]
+- Change name used in assignment grade distribution to grader's display name attribute (#5844)
```
david-yz-liu, line 4: @therealfake I keep forgetting to mention this, but please add entries to the bottom of the section (trying to keep the entries in numerical order)

david-yz-liu (approved):

PR #5853, opened 2022-02-17
Role switch test

TheTallJerry:
## Motivation and Context

- Role switching returns status 200 regardless of whether the switch was successful. 
- Role switching's error messages could use some improvement. 

## Your Changes

**Description**: 
- Modified role switching so that request status is 404 when trying to role switch to an invalid user, and 422 when trying to role switch to either the current instructor or another instructor. 
- Modified role switching so that the error messages correspond to the correct use case. 
- Added better error message for when trying to role switch to instructors. 

## Testing

- Written tests covering the entire role switching process. 
- Covered switching as invalid user, current instructor, another instructor, a student/TA. 
- Tested for the session hash and other applicable changes. 

Please help verify that the test breakdown and the added error messages make sense. 

## Checklist

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -31,7 +31,6 @@ class CoursesController < ApplicationController
 
   # Sets current_user to nil, which clears a role switch session (see role_switch)
   def clear_role_switch_session
-    puts "Instructor '#{session[:real_user_name]}' logged out from '#{session[:user_name]}'."
     MarkusLogger.instance.log("Instructor '#{session[:real_user_name]}' logged out from '#{session[:user_name]}'.")
     session[:user_name] = nil
     session[:role_switch_course_id] = nil
@@ -42,10 +41,9 @@ class CoursesController < ApplicationController
   # the perspective of another (non-instructor) user.
   def switch_role
     if params[:effective_user_login].blank?
-      session[:error] = I18n.t('main.username_not_blank')
       render partial: 'role_switch_handler',
              formats: [:js], handlers: [:erb],
-             locals: { error: session[:error] }
+             locals: { error: I18n.t('main.username_not_blank') }
       return
     end
 
@@ -53,30 +51,24 @@ class CoursesController < ApplicationController
     found_role = Role.find_by(end_user: found_user, course: current_course)
 
     if found_role.nil?
-      session[:error] = I18n.t('main.login_failed')
       render partial: 'role_switch_handler',
              formats: [:js], handlers: [:erb],
-             locals: { error: session[:error] }
+             locals: { error: I18n.t('main.login_failed') }
       return
     end
     if found_user.user_name == session[:user_name] || found_user.user_name == session[:real_user_name]
-      # error
-      session[:error] = I18n.t('main.cannot_role_switch_to_self')
       render partial: 'role_switch_handler',
              formats: [:js], handlers: [:erb],
-             # TODO: put better error message
-             locals: { error: session[:error] }
+             locals: { error: I18n.t('main.cannot_role_switch_to_self') }
       return
     end
     if found_role.instructor?
-      session[:error] = I18n.t('main.cannot_role_switch_to_instructor')
       render partial: 'role_switch_handler',
              formats: [:js], handlers: [:erb],
-             locals: { error: session[:error] }
+             locals: { error: I18n.t('main.cannot_role_switch_to_instructor') }
       return
     end
 
-    # TODO: this block won't be reached, because current instructor's role is also instructor
     # Check if an instructor trying to login as the current user or themselves
 
     log_role_switch found_user
@@ -85,12 +77,11 @@ class CoursesController < ApplicationController
 
     session[:redirect_uri] = nil
     refresh_timeout
-    session[:error] = nil
     # All good, redirect to the main page of the viewer, discard
     # role switch modal
     render partial: 'role_switch_handler',
            formats: [:js], handlers: [:erb],
-           locals: { error: session[:error] }
+           locals: { error: nil }
   end
 
   def role_switch
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -54,10 +54,11 @@ class CoursesController < ApplicationController
     if found_role.nil?
       render partial: 'role_switch_handler',
              formats: [:js], handlers: [:erb],
-             locals: { error: I18n.t('main.login_failed') },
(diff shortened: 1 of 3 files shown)


TheTallJerry:
I've run i18n-tasks normalize in my container before pushing - what else should I have done so that this i18n test wouldn't have failed?

app/controllers/courses_controller.rb
```diff
@@ -43,27 +43,35 @@ def switch_role
...
-    if found_role.nil? || found_role.instructor?
+    if found_role.nil?
       render partial: 'role_switch_handler',
              formats: [:js], handlers: [:erb],
-             locals: { error: Settings.validate_user_not_allowed_message || I18n.t('main.login_failed') }
+             locals: { error: Settings.validate_user_not_allowed_message || I18n.t('main.login_failed') },
+             status: :not_found
       return
     end
-
```
david-yz-liu, line 59: Revert this change. In general, do not make whitespace changes.

app/controllers/courses_controller.rb
```diff
@@ -43,27 +43,35 @@ def switch_role
...
-    # Check if an instructor trying to login as the current user or themselves
+    # check if the current instructor is trying to role switch as themselves
```
david-yz-liu, line 61: Capitalize comments

app/controllers/courses_controller.rb
```diff
@@ -43,27 +43,35 @@ def switch_role
...
     if found_user.user_name == session[:user_name] || found_user.user_name == session[:real_user_name]
```
david-yz-liu, line 62: Because the `role_switch` policy checks that the user is currently not role-switched, the check against `session[:user_name]` can be removed.

app/controllers/courses_controller.rb
```diff
@@ -43,27 +43,35 @@ def switch_role
...
-      # error
       render partial: 'role_switch_handler',
              formats: [:js], handlers: [:erb],
-             # TODO: put better error message
-             locals: { error: I18n.t('main.login_failed') }
+             locals: { error: I18n.t('main.cannot_role_switch_to_self') },
+             status: :unprocessable_entity
+      return
+    end
+    # otherwise, check if the current instructor is trying to role switch as other instructors
```
david-yz-liu, line 69: Insert a blank line here

spec/controllers/courses_controller_spec.rb
```diff
@@ -1,6 +1,85 @@
 describe CoursesController do
   let(:instructor) { create :instructor }
   let(:course) { instructor.course }
+
+  describe 'role switching methods' do
+    let(:end_user) { create :end_user }
```
david-yz-liu, line 6: You don't need to define this here. Instead, use a `let` within different context blocks below to define it as an instructor/TA/student. For example, something like

```ruby
context 'when switching to the current instructor' do
  let(:end_user) { instructor }
  it ...
```

Specifically, the `let` statement for `subject` can appear *before* the `let` statement for `end_user`, even if `subject` is defined using `end_user`.

spec/controllers/courses_controller_spec.rb
```diff
@@ -1,6 +1,85 @@
...
+    let(:subject) do
+      post_as instructor, :switch_role, params: { 'id' => course.id, 'effective_user_login' => end_user.user_name }
```
david-yz-liu, line 8: In general use symbol keys, not strings. So `id: course.id` instead of `'id' => course.id`, for example.

spec/controllers/courses_controller_spec.rb
```diff
@@ -1,6 +1,85 @@
...
+    end
+    describe '#switch_role' do
+      let(:second_instructor) { create :instructor }
```
david-yz-liu, line 11: Set the course association for instructors/TAs/students you create here, to be explicit that they are part of the same course.

(Unless you are testing that they aren't. In fact, you should add tests checking that `:forbidden` status is returned if trying to switch role to a role in a different course.

spec/controllers/courses_controller_spec.rb
```diff
@@ -1,6 +1,85 @@
...
+      let(:temp_user) { create :end_user }
+
+      it 'fails when no username is provided ' do
+        post_as instructor, :switch_role, params: { 'id' => course.id }
+        expect(response).to have_http_status(:not_found)
+      end
+      it 'fails when an invalid username is provided' do
+        User.delete(temp_user)
```
david-yz-liu, line 19: You can do `temp_user.destroy` here.

spec/controllers/courses_controller_spec.rb
```diff
@@ -1,6 +1,85 @@
...
+        post_as instructor, :switch_role, params: { 'id' => course.id, 'effective_user_login' => temp_user.user_name }
+        expect(response).to have_http_status(:not_found)
+      end
+      it 'fails when switching role as the current instructor' do
+        end_user.roles << instructor
```
david-yz-liu, line 24: Remove this line; see above comment about using `let` to define `end_user`

spec/controllers/courses_controller_spec.rb
```diff
@@ -1,6 +1,85 @@
...
+        expect(subject).to have_http_status(:unprocessable_entity)
+      end
+      it 'fails when switching role as another instructor in the course' do
+        end_user.roles << second_instructor
+        expect(subject).to have_http_status(:unprocessable_entity)
+      end
+      context 'when switching to a student' do
```
david-yz-liu, line 31: Also check for the response status and `session[:user_name]` here (and for the TA tests below)

spec/controllers/courses_controller_spec.rb
TheTallJerry: We currently do not have an option of returning the :forbidden or 403 status. Right now the routine is: 
- First we accquire the username. If it's blank we say username can't be blank and return a 404. 
- Then we try to find the user in the db + find their role in the course. If their role is nil we say login failed and return a 404
- Then we check whether they are logging in as themselves or another instructor in the course. Either case will be 422 but we'll give different messages
- Finally, we return 200 and do the stuff afterwards. 
I'm adding the tests for if they are switching to valid users not in the course - but let me know how you would like the status to be handled (I can either keep using 404 as their role will be nil as well, or I can modify the controller to use 403 with an extra if block).

david-yz-liu: @TheTallJerry ah yes I see, a `:not_found` is totally fine. Thanks for the explanation. You can finalize the tests to check for `:not_found` (I guess just remove the parts you commented out.)

david-yz-liu (approved):

PR #5854, opened 2022-02-17
Update criteria marks

TheTallJerry:
## Motivation and Context
For assignments' rubric criteria, if a change involves changing the criteria' levels A and B such that A's new mark = B's old mark, then the change would fail, despite being a legal change. 

## Your Changes

**Description**:
Reviewing the request on the (criteria) controller level. If the request contains unique marks for each level, then the criteria of validating the uniqueness of marks is skipped. This is done by adding an attr_accessor in the Level module, acting as a conditional for the uniqueness validation. 

## Testing

I have run the test suite. I have also done manual test in the browser with the previously mentioned changes. 

## Checklist


app/models/rubric_criterion.rb
```diff
@@ -10,6 +10,45 @@ def self.symbol
...
+        end
+        s1.add(val[:mark])
+      end
+      if should_skip_marks
+        rubric_criterion_params[:levels_attributes].each do |_key, val|
+          val[:skip_marks_validation] = true
+        end
+      end
+
+      # check if there's a dup in names
```
david-yz-liu, line 33: Merge this loop with the previous one. (A bit simpler after switching to one attribute instead of two.)

spec/models/rubric_criterion_spec.rb
```diff
@@ -16,6 +17,55 @@
     it { is_expected.to validate_presence_of(:levels) }
     it { is_expected.to have_one(:course) }
 
+    context 'Helper method #update_levels' do
```
david-yz-liu, line 20: This is actually a good time to use a `describe` block (even though the rest of this file doesn't seem to). Move this code outside of the outer context, and use `describe #update_levels` instead.

Also based on my above feedback, don't check the return values; instead, check whether the method properly updates the associated level objects or not.

david-yz-liu (approved):

PR #5855, opened 2022-02-17
add check for criteria under different assignment in #update_position in criteria_controller

Sophieeeeeeeee:
## Motivation and Context

#update_position in criteria_controller does not check if all criteria belongs to the current assignment. Implementation is added. 

## Your Changes

**Description**:
Hence transaction is added outside of the block to trigger rollback if there exists invalid criteria not under current assignment. 

## Testing

Added testing codes in criteria_controller_spec. All tests pass.

## Checklist


app/controllers/criteria_controller.rb
```diff
@@ -114,10 +114,21 @@ def update
   # Handles the drag/drop criteria sorting.
   def update_positions
     @assignment = Assignment.find(params[:assignment_id])
+    invalid_criteria = false
 
     ApplicationRecord.transaction do
-      params[:criterion].each_with_index do |id, index|
-        Criterion.update(id, position: index + 1) unless id.blank?
+      Criterion.transaction(requires_new: true) do
```
david-yz-liu, line 120: It's good that you're using a transaction here, and in fact we don't need to have a nested one. The outer transaction is a bit too broad; so you can remove this line and replace `ApplicationRecord` with `Criterion` in the previous line.

app/controllers/criteria_controller.rb
```diff
@@ -114,10 +114,21 @@ def update
...
+        params[:criterion].each_with_index do |id, index|
+          Criterion.update(id, position: index + 1) unless id.blank?
+          # check if any criteria invalid
+          if Criterion.find(id).assessment_id != @assignment.assessment_id
```
david-yz-liu, line 124: This part does work but it's a bit redundant. Instead it's simpler to modify the query on the line above to use an association query (something like `@assignment.criteria.find(id)...`)

This will raise an error if the criterion isn't found for that assignment, and then you can catch that error (using a `rescue`) and handle it using the same code you have below.

config/locales/views/criteria/en.yml
```diff
@@ -8,6 +8,7 @@ en:
     description_hint: You may use Markdown in this field. Save changes to see a preview.
     editing_criteria: Editing criteria
     errors:
+      different_assignment: 'There exists criteria under different assignment:'
```
david-yz-liu, line 11: This message should be displayed when the criterion id is for a criterion of a different assignment, *or* an id that doesn't exist in the database at all. That means the wording should be a bit different, something like "Could not find criteria for this assignment with the given ids". With this in mind, please also pick a better name for the key for this string (instead of `different_assignment`).

Also, don't end the message with a colon (the ids don't automatically display). Using no punctuation is okay, this is consistent with how we display other flash error messages.

spec/controllers/criteria_controller_spec.rb
```diff
@@ -1031,4 +1031,52 @@
...
     end
   end
+
+  context '#update_position' do
+    context 'when there exists criteria not under current assignment' do
+      let(:assignment2) { create :assignment }
+      let(:rubric_criterion) do
+        create(:rubric_criterion,
+               assignment: assignment,
+               position: 1,
```
david-yz-liu, line 1041: You don't need to specify a name for the criteria, as these will be automatically generated by the factory. In general it's good to omit specifying attributes that don't actually matter to the test case, since it keeps the test code shorter. (In fact, if you remove this attribute, each `let` statement should fit onto a single line!)

spec/controllers/criteria_controller_spec.rb
```diff
@@ -1031,4 +1031,52 @@
       let(:params) { { course_id: course.id, assignment_id: assignment.id } }
     end
   end
+
+  context '#update_position' do
```
david-yz-liu, line 1035: This should be `describe` instead of `context` (we use `describe` to introduce a test block for a particular method). Also, the method is called `update_positions`, not `update_position`.

However, I notice that there is already a test for `update_positions` up above on on line 740; please move the tests you added there. You can use the `describe` line as the outer block, and nest the existing test underneath it, together with the new ones you're adding. This way we keep all of the tests for the same method together.

app/controllers/criteria_controller.rb
```diff
@@ -115,9 +115,15 @@ def update
   def update_positions
     @assignment = Assignment.find(params[:assignment_id])
 
-    ApplicationRecord.transaction do
+    Criterion.transaction(requires_new: true) do
```
david-yz-liu, line 118: You can remove the `requires_new: true`

spec/controllers/criteria_controller_spec.rb
```diff
@@ -1031,4 +1074,43 @@
       let(:params) { { course_id: course.id, assignment_id: assignment.id } }
     end
   end
+
+  describe '#update_positions' do
```
david-yz-liu, line 1078: I think you meant to move these tests

spec/controllers/criteria_controller_spec.rb
```diff
@@ -720,7 +720,9 @@
         before(:each) do
           post_as instructor,
                   :edit,
-                  params: { course_id: course.id, id: rubric_criterion.id },
+                  params: {
```
david-yz-liu, line 723: Revert this change

app/controllers/criteria_controller.rb
```diff
@@ -115,9 +115,15 @@ def update
...
       params[:criterion].each_with_index do |id, index|
-        Criterion.update(id, position: index + 1) unless id.blank?
+        begin
```
david-yz-liu, line 120: You don't need a `begin` here. Instead, you can attach the `rescue StandardError` to the end of the transaction `do` block, like here:

https://github.com/MarkUsProject/Markus/blob/master/app/controllers/criteria_controller.rb#L151-L157

app/controllers/criteria_controller.rb
```diff
@@ -115,9 +115,15 @@ def update
...
+          @assignment.criteria.find(id)
```
david-yz-liu, line 121: `find` returns a criterion. Save it in a variable so you can call `update` on it on the following line.

david-yz-liu (approved):

Issue #5830, opened 2022-02-04
AdminUser views


mishaschwartz:
Version 2.0 introduced an AdminUser that currently only has permission to do admin-like things through the UI.
We should also give AdminUser users abilities to do things in the UI:

- view all courses as if they were an instructor
   - will require updating policies
   - they should not show up in the Instructors table though
- have a view only visible/accessible to admin users where they can do the following through the UI:
   - manage (CRUD) courses
   - manage (CRUD) EndUser users
   - manage instance specific settings... right now there are a bunch of configuration settings that are specific to MarkUs that are set by editing [a configuration file](https://github.com/MarkUsProject/Wiki/blob/master/Configuration.md#markus-settings). These settings could also be stored in the database instead and managed through the UI.
      - note that we can only do this for settings that don't require a server restart to take effect (ie. those that are used by one of the initializers).

due date: to be completed by mid April (so that it can be included in version 2.1.0)

PR #5898, opened 2022-03-09
Created a new route for admins to manage courses and users

anoverga17:
## Motivation and Context

This pull request is the first step in having a view only visible/accessible to admin users where they can manage users and courses through the UI as described in issue #5830. This pull request aims to have a route that redirects to a blank page where the content will be added in a later PR.

## Your Changes

**Description**:
- Added new settings route in main controller to access site settings
- Added new policy so that only admin users can access site settings
- Added a link in the MarkUs dashboard header that uses the route to get to a new page. This page is currently blank.

## Testing

Manual testing of the feature was done by manually following the link of the new route. 

## Checklist

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -74,6 +74,10 @@ Rails.application.routes.draw do
     match '*path', controller: 'main_api', action: 'page_not_found', via: :all
   end
 
+  namespace :admin do
+    resources :courses, only: [:index]
+  end
+
   resources :courses, only: [:show, :index] do
     member do
       get 'clear_role_switch_session'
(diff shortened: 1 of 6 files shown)


config/routes.rb
```diff
@@ -446,6 +446,7 @@
   resources :main, only: [] do
     collection do
       post 'logout'
+      get 'settings'
```
anoverga17, line 449: I'd like an opinion if placing the route here makes sense. Most of the functionality will be provided through the REST API, meaning I was unsure whether to put it here (since this feature is very general), in the api namespace, or in it's own resource entirely.

david-yz-liu: @anoverga17 it's more standard to use a separate `admin/` namespace. See https://guides.rubyonrails.org/routing.html#controller-namespaces-and-routing.

So we'd have `admin/courses/`, `admin/courses/1/edit`, etc.

PR #5906, opened 2022-03-15
Fixed Test Results Summary not being able to be sorted by date

### LINKED ISSUE #5789: Test Results Summary not sorted by date
Here is a screenshot to illustrate

test results summary not sorted by date

anoverga17:
## Motivation and Context

This pull request aims to resolve #5789 and fixes a bug where automated test result batches were not being sorted by date properly. 

## Your Changes

**Description**:
- Modified `dateSort` method to return an integer (since ReactTable cannot sort boolean values)
- Removed `defaultSortMethod` and that states that only it used from batch test run table to allow default/date Sort to be used instead.

## Testing

Manual testing of the bug fix was done via the web interface. Automated tests were run both as a student and autotest user and the results of the tests were manually checked via the UI whether or not the results were being sorted correctly.

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,6 +9,7 @@
 - Replace uses of first name and last name attributes with display name (#5844)
 - Fix bug for rubric criteria level's mark update where one of the marks = an old mark (#5854)
 - Upgrade to Rails 7 (#5885)
+- Fix bug when sorting batch test runs where sorting by date was not working (#5906)
 
 ## [v2.0.9]
 - Fix bug when downloading all automated test files where the files were saved to a sub directory (#5864)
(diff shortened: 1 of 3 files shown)


david-yz-liu (commented):
Overall I think you identified the correct problem in the `dateSort` function (returns boolean instead of numbers), but since that function is only being used as a table sort method, it's better to fix the problem at the source rather than just for the batch test run table.

So I'd recommend changing `dateSort` to return a number. You can actually use the subtraction operator between dates to ensure that a number is returned---seems obvious in retrospect, but I only learned about this trick from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#description.

Oh and after making that change, I don't think we need the `defaultSortMethod` at all---looks like it's trying to override the sort method on the `created_at` column, maybe because that wasn't working properly originally. But after fixing `dateSort` hopefully that column will just sort fine.

david-yz-liu (approved):

PR #5910, opened 2022-03-17
results: Eagerly render all result tabs (so that they only mount once)

### LINKED ISSUE #5603: Bug report: new remark request text not loaded when switching tabs
Reporting a bug noticed when using MarkUs. When submitting a remark request, although entered text is saved on the server when I press the save button, it is not maintained on the local session when switching tabs. So when I switch tabs, the state of the textbox is reset to what it was when the app loaded, prior to me entering text into the textbox. 
https://user-images.githubusercontent.com/34668740/141234298-d1fab898-ed0c-4a86-92fd-1fc4a4d662a1.mp4

david-yz-liu:
## Motivation and Context

This fixes #5603. The issue was that by default the react-tabs library we use to display tabs triggers a React component mount/unmount every time tabs are switched. In the case of the remark request form, this caused previous text to be lost after switching away from the tab.

## Your Changes

**Description**: Use the [`forceRender`](https://github.com/reactjs/react-tabs#forcerender-boolean) prop to ensure remark request tab is loaded just once on initial page load.

I did the same for the test runs and feedback files tabs---they do trigger additional requests to fetch their data, which also makes loading those tabs a bit faster.

- [ ] Other (please specify):

## Testing

Tested in the web interface.

In the future, could do a cached lazy loading of tabs as described by https://github.com/reactjs/react-tabs/issues/126#issuecomment-466240855

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -11,6 +11,7 @@
 - Upgrade to Rails 7 (#5885)
 - Modified login to allow admin users to login via the UI (#5897)
 - Fix bug when sorting batch test runs where sorting by date was not working (#5906)
+- Ensure tabs in result view do not reload when switching between them (#5910)
 
 ## [v2.0.9]
 - Fix bug when downloading all automated test files where the files were saved to a sub directory (#5864)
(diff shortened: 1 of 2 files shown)


pretendWhale (approved):
@david-yz-liu Looks good! The only thing to note is that because the tests tab does not re-render on selection, switching between the tests tab and other tabs won't load new test results. To get new test results after running tests for a particular submission, you do have to reload the entire page. But the flash message after running tests tells you to do that.

PR #5921, opened 2022-03-25
Added view for admins to edit users

anoverga17:
## Motivation and Context

As part of issue #5830, admins should be able to manage users. As part of this, they should be able to edit users via the UI. This pull request thus aims to add an admin exclusive view that allows admins to edit the attributes of a user after they have been created in the database.

## Your Changes

**Description**:
- Added edit user route and controller method
- Added edit user view
- Added form to update user attributes
- Added update user route and controller method
- Added links to edit user view to the admin users list
- Added tests for changes

These changes create the following new view:

Admin Edit User

## Testing

Manual testing was done through the web interface. As an admin, a user was selected and edited accordingly. The user was then checked through the UI and database to ensure it was updated properly and the correct views were being loaded.
Rspec tests were also added to ensure the users were being updated appropriately.

## Checklist

https://github.com/MarkUsProject/Wiki/pull/102

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -1,0 +1,16 @@
+module Admin
+  class UsersController < ApplicationController
+    DEFAULT_FIELDS = [:id, :user_name, :email, :id_number, :type, :first_name, :last_name].freeze
+    before_action { authorize! }
+
+    respond_to :html
+    layout 'assignment_content'
+
+    def index
+      respond_to do |format|
+        format.html
+        format.json { render json: Course.order(:created_at).to_json(only: DEFAULT_FIELDS) }
+      end
+    end
+  end
+end
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -19,7 +19,11 @@ module Admin
 
     def update
       @user = visible_users.find_by(id: params[:id])
-      @user.update(user_params)
+      if @user.admin_user?
+        @user.update(admin_user_params)
+      else
+        @user.update(end_user_params)
+      end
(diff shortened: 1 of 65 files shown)


app/controllers/admin/users_controller.rb
```diff
@@ -13,6 +13,17 @@ def index
       end
     end
 
+    def edit
+      @user = visible_users.find_by(id: params[:id])
```
david-yz-liu, line 17: This can be simplified to `@user = record`, since the `id` param is automatically used to initialize a special `record` variable. Same below (and possibly in the courses controller---I didn't think of this when reviewing that earlier PR).

app/views/admin/users/_form.html.erb
```diff
@@ -0,0 +1,43 @@
...
+                     size: 35 %>
+    <%= f.label :email,
+                User.human_attribute_name(:email) %>
+    <%= f.text_field :email,
+                     size: 35 %>
+    <%= f.label :type,
+                User.human_attribute_name(:type),
+                class: 'required' %>
+    <%= f.select :type,
+                 [[EndUser.model_name.human, 'EndUser'], [AdminUser.model_name.human, 'AdminUser']],
```
david-yz-liu, line 35: Use `<class>.name` to get the string, rather than repeating the string literal.

app/controllers/admin/users_controller.rb
anoverga17: The `record` variable was used in the courses controller. I just thought it was better style here to select from all non autotest users (as opposed to from all Users which is what `record` queries from). Not a big deal either way though. Will make this change.

david-yz-liu (approved):

PR #5923, opened 2022-03-29
Add view for admins to create users

anoverga17:
## Motivation and Context

As part of issue #5830, admins should be able to manage users. As part of this, they should be able to create users via the UI. This pull request thus aims to add an admin exclusive view that allows admins to create new end and admin users using the UI. 

## Your Changes

**Description**:
- Added new user route and controller method
- Added new user view
- Added create user route and controller method
- Added link to create user from the users list page
- Added tests for changes

These changes create the following new view:

admin_user_create

## Testing

Manual testing was done through the web interface. As an admin, an Admin and End User were both created through the UI. The user was then checked through the UI and database to ensure it was updated properly and the correct views were being loaded while doing so.
Rspec tests were also added to ensure the users were being updated appropriately.

## Checklist

https://github.com/MarkUsProject/Wiki/pull/102


app/controllers/admin/users_controller.rb
```diff
@@ -13,6 +13,25 @@ def index
...
     end
 
+    def new
+      @user = User.new
+    end
+
+    def create
+      user_params = params.require(:user)
+      user_type = user_params[:type].camelize.downcase
+      case user_type
```
anoverga17, line 23: Unsure if this is the best way to check if the type is valid

app/controllers/admin/users_controller.rb
```diff
@@ -13,6 +13,27 @@ def index
...
+      when 'enduser'
+        user_params[:type] = 'EndUser'
+      when 'adminuser'
+        user_params[:type] = 'AdminUser'
+      else
+        user_params[:type] = nil
+      end
+      @user = User.create(user_params.permit(*DEFAULT_FIELDS))
+      respond_with @user do |format|
+        format.html { redirect_to admin_users_path }
```
anoverga17, line 33: Overriding the response format like this seems to resolve the failing tests. Not sure why.

anoverga17: Resolved. It took embarrassingly too long, but after leaving and coming back to this problem I finally realize tests failed due to errors with the test itself as opposed to the code.

app/controllers/admin/users_controller.rb
david-yz-liu: The `case` is okay, but don't do the `camelize` and `downcase` stuff, just ensure that the input values are the literal names of the classes (`AdminUser`/`EndUser`).

spec/controllers/admin/users_controller_spec.rb
```diff
@@ -98,6 +124,70 @@
...
+          }
+        }
+      end
+      let(:invalid_params) do
+        {
+          user: {
+            user_name: 'notValidUser',
+            email: 'sample@sample.com',
+            id_number: 100_678_901,
+            type: 'Not a valid type',
```
david-yz-liu, line 153: Make this a separate test case (do not combine with other invalid attributes)

app/controllers/admin/users_controller.rb
```diff
@@ -13,6 +13,25 @@ def index
...
+      user_type = user_params[:type].camelize.downcase
+      case user_type
+      when 'enduser'
+        user_params[:type] = 'EndUser'
+      when 'adminuser'
+        user_params[:type] = 'AdminUser'
+      else
+        user_params[:type] = nil
+      end
+      @user = User.create(user_params.permit(*DEFAULT_FIELDS))
```
david-yz-liu, line 31: This doesn't need to be an instance variable, you can use `user` instead of `@user`

anoverga17: Removing the `@user` seems to create errors with regards to flash interpolation

david-yz-liu (approved):

PR #5948, opened 2022-04-04
Modified model associations so that a role belongs to a user

anoverga17:
## Motivation and Context

In order to enable admins to view the same pages as instructors, it was decided admins would get their own dedicated role. As roles are currently associated with end users, admin users cannot receive a role. This pull request therefore makes modifications to the role model and its sub classes to allow any user to receive a role.

## Your Changes

**Description**:
- Update `Role` model to belong to a user
- Removed `has_many :roles` association in end_user
- Added `has_many :roles` association to user
- Modified every query for roles to refer to a user instead of an end user.
- Added a validation to only allow End Users to receive TAs and Student roles

Note: Currently non end users can receive the instructor role with these changes. This can only be done via the API as instructor creation manually via the UI is still restricted to EndUsers. Tests for this case will be added in the admin role PR as it would be more relevant to figure out who can receive the instructor role in that PR.

## Testing

Manual testing via the UI was done to ensure no major features of MarkUs were broken. Student, Instructor, and Grader management were tested more heavily as these changes affects those roles.
Rspec tests were also modified and run (ensuring only the relevant references to end_user were replaced) to ensure all features work correctly as intended.

## Checklist


app/models/end_user.rb
```diff
@@ -1,8 +1,5 @@
 # Subclass of User that can be associated with Roles
```
david-yz-liu, line 1: This comment should be updated to something to something a bit more general, like "for regular users".

david-yz-liu (commented):
@anoverga17 great work! I just left one small comment

anoverga17:
Failing test unrelated to this PR

PR #5950, opened 2022-04-05
Dark mode

therealfake:
## Motivation and Context

The current dark mode color scheme could use some updating as the colors are a bit unusual and not as contrast. 

## Your Changes

**Description**:
- Updated color scheme of overall dark mode
- Added properties to React chart to fix incoherent styling applied
- Added dark mode compatible icons for dropdown menu, which would've been hard to see if unchanged.

## Testing

All tests pass.

- Currently, the values in ``chart_helpers.js`` are hardcoded values for the color variables ``--primary_one`` and ``--gridline``. This is not good practice. I have tried doing ``document.documentElement.style.getPropertyValue("--primary_one")`` and importing variables from ``app/assets/stylesheets/common/_constants.scss``. None of these approaches have worked and would like suggestions on how to proceed.

## Checklist


app/assets/javascripts/Components/Helpers/chart_helpers.js
```diff
@@ -2,18 +2,34 @@ import * as I18n from "i18n-js";
 
 export const chartScales = {
   x: {
+    grid: {
+      borderColor: "#98c7ff", // --primary_one
```
david-yz-liu, line 6: @therealfake I looked into this a little bit more. The `document.documentElement.style.getPropertyValue("--primary_one")` is correct; the issue is that these values are set using the "light" theme as default, and then mutated in the `set_theme` function when the page loads. But because `chartScales` is a constant, it is initialized using the initial values.

To fix this, you can turn `chartScales` into a *function* that returns an object; then whenever `chartScales` is called, it'll return the colours based on the current theme values. Note that you'll need to update _everywhere_ `chartScales` is used, not just the assignment chart.

therealfake: Oh, interesting. I converted chartScales to a function and all graphs are coherently styled in dark mode now.

david-yz-liu (approved):

david-yz-liu:
@therealfake one thing I just double-checked: there's one more usage of `chartScales` in `grade_entry_form_chart.jsx` that needs to be updated.

david-yz-liu (approved):

PR #5959, opened 2022-04-13
Added a bulk upload of end users to admin users page

anoverga17:
## Motivation and Context

A common task admins will find is creating several end users. Since the new admin views allows admins to perform tasks through the UI, it would be convenient to give admins the option to upload and create several users from a list. This pull request aims to give admins this feature.

## Your Changes

**Description**:
- Added a new heading button to upload end users
- Added a modal to submit end users
- Added a new `UploadUsersJob` that uploads users from a csv file
- Added an `EndUser::CSV_ORDER` to define what each column in the csv file corresponds to
- Added tests for changes

## Testing

Manual testing through the UI was done to ensure the feature worked properly. A sample csv file of users were uploaded and then verified to ensure they were created.
Rspec tests were added to ensure the feature worked properly on a variety of parameters and file data as well as to ensure the new upload users job was working properly.

## Checklist

https://github.com/MarkUsProject/Wiki/pull/110


config/locales/views/admin/en.yml
```diff
@@ -6,3 +6,5 @@ en:
     hidden: Hidden
     visible: Visible
   markus_administration: MarkUs Administration
+  users:
+    admin_csv_upload_help_html: For more information about this feature and the<br>file format(s) supported, please visit <a href="https://github.com/MarkUsProject/Wiki/blob/%{markus_version}/Admin-Guide.md#bulk-uploading-of-end-users">this page</a>.
```
anoverga17, line 10: Not sure if this should link to the section within the Admin Guide describing this feature and its expected format or if I should create a new `Admin-Guide--Importing-and-Exporting-Data.md` that describes the expected csv file format and link to that.

app/controllers/admin/users_controller.rb
```diff
@@ -35,6 +35,24 @@ def update
       respond_with @user, location: -> { edit_admin_user_path(@user) }
     end
 
+    def upload
+      begin
+        data = process_file_upload
+      rescue Psych::SyntaxError => e
```
david-yz-liu, line 41: `Psych` is only used for YML files, which we aren't supporting here. So you can remove this case.

app/jobs/upload_users_job.rb
```diff
@@ -0,0 +1,24 @@
...
+  def self.on_complete_js(_status)
+    'window.location.reload.bind(window.location)'
+  end
+
+  def perform(user_class, data, encoding)
+    progress.total = data.lines.count
+    user_class.transaction do
+      MarkusCsv.parse(data, encoding: encoding, skip_blanks: true, row_sep: :auto) do |row|
+        user_data = {}
+        USER_FIELDS.each_with_index do |field, index|
```
david-yz-liu, line 13: You don't need a separate variable for these fields. Instead, rely on `EndUser::CSV_ORDER` to determine the order of the values in the row.

You can use the `zip` method to iterate over `EndUser::CSV_ORDER` and `row` in parallel.

app/jobs/upload_users_job.rb
```diff
@@ -0,0 +1,24 @@
...
+          row_index = EndUser::CSV_ORDER.index(field) || index
+          user_data[field] = row[row_index]&.strip
+        end
+        next if user_data[:user_name].blank? || user_data[:last_name].blank? || user_data[:first_name].blank?
```
david-yz-liu, line 17: This check isn't necessary. If these fields are blank there will be a validation error below, which will be reported as an error by the job---and displayed as a flash message in the UI, which you should double-check to see.

app/jobs/upload_users_job.rb
```diff
@@ -0,0 +1,24 @@
...
+        user = user_class.find_or_initialize_by(user_data)
```
david-yz-liu, line 18: Let's allow *updates* of user information through CSV upload, keyed by `user_name`. So if the CSV contains a row of data for a username that already exists, MarkUs should *update* the existing user with that name using the values in the row, even if they're different from what the user currently has.

So then here, do a `find_or_initialize_by` just on the user name, and then update with the returned objects with other attributes.

Also, please update the tests and documentation to be consistent with this updating behaviour.

config/settings.yml
```diff
@@ -56,6 +56,13 @@ student_csv_order:
   - id_number
   - email
 
+end_user_csv_order:
```
david-yz-liu, line 59: This is good. Any time the settings are updated, please also make sure to update the settings validations found in `config/initializers/config.rb`.

david-yz-liu (commented):
@anoverga17 good work! One thing I forgot to mention: there are examples of the file upload formats found under the `doc/import_samples` folder. Please add a new one for the end user upload.

app/jobs/upload_users_job.rb
anoverga17: I have made this updating behaviour work. It was implemented a little differently than you described though. This is because using the `find_or_initialize_by` just on the user name doesn't work since the `after_initialize :set_display_name` prevents users from being initialized without a first or last name. 
I instead did a find_by user name or initialize by all user attributes. Let me know if you prefer a different approach to working around this.

david-yz-liu (approved):

PR #5822, opened 2022-02-03
Added The Ability to Upload URLs

anoverga17:
This pull request adds the ability to submit URLs

## Motivation and Context

Due to the fact that some files are unreasonably large to be uploaded to MarkUs, the need to upload and view them through an external service would allow graders to still view these files and still grade the submission. 
Currently, due to ease of accessing the relevant files, these external services are YouTube and Google Drive/Docs.

This pull request also resolves #5014 

## Your Changes

**Description**:
- Added the ability for instructors to enable/disable the ability to upload/preview urls (files with a `.markusurl` extension)
- Added a URL submit button and modal to the repository/submissions view if URLs are enabled
- Added a preview for URLs that displays the URL and it's related content if URLs are enabled
    - Annotation support for URLs is not available

For more information please see the relevant pull requests

**Type of change** (select all that apply):

- [x] New feature (non-breaking change which adds functionality)

## Testing

Manual testing of the feature was done through the web interface. Various URLs were uploaded through the new url submit modal or by manually submitting `.markusurl` files. These were then viewed as an instructor/student in the submission repository and result views to ensure the URLs were displaying what was intended. These views were then disabled by disallowing url submits to ensure that in that case, they were to be treated as normal files. 

Rspec tests were also added to ensure the correctness of the feature in the backend.

## Checklist

- [x] I have performed a self-review of my own code.
- [x] I have verified that the pre-commit.ci checks have passed. 
- [x] I have verified that the CI tests have passed. 
- [x] I have reviewed the test coverage changes reported on Coveralls. 
- [x] I have added tests for my changes. 
- [x] I have updated the Changelog.md file. 
- [x] I have made changes to the documentation and linked to that pull request below. 

### Pull request to make documentation changes (if applicable)

https://github.com/MarkUsProject/Wiki/pull/90

PR #5960, opened 2022-04-19
Reverted use of "Submit File" back to "Upload File" for non submission views

anoverga17:
## Motivation and Context

After changing the wording of 'upload file' to 'submit file' in pull request #5822,  non submission related views such as starter file uploading and autotest file uploading also changed to say 'submit file'. This does not make sense for those views and thus, this pull request aims to only restrict use of 'submit file' to submission related file manager views.

## Your Changes

**Description**:
- Refactor use of 'submit_the' locale to function for ease of switching to 'upload_the' in file manager
- Add new prop to submission file manager that enables change of text from upload file to submit file

## Testing

Manual testing through the web interface was done. Various non submission related views were checked to make sure they read 'upload file'. In addition, submission related views were checked to ensure they read as 'submit file'

## Checklist


app/assets/javascripts/Components/markus_file_manager.jsx
```diff
@@ -38,6 +38,11 @@ class RawFileManager extends RawFileBrowser {
     }
   };
 
+  upload_or_submit_file = () => {
```
david-yz-liu, line 41: This function name is a bit misleading, it sounds like this is the action that causes a file to be uploaded.

Add the suffix `_label` so that it's clear what's being returned.

david-yz-liu (approved):

PR #5967, opened 2022-04-28
Add Rails Performance to Admin Dashboard

anoverga17:
## Motivation and Context

When creating the admin views, it was decided that the new Admin Dashboard would be a good place to add monitoring tools. This pull request adds the [rails performance](https://github.com/igorkasyanchuk/rails_performance) which is a gem that provides simple monitoring to MarkUs.

## Your Changes

**Description**:
- Added and configured Rails Performance Gem
- Added configuration settings for new gem
- Modified admin dashboard to display dashboard links nicer
- Added rspec tests to check for user permissions

This pull request makes the following change to the MarkUs UI:
Admin Dashboard List

## Testing

Manual testing was done through the UI to ensure the gem works and all its features worked properly. The gem was enabled and accessed as admins and non admin users to ensure the proper permissions were enabled and the views for the gem works. 

In addition, rspec tests were added to further test that only admins can access the new dashboard.

## Checklist

https://github.com/MarkUsProject/Wiki/pull/116


config/initializers/rails_performance.rb
```diff
@@ -0,0 +1,40 @@
...
+      end
+    end
+  end
+
+  if defined?(RailsPerformance)
+    RailsPerformance.setup do |config|
+      config.enabled = Settings.rails_performance.enabled
+      config.redis = Redis::Namespace.new(Rails.root.to_s)
+      config.duration = Settings.rails_performance.duration.minutes
+      config.mount_at = '/admin/performance'
```
david-yz-liu, line 36: The [rails_performance documentation](https://github.com/igorkasyanchuk/rails_performance#usage) seems to indicate that we can omit this setting if we manually mount the server in our `routes.rb` file.

config/initializers/rails_performance.rb
```diff
@@ -0,0 +1,40 @@
...
+                 status: :forbidden, layout: false
+        end
+      end
+    end
+  end
+
+  if defined?(RailsPerformance)
+    RailsPerformance.setup do |config|
+      config.enabled = Settings.rails_performance.enabled
+      config.redis = Redis::Namespace.new(Rails.root.to_s)
```
david-yz-liu, line 34: 1. Let's go with the default name (`Redis::Namespace.new("#{Rails.env}-rails-performance")`)
2. I think we need to pass in the MarkUs Redis instance as the `redis` keyword argument (as illustrated in the documentation). But, we can't yet assume the MarkUs Redis instance has been initialized when we run this file. Instead, add this specific configuration setting to the `redis.rb` initializer and use `Redis.current` as an argument to the `Redis::Namespace.new` (we do something similar with `Resque.redis` in that file already).

config/initializers/rails_performance.rb
anoverga17: You can. This was just added for redundancy. I will remove it.

anoverga17: Actually after testing it a bit more, I have found that it can be omitted, though I had to change the mounting point to the default `/rails/performance` if thats alright.

I found if I changed the mount point for the gem like this:
```ruby
namespace :admin do
  ...
  mount RailsPerformance::Engine, at: '/performance', as: 'performance'
end
```
the links for each tab in the gem redirects to `<root_path>/rails/performance/<tab_name>` and are thus broken.

I think this has something to do with the admin namespace because when I move `mount RailsPerformance::Engine, at: '/performance', as: 'performance'` outside the admin namespace and rename it as `mount RailsPerformance::Engine, at: 'admin/performance', as: 'performance'` the gem works fine again.

david-yz-liu (approved):

PR #5968, opened 2022-04-29
peer-review: several bug fixes and new peer review submission table

david-yz-liu:
## Motivation and Context

This pull request makes a number of fixes to the peer review functionality in MarkUs.

## Your Changes

**Description**:

1. Enable creation of peer review assignments with a new "Create Peer Review Assignment" button.
2. Create new "submissions" table to show the marking state of all peer reviews.
3. Enable peer reviewers to use one-time only annotations (but not annotation categories).
4. Display peer review ids consistently to students (this still maintains anonymity unless the reviewer and reviewee directly communicate).
5. A number of small regression bug fixes (e.g., having to do with course ids).
6. Display peer review assignments as separate entries in assignment tables and the assignment "quick links" dropdown menu. (Previously, they were grouped together with their parent assignment.)
7. Fixed bugs when assigning reviewers randomly in cases when some peer review assignments already exist, and when the total number of reviews is not divisible by the number of submissions to review.

## Testing

Tested in the UI. Will add additional tests.

## Checklist

https://github.com/MarkUsProject/Wiki/pull/114


pretendWhale (changes requested):
@david-yz-liu I think most of the changes here look good, but when testing I noticed a couple of bugs:

- When the number of reviewers is > than the number of reviews requested, the expected `UnableToRandomlyAssignGroupException`. Instead, the request fails and returns a 500 error with `ERROR:  null value in column "reviewer_id" violates not-null constraint`
- When creating a peer review assignment via the new Create Peer Review button, the parent assignment's `has_peer_review` property is not updated to true. This means that the `review?` policy on results fails, and it also prevents the pr assignments from showing in the assignment table. I think we should set this property, but we could also use the `assignment.has_peer_review_assignment?` method in the policy/views instead.

pretendWhale (changes requested):
@david-yz-liu Just left a couple of notes inline about things that I noticed in testing, and then a couple of issues described below. 

The main issues I noticed were with criteria and annotations when doing peer reviews. 

- *Criteria:*  When a student is completing a peer review, the criteria shown in the `marks_panel` are taken from the peer review assignment, not the parent assignment. However, if the parent assignment has criteria and the submission has not been graded by the instructor/a TA, attempting to set the peer review to complete fails with a `criterion_incomplete_error` (which is not displayed to reviewers).

- *Annotations:* annotations seem to be interacting with the parent assignment. When doing a peer review, annotation categories and annotations created by the instructor are pulled from the parent assignment, not the peer review assignment. Similarly, one time annotations are stored on the parent result, not the pr result. 

Additionally semi-related to this: when attempting to save a pre-configured annotation, student reviewers are unable to do so (I suspect because the `add_existing_annotation?` policy only looks at `is_ta?` and `is_instructor?`).

PR #6001, opened 2022-05-13
Issue 5905 and related UI changes

### LINKED ISSUE #5905: Previewing files
Currently when a student submits a file and then wants to preview the PDF, clicking on the file name will download it. So to avoid that, we have to click anywhere else on that row:

Screen Shot 2022-03-15 at 1 24 09 PM

I feel like that design isn't intuitive, I only know this from using MarkUs frequently but it wasn't obvious at the start.

TheTallJerry:
## Motivation and Context
Resolves #5905 where the current submissions UI is reported to be unintuitive. 

## Your Changes

**Description**:
- Explicit download icon for each individual file. 
- Hover effect for current submission that the mouse is over at, with different effects in light vs dark mode, download icon vs the rest of the row. 
- Changed download-all option’s icon. 
- Scrollbar effect to account for header's overflow in mobile. 

## Testing

I've done manual testing within browsers (chrome, firefox), switching between ligh vs dark mode, desktop vs mobile view. 
### Current Mobile: 
current-submission-mobile

### Current Desktop: 
current-submission-desktop

### Changed Mobile: 
mobile

### Changed Desktop Light: 
light not selected
light selected

### Changed Desktop Dark: 
dark not selected
dark selected

(not showing dark/light in mobile due to similarity) 
## Checklist


app/assets/javascripts/Components/markus_file_manager.jsx
```diff
@@ -298,7 +298,7 @@ class FileManagerFile extends FileRenderers.RawTableFile {
     } else {
       icon = <i className="fa fa-file-o" aria-hidden="true" />;
     }
-
```
david-yz-liu, line 301: Keep this blank line

app/assets/javascripts/Components/markus_file_manager.jsx
```diff
@@ -298,7 +298,7 @@ class FileManagerFile extends FileRenderers.RawTableFile {
...
+    let download_icon = <i className="fa fa-download-o" aria-hidden="true" />;
```
david-yz-liu, line 301: Add a `title` attribute here so that if a user hovers over it, they'll know what it does.

app/assets/javascripts/Components/markus_file_manager.jsx
```diff
@@ -401,6 +402,7 @@ FileManager.defaultProps = {
     FolderOpen: <i className="fa fa-folder-open-o" aria-hidden="true" />,
     Delete: <i className="fa fa-trash-o" aria-hidden="true" />,
     Loading: <i className="fa fa-circle-o-notch fa-spin" aria-hidden="true" />,
+    Download: <i className="fa fa-download-o" aria-hidden="true" />,
```
david-yz-liu, line 405: I don't think adding this is necessary (I think these props are just for the React component library we're using, not our own custom actions/icons).

app/assets/javascripts/Components/markus_file_manager.jsx
```diff
@@ -332,7 +332,8 @@ class FileManagerFile extends FileRenderers.RawTableFile {
       name = (
         <a href={this.props.url || "#"} download={this.getName()}>
           {icon}
-          {this.getName()}
+          <span>{this.getName()}</span>
+          {download_icon}
```
david-yz-liu, line 336: So in terms of functionality, I want *only* the icon to trigger the download, but not the filename or file icon (`{icon}`). Those parts should be moved outside the `a` tag. But, you can keep the same positioning of the icon to the right of the filename.

app/assets/stylesheets/common/core.scss
```diff
@@ -439,6 +439,10 @@ th.required::after {
   vertical-align: middle;
 }
 
+.fa-download-o::before {
+  content: asset-url('icons/download.png');
+  vertical-align: middle;
+}
```
david-yz-liu, line 445: insert a newline after this line

app/assets/stylesheets/common/core.scss
```diff
@@ -439,6 +439,10 @@ th.required::after {
   vertical-align: middle;
 }
 
+.fa-download-o::before {
+  content: asset-url('icons/download.png');
```
david-yz-liu, line 443: Okay, so it looks like we already have a `download-file.png` icon used in some places. But, the icon you added is a true icon size, which is good. You can go ahead and *replace* the existing `download-file.png` file with the icon you added, and then here use `icons/download-file.png` instead.

vendor/assets/stylesheets/react-keyed-file-browser.css
```diff
@@ -23,7 +23,8 @@ div.rendered-react-keyed-file-browser div.action-bar {
     flex-grow: 1;
     min-width: 300px;
     margin-left: 10px;
-    white-space: nowrap; }
+    white-space: nowrap;
```
david-yz-liu, line 26: Good detective work finding this file, but files under `vendor` shouldn't be changed directly (they're copies of releases from the external libraries we use).

Instead, check out `core.scss` starting at line 431, and add your CSS changes there.

app/assets/stylesheets/common/core.scss
david-yz-liu: Oh also, please modify the file manager "Download All" action to use this icon instead of the "Save disk" icon it currently uses.

Changelog.md
```diff
@@ -32,6 +32,7 @@
 - Added background job to clean tmp folder (#5963)
 - Removed consideration of due date from scanned exam (#5964)
 - Added Exception Notification Gem (#5965)
+- Resolved issue 5905 unintuitive submissions UI design (#6001)
```
david-yz-liu, line 35: For new features (not bug fixes), Changelog descriptions can mention issues, but should really focus on what the change actually was, not what problem was solved. (i.e., say the "what", not the "why")

app/assets/stylesheets/common/core.scss
david-yz-liu: And one last thing: it would be good to add a hover effect to this icon as well, to distinguish it from the rest of the row when a user hovers over it. Maybe add a slight background colour or something, not sure.

TheTallJerry: Using a slightly stronger color for the icon hover over (relative to the row's hover)

TheTallJerry: Updated PR description with the new changes (changed download all icon, hover over background color change for both the row item & the download icon)

david-yz-liu (commented):
@TheTallJerry I forgot to mention, it would be great if you added some tests for your changes! (Building on the great JS testing work you did earlier, of course.) You don't need to add tests for the whole component, but should add tests for clicking on the row and clicking on the download button to make sure the correct requests are made.

app/assets/javascripts/Components/markus_file_manager.jsx
```diff
@@ -299,6 +299,19 @@ class FileManagerFile extends FileRenderers.RawTableFile {
       icon = <i className="fa fa-file-o" aria-hidden="true" />;
     }
 
+    let download_icon = (
```
david-yz-liu, line 302: Please include a source for this SVG (as a comment)

app/assets/javascripts/Components/markus_file_manager.jsx
```diff
@@ -299,6 +299,19 @@ class FileManagerFile extends FileRenderers.RawTableFile {
...
+      <svg
+        className="submission-download"
```
david-yz-liu, line 304: The file manager is actually used in other places beyond submissions (for example, assignment starter files). So we should keep the naming generic: something like `file-download`, for example.

app/assets/javascripts/theme_colors.js
```diff
@@ -22,6 +22,9 @@ export const themes = {
     sharp_line: "#e5e5e5",
     strings_color: "var(--severe_error)",
     sub_menu: "var(--primary_one)",
+    submission_download_icon_fill: "var(--sharp_line)",
```
david-yz-liu, line 25: As above, rename to something like `file_download_icon_fill`. Same for the other constants you added.

app/assets/stylesheets/common/core.scss
```diff
@@ -474,10 +474,30 @@ th.required::after {
...
 
+div.rendered-react-keyed-file-browser div.files table td.name div * {
+  padding-right: 0.5rem;
+}
+
+div.rendered-react-keyed-file-browser div.files tbody tr:hover {
+  background-color: $submission-row-hover;
+}
+
+.submission-download {
```
david-yz-liu, line 493: In addition to the fill, add the rule `vertical-align: bottom;` (it'll make the vertical alignment for the icon a bit more centered with respect to the text)

jest_after_env_setup.js
```diff
@@ -18,3 +18,19 @@ import {configure} from "enzyme";
 import Adapter from "enzyme-adapter-react-16";
 
 configure({adapter: new Adapter()});
+
+// Worker
```
david-yz-liu, line 22: It took me a while to figure out why you added this. But, similar to how you added `jest-canvas-mock`, let's also add https://github.com/developit/jsdom-worker. Seems to be not super active, but better than these mocks.

(See also this comment: https://github.com/jsdom/jsdom/issues/1721#issuecomment-1079407658)

webpack.development.js
```diff
@@ -7,5 +7,6 @@ module.exports = merge(common, {
   watch: true,
   watchOptions: {
     ignored: /node_modules/,
+    poll: true,
```
david-yz-liu, line 10: Please revert (I haven't had a chance to test this yet)

app/assets/javascripts/Components/__tests__/submission_file_manager.test.jsx
```diff
@@ -0,0 +1,90 @@
...
+      key: "deferred-process.java",
+      modified: 1652577324,
+      relativeKey: "deferred-process.java",
+    },
+  ];
+
+  beforeEach(() => {
+    // Unlike FileManager, files are stored in SubmissionFileManager's states, which are set when the component mounts
+    // and calls fetchData. As a result, we need to mock that fetch to return our data.
+    // We need to mock "twice" (i.e. two promises) because of how fetch works.
```
david-yz-liu, line 41: Okay I looked into this a bit, and even though what you have is find for this case, I'd rather use a library like https://github.com/jefflau/jest-fetch-mock (which is well-maintained) to do this mocking. Please introduce that in this PR as well, so that we have a good example of its use for the future. :)

david-yz-liu (approved):

PR #6003, opened 2022-05-13
string interpolation

WeissSa:
Migrated the set_release_on_results from submissions_controller to a class method of result.rb and added testing.

## Motivation and Context
This helps group related code together and adds testing for the migrated method.

## Your Changes
**Description**:
I first copy-pasted the method over. Then I changed how it was called so it references the class. Then I realized there were scope issues with the flashing of the error message so I changed the original method to raise an error and then inside submissions_controller the error is caught and the message is read in order to determine what type of error it is.

I also added some basic testing for the function.

## Testing
I ran the rspec suite and manually tested the feature to make sure the error flashed. I also added new testing.

I'm not 100% sure my rspec tests are amazing. I also think that there might be a better way to handle the error, but I could not figure out what that is.

## Checklist

### DIFF
--- a/app/controllers/submissions_controller.rb
+++ b/app/controllers/submissions_controller.rb
@@ -659,9 +659,23 @@ class SubmissionsController < ApplicationController
       changed = if assignment.is_peer_review?
                   set_pr_release_on_results(groupings, release)
                 else
-                  set_release_on_results(groupings, release)
+                  begin
+                    Result.set_release_on_results(groupings, release)
+                  rescue StandardError => e
+                    error_message, group_names = e.message.split
+                    if error_message == 'no_submission'
+                      flash_now(:error, I18n.t('submissions.errors.no_submission', group_name: group_names))
+                    elsif error_message == 'not_complete'
+                      flash_now(:error, t('submissions.errors.not_complete', group_name: group_names))
+                    elsif error_message == 'not_complete_unrelease'
+                      flash_now(:error, t('submissions.errors.not_complete_unrelease', group_name: group_names))
+                    end
+                  end
                 end
-
+      if changed.is_a? Array
+        # Changed will only be an array if an error message flashed
+        changed = 0
+      end
       if changed > 0
         assignment.update_remark_request_count
 
--- a/spec/models/result_spec.rb
+++ b/spec/models/result_spec.rb
@@ -165,4 +165,16 @@ describe Result do
       end
     end
   end
+
+  describe '.set_release_on_results' do
+    let!(:assignment) { create :assignment_with_criteria_and_results }
+    it 'should raise StandardError with message not_complete error if result has not been completed' do
+      result = assignment.ungraded_submission_results[0]
+      expect { result.set_release_on_results(grouping, true) }.to raise_error(StandardError)
+    end
+    it 'should raise a StandardError with message no_submission error if the grouping does not have a submission' do
+      grouping = assignment.unassigned_groupings[0]
+      expect { Result.set_release_on_results(grouping, true) }.to raise_error(StandardError)
+    end
+  end
 end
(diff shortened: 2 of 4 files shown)


app/controllers/submissions_controller.rb
```diff
@@ -659,9 +659,23 @@ def update_submissions
       changed = if assignment.is_peer_review?
                   set_pr_release_on_results(groupings, release)
                 else
-                  set_release_on_results(groupings, release)
+                  begin
+                    Result.set_release_on_results(groupings, release)
+                  rescue StandardError => e
+                    error_message, group_names = e.message.split
+                    if error_message == 'no_submission'
+                      flash_now(:error, I18n.t('submissions.errors.no_submission', group_name: group_names))
```
david-yz-liu, line 667: You can use `I18n.t` inside model methods. This means that you can raise an exception with the exact message you want to display, and then do something like a `flash_now(:error, e.message)` here in the controller.

david-yz-liu: Oh, and in order to make `changed = 0`, I believe you can just put in a `0` at the end of the `rescue` block, and then the result of "evaluating" that block will be `0`, which gets assigned to `changed`.

david-yz-liu (commented):
@WeissSa good start, I left a couple of comments.

For your tests:

1. You should add checks for the content of the error message itself (comparing to the correct `I18n.t` string).
2. You should add tests for the "correct" path: for when there are results that get their `released_to_students` attribute changed from true to false, or vice versa. This is where the *factories* will come in, to help you create the result objects (and the necessary associated objects, like submissions and groupings) so that you can test how calling this method mutates those objects.

spec/models/result_spec.rb
```diff
@@ -165,4 +165,41 @@
...
+
+  describe '.set_release_on_results' do
+    let!(:assignment) { create :assignment_with_criteria_and_results }
+    let!(:released) { create :released_result }
+    it 'should raise StandardError with message not_complete error if result has not been completed' do
+      create(:incomplete_result)
+      begin
+        Result.set_release_on_results(assignment.groupings, true)
+      rescue StandardError => e
+        expect(e.message).to start_with(I18n.t('submissions.errors.not_complete'))
```
david-yz-liu, line 177: You can check error messages using `raise_error`. See https://relishapp.com/rspec/rspec-expectations/v/2-8/docs/built-in-matchers/raise-error-matcher#match-message-with-a-string

spec/models/result_spec.rb
```diff
@@ -165,4 +165,41 @@
       end
     end
   end
+
+  describe '.set_release_on_results' do
+    let!(:assignment) { create :assignment_with_criteria_and_results }
+    let!(:released) { create :released_result }
```
david-yz-liu, line 171: I don't think this is doing what you think it's doing. The result being created here won't be associated with the `assignment` you created above. Same with `create(:incomplete_result)` below. Instead, it's creating a completely different result.

You can check out what the `assignment_with_criteria_and_results` factory is doing, but I'm guessing it's creating some incomplete results all by itself.

spec/models/result_spec.rb
```diff
@@ -165,4 +165,41 @@
...
+        expect(e.message).to start_with(I18n.t('submissions.errors.no_submission'))
+      end
+    end
+    it 'should keep a released submission released' do
+      Result.set_release_on_results(assignment.groupings, true)
+      expect(released.released_to_students).to be(true)
+    end
+    it 'should unrelease a released submission' do
+      Result.set_release_on_results(assignment.groupings, false)
+      expect(released.released_to_students).to be(true)
```
david-yz-liu, line 202: This isn't really what you wanted to test, it's probably a copy and paste error.

But the fact that the test is passing is a sign that the test isn't set up correctly---see my comment about factories above. This would actually be a time to use the `released_result` factory, and then use associations to go from that result object to its corresponding grouping.

david-yz-liu (approved):

PR #6009, opened 2022-05-21
Added redirect for students and TAs

WeissSa:
Added a redirect so that students and TAs do not get a 403 error upon going to <course>/courses/:id

## Motivation and Context
This solves a 403 error upon accessing courses through Quercus (I think)

## Your Changes
I added a redirect to the root_url in courses_controller. This will redirect unauthorized users to the root_url. Which for students and TA's is the assignments page and for unauthorized users should be the login page.

Also fixed tests
**Description**:

## Testing
I tried it as a student, instructor, and TA as well as as a user who is not logged in. It also fails 2 tests (which is good, since those tests expect a 403 error).

Should I remove the tests or modify the tests which expect 403 error.

## Checklist

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -25,9 +25,6 @@ class CoursesController < ApplicationController
   def edit; end
 
   def update
-    if current_role.student? || current_role.ta?
-      raise status => 403
-    end
     @current_course.update(params.require(:course).permit(:is_hidden))
     respond_with @current_course, location: -> { edit_course_path(@current_course) }
   end
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -25,13 +25,16 @@ class CoursesController < ApplicationController
   def edit; end
 
   def update
+    if current_role.student? || current_role.ta?
+      raise status => 403
+    end
     @current_course.update(params.require(:course).permit(:is_hidden))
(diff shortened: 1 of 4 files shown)


spec/controllers/courses_controller_spec.rb
```diff
@@ -172,7 +172,7 @@
 
     context 'updating course visibility' do
       context 'as an authorized instructor' do
-        it 'responds with success on update' do
+        it 'responds with 302 on update' do
```
david-yz-liu, line 175: this shouldn't be changed

david-yz-liu (commented):
In terms of testing: the changes you made to the `course_policy_spec.rb` file are good. In addition to this, you should add some new tests to `courses_controller_spec.rb` for your updates. You can use `expect(response).to redirect_to <path>` to assert that the route redirects to the correct location.

app/controllers/courses_controller.rb
```diff
@@ -25,15 +25,22 @@ def index
   def edit; end
 
   def update
+    if current_role.student? || current_role.ta?
```
david-yz-liu, line 28: @WeissSa You can revert this change as well (since you've now removed the change to the corresponding test)

WeissSa: Oops. Thought I did but must have forgotten to add it.

david-yz-liu (approved):

PR #6012, opened 2022-05-27
Added Capybara Gem

### LINKED ISSUE #5832: Testing UI interactions
We currently do not have any tests in place that test the UI directly (other than some unittests that make sure the right views are rendered). 

As a first step we should enable this type of testing using the [capybara](https://github.com/teamcapybara/capybara#using-capybara-with-rspec) gem and specifically its [rspec integration](https://github.com/teamcapybara/capybara#using-capybara-with-rspec)

1. get rspec tests with capybara working
2. add a few tests as a proof of concept
3. discuss with other developers which views should be tested in this way (ideally all of them, but since this will take a while to write them there should be some sort of priority order)
4. start writing those tests (feel free to split them up over several PRs as needed)

due date: steps 1-3 should be completed by the end of summer 2022, step 4 will probably take a while to complete so get as much done as you can before the end of the summer but you should not expect to have everything finished

anoverga17:
## Motivation and Context

This pull request aims to add and test the Capybara gem. This gem adds rspec functionality that allows us to test UI interactions and resolves #5832

## Your Changes

**Description**:
- Imported and configured Capybara Gem
- Imported and configured Selenium Webdriver
- Added test for various users logging into MarkUs
- Configured CI tests to run Selenium Capybara tests using chrome

- [x] New feature (non-breaking change which adds functionality)\

## Testing

Passing rspec login tests were added for the UI. These tests were then tested on a machine running Ubuntu 20.04 as well as on a machine running Windows/WSL 2 to ensure the settings for running the new UI tests worked in a variety of environments.

## Running Tests
In order to run Capybara tests locally, the following needs to be done (will be explained further in documentation PR)
1. Download and install [Chrome](https://support.google.com/chrome/answer/95346?hl=en&co=GENIE.Platform%3DDesktop) and [ChromeDriver](https://chromedriver.chromium.org/downloads) on your machine
2. Run `docker-compose run -p 3434:3434 --rm rails bash` to expose port 3434 for testing
3. Run `RAILS_RELATIVE_URL_ROOT=/ rspec spec/system/` in the created docker rails bash terminal to run tests
4. Optionally add `DISABLE_HEADLESS_UI_TESTING=true` to environment variables if you wish to see the tests run in a browser window

## Checklist

https://github.com/MarkUsProject/Wiki/pull/122


spec/system/main/login_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    click_link(I18n.t('main.login_with', name: Settings.local_auth_login_name ||
+      I18n.t('main.local_authentication_default_name')), href: nil)
+    # Enter username and password
+    fill_in('user_login', with: user_name, id: 'user_login')
+    fill_in('user_password', with: 'x', id: 'user_password')
+    # Login to MarkUs
+    click_button(I18n.t('main.log_in'), name: 'commit')
+  end
+
+  context 'Instructor' do
```
david-yz-liu, line 17: So in terms of the actual tests, there isn't a need here to distinguish between instructor/ta/student. For the purposes of login and checking redirection, they're all the same `StandardUser`, so you should just create one test for that user type.

spec/spec_helper.rb
```diff
@@ -80,6 +90,22 @@
   # instead of true.
   config.use_transactional_fixtures = true
 
+  config.before type: :system do
+    # Disable UI system tests by default when testing locally
+    skip('UI system testing is currently disabled') unless ENV.fetch('ENABLE_UI_TESTING', nil) == 'true'
```
david-yz-liu, line 95: An environment variable isn't a very elegant way of handling this. Since these tests are different, we can expect developers to run them separately from the main test suite.

You can configure rspec to ignore some paths by default, adding to a `.rspec` config file. See https://relishapp.com/rspec/rspec-core/v/3-11/docs/configuration/pattern or https://relishapp.com/rspec/rspec-core/v/3-11/docs/configuration/exclude-pattern (you'll need to experiment). On GitHub actions run two separate sets of tests, one with just the default `rspec` (everything except the system tests), and then one for the system tests.

spec/spec_helper.rb
```diff
@@ -80,6 +90,22 @@
...
+
+    # Configure the selenium webdriver for system tests
+    browser = ENV.fetch('DISABLE_HEADLESS_UI_TESTING', nil) == 'true' ? :chrome : :headless_chrome
+    driven_by :selenium, using: browser, screen_size: [1400, 1400],
```
david-yz-liu, line 99: This should be configured once globally, not in a `before` for each test. You should consult the Capybara documentation for how to do this.

docker-compose.yml
```diff
@@ -59,6 +59,10 @@ services:
       - PGDATABASETEST=markus_test
       - MARKUS__REPOSITORY__SSH_URL=ssh://git@localhost:2222/~/csc108
       - MARKUS__REPOSITORY__MARKUS_GIT_SHELL=/markus-git-shell.sh
+      - CAPYBARA_SERVER_PORT=${CAPYBARA_SERVER_PORT:-3434}
+      - CAPYBARA_SERVER_HOST=${CAPYBARA_SERVER_HOST:-0.0.0.0}
+    extra_hosts:
```
david-yz-liu, line 64: Is adding this key necessary?

anoverga17: Yes, I found that it is. Capybara otherwise, could not connect to the Chromedriver running on my machine. My understanding is that for Linux systems this hostname mapping allows the container to connect to my local machine.

spec/spec_helper.rb
```diff
@@ -80,6 +100,11 @@
   # instead of true.
   config.use_transactional_fixtures = true
 
+  config.before type: :system do
+    # Override the default driver used by rspec system tests
+    driven_by :selenium_remote_chrome
```
anoverga17, line 105: While I have moved what driver configuration I could globally, according to the rspec [system test documentation](https://relishapp.com/rspec/rspec-rails/v/5-1/docs/system-specs/system-spec) this `driven_by :selenium_remote_chrome`
setting is still needed to override the default driver
As mentioned [here](https://github.com/rails/rails/issues/34379) this behaviour is intended and I don't think is likely to change.

If we wanted this configured globally too, using the older [feature type](https://relishapp.com/rspec/rspec-rails/v/5-1/docs/feature-specs/feature-spec) rspec tests I think would work though my understanding is that it has slightly less functionality than system tests.

anoverga17, line 105: While I have moved what driver configuration I could globally, according to the rspec [system test documentation](https://relishapp.com/rspec/rspec-rails/v/5-1/docs/system-specs/system-spec) this `driven_by :selenium_remote_chrome`
setting is still needed to override the default driver
As mentioned [here](https://github.com/rails/rails/issues/34379) this behaviour is intended and I don't think is likely to change.

If we wanted this configured globally too, there are two alternatives: 
1. using the older [feature type](https://relishapp.com/rspec/rspec-rails/v/5-1/docs/feature-specs/feature-spec) rspec tests (these have less features though)
2. Overriding the base rails ActionDispatch::SystemTestCase class to use the custom driver (I'm not sure though if this is a good idea)

I would like your opinion though.

david-yz-liu: @anoverga17 I see, that's a bit disappointing but makes sense. You can keep this `config.before` block.

spec/spec_helper.rb
```diff
@@ -41,6 +43,21 @@
   exit 1
 end
 
+Capybara.register_driver :selenium_remote_chrome do |app|
+  chrome_options = Selenium::WebDriver::Chrome::Options.new(args: ['--no-sandbox', '--disable-gpu',
+                                                                   '--window-size=1400,1400'])
+  chrome_options.add_argument('--headless') unless ENV.fetch('DISABLE_HEADLESS_UI_TESTING', nil) == 'true'
+  Capybara::Selenium::Driver.new(app, browser: :remote, url: 'http://localhost:9515', capabilities: [chrome_options])
```
david-yz-liu, line 50: `localhost` doesn't work for me, so please change this back to `host.docker.internal`.

Or, you can help debug the following error:

```
Failure/Error: TCPSocket.open(conn_addr, conn_port, @local_host, @local_port)

          Errno::EADDRNOTAVAIL:
            Failed to open TCP connection to localhost:9515 (Cannot assign requested address - connect(2) for "localhost" port 9515)
```

david-yz-liu: @anoverga17 ah never mind, I rebuilt the docker containers and then the tests worked fine with `localhost`.

david-yz-liu (approved):

Issue #5953, opened 2022-04-09
Enable skipping "Not Collected" submissions when switching between assignments


jeffchen006:
Suppose we have the following submission summary:

| Group      | Marking State | 
| --- | ----------- |
| group_001   | In progress       | 
| group_002   | Not Collected     |
| group_003  | In progress |  

When I was trying to access next submission (shift + ->) from group_001, or to access previous submission (shift + <-) from group_003. Markus gives me a warning "There are no more results in this direction!". 

Obviously it is because group_002 did not submit anything so it blocks our "go to next/previous submission" operations. But it is somehow counterintuitive as there are other submissions before and after group_002. 

Is it possible to skip any submission which is "Not collected" when clicking the "Previous"/"Next" green arrows?

PR #6013, opened 2022-05-31
Issue - 5953

WeissSa:
Added ability to skip students without any students without results and refactored code.

## Motivation and Context
It allows TAs and Instructors to more navigate easier. Also solves issue #5953.

## Your Changes
Refactored next_group method so that it takes the closest group with a submission. Also made it so some of that logic was in the grouping model.

## Testing
Tested manually then ran rspec suite.

## Checklist

### DIFF
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -320,11 +320,8 @@ class ResultsController < ApplicationController
                               .where(assignment: assignment)
                               .joins(:group)
                               .order('group_name')
-      if params[:direction] == '1'
-        next_grouping = groupings.where('group_name > ?', grouping.group.group_name).first
-      else
-        next_grouping = groupings.where('group_name < ?', grouping.group.group_name).last
-      end
+      going_right = params[:direction] == '1'
+      next_grouping = grouping.get_next_grouping(going_right, groupings)
       next_result = next_grouping&.current_result
     elsif result.is_a_review? && current_role.is_reviewer_for?(assignment.pr_assignment, result)
       assigned_prs = current_role.grouping_for(assignment.pr_assignment.id).peer_reviews_to_others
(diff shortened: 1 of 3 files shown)


WeissSa:
Unsure why that test is failing. Does not seem to be related

david-yz-liu (commented):
This is good work. My biggest overall comment is that you should move the query for groupings into the method as well. Now that the query is student whether the current role is an instructor or TA: this is typically dealt with in other places by passing the current role object into the method.

Then in the controller method, have one branch for the peer review stuff, and the else branch be for calling the new method.

app/models/grouping.rb
```diff
@@ -694,6 +694,23 @@ def access_repo
     end
   end
 
+  def get_next_grouping(going_right, current_role)
```
david-yz-liu, line 697: Okay, just a couple of quick cleanup things now!

1. Add some documentation for this function.
2. `going_right` isn't a great name. The standard convention for this sort of behaviour is to name it `reversed` (so that `true` causes the method to go backward, and `false` goes forward).
3. Also, it would be good to make `going_right` the second parameter. In terms of function design, we try to order parameters in (rough) order of importance. Typically boolean flags are not very important (and in this case it's less important than the role needed to get the groupings).

app/controllers/results_controller.rb
```diff
@@ -335,12 +324,8 @@ def next_grouping
       end
       next_result = Result.find_by(id: next_grouping&.result_id)
     else
-      groupings = assignment.groupings.joins(:group).order('group_name')
-      if params[:direction] == '1'
-        next_grouping = groupings.where('group_name > ?', grouping.group.group_name).first
-      else
-        next_grouping = groupings.where('group_name < ?', grouping.group.group_name).last
-      end
+      reversed = params[:direction] == '1'
```
david-yz-liu, line 327: This is the opposite of the expected behaviour

WeissSa: Very true. Will fix

WeissSa:
Failed test seems to be unrelated (flaky?)

david-yz-liu (approved):

PR #6044, opened 2022-06-04
Small UI changes

WeissSa:
Added some conditions to text appearing in the assignment screen and modified how grouping language is treated for individuals.

## Motivation and Context
This should add clarity for students

## Your Changes
Removed message implying students can form a group in their section when they cannot and removed grace credits being mentioned when they are useless. Also added a variety of changes when someone indicates they want to work alone or can only work alone.

- [x] Breaking change

## Testing
I tested it manually and ran rspec suite. Was not sure if there were other testing options.

1. Where are the testing files for this (if they exist). Could not find a view folder under rspec
2. I have also given the group conundrum some thought. I think there are three main options, but all of them would be a pain. I will order them from most painful to least.
    1. Replace the name group with something more general that could be a single person(i.e. company, faction, etc). Or replace all mentions of group with "working with" and then when something is individual it will just be blank 
    2. Have an if/else whenever groups are mentioned. If it is a 1 person group then either do not say the thing about groups, or replace it with something implying it is an individual user
    3. When an assignment is individual, use a different view, or split the view into 2 sections: one for individual, one for groups. This way we only have to modify the one file and the two cases would be clearly organized (even if there is some repetition) I think. This could also be a great chance to use React. Since we could make an assignment template and then input props which imply individual or groups.

## Checklist


app/views/assignments/_read.html.erb
```diff
@@ -146,6 +151,12 @@
       <% end %>
     <% end %>
 <% end %>
+<% if @assignment.group_max == 1 %>
+    <h3><%= t('groups.student_interface.group_properties') %></h3>
+    <ul>
```
david-yz-liu, line 156: You can remove the list structure and just keep the text in a `<p>` tag.

app/views/assignments/_read.html.erb
```diff
@@ -172,4 +183,20 @@
     <% end %>
   <% end %>
 <% end %>
+  <% unless @grouping.nil? %>
+    <% if allowed_to?(:access_repo?, @grouping) %>
+      <% if @grouping.student_membership_number > 1 %>
+        <h3><%= t('groups.student_interface.url_group_repository') %></h3>
```
david-yz-liu, line 189: Let's keep this header as `Repository URL` for both individual and group assignments. (You can remove the unused locale string.)

app/views/assignments/show.html.erb
```diff
@@ -32,206 +32,201 @@
...
+        <% # If student has no group, inform them
+           if @grouping.nil? %>
+          <p class='notice'>
+            <% if @assignment.submission_rule.type == GracePeriodSubmissionRule && @assignment.group_max > 1%>
+              <%= t('groups.student_interface.no_group_yet') %>
+            <% end %>
+            <% if !@assignment.scanned_exam && @assignment.past_collection_date?(@current_role.section) %>
+              <%= t('assignments.due_date.final_due_date_passed') %>
+            <% end %>
+          </p>
```
david-yz-liu, line 35: revert this change

app/views/assignments/show.html.erb
```diff
@@ -32,206 +32,201 @@
...
 
-  <div class='pane block'>
-    <h2><%= t('groups.student_interface.group_information') %></h2>
-
```
david-yz-liu, line 38: revert this change

app/views/assignments/show.html.erb
```diff
@@ -32,206 +32,201 @@
...
     </div>
   </div>
+  <% if @assignment.group_max > 1 %>
+    <div class='pane block'>
+      <h2><%= t('groups.student_interface.group_information') %></h2>
+      <div class='block-content'>
+        <% # If student has no group, inform them
+           if @grouping.nil? %>
+          <p class='notice'>
+            <% if @assignment.submission_rule.type == GracePeriodSubmissionRule && @assignment.group_max > 1%>
```
david-yz-liu, line 42: This isn't the correct change. I think you meant to add this check for the other place I commented on above.

app/views/assignments/show.html.erb
```diff
@@ -32,206 +32,201 @@
...
 
-        <% # If students are allowed to form groups and
-           # the grace period for this assignment has not yet passed,
-           # render UI elements for group creation
-           if !@assignment.past_collection_date?(@current_role.section) && @assignment.student_form_groups %>
+          <% # If students are allowed to form groups and
+             # the grace period for this assignment has not yet passed,
+             # render UI elements for group creation
+             if !@assignment.past_collection_date?(@current_role.section) && @assignment.student_form_groups &&
+               @assignment.group_max > 1 %>
```
david-yz-liu, line 99: This change isn't necessary since you added the check for the `group_max > 1` at the top level.

app/views/assignments/show.html.erb
```diff
@@ -32,206 +32,201 @@
...
-
-    <div class='block-content'>
-      <% # If student has no group, inform them
-         if @grouping.nil? %>
-        <p class='notice'>
-          <%= t('groups.student_interface.no_group_yet') %>
-          <% if !@assignment.scanned_exam && @assignment.past_collection_date?(@current_role.section) %>
-            <%= t('assignments.due_date.final_due_date_passed') %>
+          <% if @assignment.is_timed && @assignment.group_max == 1 %>
+            <p><%= t('groups.student_interface.create_individual_folder_on_start') %></p>
```
david-yz-liu, line 51: This block will no longer ever appear (because of the `group_max > 1` check you added). You can just delete this if block and string, it doesn't add too much. But make sure to keep the `elsif` block below.

app/views/assignments/show.html.erb
```diff
@@ -32,206 +32,201 @@
...
-        <ul>
-          <% @grouping.student_memberships.includes(:role).each do |member|%>
-            <%= render partial: 'assignments/member',
-                       locals: { member: member,
-                                 grouping: @grouping } %>
-          <% end %>
-        </ul>
+        <% else %>
+          <% # Since the student has a group already, we display information about the group only %>
+          <% if @assignment.group_max > 1 %>
```
david-yz-liu, line 130: As above, you no longer need this check. (Make sure to revert any relevant indentation changes as well.)

app/views/assignments/show.html.erb
```diff
@@ -32,206 +32,201 @@
...
-        <% end %>
-      <% end %>
 
-      <% unless @grouping.nil? %>
-        <h3><%= t('groups.student_interface.group_options') %></h3>
-        <% if @grouping.group.group_name == @current_role.user_name &&
-          !@assignment.past_collection_date?(@current_role.section) && @assignment.group_max > 1 %>
-          <p><%= t('groups.student_interface.working_alone') %></p>
+            <%# For each member of this group, display name and status %>
+            <% if @grouping.student_membership_number == 1 %>
```
david-yz-liu, line 152: I'm not a fan of this change, please keep the original text

app/views/assignments/show.html.erb
```diff
@@ -32,206 +32,201 @@
...
-             @grouping&.extension.nil?%>
-          <% # Student has reached the group max and is not working alone
-             if @grouping.student_membership_number < @assignment.group_max &&
-               @grouping.group.group_name != @current_role.user_name %>
-            <%= button_tag t('groups.members.invite'), id: 'invite-student'%>
-          <% elsif @grouping.student_membership_number ==
-            @assignment.group_max &&
-            @grouping.group.group_name != @current_role.user_name %>
-            <p><%= t('groups.student_interface.may_not_add_more') %></p>
+        <% if !@grouping.nil? && @assignment.group_max > 1 %>
```
david-yz-liu, line 166: again, revert

app/views/assignments/show.html.erb
```diff
@@ -32,206 +32,201 @@
...
-          </li>
-        <% elsif @assignment.invalid_override %>
-          <li><%= Assignment.human_attribute_name(:invalid_override) %>.</li>
-        <% else %>
-          <li><%= t('groups.student_interface.students_work_alone') %></li>
-        <% end %>
-      </ul>
+        <h3><%= t('groups.student_interface.group_properties') %></h3>
+        <ul>
+          <% if @assignment.section_groups_only && @current_course.sections.any? && @assignment.group_max > 1 %>
```
david-yz-liu, line 210: revert

david-yz-liu (commented):
@WeissSa overall make sure to review the relevant section of the `show.html.erb` file; you should remove all redundant checks for `@assignment.group_max > 1` from the inner parts.

app/views/assignments/_read.html.erb
```diff
@@ -40,7 +40,8 @@
     <p><%= @assignment.submission_rule.class.human_attribute_name(:description) unless @assignment.is_timed %></p>
 
     <ul>
-      <% if @grouping.nil? %>
+      <% if @grouping.nil? && @assignment.group_max > 1 &&
```
david-yz-liu, line 43: Ah, I realized here that the old check (just `@grouping.nil?`) prevented the rest of the conditions/branches below from executing when `@grouping` is `nil`. Otherwise, for example, the `@grouping.start_time.nil?` check below raises an error when `@grouping` is `nil`.

Since the purpose is to only display the message when the student needs to create a group, you should use a nested `if` here, and leave the original condition unchanged.

app/views/assignments/_read.html.erb
```diff
@@ -172,4 +184,16 @@
     <% end %>
   <% end %>
 <% end %>
+  <% unless @grouping.nil? %>
```
david-yz-liu, line 187: Okay, so in terms of positioning, move this to above the "Submissions" section (right under the if condition on line 115)

david-yz-liu (approved):

PR #6045, opened 2022-06-04
allowing file rename on upload by manipulating states

Dhairya-Khara:
Allowing students to change the name of the uploaded file to match assignment requirements.
## Motivation and Context

Previously students had to manually rename their files before uploading. With the current changes, students will have an option
to rename their files during upload.

## Your Changes

**Description**:
`submission_file_upload_modal.jsx`

- Updated `this.state` to include `renameTo`
- Created method fileRenameInputBox that returns the correct input box based on the following rules: If only required files are allowed, a selection box is used for renaming. If there are required files but students can submit additional files, a datalist field is used. Otherwise, a textbox is used for renaming

`submission_file_manager.jsx`

-  Added an optional third argument for `handleCreateFiles` to store the updated file name. It will only be used if a single file is provided. 
- Passing the property `only_required_files` , names of the `required_files`  and `uploadTarget` to `<SubmissionFileUploadModal />`

`app/controllers/submissions_controller.rb`
- Updated `populate_file_manager` to send `only_required_files` and `required_files` for the assignment

`config/locales/common/en.yml`

- Added `select_filename` that is used in the upload modal

`config/locales/views/submissions/en.yml`

- Added `one_file_allowed` and `rename_file_to` that is used in the upload modal

## Testing

`submission_file_upload_modal.text.jsx` was created to test the new changes.
It tests if the correct file renaming option is rendered, enabled/disabled based on assignment parameters. Furthermore, it also tests if the correct rename options are given based on active directory in the user interface.

## Checklist

https://github.com/MarkUsProject/Wiki/pull/124


app/assets/javascripts/Components/Modals/submission_file_upload_modal.jsx
```diff
@@ -16,18 +17,69 @@ class SubmissionFileUploadModal extends React.Component {
 
   onSubmit = event => {
     event.preventDefault();
-    this.props.onSubmit(this.state.newFiles, this.state.unzip);
+    if (this.state.newFiles.length == 1) {
```
david-yz-liu, line 20: This check isn't necessary. You can always pass `this.state.single_file_name` to the `onSubmit` prop, and let it ignore the value if there's more than one file.

app/assets/javascripts/Components/Modals/submission_file_upload_modal.jsx
```diff
@@ -16,18 +17,69 @@ class SubmissionFileUploadModal extends React.Component {
...
   toggleUnzip = () => {
     this.setState({unzip: !this.state.unzip});
   };
 
+  handleNameChange = event => {
+    this.setState({single_file_name: event.target.value});
+  };
+
   render() {
+    let fileInput;
```
david-yz-liu, line 40: Pull this code out into a helper function. Also `fileInput` isn't a good name, since it sounds like the input where users select a file. Something like `fileRenameInput` is better.

app/assets/javascripts/Components/Modals/submission_file_upload_modal.jsx
```diff
@@ -16,18 +17,69 @@ class SubmissionFileUploadModal extends React.Component {
...
+    if (this.state.newFiles.length == 1) {
```
david-yz-liu, line 41: In Javascript you should generally use `===` instead of `==` for equality checking (the former is stricter than the latter).

See e.g. https://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons

app/assets/javascripts/Components/Modals/submission_file_upload_modal.jsx
```diff
@@ -16,18 +17,69 @@ class SubmissionFileUploadModal extends React.Component {
...
+      if (this.props.onlyRequiredFiles) {
+        fileInput = (
+          <select
+            className={"select-filename"}
+            onChange={this.handleNameChange}
+            value={this.state.single_file_name}
+          >
+            <option key={"select_file"}>Please Select File Name</option>
```
david-yz-liu, line 49: This string should be internationalized, as should all other user-facing strings you added

app/assets/javascripts/Components/Modals/submission_file_upload_modal.jsx
```diff
@@ -57,6 +109,19 @@ class SubmissionFileUploadModal extends React.Component {
                 onChange={this.handleFileUpload}
               />
             </div>
+            {this.state.newFiles.length == 1 ? (
```
david-yz-liu, line 112: This should all be part of the helper function as well, so that the only new change here is including `{fileInput}`.

app/assets/javascripts/Components/__tests__/submission_file_upload_modal.test.jsx
```diff
@@ -0,0 +1,34 @@
+import {shallow} from "enzyme";
+import SubmissionFileUploadModal from "../Modals/submission_file_upload_modal";
+
+describe("For the SubmissionFileUploadModal", () => {
+  it("checks if only selection menu renders", () => {
```
david-yz-liu, line 5: These test descriptions should include an English description of the state involved, not just what you're checking.

Moreover, there are two dimensions of state that you should be considering: the `onlyRequiredFiles`/`requiredFiles` props values, *and* how many files are actually being uploaded, i.e., are in the state (0/1/2+). This is a good time to use nested `describe` blocks, each with their own `beforeEach` that you can use to set up common test state across multiple tests.

app/assets/javascripts/Components/submission_file_manager.jsx
```diff
@@ -87,15 +91,19 @@ class SubmissionFileManager extends React.Component {
...
+  handleCreateFiles = (files, unzip, single_file_name = "") => {
     if (
       !this.props.starterFileChanged ||
       confirm(I18n.t("assignments.starter_file.upload_confirmation"))
     ) {
       const prefix = this.state.uploadTarget || "";
       this.setState({showUploadModal: false, uploadTarget: undefined});
       let data = new FormData();
-      Array.from(files).forEach(f => data.append("new_files[]", f, f.name));
+      if (files.length == 1) {
```
david-yz-liu, line 102: `===`

app/controllers/submissions_controller.rb
```diff
@@ -124,6 +124,8 @@ def populate_file_manager
       end
       entries = get_all_file_data(revision, grouping, '')
     end
+    entries.append({ only_required_files: assignment.only_required_files,
```
david-yz-liu, line 127: Overall, it's not good to add an extra element of a different structure to an otherwise homogeneous list. In this case, a better approach would be to change the `json` response to be an object where one of the keys is the previous `entries`, and the other keys are `only_required_files` and `required_files`. Then the React component can extract each of these three keys in the response object.

app/controllers/submissions_controller.rb
```diff
@@ -124,6 +124,8 @@ def populate_file_manager
...
+                     required_files: assignment.assignment_properties_config['assignment_files_attributes'] })
```
david-yz-liu, line 128: `assignment_properties_config` is a heavy-duty method; while it's creative that you found it, a much more efficient approach is to use a query to explicitly extract the `filename`s of the associated `assignment_files` for this `assignment`. (Remember that you only need a list of the filenames, not ids or other attributes.)

config/locales/common/en.yml
```diff
@@ -20,6 +22,7 @@ en:
     launch: Launch MarkUs
   markus: MarkUs
   not_applicable: N/A
+  one_file_allowed: Supports only one file
```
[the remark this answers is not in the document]
david-yz-liu: Also, please update the wording to `File renaming is only supported when one file is selected.`

app/assets/javascripts/Components/Modals/submission_file_upload_modal.jsx
```diff
@@ -16,17 +17,88 @@ class SubmissionFileUploadModal extends React.Component {
...
+  };
+
+  fileRenameInputBox = () => {
+    let fileRenameInput;
+    let filesToShow;
+    if (this.props.uploadTarget) {
+      let filenames = this.props.requiredFiles.filter(
+        filename =>
+          filename.indexOf(this.props.uploadTarget) === 0 &&
+          filename.slice(this.props.uploadTarget.length).indexOf("/") === -1
```
david-yz-liu, line 42: Use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes, and note that this method takes a `position` argument.

app/assets/javascripts/Components/Modals/submission_file_upload_modal.jsx
```diff
@@ -16,17 +17,88 @@ class SubmissionFileUploadModal extends React.Component {
...
+      );
+      filesToShow = filenames.map(filename => filename.slice(this.props.uploadTarget.length));
+    } else {
+      filesToShow = this.props.requiredFiles.filter(filename => filename.indexOf("/") === -1);
```
david-yz-liu, line 46: see above comment about `includes`

config/locales/common/en.yml
```diff
@@ -5,13 +5,15 @@ en:
   all: All
   apply: Apply
   cancel: Cancel
+  change_filename: Change File Name
```
david-yz-liu, line 8: Change key to `select_filename`, and value to `Select Filename`

app/assets/javascripts/Components/Modals/submission_file_upload_modal.jsx
```diff
@@ -57,6 +129,8 @@ class SubmissionFileUploadModal extends React.Component {
                 onChange={this.handleFileUpload}
               />
             </div>
+            <h3>{I18n.t("file_name")}</h3>
```
david-yz-liu, line 132: Don't make this an `h3`. Instead, this should be a *label* for the input element (this is important for accessibility). Note that labels need to have an `htmlFor` element that matches the `id` element of the input (which can be a `select` as well).

Visually, the label should:

1. have the text `(optional) Rename file to`
2. appear beside, rather than above, the input element, like the other label/input pairs on the modal

app/assets/javascripts/Components/submission_file_manager.jsx
```diff
@@ -87,15 +91,19 @@ class SubmissionFileManager extends React.Component {
...
+  handleCreateFiles = (files, unzip, single_file_name = "") => {
     if (
       !this.props.starterFileChanged ||
       confirm(I18n.t("assignments.starter_file.upload_confirmation"))
     ) {
       const prefix = this.state.uploadTarget || "";
       this.setState({showUploadModal: false, uploadTarget: undefined});
       let data = new FormData();
-      Array.from(files).forEach(f => data.append("new_files[]", f, f.name));
+      if (files.length === 1) {
```
david-yz-liu, line 102: You need an extra check for the rename parameter being truthy (so it isn't `undefined`/`null`/empty string). A common idiom we use for this is `!!single_file_name`, which will convert the value into a boolean explicitly

app/controllers/submissions_controller.rb
```diff
@@ -124,7 +124,14 @@ def populate_file_manager
       end
       entries = get_all_file_data(revision, grouping, '')
     end
-    render json: entries
+
+    response = {
+      entries: entries,
+      only_required_files: assignment.only_required_files,
+      required_files: assignment.assignment_files.collect(&:filename)
```
david-yz-liu, line 131: This is better but can still be improved: use `pluck`, which will access the fields directly from the database. (`collect` will first instantiate the `AssignmentFile` objects and then grab their `filename` attributes, which is slower)

Also please call `sort` on the result to make sure the filenames are sorted (and test this out by creating required files with names that are out of order alphabetically)

david-yz-liu (commented):
@Dhairya-Khara great work! I left a couple more comments for you to work on, but I think we're very close.

david-yz-liu (commented):
@Dhairya-Khara okay great, just one last change: I'd like `renameTo` to start out empty rather than be automatically populated when a user selects a file. Keeping the input blank by default will make it clear that the field is optional.

david-yz-liu (approved):

PR #6049, opened 2022-06-16
Jest for marks_panel

WeissSa:
Added tests to marks_panel

## Motivation and Context
This should increase test coverage for react components.

## Your Changes
I added tests for marks_panel.

## Checklist


app/assets/javascripts/Components/__tests__/marks_panel.test.jsx
```diff
@@ -0,0 +1,425 @@
...
+      oldMark: {},
+      description: " ",
+      updateMark: jest.fn(),
+      destroyMark: jest.fn(),
+      toggleExpanded: jest.fn(),
+    };
+    wrapper = shallow(<CheckboxCriterionInput {...basicProps} />);
+  });
+  it("should toggle expand and contract upon clicking the expand/contract button", () => {
+    basicProps.toggleExpanded.mockImplementation(() => {
```
david-yz-liu, line 87: It seems to me that this can be simplified by putting this function inside the call to `jest.fn()` above. (I'm fine with it being defined on every test, even if the function is only called in some of the tests).

app/assets/javascripts/Components/__tests__/marks_panel.test.jsx
```diff
@@ -0,0 +1,425 @@
...
+      marks: [],
+      levels: [],
+      assigned_criteria: [],
+      annotations: [],
+      updateMark: jest.fn(),
+      revertToAutomaticDeductions: jest.fn(),
+      findDeductiveAnnotation: jest.fn(),
+    };
+  });
+  it("displays no criterion if marks is an empty list", () => {
```
david-yz-liu, line 35: Include a blank line before and between each test case

app/assets/javascripts/Components/__tests__/marks_panel.test.jsx
```diff
@@ -0,0 +1,425 @@
...
+  it("it delete upon the delete button being pressed", () => {
+    basicProps.unassigned = false;
+    basicProps.mark = 1;
+    wrapper.setProps(basicProps);
+
+    wrapper.find(`#checkbox_criterion_${basicProps.id}_destroy`).simulate("click");
+    expect(basicProps.destroyMark).toHaveBeenCalledWith(undefined, basicProps.id);
+  });
+  it("correctly updates mark to max_mark and 0 when clicking on respective buttons", () => {
+    basicProps.updateMark.mockImplementation((id, new_mark) => {
```
david-yz-liu, line 107: same comment as above

app/assets/javascripts/Components/Result/marks_panel.jsx
```diff
@@ -219,7 +233,25 @@ class CheckboxCriterionInput extends React.Component {
...
+CheckboxCriterionInput.propTypes = {
+  id: PropTypes.number.isRequired,
+  max_mark: PropTypes.number,
+  mark: PropTypes.number,
+
+  expanded: PropTypes.bool.isRequired,
+  unassigned: PropTypes.bool.isRequired,
+  released_to_students: PropTypes.bool.isRequired,
+
+  oldMark: PropTypes.oneOfType([PropTypes.object, PropTypes.array]),
```
david-yz-liu, line 245: This should just be `PropTypes.object` (for all three input components). You should check how you're settings these values in the tests.

app/assets/javascripts/Components/Result/marks_panel.jsx
```diff
@@ -219,7 +233,25 @@ class CheckboxCriterionInput extends React.Component {
   }
 }
 
-class FlexibleCriterionInput extends React.Component {
+CheckboxCriterionInput.propTypes = {
+  id: PropTypes.number.isRequired,
```
david-yz-liu, line 237: Yeah so in general grouping/ordering of the props is done pretty inconsistently.

I think the best approach for now is just to order these props alphabetically (and don't include any blank lines).

app/assets/javascripts/Components/Result/marks_panel.jsx
```diff
@@ -402,7 +442,31 @@ class FlexibleCriterionInput extends React.Component {
...
-class RubricCriterionInput extends React.Component {
+FlexibleCriterionInput.propTypes = {
+  expanded: PropTypes.bool.isRequired,
+  unassigned: PropTypes.bool.isRequired,
+  released_to_students: PropTypes.bool.isRequired,
+  bonus: PropTypes.bool,
+  override: PropTypes.bool,
+
+  id: PropTypes.number.isRequired,
+  mark: PropTypes.number.isRequired,
```
david-yz-liu, line 453: This isn't required (can be `null`/`undefined`). Same for rubric criteria.

app/assets/javascripts/Components/Result/marks_panel.jsx
```diff
@@ -219,7 +233,25 @@ class CheckboxCriterionInput extends React.Component {
   }
 }
 
-class FlexibleCriterionInput extends React.Component {
+CheckboxCriterionInput.propTypes = {
+  id: PropTypes.number.isRequired,
+  max_mark: PropTypes.number,
```
david-yz-liu, line 238: I think `max_mark` is required (for all three criterion input types)

david-yz-liu (commented):
@WeissSa great job! The tests themselves are very comprehensive. The only two tests I'd like you to add are for when marks are released:

1. For flexible criteria, check that there's no number input field
2. For rubric criteria, check that clicking on the level `tr`s have no effect (i.e. check that the `updateMark` function or something isn't called).

app/assets/javascripts/Components/__tests__/marks_panel.test.jsx
WeissSa: Each time I mock toggleExpanded it is in a different describe block and refers to a different basicProps. This results in it only being mocked once per test suite. I don't necessarily think this would simplify it or maybe I am misunderstanding your suggestion. For now I will leave it as-is until i get clarification

app/assets/javascripts/Components/__tests__/marks_panel.test.jsx
WeissSa: Same comment as above for me (it is only mocked once per describe)

WeissSa (commented):
I implemented most of the changes, but am confused about 2 of the comments. Let me know.

WeissSa:
Also for the 2nd test you wanted me to add, I found the ```tr```'s do call updateMark I think. Again maybe I am misunderstanding, but their onClick calls handleChange which then calls handleMark (so I added a test to make sure this does happen)

app/assets/javascripts/Components/__tests__/marks_panel.test.jsx
david-yz-liu: @WeissSa What I suggested will simplify the test cases themselves by moving the setup code into a setup function (`beforeEach`). You're correct that currently these functions are only mocked in one or two tests, but this may change in the future.

WeissSa: Ah okay! I just wanted to make sure we are on the same page then. I'll move the location of both next time I get a chance to.

WeissSa:
> > Also for the 2nd test you wanted me to add, I found the `tr`'s do call updateMark I think. Again maybe I am misunderstanding, but their onClick calls handleChange which then calls handleMark (so I added a test to make sure this does happen)
> 
> @WeissSa When the marks are released, clicking on the different rows shouldn't cause the mark to actually change. I don't know the details of how this is implemented (maybe the onClick function isn't called, or maybe the onClick function itself contains a check for whether the marks are released or not), but that's what you should check.

I think my eyes skipped the part about marks being released. I'll update the tests at the next opportunity. Sorry!

david-yz-liu (approved):

PR #6055, opened 2022-06-24
Issue 5084

WeissSa:
Fixes issue#5084

## Motivation and Context
This allows instructors to set a distribution on randomly assigning graders to groupings.

## Your Changes
Added a modal which allows instructors to assign a weighting to each grader. The groupings are then given out in order to follow the weighting as closely as possible.

## Testing
Wrote jest tests for new component and modified Grouping.spec with a new test for back-end change.

I know my change to the back-end breaks some rules and am wondering how to make it neater. It definitely works, but I get console errors. I wanted to show I did have a finished draft which is why I am creating a PR despite these errors.

## Checklist


WeissSa:
Test failing is legitimate and likely related to the console errors I am getting. Unsure how to fix.

WeissSa:
Testing passed on local!

app/models/grouping.rb
```diff
@@ -107,8 +107,22 @@ class Grouping < ApplicationRecord
...
+
+    assign_tas(grouping_ids, ta_ids, assignment) do |grouping_ids_, _|
+      ta_ids_ = []
+      # Create 1 TA for each group they are supposed to be assigned to
+      ta_ids.each do |id|
+        # Make sure TA is supposed to be assigned to gorup
+        unless weightings[id].nil?
+          (ta_ids_ << Array([id].cycle(weightings[id].to_i))).flatten!
+        end
+      end
```
WeissSa, line 125: I do think this is slightly clunky, but it is functional!

app/assets/javascripts/Components/Modals/graders_distribution_modal.jsx
```diff
@@ -0,0 +1,73 @@
...
+
+export class GraderDistributionModal extends React.Component {
+  static defaultProps = {
+    override: false,
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      override: this.props.override,
```
david-yz-liu, line 13: I don't think this state is necessary (or at least it doesn't seem to be used anywhere). If this is the case you can remove it, and in fact the overridden constructor altogether, since there's no other state.

app/assets/javascripts/Components/Modals/graders_distribution_modal.jsx
```diff
@@ -0,0 +1,73 @@
...
+    };
+  }
+
+  componentDidMount() {
+    this.props.graders.forEach(grader => {
+      !this.props.weightings[grader._id] ? (this.props.weightings[grader._id] = 1) : null;
```
david-yz-liu, line 19: Okay so overall, you shouldn't mutate `this.props` in a React component (a component should treat its props as immutable). While one alternative would be to store mutable data in `this.state`, in this particular case I don't think that's necessary.

We don't need to store the weights at all---instead, in the `onSubmit` function you can create an object that maps the grader ids to weights, and then pass that object to `this.props.onSubmit` as an argument. (Rather than accessing the props from the `onSubmit` callback function in the parent component.)

app/assets/javascripts/Components/Modals/graders_distribution_modal.jsx
```diff
@@ -0,0 +1,73 @@
...
+    });
+    Modal.setAppElement("body");
+  }
+
+  onSubmit = event => {
+    event.preventDefault();
+    this.props.onSubmit();
+  };
+
+  createGraderRow = grader => {
```
david-yz-liu, line 29: Generally use `render*` to name methods that return JSX for rendering.

app/assets/javascripts/Components/Modals/graders_distribution_modal.jsx
```diff
@@ -0,0 +1,73 @@
...
+    return (
+      <div className="flex-row-expand" key={grader.user_name}>
+        <label className="modal-inline-label">{grader.user_name}:</label>
```
david-yz-liu, line 32: Labels should always be associated with an input using the `for` attribute (see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label)

david-yz-liu, line 32: Also, in general don't use colons in headings or labels (just the user name is fine)

app/assets/javascripts/Components/Modals/graders_distribution_modal.jsx
```diff
@@ -0,0 +1,73 @@
...
+          max="100"
+          defaultValue={this.props.weightings[grader._id] || 1}
+          onChange={event => {
+            this.props.weightings[grader._id] = parseFloat(event.target.value);
+          }}
+        />
+      </div>
+    );
+  };
+  render() {
```
david-yz-liu, line 47: Use one blank line to separate methods

app/assets/javascripts/Components/Modals/graders_distribution_modal.jsx
```diff
@@ -0,0 +1,73 @@
...
+    return (
+      <Modal
+        className="react-modal dialog"
+        isOpen={this.props.isOpen}
+        onRequestClose={this.props.onRequestClose}
+      >
+        <form id="grader-form-random" onSubmit={this.onSubmit}>
+          <h2>{I18n.t("graders.weight")}:</h2>
```
david-yz-liu, line 55: See above comment about colons

app/assets/javascripts/Components/Modals/graders_distribution_modal.jsx
```diff
@@ -0,0 +1,73 @@
...
+          <div className={"modal-container-vertical"}>
+            {this.props.graders.map(grader => this.createGraderRow(grader))}
+          </div>
+          <div className={"modal-container"}>
+            <input type="submit" value={I18n.t("graders.submit")} />
```
david-yz-liu, line 60: Generally, button labels should be capitalized.

But also in this particular case, let's go with "Randomly Assign Graders" rather than the generic "Submit".

app/assets/javascripts/Components/Modals/graders_distribution_modal.jsx
```diff
@@ -0,0 +1,73 @@
...
+  onSubmit = event => {
+    event.preventDefault();
+    this.props.onSubmit();
+  };
+
+  createGraderRow = grader => {
+    return (
+      <div className="flex-row-expand" key={grader.user_name}>
+        <label className="modal-inline-label">{grader.user_name}:</label>
+        <input
```
david-yz-liu, line 33: Make these inputs required

app/assets/javascripts/Components/Modals/graders_distribution_modal.jsx
```diff
@@ -0,0 +1,73 @@
...
+          id={`input-${grader.user_name}`}
+          type="Number"
```
david-yz-liu, line 35: I think it's more standard to use lowercase for input types (e.g., `number` instead of `Number`)

david-yz-liu (commented):
@WeissSa great work, definitely on the right track! I left some inline comments, but also overall:

1. Make sure to update the changelog.
2. Make a PR to update the corresponding Wiki page to document this new feature.
3. The modal should have a brief set of instructions for the instructor, explaining the purpose of the weights and the default behaviour (something like "if all weights are 1 then groups will be evenly distributed across the graders"). And then point to the wiki for more information.

app/assets/javascripts/Components/graders_manager.jsx
```diff
@@ -19,13 +22,34 @@ class GradersManager extends React.Component {
...
+      isGradeerDistributionModalOpen: false,
+      weightings: {},
     };
   }
 
   componentDidMount() {
     this.fetchData();
   }
 
+  showDistributionModal = () => {
```
david-yz-liu, line 34: different names should be as consistent as possible; a better name here is `openGraderDistributionModal`

app/assets/javascripts/Components/graders_manager.jsx
```diff
@@ -79,20 +103,18 @@ class GradersManager extends React.Component {
...
-      alert(I18n.t("groups.select_a_group"));
-      return;
-    }
-
-    if (graders.length === 0) {
-      alert(I18n.t("graders.select_a_grader"));
-      return;
-    }
+    let graders = this.assignedGraders();
+    this.setState({
```
david-yz-liu, line 114: a separate `setState` call here isn't necessary; you can reset this state in the implementation of `fetchData`

app/assets/javascripts/Components/graders_manager.jsx
```diff
@@ -105,15 +127,38 @@ class GradersManager extends React.Component {
...
-        graders: graders,
+        graders: Object.keys(groupsPerGrader).sort(
+          (a, b) => groupsPerGrader[a] - groupsPerGrader[b]
+        ),
+        groups_per_grader: Object.values(groupsPerGrader).sort((a, b) => a - b),
       },
     }).then(this.fetchData);
   };
 
+  calculateGroupsPerGrader = (graders, groups) => {
```
david-yz-liu, line 138: This type of calculation work is normally done in the back-end, so you should move it to the main model method that's doing the heavy lifting.

david-yz-liu (commented):
(I accidentally submitted the review before I'd made all of my comments, hence the "double-review"!)

app/assets/javascripts/Components/Modals/graders_distribution_modal.jsx
david-yz-liu: @WeissSa you can do this by using plain Javascript to interact with the input elements. The easiest way to set it up in this case is to use a **ref** to refer to the form itself, then go through its input values.

See:

- https://reactjs.org/docs/uncontrolled-components.html
- https://developer.mozilla.org/en-US/docs/Web/API/FormData/FormData

app/assets/javascripts/Components/Modals/graders_distribution_modal.jsx
```diff
@@ -0,0 +1,80 @@
...
+  componentDidMount() {
+    this.props.graders.forEach(grader => {
+      this.props.setWeighting(grader._id, 1);
+    });
+    Modal.setAppElement("body");
+  }
+
+  onSubmit = event => {
+    event.preventDefault();
+    console.log(this.input); //logs: { current: null }
```
david-yz-liu, line 24: @WeissSa I looked into this and you should try accessing `this.input.current` directly. The console log does display `null`, but if you expand the object in the console (or access `current` directly) you'll get a form element. Not sure why the display seems to be inconsistent.

WeissSa: When I try logging this.input.current I get null and if I try inputting this.current.value as a FormData object then I get an error (because null cannot have properties). Is it possible I am doing something else wrong? Could it be that I think our version of React might not allow for useRef? I will push my code again if it helps.

app/assets/javascripts/Components/Modals/graders_distribution_modal.jsx
```diff
@@ -0,0 +1,78 @@
...
+
+  componentDidMount() {
+    Modal.setAppElement("body");
+  }
+
+  onSubmit = event => {
+    event.preventDefault();
+    const weightings = {};
+    const form = new FormData(this.input.current);
+    form.forEach((key, value) => {
```
david-yz-liu, line 23: Instead of this loop, you should be able to use `Object.fromEntries` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries)

app/assets/javascripts/Components/Modals/graders_distribution_modal.jsx
```diff
@@ -0,0 +1,78 @@
...
+      <div className="flex-row-expand" key={grader.user_name}>
+        <label htmlFor={`input-${grader.user_name}`} className="modal-inline-label">
+          {grader.user_name}
+        </label>
+        <input
+          id={`input-${grader.user_name}`}
+          type="number"
+          step="0.01"
+          min="0"
+          max="100"
```
david-yz-liu, line 40: No need for a `max` attribute (since these weights aren't necessarily percentages)

app/assets/javascripts/Components/graders_manager.jsx
```diff
@@ -4,6 +4,9 @@ import {Tab, Tabs, TabList, TabPanel} from "react-tabs";
 
 import {withSelection, CheckboxTable} from "./markus_with_selection_hoc";
 import {selectFilter} from "./Helpers/table_helpers";
+import {GraderDistributionModal} from "./Modals/graders_distribution_modal";
+
+import {sum} from "lodash";
```
david-yz-liu, line 9: This import is no longer necessary

app/assets/javascripts/Components/Modals/graders_distribution_modal.jsx
```diff
@@ -0,0 +1,78 @@
...
+      <Modal
+        className="react-modal dialog"
+        isOpen={this.props.isOpen}
+        onRequestClose={this.props.onRequestClose}
+      >
+        <form id="grader-form-random" onSubmit={this.onSubmit} ref={this.input}>
+          <div className={"modal-container-vertical"}>
+            <h2>{I18n.t("graders.weight")}</h2>
+            <p className="word-wrap">{I18n.t("graders.random_instruction")}</p>
+            <a href="https://github.com/MarkUsProject/Wiki/blob/release/Instructor-Guide--Assignments--Assigning-Graders.md">
```
david-yz-liu, line 60: This isn't exactly quite right, as we support linking to a specific version of the documentation. In other places in the code this is done with the `@markus_version` instance variable, however this isn't available in the pure javascript code.

The way we've handled this in other situations is to add a line like `const MARKUS_VERSION = <%= @markus_version %>;` in the view layouts, creating a global constant that you can access in Javascript code. Check out how we do this for `AUTH_TOKEN`.

app/assets/javascripts/Components/Modals/graders_distribution_modal.jsx
```diff
@@ -0,0 +1,78 @@
...
+              {I18n.t("graders.wiki")}
```
david-yz-liu, line 61: This text by itself looks a bit strange. A complete sentence would be good, e.g. "See our [Wiki] for more information." Also you can combine this with the text in the previous `p` tag.

app/controllers/graders_controller.rb
```diff
@@ -123,10 +123,10 @@ def global_actions
           end
         end
         if found_empty_submission
-          randomly_assign_graders(filtered_grouping_ids, grader_ids)
+          randomly_assign_graders(filtered_grouping_ids, grader_ids, params[:weightings])
```
david-yz-liu, line 126: I'd like to see a bit of error-checking here. Convert all of the weightings to floats here, and if one of the conversion fails, flash an error message (you'll need to create a new one) and respond with a `head :bad_request` as above. That way you can pass in a list of floats to `randomly_assign_graders` and don't need to worry about conversions in that function.

app/models/grouping.rb
```diff
@@ -107,8 +107,24 @@ class Grouping < ApplicationRecord
   # Assigns a random TA from a list of TAs specified by +ta_ids+ to each
   # grouping in a list of groupings specified by +grouping_ids+. The groupings
   # must belong to the given assignment +assignment+.
-  def self.randomly_assign_tas(grouping_ids, ta_ids, assignment)
-    assign_tas(grouping_ids, ta_ids, assignment) do |grouping_ids_, ta_ids_|
+  def self.randomly_assign_tas(grouping_ids, ta_ids, weightings_arr, assignment)
```
david-yz-liu, line 110: Make sure to update the documentation of this method as well.

app/controllers/graders_controller.rb
david-yz-liu: Oh and one more error check: all of the weightings should be `>= 0`, and their sum should be `> 0`. (We can't simply rely on client-side checking, we should check this on the server as well.)

david-yz-liu: And finally, please add some new controller test cases for these checks :)

app/models/grouping.rb
```diff
@@ -107,8 +107,24 @@ class Grouping < ApplicationRecord
...
+    end
+
+    assign_tas(grouping_ids, ta_ids, assignment) do |grouping_ids_, _|
+      ta_ids_ = []
+      # Create 1 TA for each group they are supposed to be assigned to
+      ta_ids = ta_ids.sort_by { |ta| weightings[ta] }
+      ta_ids.each do |id|
+        # Make sure TA is supposed to be assigned to gorup
+        unless weightings[id].nil?
+          (ta_ids_ << Array([id].cycle(weightings[id].to_i))).flatten!
```
david-yz-liu, line 125: Instead of using `cycle` you can repeat arrays with `*`, e.g. `[id] * 10`.

And instead of flattening the whole thing (inefficient inside a loop) you can use `flat_map` (https://ruby-doc.org/core-2.7.6/Enumerable.html#method-i-flat_map) to initialize the list, i.e. `ta_ids_ = ta_ids.flat_map do |id| ...`

config/locales/views/graders/en.yml
```diff
@@ -3,6 +3,7 @@ en:
   graders:
     actions:
       assign_grader: Assign grader(s)
+      randomly_assign: Randomly Assign Graders
```
david-yz-liu, line 6: Ah, you don't need to add this, it's fine to have the button show the text of `randomly_assign_graders` below.

app/controllers/graders_controller.rb
WeissSa: According to https://www.rubydoc.info/stdlib/core/String:to_f converting to a float cannot fail. Do you still want me to add a test just in case? Or alternatively do you want me to manually test for invalid strings?

david-yz-liu: @WeissSa aha, yes great point! Use https://ruby-doc.org/core-3.1.2/Kernel.html#method-i-Float instead (which can raise an exception if the argument doesn't represent a float)

WeissSa:
Force push was because I forgot to pull my merge.

app/controllers/graders_controller.rb
WeissSa: Oh okay! Sounds good

app/controllers/graders_controller.rb
```diff
@@ -122,11 +122,18 @@ def global_actions
...
         end
-        if found_empty_submission
-          randomly_assign_graders(filtered_grouping_ids, grader_ids)
-          flash_now(:info, I18n.t('graders.group_submission_no_files'))
-        else
-          randomly_assign_graders(grouping_ids, grader_ids)
+        begin
+          weights = params[:weightings].map { |weight| Float(weight) }
+          if weights.reduce(:+) == 0 || weights.select(&:negative?).length > 0
+            head :bad_request
```
david-yz-liu, line 128: Whenever you use a `head` or `render` method, it should immediately be followed by a `return` to stop the rest of the function from executing. Otherwise we open ourselves up to getting "double render" errors.

Also, please flash an error message here describing the problem.

app/controllers/graders_controller.rb
```diff
@@ -122,11 +122,18 @@ def global_actions
...
+          elsif found_empty_submission
+            randomly_assign_graders(filtered_grouping_ids, grader_ids, weights)
+            flash_now(:info, I18n.t('graders.group_submission_no_files'))
+          else
+            randomly_assign_graders(grouping_ids, grader_ids, weights)
+          end
+        rescue StandardError
+          head :bad_request
```
david-yz-liu, line 136: Same comments as above. Note that you can grab the exception object and then just flash the `message` of that exception.

app/models/grouping.rb
```diff
@@ -105,10 +105,22 @@ class Grouping < ApplicationRecord
...
-  # grouping in a list of groupings specified by +grouping_ids+. The groupings
+  # grouping in a list of groupings specified by +grouping_ids+.
+  # The ratio of groupings to ta is specified in +weightings_arr+
+  # which is a parallel array to +ta_ids+. The groupings
   # must belong to the given assignment +assignment+.
-  def self.randomly_assign_tas(grouping_ids, ta_ids, assignment)
-    assign_tas(grouping_ids, ta_ids, assignment) do |grouping_ids_, ta_ids_|
+  def self.randomly_assign_tas(grouping_ids, ta_ids, weightings_arr, assignment)
+    # Create a hash of TA's to the number of groups they are supposed to mark
+    total = weightings_arr.reduce(:+)
```
david-yz-liu, line 114: Oh I must have missed this earlier, but you can just do `.sum` instead of `.reduce(:+)`

config/locales/views/graders/en.yml
```diff
@@ -18,9 +18,12 @@ en:
     help: Assign graders to specific student groups and criteria. By default, graders will be assigned to all criteria.
     hide_unassigned_criteria: Only show assigned criteria to graders (if unselected, all graders will be able to see all criteria)
     no_grader_assigned: 'Default grader: course instructor(s)'
+    random_instruction: Assign a weight for randomized distribution
```
david-yz-liu, line 21: Okay, please reword to "Assign a weight for each grader for randomized assignment. With the default weights of 1, all graders will be assigned an equal number of groups." Make sure to include the punctuation.

app/assets/javascripts/Components/Modals/graders_distribution_modal.jsx
```diff
@@ -0,0 +1,79 @@
...
+        className="react-modal dialog"
+        isOpen={this.props.isOpen}
+        onRequestClose={this.props.onRequestClose}
+      >
+        <form id="grader-form-random" onSubmit={this.onSubmit} ref={this.input}>
+          <div className={"modal-container-vertical"}>
+            <h2>{I18n.t("graders.weight")}</h2>
+            <p className="word-wrap">
+              {I18n.t("graders.random_instruction")}
+              <br></br>
```
david-yz-liu, line 57: You can remove these. It's fine to have the link below as part of the same paragraph/line as the instructions.

david-yz-liu (approved):

PR #6057, opened 2022-06-27
Added new API submission route for students

anoverga17:
## Motivation and Context

As part of our integration with Jupyter Hub, we want students to be able to submit assignments through the API. This pull request aims to add an additional API route for students that allows them to submit a file. 

## Your Changes

**Description**:
- Allowed students access to API key reset button in the account settings page
- Added new route `/api/courses/:course_id/assignments/:assessment_id/submit_file`. This route is for student use only and allows them to submit a file via the API (if the instructor allows it).
- Added new `Allow students to submit through the MarkUs API` button that enables/disables this feature. This adds an `api_submit` attribute to the Assignment properties model.
    This button looks as follows:  
    api toggle button

## Testing

Rspec tests were added to thoroughly test this new API route. In addition, manual testing using curl and the markus-jupyter-hub extension was done to ensure a student could submit using the API.

When first trying to communicate with the frontend Jupyter Hub extension, I received a 
```
Access to fetch at 'http://localhost:3000/csc108/api/courses/1/assignments/1/submission_files/submit_file' 
from origin 'http://127.0.0.1:8888' has been blocked by CORS policy: Response to preflight request doesn't 
pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. 
If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with 
CORS disabled.
```
error. My research has indicated that this is because we do not currently have a CORS policy in the backend. While I managed to resolve this locally for myself by installing the `rack-cors` gem and setting an allow any origin, I'm not sure if this is a problem exclusive to development or if this will be a problem in production too. My inclination is that it might since I don't think MarkUs and Jupyter Hub are served from the same server although I am not sure.

Either way, this seems like something for another PR. Let me know what you think might be a good approach to resolve this.

## Checklist

https://github.com/MarkUsProject/Wiki/pull/125


app/controllers/api/submission_files_controller.rb
```diff
@@ -199,10 +163,91 @@ def remove_folder
...
     def implicit_authorization_target
       SubmissionFile
     end
+
+    private
+
+    # Helper that uploads a file to this particular +grouping+'s assignment repository.
+    # If +only_required_files+ is true, only required files for this grouping's assignment can be uploaded.
+    def upload_file(grouping, only_required_files: false)
+      if has_missing_params?([:filename, :mime_type, :file_content])
```
anoverga17, line 207: I know the Jupyter Extension uses the `upload` parameter and I have tested it using the parameter. However, to recycle more code and to keep with the format we have for `create` I thought it might be a good idea to keep these parameters and modify the parameters the extension sends instead.

Let me know though if you disagree with this. I can change it easily either way.

anoverga17, line 207: I know the Jupyter Extension only uses the `upload` parameter and I have tested this route using that parameter. However, to recycle more code and to keep with the format we have for `create` I thought it might be a good idea to keep these parameters and modify the parameters the extension sends instead.

Let me know though if you disagree with this. I can change it easily and it works either way.

anoverga17, line 207: I know the Jupyter Extension only uses the `upload` parameter and I have tested this route using that parameter. However, to recycle more code and to keep with the format we have for `create` I thought it might be a good idea to keep these parameters and modify the parameters the extension sends instead.

Let me know though if you disagree with this. I can change it easily and it works either way.

app/controllers/api/submission_files_controller.rb
```diff
@@ -199,10 +163,91 @@ def remove_folder
       end
     end
 
+    def submit_file
+      student = Student.find_by(user: @real_user, course: @current_course)
```
david-yz-liu, line 167: We already have the `current_role` method that computes this, so you can do `student = current_role`

config/routes.rb
```diff
@@ -53,6 +53,11 @@
           end
         end
         resources :starter_file_groups, only: [:index, :create]
+        resources :submission_files, only: [] do
+          collection do
+            post 'submit_file'
```
david-yz-liu, line 58: In terms of organization, I think this route should go under `assignments` (as a member route). The functionality is more than just submission files, since it has the potential to create a group as well.

(Also update Changelog entry accordingly.)

app/controllers/api/submission_files_controller.rb
```diff
@@ -199,10 +163,91 @@ def remove_folder
...
+      assignment = Assignment.find_by(id: params[:assignment_id], course: @current_course)
+
+      # Disable submission via API if the instructor desires to
+      unless assignment.api_submit
+        render 'shared/http_status', locals: { code: '403', message:
+          'The instructor has disabled submission via the API' }, status: :forbidden
+        return
+      end
+
+      # Reject submission if the student has pending grouping
```
david-yz-liu, line 177: This isn't a bad idea, but in this case I think it's better to just allow students to create their own group (deleting pending group invitations, which I think will happen automatically from callbacks). Remember that students may be submitting very close to the deadline, and the point of the Jupyter extension is to get rid of as much friction as possible.

app/controllers/api/submission_files_controller.rb
```diff
@@ -199,10 +163,91 @@ def remove_folder
...
+      if has_missing_params?([:filename, :mime_type, :file_content])
+        # incomplete/invalid HTTP params
+        render 'shared/http_status', locals: { code: '422', message:
+          HttpStatusHelper::ERROR_CODE['message']['422'] }, status: :unprocessable_entity
+        return
+      end
+
+      # Only allow required files to be uploaded if +only_required_files+ is true
+      if only_required_files && grouping.assignment.assignment_files.pluck(:filename).exclude?(params[:filename])
+        render 'shared/http_status', locals: { code: '403', message:
```
david-yz-liu, line 216: Use 422 for this one as well

app/policies/api/submission_file_policy.rb
```diff
@@ -4,5 +4,9 @@ class SubmissionFilePolicy < MainApiPolicy
     def index?
       real_user.autotest_user? || check?(:manage?)
     end
+
+    def submit_file?
+      role&.student? || false
```
david-yz-liu, line 9: I'm not sure what the `|| false` is for? Does returning `nil` not cause the policy to reject the request?

app/policies/user_policy.rb
```diff
@@ -21,6 +21,6 @@ def update_settings?
 
   # Only users that are instructors in at least one course
   def reset_api_key?
-    user.roles.pluck(:type).include?('Instructor')
+    user.roles.pluck(:type).include?('Instructor') || user.roles.pluck(:type).include?('Student')
```
david-yz-liu, line 24: You can make this more general and allow all `StandardUser`s to reset their API key. Also remember to update the comment above

david-yz-liu (commented):
@anoverga17 in response to your comment, I agree it's better to keep the parameters the same. I'll update the jupyter extension accordingly.

david-yz-liu (commented):
Good work, I left a couple of inline comments. But while testing I realized that another important consideration is whether the assignment is hidden or not (including taking into account section-specific visibility).

Overall it seems like we should be reusing existing policies so that this kind of logic is not duplicated. Please check out the existing policies governing file submission (not through the API) and use those.

david-yz-liu (approved):

PR #6060, opened 2022-06-29
allow lti clients to get auth tokens

pretendWhale:
## Motivation and Context

This PR implements functionality for retrieving oauth client_credentials tokens to access canvas. It also allows us to serve JWKs created from a new rake task.

## Your Changes

**Description**:
Added a get_oauth_token method to the lti_client model, and implemented a route for serving jwks. Also implements a new rake task for generating a private key.

## Testing

Tested via rails console, as well as by calling the method from a controller. Also added a simple test for correct parameters in request to canvas.

- I wasn't sure where made sense to put keys generated by the rake task. for now, I leave them under data/lti.
- in get_oauth_token, we need to create a JWT with an issuer parameterwhich is the MarkUs root_url. route_helpers are not available in models, and it wasn't clear to me how to generate the host name, so for now I've passed it as a parameter, but I feel there is likely a better way to do this.

## Checklist

### DIFF
--- a/config/settings/development.yml
+++ b/config/settings/development.yml
@@ -43,4 +43,4 @@ starter_file:
 lti:
   domains: <%= %w[host.docker.internal] %>
   token_endpoint: "http://host.docker.internal:3100/login/oauth2/token"
-  key_path: <%= "#{::Rails.root.to_s}/data/dev" %>/lti
+  key_path: <%= "#{::Rails.root.to_s}/data/dev" %>/lti/key.pem
--- a/config/settings/development.yml
+++ b/config/settings/development.yml
@@ -42,3 +42,5 @@ starter_file:
 
 lti:
   domains: <%= %w[host.docker.internal] %>
+  token_endpoint: "http://host.docker.internal:3100/login/oauth2/token"
+  key_path: <%= "#{::Rails.root.to_s}/data/dev" %>/lti
--- a/spec/models/lti_client_spec.rb
+++ b/spec/models/lti_client_spec.rb
@@ -4,33 +4,49 @@ describe LtiClient do
     it { is_expected.to validate_uniqueness_of(:client_id).scoped_to(:host) }
   end
   describe '#get_oauth_token' do
-    # let!(:lti_client) { create :lti_client }
+    let(:lti_client) { create :lti_client }
     let(:lti_deployment) { create :lti_deployment, lti_client: lti_client }
     let(:base_url) { 'http://example.com' }
-    let(:pub_jwk_key) { OpenSSL::PKey::RSA.new File.read("#{Settings.lti.key_path}/key.pem") }
+    let(:pub_jwk_key) { OpenSSL::PKey::RSA.new File.read(Settings.lti.key_path) }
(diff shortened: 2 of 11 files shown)


spec/models/lti_client_spec.rb
```diff
@@ -0,0 +1,37 @@
+describe LtiClient do
+  describe 'uniqueness_validation' do
+    subject { create :lti_client }
+    it { is_expected.to validate_uniqueness_of(:client_id).scoped_to(:host) }
+  end
+  describe '#get_oauth_token' do
+    # let!(:lti_client) { create :lti_client }
```
david-yz-liu, line 7: I'm not sure if this comment means the approach didn't work. It seems to me that we can use `let` here (`@lti_client` below doesn't seem to be anything special), which would be preferred

spec/models/lti_client_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+          body: hash_including(
+            { grant_type: 'client_credentials',
+              client_assertion_type: 'urn:ietf:params:oauth:client-assertion-type:jwt-bearer',
+              scope: 'https://purl.imsglobal.org/spec/lti-ags/scope/lineitem',
+              client_assertion: /.*/ }
+          ),
+          headers: {
+            Accept: '*/*'
+          }
+        ).to_return(status: 200, body: { access_token: 'access_token' }.to_json)
```
david-yz-liu, line 26: Can use use `:success` instead of `200`?

spec/models/lti_client_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+    context 'with a correct payload' do
+      it 'returns a correct token' do
+        # byebug
```
david-yz-liu, line 30: delete this

spec/models/lti_client_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        expect(
+          @lti_client.get_oauth_token(base_url, [scope])
+        ).to include('access_token')
```
david-yz-liu, line 33: From the documentation (https://canvas.instructure.com/doc/api/file.oauth_endpoints.html#post-login-oauth2-token) it seems like there are four keys we can be checking for. Just to be safe it would be good to check for the presence of each one, plus the scopes being the same as the scopes we sent (presumably)

david-yz-liu (commented):
@pretendWhale looks good, I just left a couple of minor comments/questions!

config/settings/development.yml
```diff
@@ -42,3 +42,5 @@ starter_file:
 
 lti:
   domains: <%= %w[host.docker.internal] %>
+  token_endpoint: "http://host.docker.internal:3100/login/oauth2/token"
+  key_path: <%= "#{::Rails.root.to_s}/data/dev" %>/lti
```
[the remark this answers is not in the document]
pretendWhale: We can probably leave supporting multiple keys until if/when we need it, but you're right - if we aren't, it makes more sense to include the file name in the settings path

david-yz-liu (approved):

PR #6062, opened 2022-06-30
changed flash message

Dhairya-Khara:
Changed flash message when user uploaded a file name with an incorrect name.
## Motivation and Context

Earlier students might have been confused why their file was rejected. This change makes it clear why their upload is being rejected.

## Your Changes

**Description**:
Changed error message. 

If a student uploads a file with an incorrect name (let's say `a1.py`) in the root directory, the error message reads:
"Filename `a1.py` is not allowed for this assignment."

If a student uploads a file with an incorrect name (let's say `a1.py`) in a subfolder (let's say `part1`), the error message reads:
"Filename `a1.py` in `part1` is not allowed for this assignment."
 

## Testing

No additional tests were added. The current rspec and jest suites were ran just in case.

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -46,6 +46,7 @@
 - Expanded summary stats UI (#6050)
 - Fix bug for submission modal where students could submit files even if no file was chosen (#6052)
 - Removed searchbar used to filter files on submissions page (#6058)
+- Changed flash message that is displayed when students upload a file with an incorrect name. (#6062)
 
 ## [v2.0.10]
 - Fix bug when sorting batch test runs where sorting by date was not working (#5906)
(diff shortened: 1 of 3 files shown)


david-yz-liu (commented):
@Dhairya-Khara good job! In terms of the message itself, I have two overall comments:

1. The path name is important because we need to handle subdirectories (and an allowed filename in one subdirectory may not be allowed in another). However we don't necessarily want to display the required path to students directly, but could split it up into (1) the filename that the student submitted, and (2) where they submitted the file.
2. It would be good for you to come up with a couple of different variations of the message and bring them to the meeting on Monday to discuss (and get more student feedback).

david-yz-liu (approved):

PR #6101, opened 2022-07-12
improved ui on error pages (404, 403)

Dhairya-Khara:
The UI for the 404 and 403 pages was updated to match the MarkUs dashboard.

## Motivation and Context

Aesthetic Improvement.

## Your Changes

**Description**:
`_markus.scss`: Added styles for the new error pages

`application_controller.rb`: Render the error page (appropriately) if logged in, redirect to root url otherwise. Controller methods also render the new `error_page.html.erb`. Furthermore, `@current_role` is passed to the views to avoid errors and ensure the headers render properly.

`_menu.html.erb`: Several things are rendered onlly if `@current_role` is not `nil`. Required for testing purposes.

`assignment_content.html.erb`: The title now `yields` if a specific title is provided.

`error_pages.html.erb`: New file for 404 and 403.

`course_policy.rb`: Checking `real_role` is not nil

`main_controller.rb`: Dynamically choosing which `layout` to render. Required to correctly ensure that the header and nav bar renders on a 404.

Here is a screenshot of the 404 page for refernece.

## Testing

No additional tests were written. The current rspec and jest suites were ran just in case.

## Checklist


david-yz-liu (commented):
@Dhairya-Khara the UI here looks great, but I think it'll be better if you create a brand new view file to preserve backwards compatibility with other uses of the `http_status.html.erb` file. That is, instead of modifying that file to make it more like the layout file we currently have, create a new file that can embed directly within the existing layouts (so in the controller methods, don't pass in `layout: false` to the `render` method).

app/views/layouts/assignment_content.html.erb
```diff
@@ -2,7 +2,7 @@
 <html lang="<%= I18n.locale %>">
 <head>
   <%= render partial: 'layouts/head' %>
-  <title><%= t(:markus) %> - <%= controller.action_name.titleize %></title>
+  <title><%= content_for?(:html_title) ? yield(:html_title) : t(:markus) + ' - ' + controller.action_name.titleize%></title>
```
david-yz-liu, line 5: `t(:markus)` should always appear. But you can use the `yield` to replace the controller action name :+1:

david-yz-liu: Also, there should always be a space inside `<%=` and `%>` (fix throughout)

app/views/shared/_assignments_dropdown_menu.html.erb
```diff
@@ -28,17 +28,19 @@
 
 <li class='section'><%= Assignment.model_name.human.pluralize %></li>
 
-<% assignments = @current_role.visible_assessments(assessment_type: 'Assignment') %>
-<% grade_entry_forms = @current_role.visible_assessments(assessment_type: 'GradeEntryForm') %>
+<% if @current_role%>
```
david-yz-liu, line 31: I don't think this is the right place to make this check (a bit too far down the tree). Better to ensure the entire "sub_menu" doesn't render at all when `@current_role` is `nil`. I see in the `_sub_menu.html.erb` there's a policy check that should include this using `allowed_to?` so I'm not sure if this is an issue, but feel free to follow up with me about this on Slack with a more detailed error description. (But regardless you should revert the changes in this file.)

app/views/shared/error_page.html.erb
```diff
@@ -0,0 +1,6 @@
+<% content_for(:html_title) { code + ' ' + HttpStatusHelper::ERROR_CODE[code]}%>
+<div class="error-pages-container">
+  <h1><%= code + ' ' + HttpStatusHelper::ERROR_CODE[code] %></h1>
+  <p>Uh oh! <%= message %></p>
+  <%= link_to I18n.t("menu.go_home"), root_path, class: "button" rescue nil %>
```
david-yz-liu, line 5: You don't need the `rescue nil`

app/views/shared/error_page.html.erb
```diff
@@ -0,0 +1,6 @@
+<% content_for(:html_title) { code + ' ' + HttpStatusHelper::ERROR_CODE[code]}%>
+<div class="error-pages-container">
+  <h1><%= code + ' ' + HttpStatusHelper::ERROR_CODE[code] %></h1>
+  <p>Uh oh! <%= message %></p>
```
david-yz-liu, line 4: You can remove the "uh oh!" here

app/controllers/application_controller.rb
```diff
@@ -42,11 +42,14 @@ def default_url_options(_options = {})
...
 
   def page_not_found
-    render 'shared/http_status',
-           formats: [:html],
-           locals: { code: '404', message: HttpStatusHelper::ERROR_CODE['message']['404'] },
-           status: :not_found,
-           layout: false
+    if current_user
+      render 'shared/error_page',
+             formats: [:html],
```
david-yz-liu, line 47: You can delete this and rely on Rails to figure out the format (and same below)

app/assets/stylesheets/common/_markus.scss
```diff
@@ -1339,3 +1339,12 @@ canvas {
 .required-files-container {
   display: flex;
 }
+
+//Styling for the text on any error pages (for eg. 404, 403)
+.error-pages-container {
+  left: 50%;
```
david-yz-liu, line 1345: Okay, so overall I think you've made the CSS harder for yourself than you need to. You can remove all of the rules here except the `text-align` (which does the text centering).

app/views/shared/error_page.html.erb
```diff
@@ -0,0 +1,6 @@
+<% content_for(:html_title) { code + ' ' + HttpStatusHelper::ERROR_CODE[code]}%>
```
david-yz-liu, line 1: I thought about it and I think it's better to remove the code and only display the name (e.g., "Forbidden" instead of "403 Forbidden"), since the code won't be meaningful for regular users. Same with the `h1` text below.

app/controllers/application_controller.rb
```diff
@@ -42,11 +42,18 @@ def default_url_options(_options = {})
   end
 
   def page_not_found
-    render 'shared/http_status',
-           formats: [:html],
-           locals: { code: '404', message: HttpStatusHelper::ERROR_CODE['message']['404'] },
-           status: :not_found,
-           layout: false
+    if current_user
+      if current_user.admin_user?
```
david-yz-liu, line 46: You don't need to replicate the implementation of `current_role` here. You can just call the method so that it assigns to `@current_role`. But also it seems like this should only be called if `current_course` isn't nil, so you can do something like `current_role unless current_course.nil?`

app/controllers/main_controller.rb
```diff
@@ -182,4 +182,13 @@ def get_blank_message(login, password)
   def implicit_authorization_target
     OpenStruct.new policy_class: MainPolicy
   end
+
+  def resolve_layout
```
david-yz-liu, line 186: Adding this method shouldn't be necessary. You can specify the layout explicitly in a `render` call by adding `layout: ...` as an argument.

david-yz-liu (approved):

PR #6105, opened 2022-07-14
Lti user sync

pretendWhale:
## Motivation and Context

The PR adds functionality for syncing course rosters between MarkUs courses and LMS courses. We first attempt to match LTI learners to already created MarkUs users. If no match can be found, a new user is created with the student role in the associated course. Matching is on `lis_person_sourcedid`, which corresponds to canvas's SIS ID (utorid).

## Your Changes

**Description**:
Adds get_students method to lti_deployment model. Also adds `LtiServices`, a table for LTI service endpoints like the names and roles service.

- [x] Other (please specify): New LTI feature

## Testing

Tested via the rails console, as well as adding tests to the lti_deployment spec.

For local testing, add markus as a new lti dev key to canvas, and add students to a canvas course, either manually or via [SIS formatted csvs](https://canvas.instructure.com/doc/api/file.sis_csv.html) (I found this the most straightforward method and created courses.csv, users.csv, and enrollments.csv). Install markus in the course with students, launch markus, and associate the canvas course with a markus course. Then from a rails console, call the `get_students` method on the `LtiDeployment` that was created.

Notes for local testing:
- user data is not sent for users in the "pending" state. After adding users via SIS import and then adding them to a course, their status will be pending. Users can be forced out of pending state by choosing to "act as" the user and interacting with the course
- in this PR I assume `given_name` and `family_name` for each user are set for the purposes of creating new users, so these fields need to be set on canvas users when testing locally. The `name` field is used as the MarkUs `display_name1

## Checklist


app/controllers/lti_deployment_controller.rb
```diff
@@ -123,6 +123,11 @@ def redirect_login
     lti_deployment = LtiDeployment.find_or_create_by(lti_client: lti_client, external_deployment_id: deployment_id)
     session[:lti_client_id] = lti_client.id
     session[:lti_deployment_id] = lti_deployment.id
+    if decoded_token[0].key?('https://purl.imsglobal.org/spec/lti-nrps/claim/namesroleservice')
+      name_and_roles_endpoint = decoded_token[0]['https://purl.imsglobal.org/spec/lti-nrps/claim/namesroleservice']['context_memberships_url']
+      names_service = LtiService.find_or_create_by(lti_deployment: lti_deployment, service_type: 'namesroles')
```
david-yz-liu, line 128: Can you do a `find_or_initialize_by` instead, to make `url` a required field?

david-yz-liu: Oh and are you sure the service is `namesroles` and not `namesrole`? (Is there an official list somewhere?) Because the claim endpoint is `namesroleservice`, not `namesrolesservice`

app/models/lti_deployment.rb
```diff
@@ -2,4 +2,29 @@ class LtiDeployment < ApplicationRecord
   belongs_to :course, optional: true
   belongs_to :lti_client
   validates :external_deployment_id, uniqueness: { scope: :lti_client }
+
+  def get_students
+    auth_data = lti_client.get_oauth_token(['https://purl.imsglobal.org/spec/lti-nrps/scope/contextmembership.readonly'])
+    names_service = LtiService.find_by!(lti_deployment: self, service_type: 'namesroles')
```
david-yz-liu, line 8: I generally prefer using explicit associations on queries like these, e.g. `self.lti_services.find(...)` (which will raise an error if no record is found)

app/models/lti_deployment.rb
```diff
@@ -2,4 +2,29 @@ class LtiDeployment < ApplicationRecord
...
+    member_info = JSON.parse(res.body)
+    member_info['members'].each do |member|
+      if member['status'] == 'Inactive'
+        next
+      end
+      user = EndUser.find_by(user_name: member['lis_person_sourcedid'])
+      if user.nil?
+        user = EndUser.create!(user_name: member['lis_person_sourcedid'], first_name: member['given_name'],
+                               last_name: member['family_name'], display_name: member['name'])
+        Student.create!(user: user, course: course)
```
david-yz-liu, line 25: I think we should be creating `Student` roles for users that do exist but aren't yet registered in the course. (And add test for this)

app/models/lti_deployment.rb
```diff
@@ -2,4 +2,29 @@ class LtiDeployment < ApplicationRecord
...
+    names_service = LtiService.find_by!(lti_deployment: self, service_type: 'namesroles')
+    membership_uri = URI(names_service.url)
+    membership_uri.query = URI.encode_www_form(role: 'http://purl.imsglobal.org/vocab/lis/v2/membership#Learner')
+    req = Net::HTTP::Get.new(membership_uri)
+    req['Authorization'] = "#{auth_data['token_type']} #{auth_data['access_token']}"
+    http = Net::HTTP.new(membership_uri.host, membership_uri.port)
+
+    res = http.request(req)
+    member_info = JSON.parse(res.body)
+    member_info['members'].each do |member|
```
david-yz-liu, line 17: Okay so overall this loop seems like it's going to be pretty slow when the classes get large (it's a form of N+1 querying). We have an opportunity to use some bulk ActiveRecord operations to speed things up. (I don't think you've done much of this, but it's a good experience.)

Check out https://edgeapi.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-upsert_all, and in particular note that you can get the method to return the `id`s of the inserted/updated records.

app/models/lti_service.rb
```diff
@@ -0,0 +1,5 @@
+class LtiService < ApplicationRecord
+  belongs_to :lti_deployment
+  validates :service_type, format: { with: /\Anamesroles\z/ }
```
david-yz-liu, line 3: Seems like we want an `inclusion: [...]` to better generalize to other services

db/migrate/20220707182822_create_lti_services.rb
```diff
@@ -0,0 +1,12 @@
+class CreateLtiServices < ActiveRecord::Migration[7.0]
+  def change
+    create_table :lti_services do |t|
+      t.references :lti_deployment, null: false, foreign_key: true
+      t.string :service_type
```
david-yz-liu, line 5: This should be non-null

db/migrate/20220707182822_create_lti_services.rb
```diff
@@ -0,0 +1,12 @@
...
+      t.string :url
```
david-yz-liu, line 6: This should also be non-null I think (see above comment)

app/models/lti_service.rb
```diff
@@ -0,0 +1,5 @@
+class LtiService < ApplicationRecord
+  belongs_to :lti_deployment
```
david-yz-liu, line 2: you should add the corresponding `has_many` to `LtiDeployment`

spec/factories/lti_services.rb
```diff
@@ -0,0 +1,7 @@
+FactoryBot.define do
+  factory :lti_service do
+    lti_deployment { association(:lti_deployment) }
```
david-yz-liu, line 3: you can just write `lti_deployment`, by default the association will be generated automatically

spec/factories/lti_services.rb
```diff
@@ -0,0 +1,7 @@
...
+    service_type { 'namesroles' }
```
david-yz-liu, line 4: I think it's better to move this attribute into a child factory (so `:lti_service` becomes "abstract")

app/models/lti_deployment.rb
```diff
@@ -2,4 +2,29 @@ class LtiDeployment < ApplicationRecord
...
+      if member['status'] == 'Inactive'
+        next
+      end
+      user = EndUser.find_by(user_name: member['lis_person_sourcedid'])
+      if user.nil?
+        user = EndUser.create!(user_name: member['lis_person_sourcedid'], first_name: member['given_name'],
```
david-yz-liu, line 23: Oh I forgot to mention this earlier, but it would be good to store the student numbers and email addresses as well, especially if we're moving to using Canvas as the source for class lists. 

(The emails should be pretty clear but the student numbers may not be obvious in the Canvas fields and U of T might do something idiosyncratic. If you can't figure it out, please ask the ACT folks.)

app/controllers/lti_deployment_controller.rb
pretendWhale: I think you're right, it should be `namesrole` in line with the claim, as the service is the [Names and Role Provisioning service](https://www.imsglobal.org/spec/lti-nrps/v2p0). There isn't really a master list, but I've been working from the claims/scopes canvas provides on their [LTI config page](https://canvas.instructure.com/doc/api/file.lti_dev_key_config.html#anatomy-of-a-json-configuration)

app/models/lti_deployment.rb
pretendWhale: Email was straightforward. I've asked the ACT folks about student number and am waiting to hear back

david-yz-liu (approved):

PR #6107, opened 2022-07-16
Issue 5611

### LINKED ISSUE #5611: Add API route for tags
- CRUD routes for tags
- also un/assign tags to a grouping

WeissSa:
Add CRUD API routes for tags. 

## Motivation and Context
Fixes #5611 

## Your Changes
- Added CRUD routes and logic for tags

## Testing
Writing tests for api/tags_controller and potentially policy/api/tags_controller (if a policy file is necessary)

Currently all my tests give me 302 instead of 200. Not sure why this is the case

## Checklist


WeissSa:
unrelated test failure. This is ready for review :+1:

app/controllers/api/groups_controller.rb
```diff
@@ -281,6 +281,30 @@ def update_marking_state
       end
     end
 
+    def add_tag
+      tag = Tag.find_by(id: params[:tag_id])
+      grouping = Grouping.find_by(id: params[:grouping_id])
+      if tag.nil? || grouping.nil?
+        render 'shared/http_status', locals: { code: '404', message: 'User was not found' }, status: :not_found
```
david-yz-liu, line 288: This message should be updated and internationalized (and throughout)

app/controllers/api/groups_controller.rb
```diff
@@ -281,6 +281,30 @@ def update_marking_state
       end
     end
 
+    def add_tag
+      tag = Tag.find_by(id: params[:tag_id])
+      grouping = Grouping.find_by(id: params[:grouping_id])
```
david-yz-liu, line 286: The other methods here use `self.grouping` (there's a helper method defined at the bottom of this file), and you should do the same here. Note that this requires a different configuration of params (`assignment_id` to specify the assignment, and then a `Group` id (not `Grouping` id), which is used to define the `record` for the controller

app/controllers/api/groups_controller.rb
```diff
@@ -281,6 +281,30 @@ def update_marking_state
       end
     end
 
+    def add_tag
+      tag = Tag.find_by(id: params[:tag_id])
```
david-yz-liu, line 285: Tags are assignment-specific. So, you should find the grouping first, so that you can query to find tags associated only with the assignment corresponding to the grouping (rather than an arbitrary tag from a potentially different assignment).

app/controllers/api/groups_controller.rb
```diff
@@ -281,6 +281,30 @@ def update_marking_state
...
+        render 'shared/http_status', locals: { code: '404', message: 'User was not found' }, status: :not_found
+      else
+        grouping.tags << tag
+        render 'shared/http_status', locals: { code: '200', message:
+          HttpStatusHelper::ERROR_CODE['message']['200'] }, status: :ok
+      end
+    end
+
+    def remove_tag
+      tag = Tag.find_by(id: params[:tag_id])
```
david-yz-liu, line 297: Same comments as above

app/policies/api/tag_policy.rb
```diff
@@ -0,0 +1,10 @@
+module Api
+  # Policies for Api::Tag Controller
+  class TagPolicy < MainApiPolicy
+    alias_rule :create?, :update?, :destroy, to: :index
+    def index?
+      # has at least one instructor role
+      real_user.is_a?(EndUser) && real_user.roles.pluck(:type).include?('Instructor')
```
david-yz-liu, line 7: This is too broad; you need to check that the current user's role _for the current course_ is `Instructor`. You should check how other API policies do this with the `role` variable

app/controllers/api/tags_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+        format.xml do
+          render xml: tag_info
+        end
+        format.json do
+          render json: tag_info
+        end
+      end
+    end
+
+    def edit
```
david-yz-liu, line 32: You didn't define this route (and don't need it), so this method can be removed

app/controllers/api/tags_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+      @tag = Tag.find(params[:id])
+    end
+
+    # Creates a new instance of the tag.
+    def create
+      params.permit(:name, :description)
+      begin
+        new_tag = Tag.new(name: params[:name], description: params[:description],
+                          course: Course.find_by(id: params[:course_id]),
```
david-yz-liu, line 41: One of the modules we include in all controllers is `SessionHandler`, which defines a `current_course` helper method you should call here

app/controllers/api/tags_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+                          role: current_role, assessment: Assessment.find_by(id: params[:assignment_id]))
```
david-yz-liu, line 42: Similar to the above comment, you should use associations here to only query for assessments associated with the current course. Same with the `Grouping` query down below.

app/controllers/api/tags_controller.rb
```diff
@@ -0,0 +1,85 @@
...
+      end
+    end
+
+    def edit
+      @tag = Tag.find(params[:id])
+    end
+
+    # Creates a new instance of the tag.
+    def create
+      params.permit(:name, :description)
```
david-yz-liu, line 38: The way we typically do this in other controllers is to define a helper method like `tag_params` to specify what param values we allow. Then in the `Tag.new` (or `update`), you can unpack these params into their key value pairs with the syntax `Tag.new(..., **tag_params)` so that you don't need to manually specify the `name` and `description` fields in the `new`/`update`.

david-yz-liu (commented):
@WeissSa good job, and very thorough testing. I left a couple of overall comments for you to work on!

app/controllers/api/groups_controller.rb
WeissSa: I'm having an issue with this where I am sending ```id: group.id``` to the controller and the record is still nil resulting in self.grouping throwing an error. I have also tried sending ```group_id: group.id```. Do you know what might be going wrong?

I'm not quite sure what is going wrong. Will push for visibility.

david-yz-liu: @WeissSa there's a problem with the tests. Look carefully at the definitions of the group and grouping factories (their definitions tell you what attributes you must/can specify when creating them in the test files).

WeissSa: @david-yz-liu sorry that was a typo in my ```add_tag``` tests. I will definitely fix it, but the issue is that I get the message:
```Api::GroupsController An authenticated request requesting PUT add_tag should throw a 404 if the tag id is not found
     Failure/Error: @grouping ||= record.grouping_for_assignment(assignment.id)

     NoMethodError:
       undefined method `grouping_for_assignment' for nil:NilClass```
When running my remove_tag (or my newly updated add_tag)
I have tried using both id, and group_id to set the record so I am unsure why it is nil. Thanks!

Comment (david-yz-liu) [reply app/controllers/api/groups_controller.rb]: I would set a breakpoint somewhere to check the params manually. If `record` is `nil` then likely the id in the params is incorrect, possibly because the test objects are not being successfully created. And again, review the factory files.

**Comment by WeissSa** [comment]
Fixed the tests after the meeting I think this is ready for review :sunglasses:

File: app/controllers/api/tags_controller.rb [hunk 10]
```diff
@@ -0,0 +1,87 @@
+module Api
+  # Api controller for tags
+  # Uses Rails' RESTful routes (check 'rake routes' for the configured routes)
+  class TagsController < MainApiController
+    DEFAULT_FIELDS = [:id, :name, :description, :creator, :use].freeze
```
david-yz-liu, line 5: You can just delete this

app/controllers/api/tags_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+
+    def index
+      @assignment = Assignment.find_by(id: params[:assignment_id])
```
david-yz-liu, line 8: For the API routes you generally don't need to use instance variables because you aren't going to use these variables in view files. Here you should use `assignment` instead of `@assignment`.

app/controllers/api/tags_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+        end
+        format.json do
+          render json: tag_info
+        end
+      end
+    end
+
+    # Creates a new instance of the tag.
+    def create
+      params.permit(:name, :description)
```
david-yz-liu, line 34: You can delete this line (the `tag_params` below is fine)

app/controllers/api/tags_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+        tag.destroy!
+        render 'shared/http_status',
+               locals: { code: '200', message: HttpStatusHelper::ERROR_CODE['message']['200'] }, status: :ok
+      end
+    end
+
+    private
+
+    def tag_params
+      { name: params[:name] || '', description: params[:description] || '' }
```
david-yz-liu, line 84: Don't specify default values. If the `:name` or `:description` fields aren't specified, they won't be included in the `create`/`update`, which is fine. (In the `create` case a validation error should be raised, which is also fine.)

app/controllers/api/tags_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+      end
+    end
+
+    # Creates a new instance of the tag.
+    def create
+      params.permit(:name, :description)
+      begin
+        assignment = current_course.assignments.find_by(id: params[:assignment_id])
+        new_tag = Tag.new(course: current_course, role: current_role, assessment: assignment, **self.tag_params)
+        if new_tag.save && params[:grouping_id]
```
david-yz-liu, line 38: I see, so here we want to raise an error if the tag is invalid (e.g. if the tag name/description is omitted). So you can turn the previous line into `Tag.create!`, which will save the object (so no need for the `new_tag.save` here) raise an error if any validations fail.

Make sure to add tests for this as well.

app/controllers/api/tags_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+      rescue StandardError => e
+        render 'shared/http_status', locals: { code: '422', message: e.to_s }, status: :unprocessable_entity
+      else
+        render 'shared/http_status',
+               locals: { code: '201', message: HttpStatusHelper::ERROR_CODE['message']['201'] }, status: :created
+      end
+    end
+
+    def update
+      tag = Tag.find(params[:id])
```
david-yz-liu, line 51: use `find_by(id: ...)` instead of `find`; the latter raises an error, while the former will return `nil`, which is what you want here

app/controllers/api/tags_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+      if tag.nil?
+        render 'shared/http_status', locals: { code: '404', message: I18n.t('tags.not_found') }, status: :not_found
+      else
+        begin
+          name = params[:name]
+          desc = params[:description]
+
+          # not using tag_params so it can default to the previous name/description if nil
```
david-yz-liu, line 59: use `tag_params` (and see my comment about that method)

app/policies/api/tag_policy.rb
```diff
@@ -0,0 +1,11 @@
+module Api
+  # Policies for Api::Tag Controller
+  class TagPolicy < MainApiPolicy
+    alias_rule :create?, :update?, :destroy, to: :index
+
+    def index?
+      # has at least one instructor role
```
david-yz-liu, line 7: this comment needs to be updated

app/policies/api/tag_policy.rb
```diff
@@ -0,0 +1,11 @@
+module Api
+  # Policies for Api::Tag Controller
+  class TagPolicy < MainApiPolicy
+    alias_rule :create?, :update?, :destroy, to: :index
```
david-yz-liu, line 4: the conventional "default" rule is `manage?` (use that instead of `index`)

spec/controllers/api/groups_controller_spec.rb
```diff
@@ -587,5 +598,70 @@
...
+      let(:grouping) { create :grouping, group: group, tags: [tag], assignment: assignment }
+
+      before do
+        request.env['HTTP_ACCEPT'] = response_type
+      end
+
+      it 'should let the user remove a tag' do
+        put :remove_tag, params: { id: grouping.group.id, assignment_id: assignment.id,
+                                   course_id: course.id, tag_id: tag.id }
+        expect(response.status).to be(200)
```
david-yz-liu, line 613: use `have_http_status` when checking HTTP status codes (throughout the tests you added)

WeissSa:
Test fail is unrelated I think

david-yz-liu (approved):

PR #6110, opened 2022-07-19
Replace unicode char not available in most fonts with one that is

mishaschwartz:
## Motivation and Context

The dropdown arrow for the menu at the top right of the screen was rendered in a font that didn't support that unicode character. This PR replaces that character with one that has been around since unicode 1.1 so it should be supported by pretty much everything (https://www.fileformat.info/info/unicode/char/25be/index.htm)

Other options include using an image:

```
<%= image_tag('icons/bullet_arrow_down.png') %>
```

or using one of the jquery-ui icons (I don't love this idea because then we make every page on markus dependant on jquery)

## Your Changes

**Description**:

- changed a unicode character

## Testing

- viewed on chrome, firefox, safari

## Checklist

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -13,7 +13,7 @@
           <strong>(<%= @real_user.display_name %>)</strong>
         <% end %>
           <span id='user-id'>
-           <%= @current_user.display_name %> <%= "\u25be" %>
+           <%= @current_user.display_name %> <%= "\u25bc" %>
         </span>
 
           <ul>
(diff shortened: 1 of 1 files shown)


app/views/layouts/_header.html.erb
```diff
@@ -13,7 +13,7 @@
           <strong>(<%= @real_user.display_name %>)</strong>
         <% end %>
           <span id='user-id'>
-           <%= @current_user.display_name %> ⯆
+           <%= @current_user.display_name %> <%= "\u25be" %>
```
david-yz-liu, line 16: Good catch! But can you use `\u25bc` instead (https://www.fileformat.info/info/unicode/char/25bc/index.htm), it's from the same Unicode version and is bigger, which I prefer.

mishaschwartz: Sure, I avoided it initially because it's the same height as the font so it looks top heavy to me next to the other characters.

david-yz-liu (approved):

PR #6112, opened 2022-07-19
Explicit inactive description for inactive students/groups

TheTallJerry:
## Motivation and Context

Improves user exprience on the viewing/filtering of inactive students and groups

## Your Changes

**Description**:

#### When option deselected, inactive students/groups are not shown
loading

#### When option selected, inactive students/groups are shown
shown

#### Modified the search function to be able to filter inactive students
search

#### Tooltip to show number of inactive student(s)/group(s)
tooltip

## Testing

- Mannual testing on chrome and firefox. 
- Mannual inspection of browser console to make sure no errors were present. 

The horizontal space might run out. I tried enabling scrolling which is the usual solution, but for some reason the ellipsis text-overflow (i.e. 3 dots at the end of long text) for the students table, carried over to the scrolling: 
size
This change is not up yet, planning to delay until we figure out the solution.

## Checklist

https://github.com/MarkUsProject/Wiki/pull/133

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -56,7 +56,7 @@
 - Sorted courses on the dashboard. (#6099)
 - Added summary statistics for criteria (#6100)
 - Pass group name and starter files to the autotester when running tests (#6104)
-- Added explicit status and filtering for inactive students and groups in grouping page. (#6112)
+- Added explicit status and filtering for inactive students and groups in assignment groups page. (#6112)
 
 ## [v2.0.10]
 - Fix bug when sorting batch test runs where sorting by date was not working (#5906)
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,7 @@
 
 ## [unreleased]
 - Remove unmaintained locales (#5727)
+- Introduce standalone ruby script as an alternative method to checking for repository access (#5736)
 - Added the ability to submit URLs (#5822)
 - Switch Javascript bundling to jsbundling-rails gem, and update to webpack v5 (#5833)
 - Remove group name displayed attribute from assignment properties table (#5834)
@@ -53,7 +54,7 @@
 - Changed flash message that is displayed when students upload a file with an incorrect name. (#6062)
 - Added emoji annotations for graders and removed `control+click` quick annotations. (#6093)
 - Sorted courses on the dashboard. (#6099)
-- Introduce standalone ruby script as an alternative method to checking for repository access (#5736)
+- Added explicit status and filtering for inactive students and groups in grouping page. (#6112)
 
 ## [v2.0.10]
 - Fix bug when sorting batch test runs where sorting by date was not working (#5906)
--- a/app/assets/javascripts/Components/groups_manager.jsx
+++ b/app/assets/javascripts/Components/groups_manager.jsx
@@ -243,8 +243,8 @@ class GroupsManager extends React.Component {
           createAllGroups={this.createAllGroups}
           createGroup={this.createGroup}
           deleteGroups={this.deleteGroups}
-          hiddenStudentsCount={this.state.loading ? "" : this.state.hidden_students_count}
-          hiddenGroupsCount={this.state.loading ? "" : this.state.inactive_groups_count}
+          hiddenStudentsCount={this.state.loading ? null : this.state.hidden_students_count}
+          hiddenGroupsCount={this.state.loading ? null : this.state.inactive_groups_count}
           showHidden={this.state.show_hidden}
           updateShowHidden={this.updateShowHidden}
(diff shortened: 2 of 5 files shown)


app/assets/javascripts/Components/groups_manager.jsx
```diff
@@ -9,9 +9,11 @@ class GroupsManager extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      graders: [],
```
TheTallJerry, line 12: This was an unused state, so removed.

Changelog.md
```diff
@@ -2,6 +2,7 @@
 
 ## [unreleased]
 - Remove unmaintained locales (#5727)
+- Introduce standalone ruby script as an alternative method to checking for repository access (#5736)
```
TheTallJerry, line 5: Reordered in order of PR number

config/locales/models/users/en.yml
```diff
@@ -13,6 +13,7 @@ en:
         full_name: Name
         grace_credits: Grace Credits
         id_number: ID number
+        inactive: inactive
```
david-yz-liu, line 16: This is a bit awkward, but the keys here should match the actual attribute names, so in this case it would be `hidden` instead of `inactive` as the _key_. The text can stay as "inactive" though, since that's the text we display to the user

config/locales/models/groupings/en.yml
```diff
@@ -3,4 +3,6 @@ en:
   activerecord:
     attributes:
       grouping:
+        inactive_groups: inactive group(s)
```
david-yz-liu, line 6: For newer strings, we're trying to prefer using `count` to avoid the awkward `(s)`. See https://github.com/fnando/i18n-js/tree/v3#pluralization for details

david-yz-liu (commented):
@TheTallJerry great work! I just left a couple of pretty minor comments. Regarding the issue of horizontal space, I think it's okay to leave it for now (but I'll spend a bit of time playing around). The users will be able to resize the columns, and this is a bit of a niche case.

app/assets/javascripts/Components/groups_manager.jsx
```diff
@@ -43,9 +45,15 @@ class GroupsManager extends React.Component {
       this.studentsTable.resetSelection();
       this.groupsTable.resetSelection();
       this.setState({
-        groups: res.groups,
+        groups: res.groups.map(group => ({
+          ...group,
+          inactive: group.members.every(member => member[2]),
```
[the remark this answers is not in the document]
TheTallJerry: It actually does make a huge difference - this is like an improvement from 2n to n. Still O(n) but the exact time is much better.

config/locales/models/groupings/en.yml
```diff
@@ -3,4 +3,12 @@ en:
   activerecord:
     attributes:
       grouping:
+        inactive_groups:
+          one: 1 inactive group
+          other: "%{count} inactive groups"
```
TheTallJerry, line 8: The quotes are needed because the string starts with an interpolation

https://stackoverflow.com/a/43702790

app/assets/javascripts/Components/groups_manager.jsx
```diff
@@ -340,7 +376,11 @@ class RawGroupsTable extends React.Component {
       },
       filterMethod: (filter, row) => {
         if (filter.value) {
-          return row._original.members.some(member => member[0].includes(filter.value));
+          return row._original.members.some(member =>
+            `${member[1]}${
```
david-yz-liu, line 380: I think you can use `member.display_label` here as well?

app/assets/javascripts/Components/groups_manager.jsx
```diff
@@ -570,6 +639,14 @@ const StudentsTable = withSelection(RawStudentsTable);
 
 class GroupsActionBox extends React.Component {
   render = () => {
+    var showHiddenTooltip = "";
+    if (this.props.hiddenStudentsCount && this.props.hiddenGroupsCount) {
```
david-yz-liu, line 643: I'm not sure if you mean to exclude 0 here or just `""` values, but I think we *should* show the tooltips when one or both of these counts are 0.

But also `null` is a better default value for these props than `""` (what you're setting above)

config/locales/models/groupings/en.yml
```diff
@@ -3,4 +3,12 @@ en:
...
+          zero: 0 inactive group
```
david-yz-liu, line 9: "group" should be plural for 0 (and same with "student" below)

david-yz-liu (approved):

PR #6114, opened 2022-07-21
Fix flaky tests with automated test files

anoverga17:
## Motivation and Context

There are some rspec automated test tests which involve files that sometimes result in failure. This is likely the same issue as described in pull request #5813. This pull request thus aims to make the same changes to other flaky automated test tests

## Your Changes

**Description**:
- Rearranged order of file cleanup for certain flaky tests (these tests which the autotest_files_dir are now responsible for cleaning up the file first rather than rely on previous tests to clean after themselves)

## Testing

Testing was done via the github CI and was rerun multiple times to be confident the tests fixed here did not happen again.

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -54,7 +54,9 @@
 - Added emoji annotations for graders and removed `control+click` quick annotations. (#6093)
 - Sorted courses on the dashboard. (#6099)
 - Introduce standalone ruby script as an alternative method to checking for repository access (#5736)
+- Added summary statistics for criteria (#6100)
 - Pass group name and starter files to the autotester when running tests (#6104)
+- Disabled admin editing of course name and allowed instructors to edit display name (#6111)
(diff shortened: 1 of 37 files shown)


david-yz-liu (approved):
@mishaschwartz please feel free to merge this in after you've reviewed it

mishaschwartz (changes requested):
I worry that by cleaning up before the test, not afterwards; we're leaving test artifacts (files on disk mostly) that could affect tests that run after these ones that are not in these files. 

Can we clean up both before and after the tests run. It might be overkill but I think it's the safer thing to do to avoid flaky tests. 

So please leave all the changes you've made as they are. But also put back the `FileUtils.rm_rf` calls in the `after :each` blocks.

mishaschwartz (approved):

PR #6117, opened 2022-07-23
Fixing drag and drop to submit files

Dhairya-Khara:
Making the header of the submission manager 'droppable' by manually adding `DropTarget` and changing parent class of the component. Furthermore, functionality for dropping files on a folder is also implemented.

## Motivation and Context

Earlier, students could not submit files using drag and drop if no other file/folder existed. It also makes sense to upload a file in a folder if a student drags a file onto a folder.

## Your Changes

**Description**:
`markus_file_manager.jsx`:

- Implemented Drag and Drop for Header
- Created component called `RawFileManagerFolder` to make drag and dropping on folders work with the current implementation of the libraries.

`submission_file_manager.jsx`:

- Modified `handleCreateFiles` to accept `path` as argument. 

`starter_file_manager.jsx`, `autotest_manager.jsx`:
- Modified `handleCreateFiles` to accept `path` as argument. 
- Passing `onCreateFiles` to `FileManager` to enable drag and drop functionality. 

`file_upload_modal`, `submission_file_upload_modal`
- Accounting for the new path argument (passing undefined for that)

## Testing

No additional tests were added. The old rspec and jest suites were ran just in case.

## Checklist


david-yz-liu (commented):
@Dhairya-Khara given that the component you're talking about actually is a table header, you shouldn't change its parent class since we should be sticking with the library as closely as possible. I think you need to do more research into the libraries themselves.

david-yz-liu (commented):
@Dhairya-Khara great research! One general issue I noticed when testing is that if there is a file in a folder (e.g. `my_folder/a1.py`), then if the user drags a file on top of `a1.py` (not the folder, the file itself), then the file gets added to the top-level, rather than inside the folder `my_folder`. Seems like the drag target path isn't being found/stored correctly. You might find https://github.com/uptick/react-keyed-file-browser/blob/master/src/base-file.js#L191-L207 useful.

app/assets/javascripts/Components/markus_file_manager.jsx
```diff
@@ -250,16 +256,38 @@ class FileManagerHeader extends Headers.TableHeader {
...
       return header;
     }
   }
 }
 
+class RawFileManagerFolder extends FolderRenderers.RawTableFolder {
+  componentDidUpdate(prevProps) {
+    if (prevProps.isOver !== this.props.isOver) {
+      if (this.props.isOver) {
+        console.log(this.props.fileKey);
```
david-yz-liu, line 270: Remove this line

david-yz-liu (commented):
@Dhairya-Khara great work! I just left one small cleanup comment

david-yz-liu (dismissed):

david-yz-liu (commented):
@Dhairya-Khara sorry I didn't think about this earlier, but the file manager component is used in other places beyond just student submission files. This means that you'll need to update the passed props in all places where the component is used.

david-yz-liu (approved):

Issue #5831, opened 2022-02-04
Statistics page


mishaschwartz:
Right now we show general statistics about individual assignments on the dashboard page but it would be interesting to show course wide statistics in a separate view (visible to instructors).

Where to find ideas of what sort of statistics to show on this page:

- Look at other LMS software and see what they do (canvas, blackboard, moodle, etc.)
- Talk to instructors and other MarkUs users about what would be useful for them
- Talks to other MarkUs developers about what sort of statistics are available

due date: end of summer 2022 (so it can be included in version 2.2.0)

anoverga17:
This issue is now resolved. Rather than creating a general course wide statistics survey, results from a MarkUs survey conducted showed instructors would rather want more specific and clearer summary statistics. Now that these summary statistics are enhanced this issue is now resolved. 

The only thing statistics related left that may require a bit of thought is user analytics which are described in a separate pull request [here](https://github.com/MarkUsProject/Markus/issues/6190)

PR #6118, opened 2022-07-26
Added new summary statistics display for grade entry forms and grade entry column items

anoverga17:
## Motivation and Context

As part of issue #5831, part of the survey feedback given by instructors was to improve how grade entry forms are shown. This pull request therefore aims to refactor the grade entry form to make use of the reworked assignment chart layout to provide more detail about grade entry form statistics and provide a dedicated summary statistics page for grade entry forms.

## Your Changes

**Description**:
- Refactored shared grade entry form layout with assignment chart to an assessment chart component
- Added PropTypes to assessment chart component
- Passed assignment chart specific components and links as props
- Passed grade entry form specific components and links as props
- Added new grades tablist and summary statistics tab that shows an enlarged version of grade entry chart
- Generalized naming of criteria / column distribution data to just grade breakdown distribution data for front end use
- Updated rspec tests

These changes add the following new grade entry summary view:
Grade Entry Form Summary Tab

In addition the dashboard summary for grade entry forms has been slightly modified to now look like:
Grade Entry Form Dashboard Summary

## Testing

Manual testing was done via the UI to ensure the correct content was loaded properly. Rspec tests were also updated to check for errors with minor reworking to the naming of the grade distribution data provided.

## Checklist

https://github.com/MarkUsProject/Wiki/pull/136


mishaschwartz (commented):
I think overall this will work. However, the more if/else statements I see in the components the more I think that a cleaner option would be to create an abstract AssessmentChart component that contains all the common stuff and then have an AssignmentChart and GradeEntryForm chart that inherit from the AssessmentChart and override/implement the bits that are specific to that type of assessment.

This would reduce the number of if blocks by a lot and would clean up the code quite a bit in my opinion. What do you think?

anoverga17:
> I think overall this will work. However, the more if/else statements I see in the components the more I think that a cleaner option would be to create an abstract AssessmentChart component that contains all the common stuff and then have an AssignmentChart and GradeEntryForm chart that inherit from the AssessmentChart and override/implement the bits that are specific to that type of assessment.
> 
> This would reduce the number of if blocks by a lot and would clean up the code quite a bit in my opinion. What do you think?

Huh, that certainly would clean things up significantly. I never really considered that react components could be abstract but if they could that certainly would be quite clean. I'll give that a try.

mishaschwartz:
> I never really considered that react components could be abstract

Yeah as long as we're using the class based react syntax we can use all our OOP tricks!

anoverga17:
> > I never really considered that react components could be abstract
> 
> Yeah as long as we're using the class based react syntax we can use all our OOP tricks!

So I've tried it out and I just learned through [here](https://reactjs.org/docs/composition-vs-inheritance.html) that inheritance is not something React supports nor encourages. I have thus reworked the hierarchical classes it to the more 'react' way of composition. Let me know what you think of this.

mishaschwartz:
That's fine. Feel free to organize the code in a better way if that makes more sense. My only other comment is that I would not put so much logic in the render function (like [here](https://github.com/MarkUsProject/Markus/blob/02e549c679acbbfab9d1c639d04731311245b1f4/app/assets/javascripts/Components/assignment_chart.jsx)). Especially code that makes requests to the backend should be organized in its own function on the class.

mishaschwartz (commented):
I understand that this change is copying over a lot of code from previous work but there are some organizational things that I would really like to see changed. @david-yz-liu please let me know if these are not priorities for this PR though.

- data fetching should be done in the parent component, stored in its state and then the relevant data should be passed to child components as props. In this case, we want the AssignmentChart and GradeEntryChart components to be calling fetchData and then passing the info to AssessmentChart (not having AssessmentChart pass data back up to the parent components with a callback function)
- a lot of things that are fragments can and should be their own components. For example, the [assessment_graph](https://github.com/MarkUsProject/Markus/pull/6118/files#diff-6ea6452429a727eb6bab6948d5967fdc8847bda5c66d568faf9114d344ec6f35R83) variable here can be it's own component. 
- Data that doesn't change should not be stored in the state of a component. Either use a const or a prop with a default value.
- On the dashboard page, the two charts should really be the same size. I know that the top one has additional stats information on the side but can we put those in columns or make use of flexboxes or something so that the chart can grow with the page width and the secondary chart doesn't look so large in comparison?

david-yz-liu:
@anoverga17 @mishaschwartz thanks! Please see my comments below.

> * data fetching should be done in the parent component, stored in its state and then the relevant data should be passed to child components as props. In this case, we want the AssignmentChart and GradeEntryChart components to be calling fetchData and then passing the info to AssessmentChart (not having AssessmentChart pass data back up to the parent components with a callback function)

Yeah that's good for this PR :+1:

> * a lot of things that are fragments can and should be their own components. For example, the [assessment_graph](https://github.com/MarkUsProject/Markus/pull/6118/files#diff-6ea6452429a727eb6bab6948d5967fdc8847bda5c66d568faf9114d344ec6f35R83) variable here can be it's own component.

This is necessary for this PR. We can discuss this on a case by case basis later, but overall it's always a design decision about what to pull out into a separate component.

> * Data that doesn't change should not be stored in the state of a component. Either use a const or a prop with a default value.

Yes this is good for this PR :+1:

> * On the dashboard page, the two charts should really be the same size. I know that the top one has additional stats information on the side but can we put those in columns or make use of flexboxes or something so that the chart can grow with the page width and the secondary chart doesn't look so large in comparison?

I agree that it would be nice for the graphs to be the same size, but I'll review the UI later as well. So for now @anoverga17 please note but don't change the CSS.

anoverga17:
Thanks @mishaschwartz @david-yz-liu I'll note that and get to it!

app/assets/javascripts/Components/grade_entry_form_chart.jsx
```diff
@@ -68,56 +65,37 @@ export class GradeEntryFormChart extends React.Component {
...
-          width="400"
-          height="350"
-        />
-      </React.Fragment>
+          </React.Fragment>
+        }
+        secondary_grade_distribution_title={I18n.t(
+          "grade_entry_forms.grade_entry_item_distribution"
+        )}
+        secondary_grade_distribution_data={this.state.column_grade_distribution.data}
```
david-yz-liu, line 97: @anoverga17 great work! I have to jump into a meeting so I'm giving only two high-level comments in this review. The first is here: grade entry form *columns* are analogous to assignment *criteria*, and so should be displayed in the same way. In other words you should compute the same stats for each column like you did for each criterion, and display the column grade information by passing the data to the `criteria` prop rather than as "`secondary_grade_distribution`". You can use props to customize the title/labels of "criteria" versus "columns".

In terms of the Dashboard: I think the column grades are being shown in the dashboard graphs, which is maybe why you chose to use the "`secondary_grade_distribution`" props for that. To preserve the current UI, maybe you can use a `showCriteriaSummary` prop or something.

(As a note, grade entry forms *also* assign graders to students, so the stats summary could display grader information analogous to that of assignments. But in the interest of time you don't need to add that in this PR.)

app/controllers/grade_entry_forms_controller.rb
```diff
@@ -38,6 +38,13 @@ def update
     GradeEntryStudent.refresh_total_grades(@grade_entry_form.grade_entry_students.ids)
   end
 
+  def summary
```
david-yz-liu, line 41: My second comment is that overall I'd like to see the statistics displayed in a tab alongside the existing grades table; this is the direct analogue to the Assignment summary tab.

So to be clear, the "Grades" view can have a "Grades" tab and "Summary Statistics" tab

app/assets/javascripts/Components/assessment_chart.jsx
```diff
@@ -0,0 +1,215 @@
+import React from "react";
+import {Bar} from "react-chartjs-2";
+import PropTypes from "prop-types";
+import {chartScales} from "./Helpers/chart_helpers";
+import ReactTable from "react-table";
+
+class AssessmentChart extends React.Component {
```
anoverga17, line 7: With the way I redesigned this, I wonder if it might be better to split the assessment graph and grade breakdown graph into two separate helper components.

Changelog.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [unreleased]
+- Refactored grade entry form component to make use of assignment chart (#6118)
```
david-yz-liu, line 4: I think this entry could be improved a bit. The high-level info for users is we're now displaying summary statistics for grade entry forms, plus that for individual columns. The fact that you've refactored the code to reduce duplication is more of an internal consideration.

app/assets/javascripts/Components/Assessment_Chart/fraction_statistic.jsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+export class FractionStat extends React.Component {
```
david-yz-liu, line 3: The name of the component should be the same as the file name. I'm fine with `FractionStat`, so you should rename the file to `fraction_stat.jsx`

david-yz-liu (commented):
@anoverga17 great job! I just left a couple of comments for cleanup.

david-yz-liu (approved):

PR #6119, opened 2022-07-30
jest for SubmissionSelector component

WeissSa:
Added testing for SubmissionSelector

## Motivation and Context
To increase test coverage on the React components.

## Your Changes
Added testing and proptype validation on SubmissionSelector.

## Checklist


app/assets/javascripts/Components/Result/submission_selector.jsx
```diff
@@ -140,3 +144,25 @@ export class SubmissionSelector extends React.Component {
     );
   }
 }
+
+SubmissionSelector.propTypes = {
+  assignment_max_mark: PropTypes.number,
```
david-yz-liu, line 149: @WeissSa this prop type is logically correct, but currently is flagged as an error in the actual web application. Why don't you dig into this a little bit and modify the React components as necessary to ensure the prop value is always a number?

WeissSa: Should be fixed I think.

app/assets/javascripts/Components/Result/result.jsx
```diff
@@ -714,7 +714,7 @@ class Result extends React.Component {
           submission_id={this.state.submission_id}
           assignment_id={this.state.assignment_id}
           grouping_id={this.state.grouping_id}
-          assignment_max_mark={this.state.assignment_max_mark}
+          assignment_max_mark={parseFloat(this.state.assignment_max_mark)}
```
david-yz-liu, line 717: @WeissSa This does work but I don't think this is the root of the issue. Either the data from the server isn't being sent as a number (so we should fix it on the controller end) or it is, but then is being converted into a string when stored as state in this component.

app/models/assignment.rb
```diff
@@ -267,7 +267,7 @@ def display_for_note
 
   # Returns the maximum possible mark for a particular assignment
   def max_mark(user_visibility = :ta_visible)
-    criteria.where(user_visibility => true, bonus: false).sum(:max_mark).round(2)
+    Float(criteria.where(user_visibility => true, bonus: false).sum(:max_mark).round(2))
```
david-yz-liu, line 270: @WeissSa Okay this is interesting! The `.sum` here is returning a `BigDecimal`, which gets represented as a string when encoded into JSON. Since we're rounding to two decimal places I'm okay returning a `Float` here like you're doing, but please explain in the method docstring that we're converting the `BigDecimal` to a `Float`.

WeissSa: Will do 🖖

david-yz-liu (approved):

PR #6138, opened 2022-08-01
Issue-5394 -  create_modify_annotation_panel_modal

### LINKED ISSUE #5394: Markdown/Latex preview should be in a separate tab
Previewing markdown text is currently displayed below the text. Instead the preview should be in a separate tab (similar to the way github does it). 

This will give the advantage of displaying the preview on demand which should cut down on clutter on the page.

WeissSa:
Fixes #5394 

## Motivation and Context
This allows for a nicer view of latex text in a separate tab.

## Your Changes
Used react-tabs to add 2 tabs to the modal, 1 writing and 1 for preview (just like on Github). The logic of the page is the same, but the tabs make it so that the preview/writing view not-intended for the current tab is set to display-none.

## Testing
Since not much code/functionality actually changed I am unsure how much testing to add as of right now. If asked for I will write a Jest file for the annotation_panel_modal

+1 on unsure testing

## Checklist


david-yz-liu (commented):
@WeissSa okay great start, but we can go even further. Currently the preview div is updated in real-time as the user types text; but by putting it in a separate tab, we can delay the rendering until the user clicks on the preview tab. So what you should do is create a separate React component for the preview, and pass in the contents to display as a prop. Then that React component can be in charge of triggering the markdown and MathJax rendering when it's mounted/updated.

(This will also allow you to remove the existing event listener that's doing the real-time updating of the preview.)

WeissSa:
I think this is the implementation you're looking for. By removing forceRenderTabPanel from the tabs, it makes it so that the new component is only mounted when we are present in the tab. This means it is not updating in real time.

app/assets/javascripts/Components/markdown_preview.jsx
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+export default class MarkdownPreview extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  componentDidMount = () => {
```
david-yz-liu, line 9: There's something missing here compared to the old `updatePreview` method. Did you try testing the preview with annotations that use markdown syntax?

app/assets/javascripts/Components/Modals/create_modify_annotation_panel_modal.jsx
```diff
@@ -199,108 +193,114 @@ class CreateModifyAnnotationPanel extends React.Component {
...
-
-                <h3>{I18n.t("preview")}</h3>
-
-                <input
-                  type="hidden"
-                  id="annotation_text_id"
-                  name="annotation_text_id"
-                  value={this.state.annotation_text_id}
-                />
+        <Tabs>
```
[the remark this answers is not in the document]
WeissSa: I think both of these should be fixed now (at least it's working on my local) @david-yz-liu

WeissSa:
Test failing is a flaky test

app/assets/javascripts/Components/Modals/create_modify_annotation_panel_modal.jsx
```diff
@@ -199,108 +193,116 @@ class CreateModifyAnnotationPanel extends React.Component {
...
-                    type="hidden"
-                    id="new_annotation_category"
-                    name="category_id"
-                    value={I18n.t("annotation_categories.one_time_only")}
-                  />
-                ) : (
-                  <div className="inline-labels">
-                    <label htmlFor="new_annotation_category">
-                      {I18n.t("activerecord.models.annotation_category.one")}
+        <Tabs disableUpDownKeys>
```
david-yz-liu, line 196: The reason I commented on this line in particular is that I wanted you to move it. So to be clear, you should move `Tabs` so that it's inside the `div` within the `form` itself.

app/assets/javascripts/Components/Modals/create_modify_annotation_panel_modal.jsx
```diff
@@ -199,108 +193,116 @@ class CreateModifyAnnotationPanel extends React.Component {
...
-                    id="new_annotation_content"
-                    name="content"
-                    placeholder={I18n.t("results.annotation.placeholder")}
-                    value={this.state.content}
-                    onChange={this.handleChange}
-                    rows="8"
-                    autoFocus={true}
-                  />
-                </label>
-
```
david-yz-liu, line 225: Revert this change

app/assets/javascripts/Components/markdown_preview.jsx
```diff
@@ -0,0 +1,31 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+export default class MarkdownPreview extends React.Component {
+  componentDidMount = () => {
+    const target_id = "annotation_preview";
+    MathJax.Hub.Queue(["Typeset", MathJax.Hub, target_id]);
+  };
+
+  componentWillUnmount = () => {
```
david-yz-liu, line 10: I think I know what you're trying to do here, but it's not the right way to solve the problem. You can pass the `forceRender` prop to the "write" tab so that it's always rendered, to ensure that the event listeners for the autocomplete stay. (That's probably the simplest approach. Another option would have been to make the text input its own react component, and have the event listeners registered within that component.)

app/assets/javascripts/Components/markdown_preview.jsx
```diff
@@ -0,0 +1,31 @@
...
+    setTimeout(this.props.updateAnnotationCompletion(), 300);
+  };
+
+  render() {
+    return (
+      <div
+        id="annotation-preview"
+        style={{
```
david-yz-liu, line 18: Remove this style. The preview `div` had its own class with some CSS attached to it (which you should keep).

david-yz-liu (approved):
@WeissSa please also update the Wiki (specifically, this page: https://github.com/MarkUsProject/Wiki/blob/master/Instructor-Guide--Assignments--Marking--Grading-View.md#annotations)

PR #6139, opened 2022-08-02
text viewer: Ensure syntax highlighter only applies to local component

david-yz-liu:
## Motivation and Context

The current syntax highlighting implementation (in `app/assets/javascripts/Components/Result/text_viewer.jsx`) assumes that only one text viewer component is rendered. We can see this in two ways:

1. `dp.SyntaxHighlighter.HighlightAll` performs syntax highlighting for all `pre` tags with `name="code"` in the entire document. (This function doesn't have a way to limit its search to a subtree of the DOM.)
2. The `this.highlight_root` attribute was being assigned to `document.getElementsByClassName("dp-highlighter")[0]`.

Because of this, when there were both plaintext submission files and feedback files in the grading view, the plaintext was being "syntax highlighted" multiple times, resulting in duplicated code appearing. (In the dev environment this is noticeable for the short submission files like `hello.py`.)

## Your Changes

**Description**: Modified the `TextViewer` component to ensure syntax highlighting is only applied to the content of that component.

## Testing

Tested in the UI. Verified that various grading actions (e.g. switching between submission files and feedback files, adding annotations) doesn't result in multiple syntax highlighted text appearing.

It would be good in the future to switch to a React library for syntax highlighting like https://github.com/react-syntax-highlighter. That would free us from manually controlling the syntax highlighting (which is particularly awkward from within React because the SyntaxHighlighter library manipulates the DOM directly). I'll make an issue for this.

## Checklist

### DIFF
--- a/app/assets/javascripts/Components/Result/text_viewer.jsx
+++ b/app/assets/javascripts/Components/Result/text_viewer.jsx
@@ -1,8 +1,6 @@
 import React from "react";
 import {render} from "react-dom";
 
-const RAW_TEXT_DIV_ID = "code";
-
 export class TextViewer extends React.Component {
   constructor(props) {
     super(props);
@@ -12,13 +10,12 @@ export class TextViewer extends React.Component {
     };
     this.highlight_root = null;
     this.annotation_manager = null;
+    this.raw_content = React.createRef();
   }
 
   componentDidMount() {
     if (this.props.content) {
-      this.ready_annotations(RAW_TEXT_DIV_ID);
-      this.props.annotations.forEach(this.display_annotation);
-      this.scrollToLine(this.props.focusLine);
+      this.ready_annotations();
     }
   }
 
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/Components/Result/text_viewer.jsx
```diff
@@ -131,7 +142,7 @@ export class TextViewer extends React.Component {
             </a>
           </div>
         </div>
-        <pre id={RAW_TEXT_DIV_ID} name={RAW_TEXT_DIV_ID} className={this.props.type}>
+        <pre name={preElementName} ref={this.raw_content} className={this.props.type}>
```
mishaschwartz, line 145: Just to clarify.... the `SyntaxHighlighter.HighlightAll` looks for an element with the same "name" attribute as its first argument. The fact that we were giving this `pre` element the same "id" value was just a coincidence?

And we're not setting the id now because it's not needed?

mishaschwartz (approved):

PR #6145, opened 2022-08-04
Student table edit students

TheTallJerry:
## Motivation and Context

We should make the student table's dropdown and student table's edit option's functionalities consistent. 

## Your Changes

**Description**:
- Allows instructors to bulk update students' sections in student table
#### Current student table dropdowns
old
#### Updated student table dropdowns
new

- Allows instructors to update individual student's status (inactive / active) from the edit page, accessible via the "edit" option in student table
#### Current student edit page
old student edit
#### Updated student edit page
new student edit

## Testing

Ran rspec and Jest tests. 

## Checklist

https://github.com/MarkUsProject/Wiki/pull/142


app/assets/javascripts/Components/student_table.jsx
```diff
@@ -241,6 +241,9 @@ class StudentsActionBox extends React.Component {
             onChange={this.inputChanged}
             data-testid={"student_action_box_update_section"}
           >
+            <option key={"none"} value={""}>
+              {I18n.t("students.instructor_actions.no_section")}
```
david-yz-liu, line 245: This string is good, and I think we should use it in the student form as well (rather than using an empty string)

app/views/students/_student_form.html.erb
```diff
@@ -21,6 +21,11 @@
                   remote: true,
                   class: 'button inline-button' %>
     </span>
+    <%= f.label :hidden, t('students.update_status') %>
```
david-yz-liu, line 24: To match the other labels on the form this string should just be "Status"

app/views/students/_student_form.html.erb
```diff
@@ -21,6 +21,11 @@
...
+    <span id='set-active'>
```
david-yz-liu, line 25: I'm not sure why you included a `span` here, the layout should be okay with just the `select` tag (I think)?

david-yz-liu (commented):
@TheTallJerry nice job.

1. You should update the Changelog and the wiki.
2. You should add tests for the new features on the students table (setting section to nil) and student form (updating `hidden` status)

app/views/students/_student_form.html.erb
TheTallJerry: We have styling in place that will fill the select with more space without the span tag. Removing the tag in this case would look like:
Capture
I feel it's easier to just have the tag here.

spec/controllers/students_controller_spec.rb
```diff
@@ -41,6 +59,61 @@
       end
     end
 
+    describe '#bulk_modify' do
```
TheTallJerry, line 62: I realized bulk_modify is actually entirely untested, so in addition to tests for bulk setting sections, I added tests for bulk setting everything else too.

david-yz-liu (approved):
@TheTallJerry thanks!

PR #6154, opened 2022-08-07
Issue-5394 autosave_text_form

### LINKED ISSUE #5394: Markdown/Latex preview should be in a separate tab

WeissSa:
## Motivation and Context
Fixes #5394 

## Your Changes
Changed autosave_text_form to use the new  preview/write tab system.

Extracted markdown-tabs to new component.

## Testing
Manually tested and wrote tests or new extracted component.

I daisy-chained this off my other PR which seems to have been a bad idea. I will handle merge conflicts after I open this.

## Checklist

- [ ] I have made changes to the documentation and linked to that pull request below.


app/assets/javascripts/Components/markdown_edit.jsx
```diff
@@ -0,0 +1,78 @@
...
+import {Tab, Tabs, TabList, TabPanel} from "react-tabs";
+
+import PropTypes from "prop-types";
+
+export default class MarkdownEdit extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      content: this.props.content,
```
david-yz-liu, line 12: Okay so overall it's not great to duplicate props and state values. You can remove this state and have this component rely entirely on its `props.content` value, and pass `this.props.handleChange` directly to the underlying `textarea` field.

app/assets/javascripts/Components/markdown_edit.jsx
```diff
@@ -0,0 +1,78 @@
...
+            <ul className="tags" key="annotation_completion" id="annotation_completion">
+              <li className="annotation_category" id="annotation_completion_li">
+                <p id="annotation_completion_text"></p>
+                <div>
+                  <ul id="annotation_text_list"></ul>
+                </div>
+              </li>
+            </ul>
+          </div>
+          <input
```
david-yz-liu, line 53: This `input` shouldn't be in this component (and it doesn't need to be, it just needs to be inside the `form` to be detected for submission. Removing it here will also let you get right of the `annotation_text_id` prop.

app/assets/javascripts/Components/markdown_edit.jsx
```diff
@@ -0,0 +1,78 @@
...
+  handleChange = event => {
+    this.setState({content: event.target.value});
+    this.props.handleChange(event);
+  };
+
+  render() {
+    return (
+      <Tabs disableUpDownKeys>
+        <TabList>
+          <Tab id="write-tab">{I18n.t("write")}</Tab>
```
david-yz-liu, line 26: I don't think setting `id` on React components (instead of HTML elements) has any effect, though maybe I could be wrong. But in any case if you aren't using these ids you should remove them.

app/assets/javascripts/Components/markdown_edit.jsx
```diff
@@ -0,0 +1,78 @@
...
+          <Tab id="preview-tab">{I18n.t("preview")}</Tab>
+        </TabList>
+        <TabPanel forceRender>
+          <label>
+            <textarea
+              required={true}
+              id="new_annotation_content"
```
david-yz-liu, line 33: Part of making this component more reusable is removing reference to specific applications (e.g. "annotations"). For now you can make a prop that allows an id to be set, since it does look like the `new_annotation_content` id is necessary for the annotation modal autocomplete, at least for now.

app/assets/javascripts/Components/markdown_edit.jsx
```diff
@@ -0,0 +1,78 @@
...
+              placeholder={I18n.t("results.annotation.placeholder")}
+              value={this.state.content}
+              onChange={this.handleChange}
+              rows="8"
+              autoFocus={true}
+            />
+          </label>
+
+          <div className={this.props.show_autocomplete ? "" : "hidden"}>
+            <ul className="tags" key="annotation_completion" id="annotation_completion">
```
david-yz-liu, line 44: Similarly, it's good to have this autocompletion subcomponent here since it might be nice to use it more in the future, but its classes should be renamed to avoid specific mention of annotations. Also remove the `id`s here, since `id`s need to be unique but in principle we could have more than one instance of this component rendered on the page at a time.

david-yz-liu (commented):
@WeissSa great work! I left some comments to make the reusable component even more... reusable. :)

app/assets/javascripts/Components/markdown_edit.jsx
WeissSa: The id is set on the second tab for testing. I set it on the first tab as well for consistency. I will remove it form the first tag.

app/assets/javascripts/Components/markdown_edit.jsx
WeissSa: I removed all unnecessary ids and made the necessary ones props :+1:

app/assets/javascripts/Components/markdown_editor.jsx
```diff
@@ -69,10 +49,13 @@ export default class MarkdownEdit extends React.Component {
...
+MarkdownEditor.propTypes = {
   annotation_text_id: PropTypes.string,
   content: PropTypes.string.isRequired,
   handleChange: PropTypes.func.isRequired,
   show_autocomplete: PropTypes.bool,
+  text_area_id: PropTypes.string,
+  auto_completion_text_id: PropTypes.string,
+  auto_completion_list_id: PropTypes.string,
   updateAnnotationCompletion: PropTypes.func,
 };
```
WeissSa, line 61: I think it could be worth adding some documentation to clarify when the props are mandatory if show_autocomplete is on. What do you think?

app/assets/javascripts/Components/markdown_editor.jsx
```diff
@@ -0,0 +1,61 @@
...
+      </Tabs>
+    );
+  }
+}
+
+MarkdownEditor.propTypes = {
+  annotation_text_id: PropTypes.string,
+  content: PropTypes.string.isRequired,
+  handleChange: PropTypes.func.isRequired,
+  show_autocomplete: PropTypes.bool,
```
david-yz-liu, line 56: let's make this required

app/assets/javascripts/Components/markdown_editor.jsx
```diff
@@ -0,0 +1,61 @@
...
+            updateAnnotationCompletion={this.props.updateAnnotationCompletion}
+          />
+        </TabPanel>
+      </Tabs>
+    );
+  }
+}
+
+MarkdownEditor.propTypes = {
+  annotation_text_id: PropTypes.string,
```
david-yz-liu, line 53: You should remove this prop altogether, it's not being used anywhere. (And don't forget to check the places this component is being used.)

david-yz-liu (commented):
@WeissSa okay great, I left a couple more comments. Regarding your question about when `show_autocomplete` is true, it's okay to have the id fields be blank, so those props aren't required.

david-yz-liu (approved):
@WeissSa great work!

PR #6156, opened 2022-08-10
initial implementation of lti1.3 ags

pretendWhale:
## Motivation and Context

This PR aims to enable support for the LTI 1.3. Assignment and Grades services as a tool provider. 

## Your Changes

**Description**:
Adds new `lti_line_item` model, for line items (i.e. gradebook entries), as well as supporting a line item `lti_service`. Also adds a new section in the assignment form for line item creation and a modal for display on the assignment summary page which will sync grades with a previously created line item. 

Still have some todos: add appropriate error messaging, add lti sync functionality to grade entry forms, and adding tests.

- [x] Other (please specify): LTI features

## Testing

Tested in web browser with local canvas install. Tests to be added.

## Checklist


app/assets/javascripts/Components/Modals/send_lti_grades_modal.jsx
```diff
@@ -0,0 +1,96 @@
...
+
+class LtiGradeModal extends React.Component {
+  static defaultProps = {
+    override: false,
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      deployments: this.props.lti_deployments,
```
mishaschwartz, line 12: If you're not updating this value ever there is no need to store it in the state. Just refer to the prop directly

app/assets/javascripts/Components/Modals/send_lti_grades_modal.jsx
```diff
@@ -0,0 +1,96 @@
...
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  componentDidMount() {
+    Modal.setAppElement("body");
+    const deploymentsMapped = this.state.deployments.map(deployment => ({
+      deployment: deployment.id,
+      checked: true,
+    }));
+    this.setState({deploymentsChecked: deploymentsMapped});
```
mishaschwartz, line 23: This should have a default value set in the constructor. Even if you want it to be explicitly undefined, write that in the constructor so that the component is aware of the potential state keys from the beginning.

app/assets/javascripts/Components/Modals/send_lti_grades_modal.jsx
```diff
@@ -0,0 +1,96 @@
...
+  static defaultProps = {
+    override: false,
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      deployments: this.props.lti_deployments,
+    };
+    this.handleChange = this.handleChange.bind(this);
```
mishaschwartz, line 14: I'm almost certain this isn't necessary. What problem are you solving by adding this?

app/assets/javascripts/Components/Modals/send_lti_grades_modal.jsx
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  componentDidMount() {
+    Modal.setAppElement("body");
+    const deploymentsMapped = this.state.deployments.map(deployment => ({
```
mishaschwartz, line 19: Are you sure you want this to run in componentDidMount instead of componentDidUpdate? What happens if the props change while the component has already been mounted? Is that a possibility?

app/assets/javascripts/Components/Modals/send_lti_grades_modal.jsx
```diff
@@ -0,0 +1,96 @@
...
+    const deployment_id = target.name;
+    let deploymentsSettings = this.state.deploymentsChecked;
+    let setting = deploymentsSettings.find(deployment => deployment.deployment == deployment_id);
+    setting.checked = value;
+    this.setState({deploymentsChecked: deploymentsSettings});
+  };
+
+  sendLtiGrades = () => {
+    const checked = this.state.deploymentsChecked.filter(deployment => deployment.checked == true);
+    console.log(checked);
```
mishaschwartz, line 44: remove log/debugging code

app/assets/javascripts/Components/Modals/send_lti_grades_modal.jsx
```diff
@@ -0,0 +1,96 @@
...
+    const deploymentsToUpdate = checked.map(deployment => deployment.deployment);
+    console.log(deploymentsToUpdate);
```
mishaschwartz, line 46: remove log/debugging code

app/controllers/lti_deployment_controller.rb
```diff
@@ -175,6 +186,15 @@ def create_course
     redirect_to edit_course_path(new_course)
   end
 
+  def create_lti_grades
+    assessment = Assessment.find(params[:assessment_id])
```
mishaschwartz, line 190: If this path requires an assessment_id then nest the route under "assessments". 

See my comment on routes.rb

app/controllers/assignments_controller.rb
```diff
@@ -252,6 +255,13 @@ def create
 
   def summary
     @assignment = record
+    @lti_deployments = LtiLineItem.where(assessment_id: @assignment.id)
+                                  .joins(:lti_deployment)
```
mishaschwartz, line 259: this joins is unnecessary (the one below it includes the same info)

app/controllers/assignments_controller.rb
```diff
@@ -252,6 +255,13 @@ def create
...
+                                  .joins(lti_deployment: :lti_client)
+                                  .pluck_to_hash('lti_deployments.id',
+                                                 'lti_clients.host',
+                                                 'lti_deployments.lms_course_name')
+    @lti_deployments.each { |deployment| deployment.deep_transform_keys! { |key| key.to_s.split('.')[-1] } }
```
mishaschwartz, line 264: I'm pretty sure you can just use `transform_keys!` here because none of the hashes are nested

app/models/grade_entry_form.rb
```diff
@@ -48,6 +48,14 @@ def completed_result_marks
                                   .pluck(:total_grade)
   end
 
+  def released_marks
+    self.grade_entry_students.joins(:role)
```
mishaschwartz, line 52: this first joins is unnecessary

db/migrate/20220727161425_create_lti_line_items.rb
```diff
@@ -0,0 +1,14 @@
+class CreateLtiLineItems < ActiveRecord::Migration[7.0]
+  def change
+    create_table :lti_line_items do |t|
+      t.string :lti_line_item_id, null: false
+      t.references :assessment, null: false, foreign_key: true
+      t.references :lti_deployment, null: false, foreign_key: true
+
+      t.timestamps
+    end
+    add_column :lti_deployments, :lms_course_id, :integer, null: true
```
mishaschwartz, line 10: Why can the lms_course_id be null but the lms_course_name not be null?

spec/models/lti_deployment_spec.rb
```diff
@@ -143,4 +143,167 @@
...
     end
   end
+  describe '#create_or_update_lti_assessment' do
+    let(:course) { create :course }
+    let(:lti_deployment) { create :lti_deployment, course: course }
+    let(:assessment) { create :assignment, course: course }
+    let(:scope) { 'https://purl.imsglobal.org/spec/lti-ags/scope/lineitem' }
+    let(:lti_service_lineitem) { create :lti_service_lineitem, lti_deployment: lti_deployment }
+    before :each do
+      stub_request(:post, Settings.lti.token_endpoint)
```
mishaschwartz, line 153: What happened to the idea of creating a `send_request` method and just stubbing that?

mishaschwartz: Please do that instead of this. This is going to be unnecessarily difficult to maintain.

mishaschwartz (changes requested):

app/views/lti_deployment/_line_items.html.erb
```diff
@@ -0,0 +1,24 @@
+<table>
```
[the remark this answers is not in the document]
pretendWhale: Per discussion, leaving this for a later rework of the Assignment Settings page

db/migrate/20220727161425_create_lti_line_items.rb
pretendWhale: You're right, they should both be not nullable

spec/models/lti_deployment_spec.rb
pretendWhale: Per discussion this will be a separate PR

config/routes.rb
```diff
@@ -490,6 +490,7 @@
       get 'choose_course'
       post 'choose_course'
       post 'create_course'
+      post 'create_lti_grades'
```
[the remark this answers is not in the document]
pretendWhale: Per discussion this will be a separate PR

app/controllers/lti_deployment_controller.rb
pretendWhale: Per discussion this will be a separate PR

app/assets/javascripts/Components/Modals/send_lti_grades_modal.jsx
pretendWhale: Right now, the props are never modified, and shouldn't be modified, but you're right - it makes sense to do it in componentDidUpdate

pretendWhale: After thinking about it, I think this does need to be in `componentDidMount`, because it sets the initial state which is then modified by user actions (checking or unchecking the box for each lti deployment).

mishaschwartz: Ok then it should be set in the constructor as an initial value in the state

app/assets/javascripts/Components/Modals/send_lti_grades_modal.jsx
```diff
@@ -0,0 +1,93 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      deploymentsChecked: undefined,
+    };
+  }
+
+  componentDidMount() {
+    Modal.setAppElement("body");
+    const deploymentsMapped = this.props.lti_deployments.map(deployment => ({
```
mishaschwartz, line 18: Please move this logic to the constructor and set the initial state value for the `deploymentsChecked` key instead of setting it here. 

This can cause an unnecessary re-render otherwise.

mishaschwartz (changes requested):

mishaschwartz (approved):

app/models/lti_deployment.rb
```diff
@@ -45,4 +48,90 @@ def get_students
...
+  def get_assignment_marks(assignment)
+    marks = assignment.released_marks
+    mark_data = {}
+    lti_users = LtiUser.where(lti_client: lti_client)
+    marks.each do |mark|
+      result = mark.results.first
+      group_students = mark.grouping.student_memberships
+      group_students.each do |member|
+        lti_user = lti_users.find_by(user: member.role.user)
+        mark_data[lti_user.lti_user_id] = result.total_mark unless lti_user.nil?
```
david-yz-liu, line 117: This should now be changed to `result.get_total_mark` (after merge from master)

app/views/lti_deployment/_line_items.html.erb
```diff
@@ -0,0 +1,24 @@
...
+  <tr>
+    <th><%= t("lti.lti_settings") %></th>
+  </tr>
+  </thead>
+  <tbody>
+  <%= fields_for :lti_deployment do |f|  %>
+  <% @lti_deployments.each do |deployment| %>
+    <tr>
+      <td>
+
```
david-yz-liu, line 12: delete this blank line

app/views/lti_deployment/_line_items.html.erb
```diff
@@ -0,0 +1,24 @@
...
+        <%= f.label deployment.id, "#{deployment.lms_course_name} on #{URI(deployment.lti_client.host).host}", value: deployment.id %>
```
david-yz-liu, line 13: 1. The label string should be internationalized. ("on" is only English)
2. Is the `value` argument necessary? I'm not sure what it does for a `label`

app/views/lti_deployment/_line_items.html.erb
```diff
@@ -0,0 +1,24 @@
...
+        <% if deployment.lti_line_items.find_by(assessment: @assignment) %>
```
david-yz-liu, line 14: I believe you can use `exists?` here

app/views/lti_deployment/_line_items.html.erb
```diff
@@ -0,0 +1,24 @@
...
+          <%= f.check_box deployment.id, checked_value: 'true', unchecked_value: 'false', checked: true, disabled: true %>
```
david-yz-liu, line 15: I'm not sure the `checked_value`/`unchecked_value` options are necessary here, given the default values?

app/models/lti_line_item.rb
```diff
@@ -0,0 +1,5 @@
+class LtiLineItem < ApplicationRecord
+  belongs_to :assessment
```
david-yz-liu, line 2: Please also add the corresponding `has_many` association to `Assessment`

spec/factories/lti_line_items.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :lti_line_item do
```
david-yz-liu, line 2: I believe this should have an association with an `assessment` as well (which by default you could have be an `:assignment`?)

spec/models/lti_deployment_spec.rb
```diff
@@ -143,4 +143,167 @@
...
+    end
+    it 'does send grades' do
+      expect(lti_deployment.create_grades(assessment)).not_to be_empty
+    end
+  end
+  describe '#get_assignment_grades' do
+    let(:course) { create :course }
+    let(:student1) { create :student, course: course }
+    let(:lti_deployment) { create :lti_deployment, course: course }
+    # let(:lti_user) { create :lti_user, user: student1, lti_deployment: lti_deployment}
```
david-yz-liu, line 264: Please delete this line if you aren't using it

spec/models/lti_deployment_spec.rb
```diff
@@ -143,4 +143,167 @@
...
+    let(:lti_deployment) { create :lti_deployment, course: course }
+    let!(:assessment) { create :assignment_with_criteria_and_results, course: course }
+
+    let(:scope) { 'https://purl.imsglobal.org/spec/lti-ags/scope/lineitem' }
+    let(:lti_service_lineitem) { create :lti_service_lineitem, lti_deployment: lti_deployment }
+    let(:lti_line_item) { create :lti_line_item, assessment: assessment, lti_deployment: lti_deployment }
+    let(:lti_service_namesrole) { create :lti_service_namesrole, lti_deployment: lti_deployment }
+    let(:status) { 'Active' }
+    before :each do
+      Result.joins(:grouping)
```
david-yz-liu, line 207: You don't need this `.joins` because `grouping` is included in the next line. (Same as the tests further down.)

spec/models/lti_deployment_spec.rb
```diff
@@ -143,4 +143,167 @@
...
+    end
+  end
+  describe '#get_grade_entry_form_marks' do
+    let(:course) { create :course }
+    let(:student1) { create :student, course: course }
+    let(:lti_deployment) { create :lti_deployment, course: course }
+    # let(:lti_user) { create :lti_user, user: student1, lti_deployment: lti_deployment}
+    let(:assessment) { create :grade_entry_form_with_data, course: course }
+    before :each do
+      student1
```
david-yz-liu, line 293: Use `let!` instead of `let` to force the creation of this object (and `assessment` on the next line)

spec/models/lti_deployment_spec.rb
```diff
@@ -143,4 +143,167 @@
...
+            .joins(grouping: :assignment)
+            .where('assignment.id': assessment2.id).first.update!(released_to_students: true)
+      expect(lti_deployment.get_assignment_marks(assessment2)).not_to be_empty
+    end
+  end
+  describe '#get_grade_entry_form_marks' do
+    let(:course) { create :course }
+    let(:student1) { create :student, course: course }
+    let(:lti_deployment) { create :lti_deployment, course: course }
+    # let(:lti_user) { create :lti_user, user: student1, lti_deployment: lti_deployment}
```
david-yz-liu, line 290: delete this line as well

app/controllers/assignments_controller.rb
```diff
@@ -252,6 +255,12 @@ def create
 
   def summary
     @assignment = record
+    @lti_deployments = LtiLineItem.where(assessment_id: @assignment.id)
```
david-yz-liu, line 258: 1. Please put move this into the `@assignment.summary_json` method so that it's included in the JSON payload rather than passed in as props to the table component.
2. Also only send this information when the current role is an instructor.

david-yz-liu (approved):

PR #6157, opened 2022-08-10
Reorganize files stored on disk to simplify stateless containers

### LINKED ISSUE #5523: Don't store files on disk (not as git repos). Either move them into the database or make them git repos:
- autotest settings files
- autotest files
- starter files
- etc.

mishaschwartz:
## Motivation and Context

MarkUs stores files on disk (repos, starter files, exam templates, autotest settings, lti keys). When MarkUs is running in a containerized environment, these files need to be stored in volumes (or similar) mounted on the container. This allows the container running MarkUs to be stateless which in turn allows us to run multiple containers that share the same application state. For example: if we add the following snippet to the docker-compose.yml file:

```yml
rails2:
  <<: *rails
  ports:
      - '3001:3000'
```

Then we can create another container running a rails server that uses exactly the same data as the other `rails` container. The only difference being one is accessible from port 3001 instead of port 3000.

We can technically do this now without the changes made in this PR. However, our configuration files are set up in such a way that does not encourage files to be written to disk in a central location which in turn simplifies the ability to mount these files as a volume. A sysadmin installing MarkUs would need to set multiple settings (with inconsistent naming conventions: "path" vs "storage" vs "dir") in order to ensure that files are all written to disk in the same place. 

This PR updates the configuration files in order to create a "file_storage" configuration option. This option takes a single default path where all files are written to. We also allow for this default path to be optionally overridden to still allow for flexible configuration.

This PR also simplifies file storage paths by using primary ids as file paths (guaranteed to be unique) instead of potentially non-unique identifiers like name attributes.

Resolves #5523

## Your Changes

**Description**:

We previously discussed storing scanned exam files in the tmp folder but then they can't be shared between containers and so each container would need it's own copy of those files. Instead I think we should discuss using something like activestorage instead. 

## Checklist


spec/models/assignment_spec.rb
```diff
@@ -2209,7 +2209,7 @@ def grouping_count(groupings)
   describe '#starter_file_path' do
     let(:assignment) { create :assignment }
     it 'should return a path that includes the repository folder' do
```
david-yz-liu, line 2211: This description should be updated

david-yz-liu (approved):
@mishaschwartz great work. I just merged from master (and fixed the changelog entry location). I'll merge the PR in after the tests pass (or you can if you get to it first).

PR #6158, opened 2022-08-10
Changing calendar widget

Dhairya-Khara:
The old jquery plugin for date and time picking is replaced with flatpickr.
## Motivation and Context

A step in transitioning away from JQuery.

## Your Changes

**Description**:

`package.json`, `yarn.lock`: Installed flatpickr
`en.yml`: Updated the date time format to match flatpickr's requirements.
`_form.html.erb`, `_peer_review_form.html.erb`: Imported Flatpickr's CSS
`_boot.html.erb`: Replaced the old Jquery plugin's config with flatpickr
`applicaiton_webpack`: Made flatpickr's exports global
Updated `manifes.js` to use flatpickr's CSS
`date_picker.jsx`: Using flatpickr for `Datepicker` Component
`flatpickr_config`: Set config for `flatpickr`. Used globally.
`assignments_controller.rb`, `automated_tests_controller.rb`, `content_security_policy.rb` - Removed code present because jquery-ui-timepicker-addon was being used

## Testing

No additional tests were written. The current RSpec and Jest test suites were ran just in case.

## Checklist


david-yz-liu (commented):
@Dhairya-Khara nice work! I just had time for a quick high level pass, please work on these things and I'll do a more detailed review later.

1. The existing jQuery UI timepicker is also used for grade entry form dates. You should do a search in other views to replace *all* uses of the old timepicker.
2. Related, you should set a global flatpickr config in a separate javascript file, so that this configuration applies by default to all uses of the library (similar to what we do for some other libraries like `chart.js`).
3. You can then delete the timepicker addon file from the `vendor` folder, since it won't be used anywhere else.
4. In the locale file, you shouldn't modify the existing `date` keys; the `time` keys contain standard format strings that include both date and time.

david-yz-liu (approved):

david-yz-liu:
@dan1982code sorry for the delay in replying, but yes this new flatpickr widget does indeed support direct text input for datetimes!

PR #6173, opened 2022-08-16
Changed how hidden roles are treated

### LINKED ISSUE #5826: Enable hiding graders and instructors
Currently we support hiding student roles, this does not delete them from the database but hides them from the UI and sets various policies so that the student can't interact with the course. 

We should also support the ability to hide graders and instructors users:

- no database changes needed because the `hidden` column already exists on the roles table
- change grader and instructor views to enable setting the hidden value through the UI
- update policies so that hidden roles cannot interact with the given course

anoverga17:
## Motivation and Context

Currently, there are hardly any differences between a hidden student and a normal student. In addition, graders and instructors currently cannot be hidden at all. This pull request aims to significantly update how hidden users are treated and what users can be hidden. This pull request resolves #5826.

## Your Changes

**Description**:
- Added hidden toggle for grader and instructor edit page
- Added hidden graders checkbox for assignment graders tab and also marked which graders are inactive
- Modified policy and session handler to not assign a role for hidden users who visit their course page
- Modified repo permissions to deny access to hidden users
- Modified repo permissions to deny access to students when course is hidden
- Modified policy to restrict hiding graders to instructors and hiding instructors to admin
- Changed default course visibility status to visible for rspec tests
- Updated repo permissions after updating visibility of roles and courses
- Updated rspec tests

These changes make the following UI changes:

Grader list now has active filter:
Grader List Change

Grader edit has hidden active/inactive toggle:
Grader Edit Change

Instructor list now has active filter:
Instructor List Change

Instructor edit has hidden active/inactive toggle:
Instructor Edit Change

New display hidden graders checkbox and inactive grader flag (for both group and criteria sub tabs)
Grader Checkbox Change

## Testing

Manual testing via the UI was done to ensure all features worked properly. In addition rspec tests were added to ensure the modified controller and model methods were working as intended.

## Checklist

https://github.com/MarkUsProject/Wiki/pull/150


anoverga17:
Failing tests are unrelated to PR

app/models/end_user.rb
```diff
@@ -5,7 +5,7 @@ class EndUser < User
   ).map(&:to_sym).freeze
 
   def visible_courses
-    self.courses.where.not('roles.type': 'Student')
-        .or(self.courses.where('courses.is_hidden': false, 'roles.hidden': false))
+    self.courses.where('roles.hidden': false)
```
mishaschwartz, line 8: This is good but we need to go further. If the course is hidden, a student should not be able to access it at all even through the url directly.

Probably the best way to do this is to add another pre_check function in application_policy.rb that does:

```
    deny! if role.student? and role.course.is_hidden
``` 

And then skip that pre check everywhere we're currently skipping the `role_exists?` pre check. 

I guess you could also just add that line to the `role_exists?` precheck function and that would work but then maybe I'd change the name of the function from `role_exists?` to something that better describes the updated function

app/lib/repository.rb
```diff
@@ -273,15 +274,15 @@ def self.get_all_permissions
         assignment.valid_groupings.each do |valid_grouping|
```
mishaschwartz, line 274: We should not allow students who are in a hidden course to access repositories. Probably the best way to do this is to add an additional where statement in the `get_repo_auth_records` function to not consider assignments that belong to a hidden course.

db/migrate/20220815210513_add_hidden_check_to_repo_permissions_function.rb
```diff
@@ -0,0 +1,131 @@
...
+        RETURN EXISTS(
+                SELECT roles.id
+                FROM memberships
+                    JOIN roles ON roles.id=memberships.role_id
+                    JOIN groupings ON memberships.grouping_id=groupings.id
+                    JOIN groups ON groupings.group_id=groups.id
+                    JOIN assignment_properties ON assignment_properties.assessment_id=groupings.assessment_id
+                    JOIN assessments ON groupings.assessment_id=assessments.id
+                    LEFT OUTER JOIN assessment_section_properties ON assessment_section_properties.assessment_id=assessments.id
+                WHERE memberships.type='StudentMembership'
```
mishaschwartz, line 49: We should not allow students who are in a hidden course to access repositories. The best way to do this here is to join assessments table to the courses table and then add another where condition that checks if the course is hidden

mishaschwartz (changes requested):
Other things to do:

- In the students controller: check the policy before updating the student's hidden attribute
- In the instructors and tas controller: allow the hidden attribute to be set when _creating_ a new role (not just when updating)
- when the hidden value on a course gets changed, we need to update the access file as well (students should not be able to access repos in a hidden course). Do this by calling `Repository.get_class.update_permissions` in an after_update callback in the courses controller.
- add tests for:
    - checking that the repository access file is updated at the right times
    -  changes in the repository.rb file (see repository_spec.rb)

app/controllers/instructors_controller.rb
```diff
@@ -10,11 +10,17 @@ def index
...
       format.html
       format.json do
-        data = current_course.instructors
-                             .joins(:user)
-                             .where(type: Instructor.name)
-                             .pluck_to_hash(:id, :user_name, :first_name, :last_name, :email)
-        render json: data
+        instructors = current_course.instructors
+                                    .joins(:user)
+                                    .where(type: Instructor.name)
```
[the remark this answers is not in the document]
anoverga17: This is primarily so that admin roles (which are technically considered 'instructors') do not show up in the instructors table

app/policies/api/main_api_policy.rb
```diff
@@ -4,6 +4,7 @@ class MainApiPolicy < ApplicationPolicy
     default_rule :manage?
 
     skip_pre_check :role_exists?
+    skip_pre_check :view_hidden_course?
```
anoverga17, line 7: I know you mentioned skipping the new policy wherever we skip the `role_exists?` but in this case, this might lead to students being able to access a hidden course through the api. I would like your opinion if that is alright or if the precheck should not be skipped in this instance or if it should be skipped elsewhere.

mishaschwartz: added a safe navigation operator to the `:view_hidden_course?` pre-check so that it will only deny access if the role exists and is a student and the course is hidden

mishaschwartz (approved):

david-yz-liu (approved):

PR #6188, opened 2022-08-24
bugfix: marks spreadsheet grade update api

pretendWhale:
## Motivation and Context

This PR fixes a bug with the `update_grades` API route - this route attempted to join `grade_entry_students` directly to `users` rather than joining through `roles`. It also adds some basic tests for the route

## Your Changes

**Description**:
Fixed bug with incorrect associations on `update_grades` api route.

## Testing

Tested with the markus-api package, as well as adding basic tests for the route.

## Checklist

### DIFF
--- a/app/controllers/api/grade_entry_forms_controller.rb
+++ b/app/controllers/api/grade_entry_forms_controller.rb
@@ -144,7 +144,7 @@ module Api
       grade_entry_form = record
 
       grade_entry_student = grade_entry_form.grade_entry_students
-                                            .joins(role: :user)
+                                            .joins(:user)
                                             .where('users.user_name': params[:user_name])
                                             .first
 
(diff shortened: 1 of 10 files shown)


app/controllers/api/grade_entry_forms_controller.rb
```diff
@@ -144,7 +144,7 @@ def update_grades
       grade_entry_form = record
 
       grade_entry_student = grade_entry_form.grade_entry_students
-                                            .joins(:user)
+                                            .joins(role: :user)
```
david-yz-liu, line 147: I don't have a strong opinion, but to be honest I think it would be more robust to add a `through` association to `GradeEntryStudent` itself. Otherwise we may keep running into this issue for other joins in the future.

pretendWhale: You're right, it would be - I added the association and updated the places where this pattern was being used for `grade_entry_students`. The same pattern is being used in `memberships`, and could be changed the same way. I will make a separate PR to update that as well.

david-yz-liu (approved):

PR #6195, opened 2022-08-31
manage max_file_size through the UI

mishaschwartz:
## Motivation and Context

Previously the max_file_size setting allowed for a course was set in the settings yaml files and was updated only when the server restarted. This meant that updating this setting required interruptions for everyone using the instance. Also, this setting was being written to a file that was symlinked in each repo, this caused issues if the repos were mounted on a different file system where the symlink could potentially break.

This PR allows this setting to be stored on the courses table and managed through the UI. Admin users can update this value and instructor users can view it on the courses edit page. The setting is communicated to the git hooks by reading the .max_file_size file that now exists in the repository (similar to how we manage the .required_files.json file to check for required files)

## Your Changes

**Description**:

## Checklist


lib/repo/git_hooks/multihook.py
```diff
@@ -11,7 +11,6 @@
     with open(os.path.join(real_dir, 'max_file_size')) as f:
         max_file_size = f.read().strip()
     hooks_dir = os.path.join(real_dir, '{}.d'.format(hook_type))
-    env = {**os.environ, 'MAX_FILE_SIZE': max_file_size}
```
david-yz-liu, line 14: The `env` variable is used on line 19, so I'm not sure how this works?

For `subprocess.run` do we need to explicitly pass in the current environment variables if we want them to be used, or are they automatically passed if we don't specify an `env`?

lib/repo/git_hooks/client/pre-commit.d/04-file_size_too_large.py
```diff
@@ -16,7 +19,7 @@
         file_size_proc = subprocess.run(['git', 'cat-file', '-s', file_sha],
                                    stdout=subprocess.PIPE, universal_newlines=True)
         file_size = file_size_proc.stdout.strip()
-        if int(file_size) > MAX_FILE_SIZE:
```
david-yz-liu, line 19: You can delete the `MAX_FILE_SIZE` constant as well (line 4), similar to the server-side hook.

config/locales/models/courses/en.yml
```diff
@@ -5,6 +5,7 @@ en:
       course:
         display_name: Display Name
         is_hidden: Course Visibility
+        max_file_size: Maximum File Size (bytes)
```
david-yz-liu, line 8: It would be good to clarify that this is for submissions, e.g. "Maximum Submission File Size". Or shorter, "Maximum Submission Size". Since there are other types of files that can be uploaded for a course.

david-yz-liu (commented):
@mishaschwartz great work, thanks! I left a couple of comments (I know most of them are about the git hooks, which you are rewriting in a separate PR, but I still think it's a good idea to make sure the changes here to the existing git hooks are stable.)

Also, I noticed that you left the `Course#max_file_size_settings` method, which I'm pretty sure is no longer used anywhere. Was this intentional, or perhaps just something you missed?

config/locales/models/courses/en.yml
mishaschwartz: That is true, but we are also limiting the size of those files. Both autotest files and starter files cannot be larger than this value.

lib/repo/git_hooks/multihook.py
mishaschwartz: Sorry about this. I was working on the changes for this PR and the one for the hook rewrite PR at the same time so I dropped the ball on updating the hooks for this PR as needed. 

(if env=None then it inherits the parent process' environment)

david-yz-liu (approved):

PR #6196, opened 2022-08-31
Make git hooks more portable

mishaschwartz:
## Motivation and Context

Git hooks were written in python3 which required that python3 be installed on every system that could potentially run the server hooks (not just MarkUs but also a server allowing access to the git repos over ssh/https) as well as any machine running the client hooks. This seemed like an unnecessary requirement. So, this PR writes the hooks in a (hopefully) posix compliant way using shell scripts. 

This PR also:

- writes the required files as plain text instead of as a json string (easier to parse with shell scripts)
- copies each hook to each repo (instead of symlinking them)
- signals to server hooks not to fire when MarkUs is making changes using a custom environment variable (previously we were using an environment variable that was set by the git_http_backend process but not set by the git-shell used when accessing repos over ssh)
- updates client hooks so that they will still work as intended when a commit is made from a subdirectory of the repo
- write tests for git hooks

Closes #6277

## Your Changes

**Description**:

## Checklist


lib/repo/git_hooks/client/pre-commit.d/04-file_size_too_large.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env sh
+
+CHANGED_FILES=$(git diff --name-status --no-renames --diff-filter=d HEAD) # diff does not include deleted files
+MAX_FILE_SIZE=$(git show HEAD:.max_file_size)
+
+# shellcheck disable=SC2034
+echo "$CHANGED_FILES" | while read -r _status path; do
```
david-yz-liu, line 7: I tried this out when deleting a single file:

```console
$ git rm A1/hello.py
$ git commit -m "tmp"
fatal: Not a valid object name d4375a28ae7ddb7c18c90c3cd94389381c0b2ffc
e69de29bb2d1d6434b8b29ae775ad8c2e48c5391
... omitted ...
f30ebc7e481b7fe310982e2e69b25d3f55a35acd
.git/hooks/pre-commit.d/04-file_size_too_large.sh: line 10: [: : integer expression expected
```

I think the problem is that the loop here is still executing even when `$CHANGED_FILES` is "empty"? (Didn't look into this in detail, just flagging)

lib/repo/git_hooks/client/pre-commit.d/04-file_size_too_large.sh
```diff
@@ -0,0 +1,14 @@
...
+  sha=$(git ls-files --stage "$(git rev-parse --show-toplevel)/$path" | cut -d' ' -f 2)
+  size=$(git cat-file -s "$sha")
+  if [ "$size" -gt "$MAX_FILE_SIZE" ]; then
+    echo "[MARKUS] Error: The size of the uploaded file $path exceeds the maximum of $MAX_FILE_SIZE bytes."
```
david-yz-liu, line 11: These aren't changed\s you introduced, but if you don't mind making these small changes:

1. Change `[MARKUS]` to `[MarkUs]` to be consistent
2. Change "uploaded" to "modified" (since "upload" has more of a "push" connotation)

lib/repo/git_hooks/server/post-receive.d/01-show_push_time.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env sh
+
+# Do not show time if no change have been made to the master branch
+cat - | cut -d' ' -f 3 | grep -qw 'refs/heads/master' || exit 0
+
+echo "[MARKUS] Your submission has been received: $(git reflog -n 1 --date=local --format=format:%gd master)"
```
david-yz-liu, line 6: Similar to above, please modify `[MARKUS]` to `[MarkUs]` (and in the other hooks below)

lib/repo/git_hooks/client/pre-commit.d/04-file_size_too_large.sh
david-yz-liu: Update: it seems this also happens when attempting an empty commit (i.e., no files changed). I assume for the same reason, though if there is an empty commit I think we can safely skip all of the checks?

lib/repo/git_hooks/client/pre-commit.d/02-block_change_top_level.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env sh
+
+echo '[MarkUs] Checking whether top-level files/directories were created/modified...'
+
+CHANGED_FILES=$(git diff-index --name-only HEAD)
+
+if printf '%s' "$CHANGED_FILES" | grep -qv -e '/' -e '.gitignore'; then
+  echo "[MarkUs] Error: Top-level change detected for the following entries:"
+  echo "$CHANGED_FILES" | grep -v -e '/' -e '.gitignore'
+  echo "[MarkUs] You should unstage this change."
```
david-yz-liu, line 10: Okay this is a bit picky, but I realized that this hook checks for changes to top-level files regardless of whether they're staged or not.

While my preference would be to modify the hook to only check for staged files, I believe this was the previous behaviour (since you're using the same `git diff-index` as the Python version). So I'd also be okay keeping the behaviour but rewording this specific error message to say `"You should undo this change."` rather than `"You should unstage this change."`

david-yz-liu (commented):
Hey @mishaschwartz, this is great work and sorry it took me so long to review it. I've been testing it out and noted a couple of corner cases where the hook behaviour seems to have changed in the rewrite. Can you please take a look and make some updates and add tests; I'm also happy to touch base to discuss any of these tomorrow (Tuesday).

david-yz-liu:
@mishaschwartz sorry I forgot to mention, I did successfully test this out on Windows' Git Bash, so no compatibility issues there 🎉

david-yz-liu (commented):
@mishaschwartz all of the changes look great. As I was testing I did notice two more things, hopefully not too bad to address:

1. I don't think the *server-side* hook for checking top-level changes is working as intended? I disabled the local hooks and didn't seem to have a problem pushing a commit adding a new top-level file or modifying `.required`, for example.
2. One recent update I forgot to take into account: allowing spaces in required file names. 🙄 I believe this means the current use of `cut` isn't quite right. I think you can fix this by assuming the last word is `true`/`false` and everything else is the filename, or (more annoying) change the format of `.required` altogether.

david-yz-liu (approved):

PR #6209, opened 2022-09-01
Fix policies for admins

mishaschwartz:
## Motivation and Context

This PR fixes some issues related to policies for admin users:

- The `current_role` and `real_role` functions should return nil if there is no `current_course`. For admins, they returned a non-persisted AdminRole object instead because the `find_or_create_by` method failed to save a new record with an empty course.
- the pre-check in the main_admin_policy was not adding any additional benefit over the default rule and it was also causing policies in the application_policy to always pass. This meant, for example, that the role_is_switched? policy was always true for admins (which meant that they could not view their settings page for example)
- admins could view the courses/index page but could not view any courses because the adminuser model does not have a `visible_courses` method. Since, admins have a different view of the courses (admin/courses/index) the courses/index route should be disabled for admins.
- admin users could not manage subdirectories in repositories when viewing courses since they are not EndUser users. By updating the policy, now all users can manage subdirectories (as intended initially)
- similarly, admin users could not reset their api keys because they are not EndUser users.

## Your Changes

**Description**:

## Checklist


anoverga17 (approved):
Looks good! Thanks for cleaning some of this up.

david-yz-liu (approved):

PR #6224, opened 2022-09-12
filter non-existent parameters out of properties.yml

pretendWhale:
## Motivation and Context

This PR filters any parameters in the `properties.yml` file used in assignment upload/download to remove parameters that aren't present on the relevant models. This should improve compatibility of assignments between MarkUs versions. Specifically,  #5834 removed a column from the `assignment_properties` table, which makes assignment configurations downloaded from versions <2.1.0 incompatible with versions>=2.1 

## Your Changes

**Description**:
Adds a `filter_assignment_properties` method in the assignments controller which adds filters for assignment and related model attributes

## Testing

Tested via the web interface, as well as by adding non-valid attributes to the `properties.yml` fixture file and ensured all tests continue to pass.

## Checklist


mishaschwartz (changes requested):
Just adding a reminder here to add tests and try with the recursive filter function

Changelog.md
```diff
@@ -35,6 +36,7 @@
 - Fix bug where empty groups are hidden by default (#6222)
 - Fix bug where test run table did not re-render when switching results (#6223)
 - Allow admin users to manage the maximum file size setting through the UI (#6195)
+>>>>>>> master
```
mishaschwartz, line 39: remove this

Changelog.md
```diff
@@ -21,6 +21,7 @@
 - Fix bug where gzipped binary feedback files were not unzipped correctly (#6283)
 - Fix bug where remark request summary table wasn't being rendered correctly (#6284)
 - Fix bug where test results were being associated with the wrong test runs (#6287)
+- Filter out invalid properties on assignment upload (#6224)
```
mishaschwartz, line 24: Changelog entries should really describe what the change means for the user of MarkUs (whether that is an end-user, admin, or developer). This describes what has been changed but doesn't really describe how that changes things for a user.

In the case of bugfixes, I like to describe the bug that was fixed in the changelog as opposed to the mechanism used to fix it. Similarly for features, I like to describe the feature that was added instead of how it was done in the code. 

In this case, can we change this to something like "Fix bug where assignments could not be uploaded if unexpected properties were present in the uploaded file content". That's a bit wordy maybe, but something like that

mishaschwartz (changes requested):

david-yz-liu (approved):

PR #6238, opened 2022-09-19
Fix bug where groups are not retrieved in defined order

mishaschwartz:
## Motivation and Context

Groups are not retrieved in a deterministic order by an active record query which results in autotest results being assigned to the wrong test runs. 

(see comment for more details)

## Your Changes

**Description**:

## Checklist


david-yz-liu (commented):
@mishaschwartz yikes, great catch.

I think it would be worth adding a warning comment to the docs for `group_info_for_tests` to say that the group info array isn't guaranteed to be in the same order as the passed `group_ids`. That's pretty non-intuitive.

david-yz-liu (approved):

PR #6244, opened 2022-09-27
Release by url token

### LINKED ISSUE #6242: Allow results to be released through generated URLs
We have a use case for releasing grades to students through generated "magic URLs" rather than the web interface. I'll call these *release URLs* below. (This specific request is for allowed students to view final exam results, without releasing the results in bulk or having to manually release results for each individual request.)

The specs here are:

- When an instructor is setting up an assignment, provide a new configuration option for enabling the generation of release URLs.
- When an instructor is viewing submissions, provide a new option to "generate release URLs" in addition to the existing ("release marks").
- That option should export a CSV file containing the same student information as the other student export data tables (e.g. assignment marks summary), and an additional column with a link that the student could use to view their results.

    *Note*: the typical workflow would be for the instructor to send that CSV to a central office that manages student requests to view that assessment. When the student makes the request, that office can then give the student that URL for viewing, without the office staff requiring direct access to MarkUs.
- Each link should have the same permission setup as the normal `view_marks` route, except it doesn't check whether the result has been released or not. So if a student gets their link, they can view their result at any time.
- Each link shouldn't be brute-forceable (or computable given any information about the course/assessment/student). (e.g., using a random hash function or something)
- The URLs should be created with an expiry time relative to when the URL is created; each assessment should have a configuration option setting the URL expiration time.

Optional "nice to have" features/things to discuss:

- *Possible feature*: Allow instructors to view/manage these URLs in the web interface
- *Possible feature*: Allow instructors to manually expire URLs (e.g., by selecting specific rows in the submissions table)
- *Consideration*: If an instructor generates release URLs twice for the same submission, what is the expected behaviour? Possibilities:
    - Old link expired, new link generated
    - Old link returned, same expiry date
    - Old link returned, new expiry date (i.e., the configured "time to expiry" + the time when the link was re-generated)
- *Consideration*: How does this interact with results being "current" or not? If an instructor recollects a submission, does a generated release URL for an older result remain or become expired?

mishaschwartz:
## Motivation and Context

Resolves #6242

- adds an option for assignment set-up to "Only allow students to view released marks via a unique URL"
- on the submissions page, select some groups and click the "Show Release Tokens" button to open a modal
- this modal lets users update/download (as a csv) release tokens/urls for each selected group
- this modal also lets the instructor set an (optional) expiry date for the token
- tokens are generated using Rails' `has_secure_token` helper
- students can access the results if this assignment option is set to true only if: 
    - results are released
    - the expiry date of their token is nil or hasn't passed yet
    - they enter the correct token as part of the url param or they enter the correct token in the modal when navigating to the view_results page
    - note that the correct token is cached in the session so a student only has to enter it once per session
- note that the token is associated with a result so recollecting a submission will create a new result and any tokens will have to be re-distributed to the student.

## Your Changes

**Description**:

## Checklist

### DIFF
--- a/app/assets/javascripts/Components/Modals/submit_view_token_modal.jsx
+++ b/app/assets/javascripts/Components/Modals/submit_view_token_modal.jsx
@@ -1,0 +1,60 @@
+import React from "react";
+import {render} from "react-dom";
+import Modal from "react-modal";
+
+class SubmitViewTokenModal extends React.Component {
+  constructor() {
+    super();
+    this.state = {
+      isOpen: false,
+      token: null,
+    };
+  }
+
+  componentDidMount() {
+    Modal.setAppElement("body");
+  }
+
+  onSubmit = () => {
+    $.ajax({
+      url: Routes.view_token_check_course_result_path(this.props.course_id, this.props.result_id, {
+        view_token: this.state.token,
+      }),
+    }).then(
+      () => {
+        window.location = Routes.view_marks_course_result_path(
+          this.props.course_id,
+          this.props.result_id,
+          {view_token: this.state.token}
+        );
+      },
+      () => this.setState({isOpen: false, token: null})
+    );
+  };
+
+  render() {
+    return (
+      <Modal
+        className="react-modal"
+        isOpen={this.state.isOpen}
+        onRequestClose={() => this.setState({isOpen: false, token: null})}
+      >
+        <p>{I18n.t("results.view_token_submit")}</p>
+        <form onSubmit={this.onSubmit}>
+          <div className={"modal-container-vertical"}>
+            <div className={"modal-container"}>
+              <input onChange={e => this.setState({token: e.target.value})} />
+            </div>
+            <div className={"modal-container"}>
+              <input type="submit" value={I18n.t("save")} />
+            </div>
+          </div>
+        </form>
+      </Modal>
+    );
+  }
+}
+
+export function makeSubmitViewTokenModal(elem, props) {
+  return render(<SubmitViewTokenModal {...props} />, elem);
+}
--- a/app/assets/javascripts/Components/Modals/submit_view_token_modal.jsx
+++ b/app/assets/javascripts/Components/Modals/submit_view_token_modal.jsx
@@ -46,7 +46,7 @@ class SubmitViewTokenModal extends React.Component {
               <input onChange={e => this.setState({token: e.target.value})} />
             </div>
             <div className={"modal-container"}>
-              <input type="submit" value={I18n.t("save")} />
+              <input type="submit" value={I18n.t("results.submit_token")} />
             </div>
           </div>
         </form>
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -728,21 +728,28 @@ class ResultsController < ApplicationController
     data = requested_results.left_outer_joins(grouping: [:group, { accepted_student_memberships: [role: :user] }])
                             .pluck('groups.group_name',
                                    'users.user_name',
+                                   'users.first_name',
+                                   'users.last_name',
+                                   'users.email',
+                                   'users.id_number',
                                    'results.view_token',
                                    'results.view_token_expiry',
                                    'results.id')
-                            .group_by(&:first)
     header = [[I18n.t('activerecord.models.group.one'),
+               I18n.t('activerecord.attributes.user.user_name'),
+               I18n.t('activerecord.attributes.user.first_name'),
+               I18n.t('activerecord.attributes.user.last_name'),
+               I18n.t('activerecord.attributes.user.email'),
+               I18n.t('activerecord.attributes.user.id_number'),
                I18n.t('submissions.release_token'),
-               I18n.t('submissions.release_token_url'),
                I18n.t('submissions.release_token_expires'),
-               I18n.t('activerecord.attributes.group.student_memberships')]]
+               I18n.t('submissions.release_token_url')]]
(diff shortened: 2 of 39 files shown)


app/policies/result_policy.rb
```diff
@@ -52,4 +67,9 @@ def download?
       )
     )
   end
+
+  # This is a separate policy function because it reports a specific error message on failure
+  def view_with_result_token?
+    check?(:member?, record.submission.grouping) && record.view_token == view_token && !record.view_token_expired?
```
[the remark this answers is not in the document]
david-yz-liu: Okay I looked into this a bit more. The `has_secure_token :view_token` sets the token on create, so new results have one set, but old results don't. So when I was testing this PR on an existing assignment, I see a bunch of `nil`s.

It's not a deal-breaker for this semester, but I think this could be a good time to try out a slightly more sophisticated migration that sets `view_token`s for all existing results (and then puts a non-null restriction on the column). We haven't done migrations like that before but I don't think it would be too bad, and that way you get to try that out and we get an extra constraint we want in the database.

On the other hand, I'm not sure if you were originally planning for cases where `view_token` would be `nil`? (E.g., only generate tokens when results are released.) Happy to talk more!

app/views/layouts/menus/_student_sub_menu.html.erb
```diff
@@ -53,13 +53,28 @@
...
+          <!-- check if a token not required or if the current user has entered the token already this session -->
+          <li class='<%= "active" if active %>'>
+            <%= link_to Result.model_name.human.pluralize,
+                        view_marks_course_result_path(@current_course, @result) %>
+          </li>
+        <% elsif allowed_to? :view_token_check?, @result %>
+          <li class='<%= "active" if active %>'>
+            <%= link_to Result.model_name.human.pluralize, '#', id: "submit-view-token" %>
+          </li>
+          <div id="submit-view-token-modal"></div>
```
david-yz-liu, line 69: Putting the `div` here breaks the horizontal flow of the `li`s across the page (specifically the Peer Reviews tab below gets shifted to a new line). You can move the div elsewhere but keep the javascript here (or move both).

david-yz-liu (commented):
@mishaschwartz fantastic work! I left a couple of comments, and also have two general notes:

1. Releasing marks currently triggers a notification email to students (if they enabled the setting). We should disable that when `release_with_urls` is true, and make a note of this in the Wiki to warn instructors.
2. The release url allows students to view results for hidden assignments, which is good. (I imagine this will be the most common use case for instructors w/ final exams.) However, when the student goes to see the result, the sub-menu also shows tabs for "Assignments" and "Submissions". The former results in a 403 page, but the latter takes the student to their submissions page for the assignment. So the two issues here are:

    a. When the assignment is hidden please hide all of the other tabs in the sub-menu besides "Results".
    b. Prevent students from accessing the submissions page for a hidden assignment.

app/controllers/results_controller.rb
```diff
@@ -708,6 +710,52 @@ def get_test_runs_instructors_released
...
+  # Update the view token expiry date for the results whose ids are given
+  def update_view_token_expiry
+    expiry = params[:expiry_datetime]
+    updated = requested_results.map { |r| r.update(view_token_expiry: expiry) ? r.id : nil }.compact
+    render json: Result.where(id: updated).pluck(:id, :view_token_expiry).to_h
+  end
+
+  # Download a csv containing view token and grouping information for the results whose ids are given
+  def download_view_tokens
+    data = requested_results.left_outer_joins(grouping: [:group, { accepted_student_memberships: [role: :user] }])
```
david-yz-liu, line 728: Please sort by group name (matching the default sort order in the React table)

app/assets/javascripts/Components/Modals/submit_view_token_modal.jsx
```diff
@@ -0,0 +1,60 @@
...
+        onRequestClose={() => this.setState({isOpen: false, token: null})}
+      >
+        <p>{I18n.t("results.view_token_submit")}</p>
+        <form onSubmit={this.onSubmit}>
+          <div className={"modal-container-vertical"}>
+            <div className={"modal-container"}>
+              <input onChange={e => this.setState({token: e.target.value})} />
+            </div>
+            <div className={"modal-container"}>
+              <input type="submit" value={I18n.t("save")} />
```
david-yz-liu, line 49: Please introduce a new string label for this button. Something like "Continue" or "Submit" or "Submit Token".

config/locales/views/submissions/en.yml
```diff
@@ -44,7 +48,22 @@ en:
...
+      Any old tokens or URLs will be automatically expired and you will have to provide the new tokens or URLs to the students.
+
+      Would you like to continue?
     release_marks: Release Marks
+    release_token: Release Token
+    release_token_expires: Expiry Date
+    release_token_expires_null: Never
+    release_token_url: Release URL
+    release_url_with_token: |-
+      Students can access released marks only at the following url:
```
david-yz-liu, line 64: Capitalize "URL" to be consistent with the other text

david-yz-liu (commented):
@mishaschwartz great, thank you! I've spent a bunch of time testing this out and things look good. I just left a few minor inline comments, and one other thing below.

**NOTE**: The `marks.rake` seed file needs to be updated to provide a `Result.generate_unique_secure_token` for the seed result objects, which are passed to `Result.upsert_all`.

Seems like something interesting happened with `upsert`. I'll follow up with you about this offline.

david-yz-liu (approved):

PR #6285, opened 2022-10-17
Fix bug where tests groups can be deleted without deleting associated test group results

mishaschwartz:
## Motivation and Context

When test specs are updated for the autotester, previous test groups are deleted but the associated test group results are not because of a delete_all call which doesn't trigger callbacks. 

This causes additional problems when test groups are deleted but the test scripts themselves fail to update. For example, if a test group is deleted and there is a schema error, any subsequent tests that are run will have results associated with a non-existent test group. The test results will fail to save but the test run will have completed so there will be an empty test run with no results (in the instructor/TA view) and an error raised (in the student view) because of an error (comparing a nil value to a timestamp when finding the latest test results to display).

In order to fix this, I moved the code that updates the test groups into a transaction with the call to update the scripts on the autotester. If one fails, now the other is not committed/run. I also changed the delete_all call to a destroy_all call but since it is now running in a background job it shouldn't cause a problematic delay.

This means that if the test settings are changed after a test is run, previous test group results associated with now deleted test groups will be deleted as well. I believe this was the original intention of the code. Please note:

- tests results will no longer be displayed to instructors or students (since they have been deleted)
- marks that were created by running a test whose result has now been deleted will *not* be removed.
- this occurs whether the results have already been released to the student or not

Because of this... Although this probably was how the code was supposed to be, the current behaviour is not ideal. We should either:

- not delete test group results when the associated test group is deleted (not ideal because we don't keep a record of old test groups so old test group results look out of date and may cause confusion if they don't match the test settings)
- not delete old test groups when updating the test settings and instead mark them as active/inactive and show test results that are associated with an inactive test group differently in the test runs table.
- warn before deleting test groups if tests have already been run and don't allow deleting test groups if marks have been released (like we do with criteria)

DO THIS ONE:

- [x] warn before updating settings if test runs exist and then keep the old test groups. Don't differentiate between active/inactive test groups unless necessary. 

## Your Changes

**Description**:

See description

## Checklist


app/models/grouping.rb
```diff
@@ -59,6 +59,7 @@ class Grouping < ApplicationRecord
            -> {
              left_outer_joins(role: :user,
                               test_group_results: [:test_group, :test_results])
+               .where.not('test_groups.id': nil)
```
david-yz-liu, line 62: @mishaschwartz sorry for the delay in looking at this one. All of the changes look good, except this one here I think is from an earlier iteration of your changes. I believe we decided to ensure `TestGroupResult`s always had an associated test group (to avoid the "test group id is nil")?

If so, I don't think this `where` is necessary, and moreover we can change `TestGroupResult` association `belongs_to :test_group, optional: true` to remove the `optional: true`, and ideally make a database migration to make the column non-null.

But perhaps I'm not remembering correctly, or missing a use case for keeping this column as nullable?

david-yz-liu (commented):
@mishaschwartz I realized after merging from master that the new `AutotestResetUrlJob` also runs the `AutotestSpecsJob`, and so needs to be updated.

david-yz-liu (approved):

mishaschwartz:
@david-yz-liu thanks for catching that. I've added a fix in https://github.com/MarkUsProject/Markus/pull/6337

PR #6290, opened 2022-10-19
Don't cache total marks and total grades

mishaschwartz:
## Motivation and Context

Right now we are maintaining a cache of the total mark (sum of the marks/grades + bonus marks) on the Result and GradeEntryStudent objects. This was intended to make queries a bit faster when showing the totals but the outcome is that the cached value can get out of sync with the actual total if the cached value isn't updated properly. 

We were largely doing this correctly but every once in a while the update to the cached value would be missed. Instead of hunting down these bugs whenever they pop up, this PR simplifies the code by removing the cached value altogether and calculating the total marks/grades from the existing marks/grades when needed. 

This may be marginally slower in some cases because it increases the number of additional queries made to the database. But, the number of additional queries (relative to the number of queries in the old code) is constant so there should not be any noticeable slowdown for the end user. The trade off is that the code is more maintainable and simpler now.

## Your Changes

**Description**:

## Checklist


app/controllers/results_controller.rb
```diff
@@ -633,7 +633,6 @@ def add_extra_mark
     @extra_mark = @result.extra_marks.build(extra_mark_params.merge(unit: ExtraMark::POINTS))
     if @extra_mark.save
       # need to re-calculate total mark
```
david-yz-liu, line 635: Delete this comment as well

app/helpers/criteria_helper.rb
```diff
@@ -34,19 +34,18 @@ def upload_criteria_from_yaml(assignment, data)
...
       raise "#{I18n.t('criteria.errors.invalid_format')} #{crit_format_errors.join(', ')}"
     end
-    reset_results_total_mark(assignment.id)
+    reset_marking_states(assignment.id)
     successes
   end
 
   private
 
   # Resets the total mark for all results for the given assignment with id +assessment_id+.
```
david-yz-liu, line 43: Please update this comment

app/models/grade_entry_form.rb
```diff
@@ -40,48 +40,31 @@ def max_mark
...
-                                  .where(roles: { hidden: false })
-                                  .where.not(total_grade: nil)
-                                  .order(:total_grade)
-                                  .pluck(:total_grade)
+    ges_ids = self.grade_entry_students
+                  .joins(:role)
+                  .where(roles: { hidden: false })
+                  .pluck('grade_entry_students.id')
+
+    GradeEntryStudent.get_total_grades(ges_ids).values.compact.sort
```
david-yz-liu, line 48: I think here you want to do `@completed_result_marks = ...`?

db/structure.sql
```diff
@@ -2142,7 +2140,8 @@ CREATE TABLE public.test_groups (
     updated_at timestamp without time zone NOT NULL,
     assessment_id bigint NOT NULL,
     autotest_settings json DEFAULT '{}'::json NOT NULL,
-    "position" integer NOT NULL
+    "position" integer NOT NULL,
+    tester_index integer
```
david-yz-liu, line 2144: @mishaschwartz I must have missed this, but... I'm not sure what's going on with this addition? Is `master` out of date, or is this a leftover from other changes you were working on?

david-yz-liu (approved):

PR #6293, opened 2022-10-21
Lti controller abstraction

pretendWhale:
## Motivation and Context

Note: depends on #6156 . To make future support of other LMS's easier, takes all canvas-specific functionality into a canvas controller.

## Your Changes

**Description**:
Creates `canvas_controller` subclassing `lti_deployment_controller` for canvas-specific functionality, and re-arranges tests to support that. Canvas specific routes are nested under lti_deployment routes

- [x] Refactoring
- [x] Other (please specify): Modifications required for LTI support

## Testing

Tested canvas LTI workflow in the browser, as well as making controller tests into a shared example

## Checklist


mishaschwartz (changes requested):

spec/controllers/lti_deployment_controller_spec.rb
```diff
@@ -1,136 +1,7 @@
...
-        expect(session[:lti_user_id]).to eq(1)
-      end
-      it 'successfully creates a new lti object' do
-        post :redirect_login, params: { state: session.id.to_s, id_token: lti_jwt }
-        expect(LtiDeployment.count).to eq(1)
-      end
-    end
-  end
+
   describe '#choose_course', :choose_course do
```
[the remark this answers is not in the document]
pretendWhale: It can be used for [metadata](https://relishapp.com/rspec/rspec-core/docs/metadata/user-defined-metadata) - but in this case it was an artifact of trying to get a test to work before realizing I had misconfigured routes and doesn't do anything, so I've removed it.

app/controllers/lti_deployment_controller.rb
```diff
@@ -3,52 +3,13 @@ class LtiDeploymentController < ApplicationController
...
-        user_id: '$Canvas.user.id',
-        course_id: '$Canvas.course.id',
-        course_name: '$Canvas.course.name'
-      }
-    }
-
-    render json: config.to_json
-  end
+  # Default Canvas values to be overridden in each LMS controller
+  LMS_REDIRECT_ENDPOINT = '/api/lti/authorize_redirect'.freeze
```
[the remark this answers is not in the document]
mishaschwartz: Are you sure? If you can keep these as constants that would make more sense. This works for me... 

```ruby
class C
  def a
    self.class::A
  end
end

class B < C
  A = 10
end

B.new.a # => 10

```

spec/controllers/lti_deployment_controller_spec.rb
```diff
@@ -1,136 +1,7 @@
 describe LtiDeploymentController do
```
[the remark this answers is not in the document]
mishaschwartz: I think some of these are failing now. Also, please add some tests that do more than just check the response status. Do some checks that the behaviour of the actual method works as intended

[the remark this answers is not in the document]
pretendWhale: I added a couple of tests that check the `public_jwk` method returns a valid jwk and check that users are updated in `create_lti_grades`, but for the latter, it really just runs `lti_deployment.get_students` and `lti_deployment.create_grades`, which have tests in the model

spec/controllers/lti_deployment_controller_spec.rb
```diff
@@ -187,4 +61,105 @@
...
+      get :public_jwk
+      is_expected.to respond_with(:success)
+    end
+    it 'responds with success when logged in' do
+      get_as instructor, :public_jwk
+      is_expected.to respond_with(:success)
+    end
+    it 'responds with a valid jwk' do
+      pub_jwk = get :public_jwk
+      expect { JWT::JWK.new(pub_jwk.body) }.not_to raise_error
```
mishaschwartz, line 75: `JWT::JWK.new` will not raise an error as long as its argument is any string. The only thing this test is checking is that the response body is a string. 

Please rewrite this to check that the json response is well formed, the key is in the expected place, that the key is signed with the right secret... I'm sure there are other things we can be checking too

spec/controllers/lti_deployment_controller_spec.rb
```diff
@@ -187,4 +61,105 @@
...
+    end
+  end
+  describe '#create_lti_grades' do
+    let!(:course) { create :course }
+    let(:instructor) { create :instructor, course: course }
+    let!(:lti) { create :lti_deployment, course: course }
+    let(:assignment) { create :assignment_with_criteria_and_results }
+    let(:scope) { 'https://purl.imsglobal.org/spec/lti-ags/scope/lineitem' }
```
mishaschwartz, line 83: use the constant you declared in lti_deployment.rb

config/routes.rb
```diff
@@ -493,15 +493,23 @@
...
-      get 'choose_course'
-      post 'choose_course'
-      post 'create_course'
-      post 'create_lti_grades'
+    resources :lti_deployment, only: [] do
+      collection do
+        get 'public_jwk'
+        get 'choose_course'
+        post 'choose_course'
+        post 'create_lti_grades'
```
mishaschwartz, line 501: why is this not an assignment member route again?

mishaschwartz (changes requested):

spec/controllers/lti_deployment_controller_spec.rb
pretendWhale: I couldn't think of too many additional checks here beyond what you mentioned, but I added a couple of checks for making sure a `kid` param is in the json, that the algorithm is present and correct, and checked to make sure the keys verify a token signed with the key's secret.

config/routes.rb
pretendWhale: We had talked about having lti sync available for grade_entry_forms as well, and to my mind it made sense to have it be an `lti_deployment` collection route based on that. We could definitely use two different member routes though, one on assignments and one on gefs

mishaschwartz: The only reason why I mention it is because right now you're passing in the assignment_id as an additional parameter to this route. If it was an assignment route then we get a lot of the checks that the current user is permitted (or not) to access the route for free.

I would vote to make an assignment and grade entry form route and just have them both call the same helper function.

If you want, let's discuss and we can implement it in a different PR or not depending on what we decide.

spec/support/lti_controller_examples.rb
```diff
@@ -0,0 +1,136 @@
+shared_examples 'lti deployment controller' do
+  let(:instructor) { create :instructor }
+  let!(:client_id) { 'LMS defined ID' }
+  let(:target_link_uri) { 'https://example.com/authorize_redirect' }
+  let(:host) { 'https://canvas.instructure.com' }
```
mishaschwartz, line 5: if this is a shared example for all lti controller subclasses, do not use canvas specific examples

spec/support/lti_controller_examples.rb
```diff
@@ -0,0 +1,136 @@
...
+  let(:state) { 'state_param' }
+  describe '#launch' do
+    context 'when launching with invalid parameters' do
+      let(:lti_message_hint) { 'opaque string' }
+      let(:login_hint) { 'another opque string' }
+      let(:auth_url) { 'http://canvas.instructure.com:433/api/lti/authorize_redirect' }
```
mishaschwartz, line 11: if this is a shared example for all lti controller subclasses, do not use canvas specific examples

spec/support/lti_controller_examples.rb
```diff
@@ -0,0 +1,136 @@
...
+        is_expected.to respond_with(:unprocessable_entity)
+      end
+      it 'responds with unprocessable_entity if login_hint is not passed' do
+        post :launch,
+             params: { lti_message_hint: lti_message_hint, client_id: client_id, target_link_uri: target_link_uri }
+        is_expected.to respond_with(:unprocessable_entity)
+      end
+      context 'when all required params exist' do
+        before :each do
+          stub_request(:post, 'http://canvas.instructure.com:443/api/lti/authorize_redirect')
```
mishaschwartz, line 36: if this is a shared example for all lti controller subclasses, do not use canvas specific examples

spec/support/lti_controller_examples.rb
```diff
@@ -0,0 +1,136 @@
...
+        post :redirect_login, params: {}
+        is_expected.to render_template('shared/http_status')
+      end
+      it 'redirects to an error page with a mismatched state' do
+        post :redirect_login, params: { state: state, id_token: 'token' }
+        is_expected.to render_template('shared/http_status')
+      end
+    end
+    context 'with correct parameters' do
+      let(:jwk_url) { 'https://canvas.instructure.com:443/api/lti/security/jwks' }
```
mishaschwartz, line 79: if this is a shared example for all lti controller subclasses, do not use canvas specific examples

spec/support/lti_controller_examples.rb
```diff
@@ -0,0 +1,136 @@
...
+      let(:payload) do
+        { aud: client_id,
+          iss: 'https://canvas.instructure.com',
```
mishaschwartz, line 82: if this is a shared example for all lti controller subclasses, do not use canvas specific examples

spec/support/lti_controller_examples.rb
```diff
@@ -0,0 +1,136 @@
...
+          'https://purl.imsglobal.org/spec/lti/claim/deployment_id': 'some_deployment_id',
```
mishaschwartz, line 83: use your constants

spec/support/lti_controller_examples.rb
```diff
@@ -0,0 +1,136 @@
...
+          'https://purl.imsglobal.org/spec/lti/claim/context': {
```
mishaschwartz, line 84: use your constants

spec/support/lti_controller_examples.rb
```diff
@@ -0,0 +1,136 @@
...
+            label: 'csc108',
+            title: 'test'
+          },
+          'https://purl.imsglobal.org/spec/lti/claim/custom': {
```
mishaschwartz, line 88: use your constants

spec/controllers/lti_deployment_controller_spec.rb
```diff
@@ -187,4 +61,133 @@
...
+          headers: {
+            Accept: '*/*'
+          }
+        ).to_return(status: :success, body: { access_token: 'access_token',
+                                              scope: scope,
+                                              token_type: 'Bearer',
+                                              expires_in: 3600 }.to_json)
+      stub_request(:get, lti_service_namesrole.url).with(headers: { Authorization: 'Bearer access_token' },
+                                                         query: {
+                                                           role: 'http://purl.imsglobal.org/vocab/lis/v2/membership#Learner'
```
mishaschwartz, line 136: use your constants

spec/controllers/lti_deployment_controller_spec.rb
```diff
@@ -187,4 +61,133 @@
...
+                                                                 picture: 'http://example.com/picture.png',
+                                                                 given_name: Student.first.first_name,
+                                                                 family_name: Student.first.last_name,
+                                                                 lis_person_sourcedid: Student.first.user_name,
+                                                                 email: Student.first.email,
+                                                                 user_id: 'lti_user_id',
+                                                                 lti11_legacy_user_id: 'legacy_lti_user_id',
+                                                                 roles:
+                                                                   [
+                                                                     'http://purl.imsglobal.org/vocab/lis/v2/membership#Learner'
```
mishaschwartz, line 152: use your constants

app/models/lti_deployment.rb
```diff
@@ -4,11 +4,19 @@ class LtiDeployment < ApplicationRecord
...
+                 deployment_id: 'https://purl.imsglobal.org/spec/lti/claim/deployment_id' }.freeze
 
   # Gets a list of all users in the LMS course associated with this deployment
   # with the learner role and creates roles and LTI IDs for each user.
   def get_students
-    auth_data = lti_client.get_oauth_token(['https://purl.imsglobal.org/spec/lti-nrps/scope/contextmembership.readonly'])
+    auth_data = lti_client.get_oauth_token([LTI_SCOPES[:names_role]])
     names_service = self.lti_services.find_by!(service_type: 'namesrole')
     membership_uri = URI(names_service.url)
     membership_uri.query = URI.encode_www_form(role: 'http://purl.imsglobal.org/vocab/lis/v2/membership#Learner')
```
mishaschwartz, line 22: purl.imsglobal.org

mishaschwartz (changes requested):
All urls should be stored as constants. Is there a reason why there's one specific url that is not a constant?

pretendWhale:
> All urls should be stored as constants. Is there a reason why there's one specific url that is not a constant?
Fair enough, I've moved it to a new `LTI_ROLES` constant. Right now it's the only 'role' url we use/care about, so I wasn't sure if it made sense to make it a constant compared to the claims and scopes, where we need several. But I've just gone ahead and added a constant with that role as the only element of the lti_roles hash, and we can add more as we find we need them

mishaschwartz (approved):

app/assets/javascripts/Components/Modals/send_lti_grades_modal.jsx
```diff
@@ -39,7 +39,7 @@ class LtiGradeModal extends React.Component {
     );
     const deploymentsToUpdate = checked.map(deployment => deployment[0]);
     $.post({
-      url: Routes.lti_deployment_create_lti_grades_path(),
+      url: Routes.create_lti_grades_lti_deployment_index_path(),
```
david-yz-liu, line 42: This looked quite strange to me, since "collection" routes normally do not need an `index` suffix.

I think you aren't quite following the Rails naming conventions for resourceful routing:

1. In the routes.rb file, the "resources" should be pluralized (`lti_deployments`)
2. The controller filename and class name should be pluralized (`lti_deployments_controller.rb`, `LtiDeploymentsController`

I think after you make this change you should be able to use the path helper `create_lti_grades_lti_deployments_path`, which follows a more standard naming style (as with other uses of "`_index_path`" below).

app/controllers/lti_deployment_controller.rb
```diff
@@ -90,12 +46,11 @@ def redirect_login
       render 'shared/http_status', locals: { code: '422', message: I18n.t('lti.config_error') }, layout: false
       return
     end
-
     referrer_uri = URI(request.referer)
     # Get canvas JWK set
```
david-yz-liu, line 50: There are a few different comments in this file that still refer to "canvas", and should now refer to "LMS" (or whatever term the LTI standard uses, like "platform" or something).

config/routes.rb
```diff
@@ -493,15 +493,23 @@
...
-      get 'public_jwk'
-      post 'redirect_login'
-      get 'choose_course'
-      post 'choose_course'
-      post 'create_course'
-      post 'create_lti_grades'
+    resources :lti_deployment, only: [] do
+      collection do
+        get 'public_jwk'
+        get 'choose_course'
```
david-yz-liu, line 499: I'm wondering why `choose_course` is a collection route instead of a member route. The session carries the `lti_deployment_id` at this point; is there a reason that's better than keeping the `lti_deployment_id` in the URL?

pretendWhale: You're right - at the point we get to `choose_course` there's no reason not to put the `lti_deployment_id` in the url

PR #6306, opened 2022-10-28
Allow spaces in file names

### LINKED ISSUE #6249: Allow uploaded file names to include spaces
Right now, we replace spaces in uploaded files with underscores. We should allow files with spaces to be stored without a name change.

mishaschwartz:
## Motivation and Context

Resolves #6249

## Your Changes

**Description**:

## Checklist


david-yz-liu (commented):
@mishaschwartz this is good, but please also update the assignment form validation for required file inputs. Currently in `app/views/assignments_boot.js.erb`, `check_required_fields` function. I'm pretty sure you can get rid of some of that JS code and instead use the [pattern](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text#specifying_a_pattern) HTML attribute directly to get the browser to do the validation.

I tried thinking of other form inputs where the user specifies arbitrary file names. We could add this validation to autotest feedback files, but that form is validated differently so I don't think we absolutely need to. If you think of anywhere else, please update in this PR too.

mishaschwartz:
pre-commit.ci run

mishaschwartz:
pre-commit.ci run

mishaschwartz:
pre-commit.ci run

david-yz-liu (approved):

PR #6327, opened 2022-11-04
Hide net-http warnings

mishaschwartz:
## Motivation and Context

Every time rails loads we get warnings about constants already being loaded. This is caused by the error reported here:

https://github.com/ruby/net-imap/issues/16

To avoid these warnings the suggestion is to explicitly require the 'net-http' gem in our Gemfile. This appears to work.

## Your Changes

**Description**:

- [x] Other (please specify): remove warnings

I don't love the idea of having to add a new dependency just to get rid of warnings but it looks like this won't be fixed in ruby 2.7 so I've added a note to remove it once we upgrade to ruby 3.

## Checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,11 @@ gem 'rails', '~> 7.0.4'
 gem 'sprockets'
 gem 'sprockets-rails'
 
+# Adding this removes some warnings caused by double-loading of the net-protocol library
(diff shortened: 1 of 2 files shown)


mishaschwartz:
vote to close this, we can handle a few warning until we update to ruby 3 and cluttering up the gemfile is not good

Issue #6254, opened 2022-09-29
Add additional bulk collection options


mishaschwartz:
When collecting bulk submissions, we should provide the following options:

- collect all selected (already implemented)
- collect all selected that does not have a previously collected submission (already implemented)
- collect all selected that does not have a previously collected submission OR has a newer revision
    - also provide the option to select whether penalties apply (if the new revision collected is during the penalty period)
    - also provide the option to select whether to collect a revision selected after the collection date (see #6252 and #6253)

pretendWhale:
Closed via #6341

Issue #6253, opened 2022-09-29
Allow bulk collection of most recent submission


mishaschwartz:
When bulk collecting submissions, MarkUs collects the last revision from before the collection time. We should also provide the option to bulk collect the most recent revision (regardless of the collection time). 

When using this option, the instructor should also be able to choose whether or not to apply penalties (if any)

This will likely interact with the feature suggested in #6252

david-yz-liu:
Closed in #6341

PR #6341, opened 2022-11-15
allow collection before due date

### LINKED ISSUE #6252: Allow submissions to be collected before the collection time
Right now, bulk collection of submissions can only happen once the final collection date has passed. We should add an option to bulk collect submissions before the collection date. 

When collecting before the collection date, the most recent revision should be collected.

pretendWhale:
## Motivation and Context

Allow instructors to bulk collect submissions via an additional option on the collection modal, as well as choosing whether to apply late penalties. Resolves #6252, #6253, and #6254 

## Your Changes

**Description**:
Add additional `collect_current` parameter to the collect_submissions route, and a checkbox to set the parameter to true, as well as explicitly sending `apply_late_penalty` on bulk collection and adding a checkbox to control its value.

collect submissions modal

## Testing

Added additional tests for when this option is present, as well as testing the functionality works via web interface.

With this change, I don't believe that there are any cases where `apply_late_penalty` is not explicitly passed to a `SubmissionJob`. Should we modify the conditional in SubmissionJob to only check `if options[:apply_late_penalty]` rather than also checking `if options[:apply_late_penalty].nil?`

## Checklist

https://github.com/MarkUsProject/Wiki/pull/168


config/locales/views/submissions/en.yml
```diff
@@ -9,6 +9,7 @@ en:
       Would you like to continue?
     collect:
       apply_late_penalty: Apply late penalty/Deduct grace credits (Leaving this box unchecked means the submission will not be penalized, regardless of submission time.)
+      collect_current: 'Collect most recent submissions? <strong>Warning: This will collect the most recent submission, regardless of due date.</strong>'
```
mishaschwartz, line 12: The convention is to add the `_html` suffix to keys that contain html. See the `override_existing_html` key below for an example. This should be renamed to `collect_current_html` (see here for more info: https://guides.rubyonrails.org/i18n.html#using-safe-html-translations)

app/assets/javascripts/Components/Modals/collect_submissions_modal.jsx
```diff
@@ -59,6 +73,35 @@ class CollectSubmissionsModal extends React.Component {
...
+                />
+              </label>
+            </p>
+            <p>
+              <label>
+                <input
+                  type="checkbox"
+                  defaultChecked={this.state.apply_late_penalty}
+                  name="apply_late_penalty"
+                  id="apply_late_penalty"
```
mishaschwartz, line 97: Is there a reason why we need an id here but not for the other checkboxes? Was there some custom css you wanted to apply to this one?

spec/controllers/submissions_controller_spec.rb
```diff
@@ -958,6 +962,23 @@
             expect(response).to render_template(partial: 'shared/_poll_job')
           end
 
+          it 'should succeed if it is before the section due date and collect_current is selected' do
+            @assessment_section_properties.update!(due_date: 1.week.from_now)
+            allow(Assignment).to receive_message_chain(
+              :includes, :find
+            ) { @assignment }
+            expect_any_instance_of(SubmissionsController).not_to receive(:flash_now).with(:error, anything)
```
mishaschwartz, line 970: can we split this test up into two. I'd prefer to have a single expect statement per `it` block whenever possible

spec/controllers/submissions_controller_spec.rb
```diff
@@ -998,6 +1019,23 @@
             expect(response).to render_template(partial: 'shared/_poll_job')
           end
 
+          it 'should succeed if it is before the global due date but collect_current is true' do
+            @assignment.update!(due_date: 1.week.from_now)
+            allow(Assignment).to receive_message_chain(
+              :includes, :find
+            ) { @assignment }
+            expect_any_instance_of(SubmissionsController).not_to receive(:flash_now).with(:error, anything)
```
mishaschwartz, line 1027: same here, can we split this up into two tests and test for the presence of an error message in the flash hash instead?

spec/controllers/submissions_controller_spec.rb
```diff
@@ -958,6 +962,23 @@
...
+            allow(SubmissionsJob).to receive(:perform_later) { Struct.new(:job_id).new('1') }
+
+            post_as @instructor,
+                    :collect_submissions,
+                    params: { course_id: course.id, assignment_id: @assignment.id,
+                              override: true, collect_current: true,
+                              groupings: ([] << @assignment.groupings).flatten }
```
mishaschwartz, line 977: this syntax is weird. I know they use it in other tests but is there a reason why `@assignment.groupings.to_a` doesn't work?

spec/controllers/submissions_controller_spec.rb
mishaschwartz, line 970: This expect statement is checking whether flash_now is called. It's more robust to test if anything is added to the flash hash at any point. Can we move this test to after the `post_as` call and instead change the test to something like:

```ruby
expect(flash[:error]).to be_nil
```

There are examples of this in a bunch of controller specs, if you search for `expect(flash[:error])` in the source code it should show a bunch of examples

mishaschwartz (changes requested):
This looks good. Just a few comments and changes. We also need to add tests for your changes to the SubmissionsJob. Like I mentioned in our meeting, you can probably just reuse the tests when the assignment is a scanned exam.

app/assets/javascripts/Components/Modals/collect_submissions_modal.jsx
pretendWhale: I was at one point intending to use css to hide this checkbox unless the `collect_now` checkbox was checked, but I realized that didn't really make sense so I removed it and this got left in accidentally. I've removed it since it isn't actually doing anything.

mishaschwartz (approved):
Looks good. Please mark that you've updated the changelog in the PR description and this change will need an update to the Wiki as well.

The only other thing is that there is a lot of duplicate code in the tests now that should really be moved into a `before` block. I don't think it's worth it to make that change in this PR necessarily though. What do you think?

pretendWhale:
> Looks good. Please mark that you've updated the changelog in the PR description and this change will need an update to the Wiki as well.

Done, and I put up a PR on the wiki repo - thanks for pointing that out.

> The only other thing is that there is a lot of duplicate code in the tests now that should really be moved into a `before` block. I don't think it's worth it to make that change in this PR necessarily though. What do you think?

I agree - it seems fine for that to be a separate test update PR. That said, also fine with adding it here.

david-yz-liu (commented):
@pretendWhale sorry for the delay in looking at this one! The overall functionality is good, I just spent a bunch of time thinking through the UI.

Since the UI is a bit finicky, I made a small PR here: https://github.com/pretendWhale/Markus/pull/78. Can you please take a look and if you're okay with those changes merge that in?

Also I was thinking about the interaction between this and the scanned exams, which right now only support collection based on the current time (because this is the standard workflow). I think that's fine, but in this case we should update the UI so that only that option is shown in the collect submissions modal.

david-yz-liu (approved):

PR #6348, opened 2022-11-24
Make puma server workers and threads configurable

mishaschwartz:
## Motivation and Context

These are settings that should be configured by a sysadmin installing MarkUs. They can be configured to get optimum performance from the app depending on the server resources. For example, with lots of additional memory available, they may want to increase the number of workers, with lots of cpu capacity available they may want to increase the number of threads.

Note that the current defaults in settings.yml are the same defaults that we are using at the moment.

This also makes it easy to configure the number of puma workers to zero in development. This temporarily solves the issue that redis-py will not resolve docker container hostnames but only in a forked puma worker (running on docker engine in ubuntu 20.04). Until that issue has been properly debugged and resolved, this provides developers a temporary workaround (ie. they can run zero forked puma workers in development)

## Your Changes

**Description**:

- [x] Other (please specify): configuration options update

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -26,6 +26,7 @@
 - Downloading files from a single repository no longer adds an additional subdirectory to the zip archive (#6323)
 - Allow admins to role switch to instructors (#6353)
 - Fix bug where rubric grades could not be selected with the return key (#6354)
+- Allow admins to set the number of puma workers and threads from the settings files (#6348)
 
 ## [v2.1.6]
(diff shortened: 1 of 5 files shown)


david-yz-liu (commented):
@mishaschwartz this looks good. TO be complete, can you please also make the "worker timeout" configurable (Line 6-8 in `puma.rb`).

david-yz-liu (approved):

PR #6349, opened 2022-11-24
Make uid of docker user configurable

mishaschwartz:
## Motivation and Context

Docker engine handles bind mounts differently from docker desktop. Docker engine keeps all permission, including ownership, the same when mounting a file or directory from the host machine to a container. This means that the owner of the file/directory on the host machine must have the same uid as the user on the container if we want to emulate the container user being the owner of these files.

This should have no effect on developers using docker desktop since docker desktop handles bind mounts differently.

## Your Changes

**Description**:

- [x] Other (please specify):

## Checklist


mishaschwartz:
add to .dockerignore file as well

david-yz-liu (approved):

PR #6350, opened 2022-11-25
move lti sync to background job

pretendWhale:
## Motivation and Context

This change moves LTI functionality to a background job, as well as ensuring that policies already in place are followed.

## Your Changes

**Description**:
Move LTI roster and grade sync to new `lti_sync_job`, and move gradebook item creation to `lti_lineitem` job. 

- [x] Other (please specify): LTI update

## Testing

Tested the jobs via the web interface, as well as adding tests for new jobs and helper methods in `lti_helper`

## Checklist


app/controllers/lti_deployments_controller.rb
```diff
@@ -14,8 +14,9 @@ def launch
       return
     end
     nonce = rand(10 ** 30).to_s.rjust(30, '0')
-    session[:client_id] = params[:client_id]
-    session[:nonce] = nonce # Nonce should be present in the JWT sent to MarkUs, and must be verified
+    cookies.encrypted[:client_id] = { value: params[:client_id], expires: 1.hour.from_now }
```
mishaschwartz, line 17: We talked about putting all info into a single cookie. Is that something that we still want to do?

app/helpers/lti_helper.rb
```diff
@@ -0,0 +1,115 @@
+require 'mono_logger'
```
mishaschwartz, line 1: Where is this being used? Typically you don't need to explicitly require libraries under /app

app/helpers/lti_helper.rb
```diff
@@ -0,0 +1,115 @@
...
+  # Synchronize LMS user with MarkUs users.
+  # if role is not nil, attempt to create users
+  # based on the values of can_create_users and
+  # can_create_roles.
+  def roster_sync(lti_deployment, course, can_create_users: false, can_create_roles: false)
+    errors = []
+    auth_data = lti_deployment.lti_client.get_oauth_token([LtiDeployment::LTI_SCOPES[:names_role]])
+    names_service = lti_deployment.lti_services.find_by!(service_type: 'namesrole')
+    membership_uri = URI(names_service.url)
+    membership_uri.query = URI.encode_www_form(role: 'http://purl.imsglobal.org/vocab/lis/v2/membership#Learner')
```
mishaschwartz, line 12: can this URL be stored as a constant somewhere (like the others)

app/helpers/lti_helper.rb
```diff
@@ -0,0 +1,115 @@
...
+          email: user['email'],
+          lti_user_id: user['user_id'] }
+      end
+    end
+    if user_data.empty?
+      raise 'No user data returned from lms'
+    end
+    user_data.each do |lms_user|
+      markus_user = EndUser.find_by(user_name: lms_user[:user_name])
+      if markus_user.nil? && can_create_users # allowed_to?(:manage?, current_user, with: Admin::UserPolicy)
```
mishaschwartz, line 31: clean up commented out code

app/helpers/lti_helper.rb
```diff
@@ -0,0 +1,115 @@
...
+        { user_name: user['lis_person_sourcedid'].nil? ? user['name'] : user['lis_person_sourcedid'],
+          first_name: user['given_name'],
+          last_name: user['family_name'],
+          display_name: user['name'],
+          email: user['email'],
+          lti_user_id: user['user_id'] }
+      end
+    end
+    if user_data.empty?
+      raise 'No user data returned from lms'
```
mishaschwartz, line 27: If this is a message that will be visible to the end users, make it an I18n string

app/helpers/lti_helper.rb
```diff
@@ -0,0 +1,115 @@
...
+    end
+    user_data.each do |lms_user|
+      markus_user = EndUser.find_by(user_name: lms_user[:user_name])
+      if markus_user.nil? && can_create_users # allowed_to?(:manage?, current_user, with: Admin::UserPolicy)
+        markus_user = EndUser.create!(lms_user.except(:lti_user_id))
```
mishaschwartz, line 32: This will cause everything to fail if a single user cannot be created. Do you want that or do you want to create as many as possible and report those that failed?

If you want to do all or nothing then wrap this in a database transaction so that if you create 5/10 users successfully and then fail on the 6th you will roll back the other ones you've created.

app/helpers/lti_helper.rb
```diff
@@ -0,0 +1,115 @@
...
+      elsif markus_user.nil? && !can_create_users
+        errors.append("MarkUs user #{lms_user[:user_name]} not found")
```
mishaschwartz, line 34: I18n string

app/helpers/lti_helper.rb
```diff
@@ -0,0 +1,115 @@
...
+        next
+      end
+      course_role = Student.find_by(user: markus_user, course: course)
+      if course_role.nil? && can_create_roles # allowed_to?(:manage?, role, with: RolePolicy)
```
mishaschwartz, line 38: clean up commented code

app/helpers/lti_helper.rb
```diff
@@ -0,0 +1,115 @@
...
+        payload = {
+          timestamp: Time.current.iso8601,
+          scoreGiven: mark,
+          scoreMaximum: assessment.max_mark.to_f,
+          activityProgress: 'Completed',
+          gradingProgress: 'FullyGraded',
+          userId: lti_user_id
+        }
+        req.set_form_data(payload)
+        lti_deployment.send_lti_request!(req, score_uri, auth_data, scopes)
```
mishaschwartz, line 79: If this fails do you want to keep trying for the rest of the marks and report the failure or stop completely

app/helpers/lti_helper.rb
```diff
@@ -0,0 +1,115 @@
...
+    end
+  end
+
+  # Returns a hash mapping lti_user_id to marks
+  # for each released mark where the user has an lti_user_id
+  def get_assignment_marks(lti_deployment, assignment)
+    marks = assignment.released_marks
+    mark_data = {}
+    lti_users = LtiUser.where(lti_client: lti_deployment.lti_client)
+    marks.each do |mark|
```
mishaschwartz, line 90: I feel like this could be done in a single database query....

app/helpers/lti_helper.rb
```diff
@@ -0,0 +1,115 @@
...
+      result = mark.results.first
```
mishaschwartz, line 91: I'm pretty sure Mark objects have a single `result` association. Not multiple `results`

app/helpers/lti_helper.rb
mishaschwartz: Or at least make sure of the Result.total_marks method with a list of result ids that you can get in a single query

app/helpers/lti_helper.rb
```diff
@@ -0,0 +1,115 @@
...
+    mark_data
+  end
+
+  # Returns a hash mapping lti_user_id to marks
+  # for each released mark where the user has an lti_user_id
+  def get_grade_entry_form_marks(lti_deployment, grade_entry_form)
+    marks = grade_entry_form.released_marks
+    mark_data = {}
+    lti_users = LtiUser.where(lti_client: lti_deployment.lti_client)
+    marks.each do |mark|
```
mishaschwartz, line 107: Same here, I think we can do this in a query or two without the loop

app/jobs/lti_sync_job.rb
```diff
@@ -0,0 +1,20 @@
...
+    I18n.t('lti.start_grade_sync')
+  end
+
+  def self.completed_message(_status)
+    I18n.t('lti.grade_sync_complete')
+  end
+
+  def perform(lti_deployments, assessment, course, can_create_users: false, can_create_roles: false)
+    lti_deployments.each do |deployment|
+      roster_sync(deployment, course, can_create_users: can_create_users, can_create_roles: can_create_roles)
```
mishaschwartz, line 16: This method returns a list of error messages. What's the plan for those?

app/jobs/lti_sync_job.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def self.show_status(_status)
+    I18n.t('lti.start_grade_sync')
+  end
+
+  def self.completed_message(_status)
+    I18n.t('lti.grade_sync_complete')
+  end
+
+  def perform(lti_deployments, assessment, course, can_create_users: false, can_create_roles: false)
```
mishaschwartz, line 14: One thing that's nice to have is a progress message. You could have something that reports X/Y grades sent since you're doing them one by one.

mishaschwartz (dismissed):

app/jobs/lti_sync_job.rb
pretendWhale: Ended up making a single error message

app/helpers/lti_helper.rb
pretendWhale: I think we want to stop because the most likely error is probably a rate limit error, in which case subsequent request will likely also fail

app/helpers/lti_helper.rb
```diff
@@ -0,0 +1,151 @@
...
+        req.set_form_data(payload)
+        lti_deployment.send_lti_request!(req, score_uri, auth_data, scopes)
+      end
+    end
+  end
+
+  # Returns a hash mapping lti_user_id to marks
+  # for each released mark where the user has an lti_user_id
+  def get_assignment_marks(lti_deployment, assignment)
+    released_marks = assignment.released_marks
```
david-yz-liu, line 90: This variable name is misleading, the result that's stored isn't mark data, it's result data.
In addition to renaming it, I recommend making this a hash mapping lti user id to result id.

Also (about the code below this line):

1. Whenever dealing with marks, we need to include only *accepted* memberships, never pending ones. `Grouping` has a `:accepted_student_memberships` association that you can join on directly.
2. You can put all the joins on one line. Syntax like `.joins(x: { y: { z: ... } } }` should work fine.
3. Rather than iterate over `grades`, I recommend iterating over the return hash of this query. Since `grades` is itself a hash, you won't need to call `.find`, but instead just use hash lookup to get the associated total grade.

app/helpers/lti_helper.rb
```diff
@@ -0,0 +1,151 @@
...
+      [student[0], grade[1]]
+    end.to_h
+  end
+
+  # Returns a hash mapping lti_user_id to marks
+  # for each released mark where the user has an lti_user_id
+  def get_grade_entry_form_marks(lti_deployment, grade_entry_form)
+    student_data = grade_entry_form.released_marks
+                                   .joins(role: :user)
+                                   .joins(user: :lti_users)
```
david-yz-liu, line 109: As above, this should be merged into the previous line

app/helpers/lti_helper.rb
```diff
@@ -0,0 +1,151 @@
...
+                                   .where('lti_users.lti_client': lti_deployment.lti_client,
+                                          'grade_entry_students.released_to_student': true)
```
david-yz-liu, line 111: This shouldn't be necessary, the `released_marks` association should generate this condition already (but if it doesn't, that's another problem to address).

app/helpers/lti_helper.rb
david-yz-liu: (Same general comments apply to `get_grade_entry_form_marks`

app/views/assignments/create_lti_line_items.erb
```diff
@@ -0,0 +1,36 @@
...
+          <td>
+            <%= f.label deployment.id, I18n.t("lti.lti_deployment", {
+              lti_deployment_name: deployment.lms_course_name,
+              lti_host: URI(deployment.lti_client.host).host,
+            })%>
+          <td>
+            <% if deployment.lti_line_items.where(assessment: @assignment).exists? %>
+              <%= f.check_box deployment.id, checked: true, disabled: true, class: 'lti-checkbox' %>
+            <% else %>
+              <%= f.button deployment.id, checked: false, class: 'lti-checkbox' %>
```
david-yz-liu, line 23: I'm confused, shouldn't this be `f.check_box`? What does the UI look for you right now?

app/views/assignments/create_lti_line_items.erb
```diff
@@ -0,0 +1,36 @@
+<div class="table lti-table">
+  <table>
+    <thead>
+    <tr>
+      <th><%= t("lti.lti_settings") %></th>
+      <th><%= t("lti.create_line_item_checkbox")%></th>
```
david-yz-liu, line 6: In terms of style, each `%>` should have exactly one preceding space. In addition to this line, there are a couple in the file below that have 2 spaces.

app/views/assignments/create_lti_line_items.erb
```diff
@@ -0,0 +1,36 @@
...
+              lti_deployment_name: deployment.lms_course_name,
+              lti_host: URI(deployment.lti_client.host).host,
+            })%>
+          <td>
+            <% if deployment.lti_line_items.where(assessment: @assignment).exists? %>
+              <%= f.check_box deployment.id, checked: true, disabled: true, class: 'lti-checkbox' %>
+            <% else %>
+              <%= f.button deployment.id, checked: false, class: 'lti-checkbox' %>
+            <% end  %>
+          </td>
```
david-yz-liu, line 25: This tag seems to be misplaced

app/views/assignments/create_lti_line_items.erb
```diff
@@ -0,0 +1,36 @@
...
+          </td>
+        </tr>
+      <% end %>
+      </tbody>
```
david-yz-liu, line 29: The indentation starting from here onwards is off

app/views/assignments/create_lti_line_items.erb
```diff
@@ -0,0 +1,36 @@
...
+      </table>
+      </div>
+      <div>
+        <%= f.button t('lti.create_new_line_items'), class: 'large-button',
```
david-yz-liu, line 33: More standard is `f.submit`

app/views/layouts/_sub_sub_menu.html.erb
```diff
@@ -66,6 +66,12 @@
               </li>
             <% end %>
           <% end %>
+          <% if @lti_deployments %>
+          <li class='<%= "active" if controller.action_name == 'create_lti_line_items' %>'>
```
david-yz-liu, line 70: This part should be indented.

app/views/assignments/create_lti_line_items.erb
```diff
@@ -0,0 +1,36 @@
+<div class="table lti-table">
```
david-yz-liu, line 1: This file is missing the `.html` part of the extension (should be `.html.erb`)

app/views/layouts/_sub_sub_menu.html.erb
```diff
@@ -66,6 +66,12 @@
...
+            <%= link_to I18n.t('lti.lti_settings'),
+                        create_lti_line_items_course_assignment_path(@current_course, @assignment) %>
```
david-yz-liu, line 72: Okay so overall I like the idea of creating a new tab. However, for this tab please create a new assignment route called `lti_settings` to render the page. Right now that page can have a title and just include the (renamed) `_create_lti_line_items.html.erb` as a partial, but we'll likely expand on this later.

Then, you can ensure the `create_lti_line_items` route is just a POST, not a GET.

app/controllers/assignments_controller.rb
```diff
@@ -622,6 +622,30 @@ def upload_config_files
...
+
+  def create_lti_line_items
+    @assignment = record
+    @lti_deployments = LtiDeployment.where(course: @current_course)
+    if request.post? && params.key?(:lti_deployment)
+      items_to_create = params[:lti_deployment].select { |_key, val| val == '1' }.keys.map(&:to_i)
+      unless items_to_create.empty?
+        @current_job = LtiLineItemJob.perform_later(items_to_create, @assignment)
+        session[:job_id] = @current_job.job_id
+        render 'shared/_poll_job'
```
david-yz-liu, line 643: You need to `return` here as well, to prevent a double render.

app/controllers/assignments_controller.rb
```diff
@@ -622,6 +622,30 @@ def upload_config_files
...
+      end
+    end
+    render 'lti_deployments/create_lti_line_items'
```
david-yz-liu, line 646: This is referring to an old view.

david-yz-liu (commented):
@pretendWhale good work. I left a couple of comments for you to work on. I'll touch base with you tomorrow about the status of the PR as well!

app/views/assignments/lti_settings.html.erb
```diff
@@ -0,0 +1,4 @@
+<div class='no-items-error'>
```
david-yz-liu, line 1: In the controller, use `render layout: 'assignment_content'` to use that layout. That should have the `shared/flash_message` partial already included, so you won't need to include it here.

But instead, please supply a title for the view (you can see how other views do it, involves a `content_for` or something).

lib/tasks/lti_key.rake
```diff
@@ -3,6 +3,8 @@ namespace :markus do
   task lti_key: :environment do
     print('Creating new private key')
     key = OpenSSL::PKey::RSA.new(2048)
+    FileUtils.mkdir_p(File.join(Settings.file_storage.lti ||
```
david-yz-liu, line 6: It's a bit funny to recompute this when it's (presumably?) used in `LtiClient::KEY_PATH`. A better approach would be to use something like `File::dirname` to retrieve the directory path.

david-yz-liu (commented):
@pretendWhale looks great! I just left a couple more small comments

david-yz-liu (approved):

Issue #6351, opened 2022-11-25
Minor visual bug - popup animation bugging when clicking "Marks Chart"


ggggg:
### Steps to reproduce:
- Have an assignment with 6 questions, with all having full marks
- Go to the "results" tab
- Press "summary"
- Press "Marks Chart"

### Expected behaviour:
It is expected that the animation will open smoothly, but instead, it's moving around.

### Additional info:
This issue can also be observed when hovering over the marks as well.

### Platform:
Screen resolution: tested 1280x720 and 1920x1080
User-agent: "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/107.0.0.0 Safari/537.36"

### Video:

https://user-images.githubusercontent.com/25105662/204062530-dff31a72-11ab-4ce1-909a-d3baabbc9546.mp4

mishaschwartz:
@ggggg Thank you for this PR. I was able to replicate this issue and your fix looks like it works to solve the flickering problem.
However, I think that the problem is actually caused by a different part of the code. If you look at this line:

https://github.com/MarkUsProject/Markus/blob/v2.1.6/app/assets/javascripts/Components/Helpers/data_chart.jsx#L60

There's the bit where it's setting the style of the canvas rendered by the Bar component to `display: "inline-flex"`. I think that this is a holdover from before we were using the react-chartjs-2 library to render charts. If we remove this so that the line is simply:

```jsx
style={{margin: "10px"}}
```

Then the problem is fixed as well. 

I think what is happening is that setting the display to "inline-flex" permits the browser to resize the canvas element. But the react-chartjs-2 library is also dynamically resizing the canvas element and so it flickers back and forth since the browser and the js library have a slightly different idea of what the canvas size should be.

Setting overflow to hidden works as well because I think it prevents the browser from triggering a resize since it just hides the extra content instead. However, I'd rather fix the content at the "source" by not adding the unnecessary `display: "inline-flex"` style in the first place.

Does that make sense to you? Do you mind trying out both solutions and confirming whether or not they work for you as well?

Thanks!

PR #6352, opened 2022-11-25
Fixed popup chart bugging (#6351)

ggggg:
## Motivation and Context

This fixes a small visual bug (see #6351)

## Your Changes

**Description**:
Disabled overflow scrolling for the data-chart-container class as it is not needed and there is no overflow.

## Checklist


mishaschwartz:
@ggggg Thank you for this PR. I was able to replicate this issue and your fix looks like it works to solve the flickering problem.
However, I think that the problem is actually caused by a different part of the code. If you look at this line:

https://github.com/MarkUsProject/Markus/blob/v2.1.6/app/assets/javascripts/Components/Helpers/data_chart.jsx#L60

There's the bit where it's setting the style of the canvas rendered by the Bar component to `display: "inline-flex"`. I think that this is a holdover from before we were using the react-chartjs-2 library to render charts. If we remove this so that the line is simply:

```jsx
style={{margin: "10px"}}
```

Then the problem is fixed as well. 

I think what is happening is that setting the display to "inline-flex" permits the browser to resize the canvas element. But the react-chartjs-2 library is also dynamically resizing the canvas element and so it flickers back and forth since the browser and the js library have a slightly different idea of what the canvas size should be.

Setting overflow to hidden works as well because I think it prevents the browser from triggering a resize since it just hides the extra content instead. However, I'd rather fix the content at the "source" by not adding the unnecessary `display: "inline-flex"` style in the first place.

Does that make sense to you? Do you mind trying out both solutions and confirming whether or not they work for you as well?

Thanks!

mishaschwartz:
@ggggg I'd like to include this in the next release, scheduled for the end of the week. Please let me know if you have any concerns with my suggestions, otherwise I'm going to resolve this in the next few days.

ggggg:
Yeah, I think it works too!

mishaschwartz (approved):

mishaschwartz:
@ggggg would you like to be added to the [contributor's list](https://github.com/MarkUsProject/Markus/blob/master/doc/markus-contributors.txt)? If so, what name should I add?

ggggg:
I would love that, my full name is Ido Ben Haim.
Thank you!

david-yz-liu (approved):

david-yz-liu:
Thanks @ggggg!

PR #6355, opened 2022-11-28
Handle case where no python interpreter is found

mishaschwartz:
## Motivation and Context

The python.rb initializer will fail if no python interpreter is found. Since python is now optional, this should instead print a warning and continue on with the initialization of the app. 

Also, remove the default python location from the settings.yml file since the default should be that it is not set (since it is an optional argument)

## Your Changes

**Description**:

## Checklist

### DIFF
--- a/.dockerfiles/.env
+++ b/.dockerfiles/.env
@@ -1,4 +1,5 @@
 MARKUS__REDIS__URL=redis://redis:6379/
+MARKUS__PYTHON=/app/venv/bin/python
 PGUSER=postgres
 PGHOST=postgres
 PGPORT=5432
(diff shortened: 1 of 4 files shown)


.dockerfiles/.env
```diff
@@ -1,4 +1,5 @@
 MARKUS__REDIS__URL=redis://redis:6379/
+MARKUS__PYTHON=/app/venv/bin/python
```
david-yz-liu, line 2: Maybe I'm missing something here, but is this environment variable being used anywhere?

Perhaps related: I agree it makes sense to remove `python` from the `settings.yml` file, but I want to make sure it's active by default in development (in docker), or at the very least that we update the Wiki to include instructions to add this to a `settings.local.yml` or something.

david-yz-liu (approved):

PR #6385, opened 2022-12-06
in-progress fixes

pretendWhale:
## Motivation and Context

A variety of fixes required for deploying LTI in production

## Your Changes

**Description**:
- Use cookies for lti launch parameters when user is not logged in to markus
- ensure login prior to building any database records
- update tests

- [x] Other (please specify): additional lti fixes

## Testing

Tested in web browser and updated tests to account for changes

## Checklist

### DIFF
--- a/app/controllers/lti_deployments_controller.rb
+++ b/app/controllers/lti_deployments_controller.rb
@@ -8,27 +8,42 @@ class LtiDeploymentsController < ApplicationController
   before_action :check_host, except: [:choose_course]
 
   def launch
-    if params[:client_id].blank? || params[:login_hint].blank? ||
-      params[:target_link_uri].blank? || params[:lti_message_hint].blank?
+    lti_params = params[:lti_message_hint].present? ? params : JSON.parse(cookies.encrypted[:lti_params])
+    if lti_params.is_a?(Hash)
+      lti_params.symbolize_keys!
+    end
+    if lti_params[:client_id].blank? || lti_params[:login_hint].blank? ||
+      lti_params[:target_link_uri].blank? || lti_params[:lti_message_hint].blank?
       head :unprocessable_entity
       return
     end
     nonce = rand(10 ** 30).to_s.rjust(30, '0')
-    session[:client_id] = params[:client_id]
-    session[:nonce] = nonce # Nonce should be present in the JWT sent to MarkUs, and must be verified
+    unless logged_in?
+      cookies.encrypted.permanent[:lti_params] = params.to_json
+      cookies.encrypted.permanent[:lti_redirect] = request.url
+      cookies.encrypted.permanent[:lti_referrer] = request.referer
+      redirect_to root_path
+      return
+    end
+    session[:client_id] = lti_params[:client_id]
+    session[:iss] = lti_params[:iss]
     auth_params = {
       scope: 'openid',
       response_type: 'id_token',
-      client_id: params[:client_id],
-      redirect_uri: params[:target_link_uri],
-      lti_message_hint: params[:lti_message_hint],
-      login_hint: params[:login_hint],
+      client_id: lti_params[:client_id],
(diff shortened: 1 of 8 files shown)


app/controllers/lti_deployments_controller.rb
```diff
@@ -14,8 +14,10 @@ def launch
       return
     end
     nonce = rand(10 ** 30).to_s.rjust(30, '0')
-    session[:client_id] = params[:client_id]
-    session[:nonce] = nonce # Nonce should be present in the JWT sent to MarkUs, and must be verified
+    cookies.permanent.encrypted[:client_id] = { value: params[:client_id], expires: 1.hour.from_now }
```
mishaschwartz, line 17: Can all of these values be stored in a single cookie and the value be a json string containing all the relevant values? That way we only have one additional cookie to keep track of and we lessen the risk of leaving cookies undeleted.

app/controllers/lti_deployments_controller.rb
```diff
@@ -28,75 +30,119 @@ def launch
       prompt: 'none',
       state: session.id # Binds this request to the LTI response
     }
-    referrer = URI(request.referer)
-
-    auth_request_uri = URI("#{referrer.scheme}://#{referrer.host}:#{referrer.port}#{self.class::LMS_REDIRECT_ENDPOINT}")
+    referrer_uri = URI(request.referer)
+    referrer_uri = referrer_uri.to_s.remove(referrer_uri.path) if referrer_uri.path
```
mishaschwartz, line 34: Does remove always remove from the right end of the string? What if there's a query that occurs in the host name. I think it's safer to reconstruct the query from the left side instead. Something like:

```ruby
referer = URI(request.referer)
referer_host = "#{referer.scheme}://#{referer.host}"
referer_host_with_port = "#{referer_host}:#{referer.port}"
referer_host = referer_host_with_port if referer.to_s.start_with?(referer_host_with_port)
auth_request_uri = URI("#{referer_host}#{self.class::LMS_REDIRECT_ENDPOINT}")
```

app/controllers/lti_deployments_controller.rb
```diff
@@ -28,75 +30,119 @@ def launch
...
+          true, # Verify the signature of this token
+          algorithms: ['RS256'],
+          iss: cookies.encrypted[:iss],
+          verify_iss: true,
+          aud: cookies.encrypted[:client_id], # OpenID Connect uses client ID as the aud parameter
+          verify_aud: true,
+          jwks: lms_jwks # The correct JWK will be selected by matching jwk kid param with id_token kid
+        )
+      rescue JWT::DecodeError
+        cookies.delete(:iss)
```
mishaschwartz, line 77: can we put the cookie deletion in to an ensure block instead so that we can be sure they're always deleted whether the rest of the code succeeds or not

app/controllers/lti_deployments_controller.rb
```diff
@@ -28,75 +30,119 @@ def launch
...
+      end
+      if lti_params.key?(LtiDeployment::LTI_CLAIMS[:ags_lineitem])
+        grades_endpoints = lti_params[LtiDeployment::LTI_CLAIMS[:ags_lineitem]]
+        if grades_endpoints.key?('lineitems')
+          lti_data[:line_items] = grades_endpoints['lineitems']
+        end
+      end
+      lti_data[:lti_user_id] = lti_params[LtiDeployment::LTI_CLAIMS[:user_launch_data]]['user_id']
+      unless logged_in?
+        cookies.encrypted.permanent[:lti_data] = { value: JSON.generate(lti_data), expires: 1.hour.from_now }
```
mishaschwartz, line 109: same thing here. Can we store this as one cookie?

mishaschwartz (changes requested):

mishaschwartz (approved):

david-yz-liu (approved):

PR #6519, opened 2023-03-15
bugfix: destroy unused test groups

pretendWhale:
## Motivation and Context

Currently, if there is a test group associated with a criterion, but that test group is not part of an active test, `set_autotest_marks` will set the mark for that criterion to 0, regardless of the current mark.

## Your Changes

**Description**:
This PR changes `update_test_groups_from_specs` such that it will destroy any test groups which are not active tests.

- [ ] Other (please specify):

## Testing

Added a test for unused test groups.

## Checklist


david-yz-liu (commented):
@pretendWhale thanks! I looked into this a bit and dug up the old PR where I believe we decided to keep old test groups: https://github.com/MarkUsProject/Markus/pull/6285.

So, I think the proper fix is not to delete the old test groups (i.e., in `update_test_groups_from_specs`), but instead in the code for processing the test results and updating marks (i.e., somewhere in the background job that processes results).

pretendWhale:
@david-yz-liu Thanks for digging that up! Agreed, I'll close this PR and make a new one with that approach.

PR #6521, opened 2023-03-20
Allow negative bonus marks in Marks Spreadsheets

### LINKED ISSUE #6383: Allow negative bonus marks in Marks Spreadsheets
Assignments currently support positive or negative bonus marks, Marks Spreadsheets (grade entry forms) should also be able support both.

Some things to think about:

- what should the minimum value be? (set by the user in the form? negative value of the total for that column? unlimited?)
- should we extend this to regular (non-bonus) columns as well?

nina-huangg:
This PR allows the existence of negative bonus marks in the Marks Spreadsheets (grade_entry_forms). 

## Motivation and Context

This PR fixes #6383. 

## Your Changes

**Description**:
Two model files were modified:`grade.rb` and `grade_entry_student.rb`. 

In `grade.rb` - two validations were added:
1) validate that the grade is either nil or >= zero if it is a **not** bonus grade
2) validate that the grade is either nil or any value if it **is** a bonus grade

In `grade_entry_student.rb`, the `get_total_grades` method was modified to ensure that the total mark computed is either greater than zero or zero. This ensures that any negative bonus mark does not exceed the max_mark and results in a negative total mark. 

- [X] Refactoring (internal change to codebase, without changing functionality)

## Testing

In `grade_entry_form_spec.rb`, these test cases were added:
1) calculation of total mark/percentage when there is a positive bonus mark.
2) calculation of total mark/percentage when there is a negative bonus mark.
3) calculation of total mark/percentage when the negative bonus mark exceeds the max_mark.
4) calculation of total mark/percentage when all entries are bonus marks.

In `grade_spec.rb`, the tests cases to check the numericality of `grade` have been split up into two cases:
1) when grade is a bonus grade
2) when grade is not a bonus grade 

## Checklist


app/models/grade_entry_student.rb
```diff
@@ -177,7 +177,7 @@ def self.get_total_grades(ges_ids)
     grades.map do |k, v|
       ges_grades = v.map(&:last)
       ges_grades = ges_grades.all?(&:nil?) ? nil : ges_grades.compact.sum
-      [k, ges_grades]
+      [k, ges_grades.nil? ? nil : [ges_grades, 0].max]
```
david-yz-liu, line 180: I'd recommend doing the `max` on the previous line, since that already differentiates between the `nil` and non-`nil` cases

david-yz-liu (commented):
@nina-huangg good work, I left one style comment. Also please remember the feedback I gave you at the meeting (only allow negative grades for bonus mark columns), and make sure to add additional tests that differentiate between those cases.

spec/models/grade_spec.rb
```diff
@@ -1,20 +1,43 @@
 describe Grade do
-  subject do
-    create :grade_entry_form
-    student = create :student
-    create :grade, grade_entry_student: student.grade_entry_students.first
+  context 'when it is not a bonus grade' do
+    subject do
+      create :grade_entry_form
+      student = create :student
+      grade_entry_item = create :grade_entry_item, bonus: false
```
david-yz-liu, line 6: This grade entry item should be associated with the grade entry form created in line 4 (and similarly in the block below)

nina-huangg: The most updated PR now incorporates this change. 

The `grade_entry_item` when created, is associated to the `grade_entry_form` created above. 

The test case in the `course_association_shared_examples.rb` has been modified on line 8, which changes it from `subject.reload` to `subject.grade_entry_student.reload`. This change was necessary as the previous reload method reloads both the course associated with grade_entry_item and grade_entry_student, resulting in the course being the same and the test case failing. By only reloading one of the models, the associated courses will remain different and the test case can now accurately account for this scenario.

nina-huangg: The tests have been modified to remove the necessity to create a grade_entry_form in this block of the test cases. I have reverted the test file such that the creation of the grade_entry_form occurs in the first subject block, resembling the original form of the file.
The only change to this rspec file as of now would be splitting up the checking of allowed values into two cases: when it is a bonus mark and when it is not. 
This change allows the file to be more concise - with each block testing what needs to be tested clearly with no repetition.

david-yz-liu (approved):

PR #6558, opened 2023-04-27
LTI fixes: set cookie same-site to none and use generated session nonce

pretendWhale:
## Motivation and Context

This PR fixes minor issues with LTI integration not obvious in a development environment.

## Your Changes

**Description**:
- Sets LTI cookies same-site attribute to none, to allow the cookie to be passed back to markus during the LTI launch process.
- enforce ssl on requests to https endpoints
- uses a generated nonce rather than a session ID, because anonymous users do not have sessions.
- remove `check_host` action from `redirect_login` method
- reorganize tests to avoid double course creation when creating a course via LTI.

- [x] Other (please specify): LTI update

## Testing

Tested that all LTI functions operate appropriately via web interface in test instance, as well as running tests previously created.

## Checklist

### DIFF
--- a/app/controllers/lti_deployments_controller.rb
+++ b/app/controllers/lti_deployments_controller.rb
@@ -5,7 +5,7 @@ class LtiDeploymentsController < ApplicationController
   before_action :authenticate, :check_course_switch, :check_record,
                 except: [:get_config, :launch, :public_jwk, :redirect_login]
   before_action(except: [:get_config, :launch, :public_jwk, :redirect_login]) { authorize! }
-  before_action :check_host, only: [:launch, :redirect_login]
+  before_action :check_host, only: [:launch]
 
   def launch
     if params[:client_id].blank? || params[:login_hint].blank? ||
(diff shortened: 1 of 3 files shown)


david-yz-liu (commented):
@pretendWhale thanks, looks good. Two things:

1. Can you explain the "remove check_host action from redirect_login method" change?
2. It seems like there are some failing tests.

pretendWhale:
@david-yz-liu thanks for the review!

1. I ended up reverting that change and instead just ensuring the current host is one of the `known hosts`. In cases where the user launching MarkUs was not logged in to MarkUs, they are sent to the login page and then redirected back to the `redirect_login` method. I initially thought this would be fine, given that we verify the jwt canvas sends us, but we also should verify the host is whitelisted.
2. Yes, those should be fixed now. `check_host` was also not being tested correctly, so I've fixed that as well (the tests were trying to access a world-accessible route).

david-yz-liu (approved):

PR #5897, opened 2022-03-09
Modified Login to Allow Admin Users to Login

anoverga17:
## Motivation and Context

As part of issue #5830, AdminUsers will be able to do the admin like things through the UI (not just the API). As a first step to do this, they must be able to login through the UI. This pull request aims to give AdminUsers that functionality, by allowing them to log in and be redirected to a special admin only page

## Your Changes

**Description**:
- Allowed AdminUsers to be able to login
- Added new admin exclusive route that AdminUsers are taken to after logging in. This route is where an admin can manage users and courses (will be implemented in a later PR)
- Added new main_admin controller and policy to handle permissions and display of content for new admin exclusive route.
- Added a simple view for new admin exclusive route that provides a warning to logged in admins not to visit other markus web routes (lack of role for admin can cause errors)
- Added new admin translation page for all future admin related views.
- Added tests for my changes

**Type of change** (select all that apply):

- [x] New feature (non-breaking change which adds functionality)

## Testing

Manual testing of the feature was done by logging in as an AdminUser through the Web UI. In addition, rspec tests were written to ensure AdminUsers were logged in and redirected to the appropriate view.

## Checklist

- [x] I have performed a self-review of my own code.
- [x] I have verified that the pre-commit.ci checks have passed. 
- [x] I have verified that the CI tests have passed. 
- [x] I have reviewed the test coverage changes reported on Coveralls. 
- [x] I have added tests for my changes.

PR #6581, opened 2023-05-09
Admin developer task

pretendWhale:
## Motivation and Context

This removes the creation of a secondary admin user from the `db:admin` task, which is not necessary. It also renames the admin user's last name to 'user', which makes it slightly more consistent with the instructor user.

## Your Changes

**Description**:
Remove creation of second admin, and make the first and last name of the `.admin` user distinct.

- [x] Other (please specify): update rake task to create only one admin user

## Testing

Tested by running `rails db:reset` and inspecting the database afterwards, as well as logging in via the web interface.

Currently the admin user created has the `user_name` `.admin`. we could change this, or we could create a separate task to create a user with the `user_name` `admin`, which might be more intuitive as a `user_name` for an admin user in development and testing.

## Checklist

### DIFF
--- a/lib/tasks/users.rake
+++ b/lib/tasks/users.rake
@@ -4,16 +4,6 @@ namespace :db do
     puts user.api_key
   end
 
-  desc 'Create a developer Admin User'
-  task developer_admin: :environment do
-    [%w[admin test admin]].each do |admin|
-      a = AdminUser.create!(user_name: admin[0],
-                            first_name: admin[1],
-                            last_name: admin[2])
-      a.reset_api_key
-    end
-  end
-
   desc 'Create a single Instructor'
   task instructor: :environment do
     puts 'Populate database with Instructors'
--- a/lib/tasks/users.rake
+++ b/lib/tasks/users.rake
@@ -1,7 +1,7 @@
 namespace :db do
   task admin: :environment do
-    user = AdminUser.find_or_create
-    puts user.api_key
+    puts 'Populate database with Admins'
+    AdminUser.find_or_create
   end
 
   desc 'Create a single Instructor'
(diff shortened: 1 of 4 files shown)


david-yz-liu (commented):
@pretendWhale it seems like there are a few different changes made in this PR, can you give more context about the purpose of each change?

pretendWhale:
@david-yz-liu Sure! The initial goal was just to move the creation of the second admin user out of the `db:admin` task. As far as I can tell, that `test admin` user is primarily for convenience during development (allowing login as `admin` rather than `.admin`).

My initial commit moved the creation of the second user to a new task and added it to the `seeds`, to minimize changes to development. I ended up reconsidering and just removing the second user because it's not actually necessary. I don't have strong feelings about this, and I'm happy to revert back to that.

I also updated the admin user's last name to `user` because I thought it was more consistent with the naming scheme for the users generated by the `instructor` task. I also don't have strong feelings about this, but I noticed we had changed the instructor user's login to 'test instructor' in #5897 so I added it.

lib/tasks/users.rake
```diff
@@ -2,12 +2,6 @@ namespace :db do
   task admin: :environment do
     user = AdminUser.find_or_create
```
david-yz-liu, line 3: Okay, let's do the following:

1. Add a `puts` to log this action, similar to the other user types
2. Delete the `puts user.api_key` on the following line (seems like a leftover from testing or something), and so you won't need the `user` variable any more either.

david-yz-liu (commented):
@pretendWhale okay great that all makes sense, thanks. I just left one comment to further simplify the admin creation task.

I'm fine leaving the username as `.admin`, it's just something developers will need to know about, which is fine.

david-yz-liu (approved):

PR #6583, opened 2023-05-11
Web Socket -  Collect Submissions

DanielDervishi:
## Motivation and Context

- Replaces the use of polling to update the Submissions Table by using WebSockets. Now web sockets are used to communicate when the job responsible for collecting submissions is finished to the client. When the client receives this info, the client side web socket code updates the submissions table.

## Your Changes

**Description**:
- Migrated Submission Table from using polling to update after submission job completion to instead use web sockets to accomplish the same goal
- Added @rails/actioncable npm package
- Added action-cable-testing gem

## Testing

- Testing has been done on the web app assuring that the SubmissionsTable is correctly updated after submissions have been collected by manually going through the collection process
- Connection tests have been written assuring that each connection is associated with the correct user, rejecting clients that are not signed in
- CollectSubmissionChannel tests have been written assuring that when a user is signed in, they stream from the correct location
- SubmissionsJob has new tests assuring that after a user enqueues the job, they receive only one correct message and that no other user receives the same message. Further there are tests assuring that if the enqueued_user or notify_socket actions are not passed in, the server doesn't broadcast a message through the CollectSubmissionsChannel to the user.

Still to be done (Now all completed): 
- [x] Assure the introduction of Web-Sockets doesn't create any security issues
- [x] Look more into the polling code to assure that I've properly removed it 
- [x] Make sure that a user can only subscribe to the CollectSubmissionsChannel if they are actively viewing a submission page
- [x] Look more into the code for establishing connections in connection.rb (handle edge cases)
- [x] Look more into the manually_collect_and_begin grading Submission Controller method as it also interacts with the submission job
- [x] Update the web socket route in app/javascript/channels/consumer.js to be a relative path rather than absolute - almost complete.

## Checklist


app/javascript/channels/collect_submissions_channel.js
```diff
@@ -0,0 +1,24 @@
+import consumer from "./consumer";
+
+function create_collect_submissions_channel_subscription() {
```
david-yz-liu, line 3: Great! I couldn't see where this function was being called, but I think the right place to do it is in the `SubmissionTable` React component. If you do it in the `componentDidMount` method it should make the table automatically listen for updates (even from another tab).

app/javascript/channels/consumer.js
```diff
@@ -0,0 +1,6 @@
+// Action Cable provides the framework to deal with WebSockets in Rails.
+// You can generate new channels where WebSocket features live using the `bin/rails generate channel` command.
+
+import {createConsumer} from "@rails/actioncable";
+
+export default createConsumer("ws://localhost:3000/csc108/cable");
```
david-yz-liu, line 6: Here it seems like you can use [action_cable_meta_tag](https://api.rubyonrails.org/classes/ActionCable/Helpers/ActionCableHelper.html#method-i-action_cable_meta_tag) to avoid hard-coding the URL.

config/cable.yml
```diff
@@ -1,10 +1,8 @@
...
 
+production: *redis
+development: *redis
 test:
   adapter: async
-
-production:
-  adapter: redis
-  url: <%= Settings.redis.url %>
-  channel_prefix: markus_production
```
david-yz-liu, line 10: You should keep this (so merge this setting with the shared `*redis` settings.

app/channels/application_cable/connection.rb
```diff
@@ -0,0 +1,8 @@
+module ApplicationCable
+  class Connection < ActionCable::Connection::Base
+    identified_by :current_user
+    def connect
+      self.current_user = User.find_by user_name: cookies.encrypted["_markus_session_csc108"]["real_user_name"]
```
david-yz-liu, line 5: Instead of accessing the cookies directly, can you do `request.session[:user_name]`? This is what we use in `SessionHandler`, but it seems like we only have direct access to the `request` object in here?

config/application.rb
```diff
@@ -126,6 +125,8 @@ class Application < Rails::Application
         end
       end]
     end
+    config.action_cable.url =
+      "#{config.action_controller.relative_url_root}#{Settings.rails.session_store.args.path}/cable"
```
david-yz-liu, line 129: I'm not sure `#{Settings.rails.session_store.args.path}` is necessary.

But also, I'm wondering if the relative url root is also necessary if you use `config.action_cable.mount_path` instead of `url`. See https://guides.rubyonrails.org/action_cable_overview.html#in-app.

app/channels/collect_submissions_channel.rb
```diff
@@ -0,0 +1,9 @@
+class CollectSubmissionsChannel < ApplicationCable::Channel
+  def subscribed
+    stream_for current_user
```
david-yz-liu, line 3: Ah, one thing about security I forgot to mention: it would be good to enable action policy here, https://github.com/palkan/action_policy/blob/master/docs/rails.md#channels-integration. 

(You'll need to find and use the existing policy that governs whether a user can collect submissions or not.)

After you've enabled this, don't forget to add tests!

spec/controllers/submissions_controller_spec.rb
```diff
@@ -797,10 +797,13 @@
         end
 
         it 'should collect all groupings when override is true' do
+          enqueuing_user = User.find_by user_name: @instructor.user_name
```
david-yz-liu, line 800: I don't think the changes in this file are necessary? These values should be optional and not affect the rest of the functionality.

config/application.rb
DanielDervishi: When using `config.action_cable.mount_path` instead of `url` I believe it relocates the action cable server, mounting it on top of localhost:3000/csc108. 
Ex: if we did `config.action_cable.mount_path = "/csc108/cable"`, the server is now located at localhost:3000/csc108/csc108/cable. The client however attempts to connect to "localhost:3000/csc108/cable". 
When we don't specify the `mount_path`, but instead specify the `url` we avoid relocating the server and can properly target it.
In order for the client to know where the web socket is located (know about the Settings.rails.session_store.args.path section of the url), I need to specify it; otherwise it gets left out of the url that the client tries connecting to.

spec/controllers/submissions_controller_spec.rb
DanielDervishi: The collect submissions controller method currently passes the enqueuing user to the submissionsjob by default, causing the tests to expect that option in the job to be set. The changes are to account for that.

app/controllers/submissions_controller.rb
```diff
@@ -208,9 +208,11 @@ def collect_submissions
     end
     if collectable.count > 0
       @current_job = SubmissionsJob.perform_later(collectable,
+                                                  enqueuing_user: @current_user,
```
DanielDervishi, line 211: User @real_user instead.

config/application.rb
david-yz-liu: Perhaps you misunderstood what I was getting at. Try specifying `mount_path = 'cable'`. If that doesn't work push the change and I'll take a look.

spec/controllers/submissions_controller_spec.rb
david-yz-liu: I'm not sure what you mean by "causing the tests to expect that option in the job to be set". If you're getting an error when you don't include this line, you can show that.

app/channels/collect_submissions_channel.rb
```diff
@@ -0,0 +1,14 @@
+class CollectSubmissionsChannel < ApplicationCable::Channel
+  def subscribed
+    unless allowed_to?(:collect_submissions?, with: SubmissionPolicy, context: { real_user: current_user,
+                                                                                 role: current_user,
+                                                                                 real_role: current_user })
```
david-yz-liu, line 5: `real_role` shouldn't be necessary, it's marked as optional

spec/channels/application_cable/connection_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'rails_helper'
```
david-yz-liu, line 1: I'm not sure this line is necessary, I believe this file is included automatically when tests are run

app/channels/collect_submissions_channel.rb
```diff
@@ -0,0 +1,14 @@
+class CollectSubmissionsChannel < ApplicationCable::Channel
+  def subscribed
+    unless allowed_to?(:collect_submissions?, with: SubmissionPolicy, context: { real_user: current_user,
+                                                                                 role: current_user,
```
david-yz-liu, line 4: Roles are separate from users: they act as a join model between users and courses. So it's quite strange here that `current_user` (a `User`) is being passed as `role`. Are you confident the policy is working as intended here? As a comment, some of the tests below in `collect_submissions_channel_spec.rb` are stubbing `current_user` as an instructor/student (i.e., as a `Role`) rather than a `User`, so they might not be checking what you intend.

The tricky part is that you need a `course_id` in order to compute the associated role for this user. I think you can pass it as a parameter when creating the subscription from the front-end?

app/javascript/channels/collect_submissions_channel.js
```diff
@@ -0,0 +1,19 @@
...
+      // Called when the subscription is ready for use on the server
+    },
+
+    disconnected() {
+      // Called when the subscription has been terminated by the server
+    },
+
+    received(data) {
+      // Called when there's incoming data on the websocket for this channel
+      window.submissionTable.wrapped.fetchData();
```
david-yz-liu, line 15: Oh sorry I forgot to comment on this last time, but we should avoid using global variables (from `window`) when possible. In this case, it is possible to pass in the component itself (`this`) to `create_collect_submissions_channel_subscription`, so that here you can do something like `component.fetchData()`

david-yz-liu: Actually, please delete this file entirely, and instead convert this function into a method of the component itself.

app/channels/collect_submissions_channel.rb
david-yz-liu: (and to be clear, when you move this code into the React component at I mentioned below, you should have access to the props.course_id)

app/assets/javascripts/Components/submission_table.jsx
```diff
@@ -332,6 +334,26 @@ class RawSubmissionTable extends React.Component {
     }, after_function);
   };
 
+  create_collect_submissions_channel_subscription(submissionTable) {
```
david-yz-liu, line 337: Oh sorry I don't think I was clear, you shouldn't need to pass in `this` here. Since you've made this function a method of the class, you can use `this` directly.

However, you should use the "class properties" syntax for defining a method: `create_collect_submissions_channel_subscription = () => { ...`

spec/channels/collect_submissions_channel_spec.rb
```diff
@@ -0,0 +1,43 @@
+describe CollectSubmissionsChannel, type: :channel do
+  context 'when a user can collect submissions' do
+    let!(:instructor) { create :instructor }
+    let!(:current_user) { User.find_by user_name: instructor.user_name }
```
david-yz-liu, line 4: You can just do `instructor.user` here, and throughout (prefer association access over manual queries, at least for individual records)

david-yz-liu (approved):

PR #6584, opened 2023-05-11
added rspec tests for the GroupingsController#create_groups_when_stud…

khushiagl:
Added tests for the function of the Groups Controller - create_groups_when_students_work_alone

## Motivation and Context

Make sure the function is well tested.

## Your Changes

**Description**:

## Testing

Added rspec tests to check whether the appropriate job is enqueued and the correct template is rendered

## Checklist

### DIFF
--- a/spec/controllers/groups_controller_spec.rb
+++ b/spec/controllers/groups_controller_spec.rb
@@ -218,7 +218,7 @@ describe GroupsController do
       end
 
       it 'accepts a valid file' do
-        ActiveJob::Base.queue_adapter = :test
+        # ActiveJob::Base.queue_adapter = :test
         expect do
           post_as instructor, :upload, params: { course_id: course.id,
                                                  assignment_id: @assignment.id,
@@ -250,7 +250,7 @@ describe GroupsController do
 
         it 'creates groups for individual students' do
           data = students.map { |record| [record.user_name, record.user_name] }
-          ActiveJob::Base.queue_adapter = :test
+          # ActiveJob::Base.queue_adapter = :test
 
           get_as instructor, :create_groups_when_students_work_alone,
                  params: { course_id: course.id, assignment_id: assignment.id },
--- a/spec/controllers/groups_controller_spec.rb
+++ b/spec/controllers/groups_controller_spec.rb
@@ -218,7 +218,6 @@ describe GroupsController do
       end
 
       it 'accepts a valid file' do
-        # ActiveJob::Base.queue_adapter = :test
         expect do
           post_as instructor, :upload, params: { course_id: course.id,
                                                  assignment_id: @assignment.id,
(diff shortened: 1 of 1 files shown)


spec/controllers/groups_controller_spec.rb
```diff
@@ -241,6 +241,53 @@
       end
     end
 
+    describe '#create_groups_when_students_work_alone' do
+      before :each do
```
david-yz-liu, line 245: You have the right idea here, but should use `let` and `let!` instead of `before :each`. See [here](https://www.codewithjason.com/difference-let-let-instance-variables-rspec/), for example.

Also, you shouldn't hard-code specific user names. Instead, use `create_list` to create two students, and then compute the student's usernames down below for `data`.

spec/controllers/groups_controller_spec.rb
```diff
@@ -241,6 +241,53 @@
...
+        @assignment = create(:assignment)
+        # Create students
+        @student_user_names = %w[c8shosta c5bennet]
+        @student_user_names.each do |name|
+          create(:student, user: create(:end_user, user_name: name))
+        end
+      end
+
+      it 'creates groups for individual students' do
```
david-yz-liu, line 254: The tests from here down to line 278 should be wrapped in a `context` that describes the situation (the assignment has a group max of 1).

spec/controllers/groups_controller_spec.rb
```diff
@@ -241,6 +241,53 @@
...
+      end
+
+      it 'responds with appropriate status' do
+        get_as instructor, :create_groups_when_students_work_alone,
+               params: { course_id: course.id, assignment_id: @assignment.id },
+               format: 'js'
+        expect(response).to have_http_status(200)
+      end
+
+      it 'does not create groups if group_max > 1' do
```
david-yz-liu, line 280: Then here, this `it` should be turned into a `context`, and you should repeat the same test structure as above within this `context` block, so that each `expect` is in a separate test case.

david-yz-liu (commented):
@khushiagl great start, and good job finding the `        ActiveJob::Base.queue_adapter = :test`. I left a few comments for you to work on.

spec/controllers/groups_controller_spec.rb
```diff
@@ -241,6 +241,66 @@
...
+    describe '#create_groups_when_students_work_alone' do
+      # Create students
+      let!(:students) { create_list(:student, 5) }
+
+      context 'assignment.group_max = 1' do
+        let!(:assignment) { create(:assignment) }
+
+        it 'creates groups for individual students' do
+          data = []
+          students.each do |record|
```
david-yz-liu, line 253: Here try using `map` instead of `each` (you may want to look up these methods to understand the difference for creating a new list)

spec/controllers/groups_controller_spec.rb
```diff
@@ -241,6 +241,66 @@
...
+        end
+
+        it 'responds with appropriate status' do
+          get_as instructor, :create_groups_when_students_work_alone,
+                 params: { course_id: course.id, assignment_id: assignment.id },
+                 format: 'js'
+          expect(response).to have_http_status(200)
+        end
+      end
+      context 'assignment.group_max > 1' do
```
david-yz-liu, line 281: There should be a blank line between each `context` block, and between each `it` block.

david-yz-liu (approved):

PR #6591, opened 2023-05-19
Use Web Sockets to Update TestRunTable for Non-batch Run Tests

DanielDervishi:
**NOTE: Whenever I mention grader run tests, I mean tests run through the results controller.**
## Motivation and Context

Updates a student/grader's test table after the tests they run are in progress and completed.

## Your Changes

**Description**:
- Added new channel allowing for tests to communicate when they are in progress and completed to the enqueuing user (used to broadcast that a test has finished)
- Changed the message that's flashed when a student/grader runs tests to "Running tests." rather than "Running tests. Please refresh this page in a bit to view the results."

## Testing

**In the web interface**

- Manually tested that a subscription is created when a student/grader renders the test_run_table
- Manually tested that a successful test run by the student/grader triggers an update to the test_run_table when the test is in progress and completed

**Rspec tests**

- Tested that a student/grader subscribing who can run tests works fine
- Tested that a student/grader subscribing who can not run tests gets rejected
- Tested that the case where user's role can not be identified (course id not valid), subscribing fails
- Tested that when a test is not batch and is eventually completed, the enqueuing user's stream is broadcasted to
- Tested that when a test is not batch run and in progress, the enqueuing user's stream is broadcasted to
- Tested that when a test is batch run (with more than one group), at no point in the test's execution is the enqueuing user's stream broadcasted to

## Checklist


app/assets/javascripts/Components/test_run_table.jsx
```diff
@@ -18,6 +19,7 @@ export class TestRunTable extends React.Component {
 
   componentDidMount() {
     this.fetchData();
+    this.create_student_tests_channel_subscription();
```
david-yz-liu, line 22: Even though this feature started as just student tests, you've since extended it to instructor tests as well. This means you should rename the channels/functions/etc. to not mention "student" specifically. A better name for the channel is something like `TestRunsChannel`.

app/channels/student_tests_channel.rb
```diff
@@ -0,0 +1,28 @@
+class StudentTestsChannel < ApplicationCable::Channel
+  def subscribed
+    course = Course.find_by(id: params[:course_id])
+    role = Role.find_by(user: current_user, course: course)
+    if role.nil?
+      reject
+      return
+    end
+
+    assignment = Assignment.find_by(id: params[:assignment_id])
```
david-yz-liu, line 10: For these queries it would be good to use associations to further restrict the models that can be accessed. E.g., `course.assignments.find_by(...)`

app/controllers/automated_tests_controller.rb
```diff
@@ -62,8 +62,9 @@ def execute_test_run
                                                   assignment.id,
                                                   [grouping.group_id],
                                                   collected: false)
+
```
david-yz-liu, line 65: Revert this change

app/controllers/automated_tests_controller.rb
```diff
@@ -62,8 +62,9 @@ def execute_test_run
...
       session[:job_id] = @current_job.job_id
-      flash_message(:success, I18n.t('automated_tests.tests_running'))
+      flash_message(:success, I18n.t('automated_tests.student_tests_running'))
```
david-yz-liu, line 67: Similar to my above comment, you don't need to rename this I18n key

spec/channels/student_tests_channel_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+      expect(subscription).to have_stream_for(current_user)
+    end
+  end
+  context 'when the role is a student' do
+    let(:role) { create :student }
+    let(:current_user) { role.user }
+    before(:each) do
+      stub_connection(current_user: current_user)
+    end
+    context 'when the student can not run tests' do
```
david-yz-liu, line 24: change "can not" to "cannot" (throughout)

spec/factories/test_batch.rb
```diff
@@ -0,0 +1,5 @@
+FactoryBot.define do
+  factory :test_batch do
+    association :course, factory: :course
```
david-yz-liu, line 3: Look at how other model factories define a `course` association; they use a default course, which you should do here. This avoids needing to include the `before(:create)` block in the `test_runs.rb` factory you added.

david-yz-liu (approved):

PR #6596, opened 2023-05-23
Issue 6523

### LINKED ISSUE #6523: Suggestion: Button for suspending automated tests for coding assignments
This isn't a bug, but it would be really helpful to have a button for suspending automated tests for coding assignments!
Sometimes there may be errors in our code that make the Markus automated tests run for like 30+ minutes and return empty test results. I found it time-consuming waiting for those tests to finish so I can test it again after fixing one line, especially when I have the assignment due soon. If only we could suspend tests instead of refreshing the page numerous times and waiting for the test to run... We would save a lot of time and be able to start a new automated test with the fixed code more quicker...

khushiagl:
## Motivation and Context

Button for suspending automated tests for coding assignments for student interface. 
Closes #6523 

## Your Changes

**Description**:
Added the button for cancelling automated tests for the student interface which when clicked updates test status to cancelled and changed the policy of stop_tests in assignment controller to be accessible to students if student tests are enabled.

## Testing

Tested via UI and rspec tests

## Checklist

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -7,6 +7,8 @@ class AssignmentsController < ApplicationController
   responders :flash
   before_action { authorize! }
 
+  authorize :test_id, through: :test_id_param
+
   CONFIG_FILES = {
     properties: 'properties.yml',
     tags: 'tags.yml',
@@ -292,10 +294,11 @@ class AssignmentsController < ApplicationController
   end
 
   def stop_test
-    test_id = params[:test_run_id].to_i
+    test_id = test_id_param
+    @test_run = TestRun.find_by(id: test_id)
     assignment_id = params[:id]
     if current_role.student?
-      @current_job = AutotestCancelJob.perform_now(assignment_id, [test_id])
+      AutotestCancelJob.perform_now(assignment_id, [test_id])
     else
       @current_job = AutotestCancelJob.perform_later(assignment_id, [test_id])
       session[:job_id] = @current_job.job_id
@@ -988,5 +991,9 @@ class AssignmentsController < ApplicationController
       return false
     end
     true
+  end
+
+  def test_id_param
+    params[:test_run_id].to_i
   end
 end
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -294,7 +294,12 @@ class AssignmentsController < ApplicationController
   def stop_test
     test_id = params[:test_run_id].to_i
     assignment_id = params[:id]
-    @current_job = AutotestCancelJob.perform_now(assignment_id, [test_id])
+    if current_role.student?
+      @current_job = AutotestCancelJob.perform_now(assignment_id, [test_id])
+    else
+      @current_job = AutotestCancelJob.perform_later(assignment_id, [test_id])
+      session[:job_id] = @current_job.job_id
+    end
     redirect_back(fallback_location: root_path)
   end
 
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -7,7 +7,7 @@ class AssignmentsController < ApplicationController
   responders :flash
   before_action { authorize! }
 
-  authorize :test_id, through: :test_id_param
+  authorize :test_run_id, through: :test_run_id_param
 
   CONFIG_FILES = {
     properties: 'properties.yml',
@@ -294,13 +294,12 @@ class AssignmentsController < ApplicationController
   end
 
   def stop_test
-    test_id = test_id_param
-    @test_run = TestRun.find_by(id: test_id)
+    test_run_id = test_run_id_param
     assignment_id = params[:id]
     if current_role.student?
-      AutotestCancelJob.perform_now(assignment_id, [test_id])
+      AutotestCancelJob.perform_now(assignment_id, [test_run_id])
     else
-      @current_job = AutotestCancelJob.perform_later(assignment_id, [test_id])
+      @current_job = AutotestCancelJob.perform_later(assignment_id, [test_run_id])
       session[:job_id] = @current_job.job_id
     end
     redirect_back(fallback_location: root_path)
@@ -993,7 +992,7 @@ class AssignmentsController < ApplicationController
     true
   end
 
-  def test_id_param
+  def test_run_id_param
     params[:test_run_id].to_i
   end
 end
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -294,8 +294,7 @@ class AssignmentsController < ApplicationController
   def stop_test
     test_id = params[:test_run_id].to_i
     assignment_id = params[:id]
-    @current_job = AutotestCancelJob.perform_later(assignment_id, [test_id])
-    session[:job_id] = @current_job.job_id
+    @current_job = AutotestCancelJob.perform_now(assignment_id, [test_id])
     redirect_back(fallback_location: root_path)
   end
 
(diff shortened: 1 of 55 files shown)


config/routes.rb
```diff
@@ -257,6 +257,10 @@
           get 'student_interface'
           post 'execute_test_run'
         end
+
+        member do
```
david-yz-liu, line 261: It seems like you are reusing an existing route, so you shouldn't add a new route here.

app/views/automated_tests/_test_runs.html.erb
```diff
@@ -20,11 +20,18 @@
...
     run_tests_course_result_path(@current_course, @result) :
-    execute_test_run_course_assignment_automated_tests_path(@current_course, @assignment) %>
+    execute_test_run_course_assignment_automated_tests_path(@current_course, @assignment)
+
+    test_run_id = @grouping.test_runs.find_by(status: 0)&.autotest_test_id
+    route_stop = stop_test_course_assignment_path(@current_course, @assignment)
+  %>
   <div class="rt-action-box">
     <%= button_to t('automated_tests.run_tests'), route, disabled: !@authorized, remote: true %>
+    <%= button_to "Cancel tests", route_stop, method: :get, params: {test_run_id: test_run_id},
```
david-yz-liu, line 33: All user-facing strings (like `"Cancel tests"`) should go in a localization file. You should follow the example for the button on the previous line, and can read more about localization here: https://guides.rubyonrails.org/i18n.html

app/views/automated_tests/_test_runs.html.erb
```diff
@@ -20,11 +20,18 @@
...
 <%# student results page (with instructor tests released) does not need the button %>
 <% unless @current_role.student? && ran_by == 'instructor' %>
-  <% route = @current_role.instructor? ?
+  <%
+    route = @current_role.instructor? ?
     run_tests_course_result_path(@current_course, @result) :
-    execute_test_run_course_assignment_automated_tests_path(@current_course, @assignment) %>
+    execute_test_run_course_assignment_automated_tests_path(@current_course, @assignment)
+
+    test_run_id = @grouping.test_runs.find_by(status: 0)&.autotest_test_id
```
david-yz-liu, line 28: This query is on the right track, but needs some changes:

- the query should also only look for test runs that were created by the current role (i.e., current student)
- don't hard-code 0; instead, read about enums at https://guides.rubyonrails.org/active_record_querying.html#enums (since `status` is an enum field)
- you should be retrieving just the `id`, not `autotest_test_id`

app/controllers/assignments_controller.rb
```diff
@@ -294,8 +294,12 @@ def download_test_results
   def stop_test
     test_id = params[:test_run_id].to_i
     assignment_id = params[:id]
-    @current_job = AutotestCancelJob.perform_later(assignment_id, [test_id])
-    session[:job_id] = @current_job.job_id
+    if current_role.student?
+      @current_job = AutotestCancelJob.perform_now(assignment_id, [test_id])
```
david-yz-liu, line 298: `perform_now` doesn't return anything, so you don't need to assign it to a variable.

app/policies/assignment_policy.rb
```diff
@@ -28,6 +28,10 @@ def view_test_options?
     (check?(:run_tests?, role) && check?(:tests_enabled?)) || (role.student? && check?(:student_tests_enabled?))
   end
 
+  def stop_test?
```
david-yz-liu, line 31: Good, but since this method corresponds to at least one action, it's not a 'helper policy', so should be moved above the comment on line 25

david-yz-liu (commented):
@khushiagl good work (and good testing). I left a few inline comments, and have one additional general comment:

Even though you correctly updated the policy, there's an additional check that needs to be made to ensure students have the right permission to cancel tests. Take a look at the `execute_test_run` method (which is what is called when a student wants to run a test). Before actually running the test, the policy gets the student's grouping and checks whether the student can actually run tests for that grouping. You should do the same thing for `stop_test`, and additionally check that the test run the student is trying to stop is actually one they're allowed to stop! (This follows similar logic as the query you've written for getting the `test_run_id` in the first place.)

app/controllers/assignments_controller.rb
```diff
@@ -7,6 +7,8 @@ class AssignmentsController < ApplicationController
   responders :flash
   before_action { authorize! }
 
+  authorize :test_id, through: :test_id_param
```
david-yz-liu, line 10: Name this `test_run_id` and `test_run_id_param` to make it consistent with the model name

app/controllers/assignments_controller.rb
```diff
@@ -292,10 +294,15 @@ def download_test_results
   end
 
   def stop_test
-    test_id = params[:test_run_id].to_i
+    test_id = test_id_param
+    @test_run = TestRun.find_by(id: test_id)
```
david-yz-liu, line 298: It doesn't seem like this variable is ever used

app/models/student.rb
```diff
@@ -54,6 +54,11 @@ class Student < Role
     Settings.student_csv_order || %w[user_name section_name last_name first_name id_number email]
   ).map(&:to_sym).freeze
 
+  def can_cancel_test?(test_run_id)
```
david-yz-liu, line 57: Since this has to with authorization, a better place for this method is inside the `StudentPolicy` class, as a helper method.

app/models/student.rb
```diff
@@ -54,6 +54,11 @@ class Student < Role
...
+    test_run = test_runs.in_progress.where(id: test_run_id).first
```
david-yz-liu, line 58: Rather than doing a query and checking `.nil?`, look up the `exists?` ActiveRecord method

app/views/automated_tests/_test_runs.html.erb
```diff
@@ -20,11 +20,18 @@
...
 <%# student results page (with instructor tests released) does not need the button %>
 <% unless @current_role.student? && ran_by == 'instructor' %>
-  <% route = @current_role.instructor? ?
+  <%
+    route = @current_role.instructor? ?
     run_tests_course_result_path(@current_course, @result) :
-    execute_test_run_course_assignment_automated_tests_path(@current_course, @assignment) %>
+    execute_test_run_course_assignment_automated_tests_path(@current_course, @assignment)
+
+    test_run_id = @grouping.test_runs.where(role: @current_role).in_progress.first&.id
```
david-yz-liu, line 28: Any time there's a `.where` and `.first` the natural question is what happens when there's more than one (this is important to consider even if it's rare). This is a bit complicated in general, but here you can sort in descending order of `created_at` to get the most recent in-progress test.

spec/controllers/assignments_controller_spec.rb
```diff
@@ -801,6 +801,37 @@
...
+
+    context '#stop_test' do
+      context 'when student tests are allowed to cancel tests' do
+        it 'should respond with 302' do
+          assignment.assignment_properties.update!(unlimited_tokens: true)
+          allow_any_instance_of(AutotestCancelJob).to receive(:perform_now).and_return []
+          get_as role, :stop_test, params: { course_id: course.id,
+                                             id: assignment.id,
+                                             test_run_id: test_run.id }
+          expect(response.status).to eq 302
```
david-yz-liu, line 820: Use the `have_http_status` expectation method

david-yz-liu (commented):
@khushiagl great job! I left a few comments for you to work on. Additionally, there should be test cases for when the `test_run` is not `in_progress`, but has another status.

david-yz-liu (approved):

PR #6597, opened 2023-05-30
Web Sockets - Display Submission Collection Status

DanielDervishi:
## Motivation and Context

Uses web sockets instead of polling to update the flash header that displays  submission collection status after the collect submissions action has been called. Now gives more accurate status updates as a result.

## Your Changes

**Description**:
- Now broadcasts an `update_table` flag through `CollectSubmissionsChannel` after submissions have been collected, rather than `body: "sent"` in order to update the submissions table
- Added broadcasting through `CollectSubmissionsChannel` throughout the submission collection lifecycle to update the job status displayed to the user. What is broadcasted in most cases is a hash table representation of the data maintained for the given job by the activejob-status gem.
- Changed the `RawSubmissionTable`'s `createCollectSubmissionsChannelSubscription` method to now handle both submission table updating and status updating when data is received
- When data indicating an update to the submissionsJob's lifecycle is broadcasted, the client now parses that data and renders the new flash contents based off it 
- Removed polling when collecting submissions
- No longer sets `session[:job_id]` after the submissions job is enqueued through the collectSubmissionsMethod (this would cause uninteded flash messages to be rendered)
- Made sure to early return when a subscription should not be established in `CollectSubmissionsChannel`

## Testing

**Web Interface Tests**
Note that this code was run with the small addition surrounded by the red rectangle to ensure a warning message would pop up in the following video
Screenshot 2023-06-07 at 2 55 13 PM

https://github.com/MarkUsProject/Markus/assets/58835213/727c0ffe-7cf7-4cfe-9e5f-5547c992c6fc
This video displays:
- Submission collection status flashed to the user throughout the submissionsJob lifecycle
- Warning messages properly displayed to users throughout the submissionsJob lifecycle

Also separately tested that raising errors within the job results in an error message being shown to the user

**RSpec Tests**
- Tested that when a user collects submissions and one of the groups specified can actually have their submission collected, within the `SubmissionsController`, the controller broadcasts a status update as soon as the job is created
- Also tested that within the `SubmissionsJob`, we broadcast that the job's status is "working", the number of submissions collected so far and the total number of submissions to collect. We do this for each submission that has been collected
- When the job is complete, we test that the status broadcasted is "completed"
- Tested that warning messages are broadcasted despite the status of the job if present
- Tested that upon failure (due to exception), the submissionsJob broadcasts its status to the client
- Tested that when a user collects submissions and none of the groups specified can actually have their submission collected, no broadcasting occurs

## Checklist


david-yz-liu (commented):
@DanielDervishi you're definitely on the right track. My biggest pieces of advice are the following:

- in the job, it's a good idea to broadcast JSON; but instead of making up new keys, use the appropriate `status` values (see https://github.com/inkstak/activejob-status), which is what the current polling mechanism really uses
- in particular, send the progress information from the job
- then on the front-end, you don't need to make a request just to get a particular flash message to appear; instead, look for the Javascript code that actually displays flash messages and reuse that inside the React component

app/controllers/submissions_controller.rb
```diff
@@ -225,7 +225,6 @@ def collect_submissions
                      assignment_identifier: assignment.short_identifier)
       flash_now(:error, error)
     end
-    render 'shared/_poll_job'
```
david-yz-liu, line 228: This is good, but in general prefer an explicit `head :success` to send just an HTTP status rather than use whatever gets rendered by default.

app/controllers/submissions_controller.rb
```diff
@@ -213,7 +213,7 @@ def collect_submissions
                                                   collect_current: collect_current,
                                                   apply_late_penalty: apply_late_penalty,
                                                   notify_socket: true)
-      session[:job_id] = @current_job.job_id
+      CollectSubmissionsChannel.broadcast_to(@current_user, ActiveJob::Status.get(@current_job).to_h)
```
david-yz-liu, line 216: Make `current_job` a local variable rather than an instance variable (since it no longer needs to be accessed from a rendered view).

app/jobs/submissions_job.rb
```diff
@@ -43,10 +39,27 @@ def perform(groupings, apply_late_penalty: true, **options)
...
+        CollectSubmissionsChannel.broadcast_to(options[:enqueuing_user], ActiveJob::Status.get(job_id).to_h)
+      end
     end
+  rescue StandardError => e
+    status.catch_exception(e)
+    raise e
   ensure
     unless options[:notify_socket].nil? || options[:enqueuing_user].nil?
-      CollectSubmissionsChannel.broadcast_to(options[:enqueuing_user], body: 'sent')
+      message = if status&.progress == 1
```
david-yz-liu, line 51: In general don't use multiline if expressions. Instead turn this into an if statement.

app/jobs/submissions_job.rb
```diff
@@ -43,10 +39,27 @@ def perform(groupings, apply_late_penalty: true, **options)
...
+                    { status: :completed }
+                  else
+                    { status: :completed, warning_message:
+                      status[:warning_message] }
+                  end
+                else
+                  ActiveJob::Status.get(job_id).to_h
+                end
+      CollectSubmissionsChannel.broadcast_to(options[:enqueuing_user], message)
+      CollectSubmissionsChannel.broadcast_to(options[:enqueuing_user], update_table: true)
```
david-yz-liu, line 62: Seems strange to broadcast two separate messages. You can merge `message` and `{update_table: true}`

app/assets/javascripts/Components/submission_table.jsx
```diff
@@ -348,12 +350,70 @@ class RawSubmissionTable extends React.Component {
...
     );
   };
 
+  generateFlashMessageContentsUsingStatus = status_data => {
+    let message_data = {};
+
+    switch (status_data["status"]) {
+      case "failed":
+        if (!status_data["exception"] || !status_data["exception"]["message"]) {
+          message_data["error"] = "Failed";
```
david-yz-liu, line 371: Remember that all user-facing strings should be internationalized. I suspect that the previous statuses (from `activejob-status`) actually weren't, but we can do better now (especially since I never liked those terse one-word statuses anyway). Please use the following:

- `failed`, no message: `This job failed due to an unexpected error.`
- `failed`, message: `This job failed due to the following error:` (plus the error message, of course)
- `completed`: `This job is now complete.`
- `queued`: `This job is waiting to be run.`

Note that the in-progress message is fine, but should also be internationalized.

app/assets/javascripts/Components/submission_table.jsx
```diff
@@ -348,12 +350,70 @@ class RawSubmissionTable extends React.Component {
...
+        let progress = status_data["progress"];
+        let total = status_data["total"];
+        message_data["notice"] = `${progress}/${total} Submissions collected`;
+    }
+    if (status_data["warning_message"]) {
+      message_data["warning"] = status_data["warning_message"];
+    }
+    return message_data;
+  };
+  renderFlashMessages = message_data => {
```
david-yz-liu, line 392: You've done a good job extracting the relevant parts. You should move these functions into a `flash.js` file and import them from that file. Ideally also modify `ajax_events.js` to import at least the shared constants (`FLASH_KEYS`) from that file.

spec/controllers/submissions_controller_spec.rb
```diff
@@ -654,11 +654,11 @@
         grader_permission.save
       end
       context '#collect_submissions' do
-        before do
+        it('should respond with 204') do
           post_as grader, :collect_submissions,
                   params: { course_id: course.id, assignment_id: @assignment.id, groupings: [@grouping.id] }
+          expect(response.status).to eq 204
```
david-yz-liu, line 660: Please modify the tests to use the newer `have_http_status` matcher, which produces a better error message.

david-yz-liu (approved):

PR #6619, opened 2023-06-01
Changes to MarkUs API: Allow all roles to access index routes in CoursesController and AssignmentsController 

khushiagl:
## Motivation and Context

Changes to the MarkUs API (under the controllers/api folder) for improving the MarkUs-JupyterHub extension project.

## Your Changes

**Description**:
1. Modified the CoursesController#index route to accept requests from all types of roles, and used the visible_courses method to control what courses they can see.
2. Similar type of modification for the AssignmentsController#index route.

## Testing

Rspec tests for both controllers and corresponding policies.

## Checklist

### DIFF
--- a/app/controllers/api/assignments_controller.rb
+++ b/app/controllers/api/assignments_controller.rb
@@ -18,7 +18,7 @@ module Api
     # Returns a list of assignments and their attributes
     # Optional: filter, fields
     def index
-      assignments = get_collection(current_course.assessments) || return
+      assignments = get_collection(current_role&.visible_assessments&.assignments) || return
 
       respond_to do |format|
         json_response = "[#{assignments.map { |assignment| assignment.to_json(only: DEFAULT_FIELDS) }.join(',')}]"
--- a/app/controllers/api/assignments_controller.rb
+++ b/app/controllers/api/assignments_controller.rb
@@ -18,7 +18,9 @@ module Api
     # Returns a list of assignments and their attributes
     # Optional: filter, fields
     def index
-      assignments = get_collection(current_role&.visible_assessments&.assignments) || return
+      assignments = get_collection(current_role&.visible_assessments
+                                     &.assignments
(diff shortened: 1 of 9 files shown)


app/policies/api/assignment_policy.rb
```diff
@@ -12,5 +12,9 @@ def submit_file?
     def see_hidden?
       role.instructor? || role.ta? || role.visible_assessments(assessment_id: record.id).exists?
     end
+
+    def index?
+      real_user.is_a?(EndUser) && real_user.roles.exists?
```
david-yz-liu, line 17: This policy isn't strict enough. Checking `real_user.roles.exists?` doesn't check whether that user has a role in the current assignment's course. (Remember that this route is nested within `courses/`)

app/controllers/api/assignments_controller.rb
```diff
@@ -18,7 +18,8 @@ class AssignmentsController < MainApiController
     # Returns a list of assignments and their attributes
     # Optional: filter, fields
     def index
-      assignments = get_collection(current_course.assessments) || return
+      assignments = get_collection(current_role&.visible_assessments
```
david-yz-liu, line 21: This query is too permissive, though possibly related to my comment below on the policy. `current_role` should never be `nil` if the authorization was successful. Also, the `where` clause is not necessary because the `visible_assessments` method should take into account the associated `course` of the current role.

app/controllers/api/assignments_controller.rb
```diff
@@ -5,6 +5,8 @@ class AssignmentsController < MainApiController
     include AutomatedTestsHelper
     include SubmissionsHelper
 
+    authorize :course, through: :current_course
```
david-yz-liu, line 8: It's not necessary to pass this into the policy. The policy already has access to `role`, which is computed through the current course. So it should be sufficient to check that `role` is not `nil` in the policy.

PR #6620, opened 2023-06-01
Issue 5777

### LINKED ISSUE #5777: handle duplicate test result names
What should we do if there are duplicate test result names for a given test group

DanielDervishi:
## Motivation and Context

- Fixes #5777 
- Makes it so that one test causing an error does not spoil the `test_group_result`
- Improved feedback for the client on error

## Your Changes

**Description**:
- We now validate that `test_results` can not have the same name if they fall under the same `test_group_result`
- Error messages displayed to the grader now include error messages from the auto-tester and error messages from individual tests that may fail validation
- A test that causes some sort of validation error will no longer contribute to `marks_total` or `marks_earned` for a `test_group_result`
- A test that causes some sort of validation error will no longer prevent other tests from being saved and viewed by the client
- A test that causes an error no longer prevents other passing tests from being used to update the marks of an associated submission

## Testing

**Web Interface**
- Tested that tests can't have duplicate names in a `test_group_result` and that if they do, an error message is reported in the `extra_info` section of the table
- Tested that all error messages for all failing tests are present in the `extra_info` section
- Tested that when a criterion is associated with a test group, and we run tests for a given submission, despite errors in some tests, the result is updated properly

**RSpec Tests**
- Tested that a `test_result` with the same name as another within a `test_group_result` is not valid
- Tested that error messages are properly set in `test_group_result.extra_info`
- Tested that despite a test producing an error, the other tests contribute to `marks_total` and `marks_earned`
- Tested that `test_results` are created for the non-error producing tests
- Tested that a single test producing an error does not prevent the associated submission's results from being updated properly

## Checklist


app/models/test_run.rb
```diff
@@ -46,12 +45,21 @@ def update_results!(results)
               time: test['time'],
               position: i + 1
             )
+            marks_earned += test['marks_earned']
+            marks_total += test['marks_total']
+          rescue StandardError => e
+            extra_info = test_group_result.extra_info
+            test_name = test['name'].nil? ? '' : "#{test['name']} - "
+            test_group_result.update(extra_info: if extra_info.nil?
```
david-yz-liu, line 53: Don't use multi-line if expressions, use an if statement instead.

david-yz-liu (approved):

PR #6622, opened 2023-06-05
Replace jQuery AJAX Requests with Fetch Function

khushiagl:
## Motivation and Context

This PR aims to enhance the existing codebase by replacing all jQuery AJAX requests, which utilize the jQuery.get() method, with the modern JavaScript fetch() function. The fetch() function provides a built-in and standardized way to make AJAX requests, eliminating the dependency on the jQuery library for this specific functionality.

## Your Changes

**Description**:
Changed jQuery get requests to use fetch.

## Testing

test via UI

Some of the requests utilizing jQuery sent over additional javascript code, which cannot be displayed by the fetch API and hence, those requests remain unchanged.

## Checklist


david-yz-liu (commented):
@khushiagl this is good. Don't worry about adding the extra error checking right now, that's beyond the scope of this PR. Keep the logic the same as the original code :)

app/assets/javascripts/Components/Result/file_viewer.jsx
```diff
@@ -145,17 +145,27 @@ export class FileViewer extends React.Component {
             this.setFileUrl();
           });
         } else {
-          $.ajax({
-            url: this.props.selectedFileURL,
-            data: {preview: true, force_text: force_text},
+          const requestData = {preview: true, force_text: force_text};
+          const url = this.props.selectedFileURL;
+          const queryString = Object.keys(requestData)
```
david-yz-liu, line 150: Okay, so following up with my above comment, in this case because the URL isn't generated using js-routes, we do need to manually encode the query parameters to build the full URL. In these cases, you should use the newer [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams) (and the `.toString` method) to do the encoding for you.

app/assets/javascripts/Components/Modals/create_modify_annotation_panel_modal.jsx
```diff
@@ -29,17 +29,23 @@ class CreateModifyAnnotationPanel extends React.Component {
...
+
+    fetch(requestUrl, {
+      method: "GET",
+      headers: {
+        Accept: "application/json",
       },
     })
+      .then(response => {
+        if (response.ok) {
+          return response.json(); // Parse the response as JSON
```
david-yz-liu, line 46: Delete the comment here (and elsewhere you included this or similar comments); it is clear from the code

app/javascript/application_webpack.js
```diff
@@ -15,6 +15,9 @@ window.$ = window.jQuery = $;
 import * as ajax_events from "javascripts/ajax_events";
 window.ajax_events = ajax_events;
 
+import * as fetch_proxy from "javascripts/fetch_proxy";
```
david-yz-liu, line 18: You can just do `import "javascripts/fetch_proxy";` here. There isn't anything being exported, so this script is being imported for the side effect only (and should be included with the other such imports in this file)

app/assets/javascripts/Components/test_run_table.jsx
```diff
@@ -33,41 +33,49 @@ export class TestRunTable extends React.Component {
   }
 
   fetchData = () => {
-    let ajaxDetails = {};
+    let fetchDetails = {};
```
david-yz-liu, line 36: It seems like this should have the "accept JSON" header in all cases.

app/views/groups/assign_scans.html.erb
```diff
@@ -30,13 +30,20 @@
...
-      });
+      const queryString = $("#assign_student").serialize();
+      const url = "assign_student_and_next";
+      const requestUrl = `${url}?${queryString}`;
+      fetch(requestUrl, {
+        headers: {
+          Accept: "application/json",
+        }
+      })
+      .then((response) =>{
```
david-yz-liu, line 41: There's some whitespace formatting issues in this code block (note that prettier doesn't check `.erb` files, so you need to be careful with style yourself.

david-yz-liu (approved):
@khushiagl great work!

PR #6625, opened 2023-06-08
Bug Fixes - Web Sockets - Update TestRunTable for Non-Batch Run Tests

DanielDervishi:
## Motivation and Context

**Bugs this PR tackles:**
- When running tests and navigating away to another page, the "Running Tests. Please refresh ..." message would pop up
- When running a test on an empty `TestRunTable`, the table would not properly display the results of the test run once complete

**PR mentioned in title:**
- #6591

## Your Changes

**Description**:
- `AutotestRunJob` now no longer has a custom `on_complete_js` method as that would cause a misleading (only post web socket addition) flash message to pop up
- `TestGroupResults` table now updates the state variable `filteredData` right before it is rendered/re-rendered, causing the user to see the correct test results

## Testing

**Web Interface**
- Tried running a test on an empty `TestRunTable`, and now it updates properly
- Tried navigating away from the page with a `TestRunTable` after running the tests and now I no longer get the "Running Tests. Please refresh page ..." popup

## Checklist

### DIFF
--- a/app/assets/javascripts/Components/test_run_table.jsx
+++ b/app/assets/javascripts/Components/test_run_table.jsx
@@ -167,6 +167,7 @@ class TestGroupResultTable extends React.Component {
       filteredData: props.data,
     };
   }
+
   static getDerivedStateFromProps(props, _) {
     return {filteredData: props.data};
   }
--- a/app/assets/javascripts/Components/test_run_table.jsx
+++ b/app/assets/javascripts/Components/test_run_table.jsx
@@ -168,8 +168,10 @@ class TestGroupResultTable extends React.Component {
     };
   }
 
-  static getDerivedStateFromProps(props, _) {
-    return {filteredData: props.data};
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/Components/test_run_table.jsx
```diff
@@ -168,6 +168,10 @@ class TestGroupResultTable extends React.Component {
     };
   }
 
+  static getDerivedStateFromProps(props, _) {
```
david-yz-liu, line 171: You've identified the correct problem (`filteredData` isn't reset), but it's not exactly correct to update it here. `filteredData` always needs to remain in sync with `filtered` so that the filters apply (and this can occur when the state changes). Instead, you can use `componentDidUpdate` and check whether `props.data` has changed, and if so, reset the filter entirely like what's done in the constructor.

david-yz-liu (approved):

PR #6626, opened 2023-06-09
Update auto exam matching pipeline

anissingh:
## Motivation and Context

This pull request aims to integrate the new MarkUs Exam Matching pipeline with MarkUs. It attempts to increase the success rate with automatically matching student numbers.

## Your Changes

- `requirements-scanner.txt` was changed to include only the `markus_exam_matcher` package. Note that the version is specified by the commit hash, **not** branch name.
- The files in `./lib/scanner` (the original exam matching scripts) were removed and replaced with a succinct Python script that simply invokes the `run` method from the `markus_exam_matcher` package.

- [ ] Other (please specify):

## Testing

This PR was tested by creating a scanned assignment using a past exam provided to me. I then updated certain Student database entries to contain the student numbers found on past exam scans that I was given, and uploaded the scans to MarkUs.

The pipeline appears to only be working when the `SplitPDFJob` is called with `perform_now` instead of `perform_later`. As I cannot use `byebug` when `perform_later` is called, I attempted to put some `puts` statements around the Python script execution. These statements were producing output when `perform_now` was called, but not when `perform_later` is called. Because the Ruby backend is capturing the output of the Python script through the standard output, I suspect that there's some issue with this form of communication when `perform_later` is called. I'm currently thinking of changing the mode of communication to be through a temporary file, but I was wondering if there is a better solution. Maybe there's a way to get `stdout` to work in this context? Communicating results from Python scripts to Ruby through `Open3` seem to be done with the use of the standard output, and indeed this would be the easiest way given we can make it work somehow.

Edit: I've confirmed that I did not restart all the docker containers when I was originally testing, I only restarted `markus-rails-1`. Following my conversation with David yesterday, I learned that I was supposed to restart other containers as well in order to get the code to work with asynchronous jobs. This is why it began to work the next day, as I had restarted all the containers.

## Checklist


anissingh:
Okay so I'm testing today, and it seems to be working with `perform_later`. I don't quite understand why it wasn't working yesterday and it is working today.

.dockerfiles/entrypoint-dev-rails.sh
```diff
@@ -9,7 +9,9 @@ npm list &> /dev/null || npm ci
 # install python packages
 [ -f ./venv/bin/python3 ] || python3 -m venv ./venv
 ./venv/bin/python3 -m pip install --upgrade pip > /dev/null
-./venv/bin/python3 -m pip install -r requirements-jupyter.txt -r requirements-scanner.txt > /dev/null
+./venv/bin/python3 -m pip install -r requirements-jupyter.txt > /dev/null
+./venv/bin/python3 -m pip uninstall -y -r requirements-scanner.txt > /dev/null
```
david-yz-liu, line 13: @anissingh Okay I looked into this a little bit and didn't find anything conclusive, but my hypothesis is the installation of a fresh commit hash will work if the package version is bumped. Feel free to make a PR on markus-exam-matcher to try it out.

That's a better fix than uninstalling here (and is more in line with where we'll eventually end up, which is installing from the official package repository PyPI rather than GitHub).

anissingh:
I've confirmed that I did not restart all the docker containers when I was originally testing, I only restarted `markus-rails-1`. Following my conversation with David yesterday, I learned that I was supposed to restart other containers as well in order to get the code to work with asynchronous jobs. This is why it began to work the next day, as I had restarted all the containers.

.dockerfiles/entrypoint-dev-rails.sh
anissingh: @david-yz-liu Sounds good, just made the PR.

david-yz-liu: @anissingh Okay I just merged it in, please give it a try!

anissingh: It works, thanks! One thing I should point out, however, is that before integrating this package into MarkUs, `requirements-scanner.txt` contained a particular distribution of `opencv-python-headless`. Installing this package does not remove `opencv-python-headless` from the installed libraries because we switched over to `opencv-python`. Besides the fact that installing an unused library is unnecessary, it appears that it could cause errors if the versions do not align, which I do not believe they do in our case (see [here](https://stackoverflow.com/a/74245841), [here](https://stackoverflow.com/a/70898025), and [here](https://stackoverflow.com/a/71112424)). Although it doesn't seem to be dangerous in our environment, it's something that should maybe be addressed nonetheless.

david-yz-liu (approved):
Thanks for pointing out the opencv version issue, @anissingh! We don't have an automated way of removing Python requirements that are no longer needed from our dev environment, so it's just something we'll watch out for.

Issue #6638, opened 2023-06-19
Automatic Student Matching Crop Selection Disappears


anissingh:
Sometimes after having already selected (cropped) region of an exam cover page in the automatic student matching section of an exam template and clicking save, upon leaving the page and then re-entering it, the cropped region will not be displayed until the user clicks save again. This is misleading since the crop is stored in the database after the first save is clicked. See attached images for a better idea of the issue; note that a crop had already been selected and saved prior to this.

### Before Clicking Save Again
<img width="654" alt="prior" src="https://github.com/MarkUsProject/Markus/assets/79552425/e063e89e-8b1d-4226-af1f-4c5f8414c1c6">

### After Clicking Save Again
<img width="653" alt="post" src="https://github.com/MarkUsProject/Markus/assets/79552425/185516e6-c182-45bb-bdc3-eecdf4698e84">

PR #6640, opened 2023-06-20
Add Zoom Functionality For Cropping Exam Template

anissingh:
## Motivation and Context

This pull request aims to provide the user with an easier way to select the region of a scanned assignment containing the student number (or student username in the future) for automatic matching. Previously, it was difficult to accurately select the region of interest on the cover page, and it is important that the region be accurately selected to maximize the efficacy of the automatic matcher.

## Your Changes

**Description**:
Buttons were added beside the exam template image that allow the user to zoom the image in and out. The buttons have event listeners that are attached upon a `DOMContentLoad` event. These event listeners fire on a click event, and they utilize the `Jcrop` library to remove the existing `Jcrop` object and add a new one with different properties (specifically `boxWidth` and `boxHeight`) that allows the image to appear to be more zoomed in. CSS styling is used to prevent the image from overflowing and to allow for scrolling upon overflow. Some code in `_boot.js` was also refactored to adhere more closely to the Single Responsibility Principle.

## Testing

This change was tested in the web interface. The zoom buttons were tested such that clicking them appropriately increases / decreases the size of the image and preserves the crop box if it exists. Further, since the crop selection region is stored in the database in terms of `x`, `y`, `width`, and `height`, the automatic matching pipeline was tested on a small sample of 10 scanned exams consisting of empirical data. The automatic matching pipeline was able to match 9/10 exams when taking advantage of the crop zooming feature, as oppose to the previous success ratio of 8/10.

## Checklist


app/views/exam_templates/_student_info.html.erb
```diff
@@ -24,9 +24,17 @@
 
   <div id="exam-cover-display-<%= exam_template.id %>" class="<%= exam_template.automatic_parsing ? 'flex-display' : 'no-display' %>">
     <div id="exam_cover-<%= exam_template.id %>">
-      <img src="<%= show_cover_course_exam_template_path(@current_course, exam_template) %>" class="crop-target"
-           alt="<%= t(:'exam_templates.parsing.cover_page_for', id: exam_template.id) %>"
-           loading="lazy">
+      <div class="exam-crop-container">
```
david-yz-liu, line 27: It seems to me like this div can be merged into the outer one (`id="exam_cover-..."`), since I don't think the outer one is doing much as it is and you've only added one child do it

app/views/exam_templates/_student_info.html.erb
```diff
@@ -24,9 +24,17 @@
...
+        <div class="exam-crop-img-container">
+          <img src="<%= show_cover_course_exam_template_path(@current_course, exam_template) %>" class="crop-target"
+               alt="<%= t(:'exam_templates.parsing.cover_page_for', id: exam_template.id) %>"
+               loading="lazy">
+        </div>
+        <div class="exam-crop-button-container">
+          <button id="increase-crop-scale" class="crop-scale-button" type="button">+</button>
```
david-yz-liu, line 34: add a `title` attribute and set it to be the same (localized) "Zoom in" and "zoom out" strings that the image viewer uses

app/assets/stylesheets/common/_markus.scss
```diff
@@ -1366,3 +1365,35 @@ canvas {
   min-width: 250px;
   width: 50%;
 }
+
+.exam-crop-img-container {
+  margin-right: 3px;
+  max-width: 80%;
```
david-yz-liu, line 1371: Since the parent container has `display: flex` you can be a bit more specific with these widths. The button container can have `width: 40px` to match the width you set for the buttons, and then the image container can have `flex: 1` to make it fill the space

david-yz-liu (commented):
@anissingh good work. The Javascript changes all look good, though using global variables isn't idea it's okay for now. I left some style comments, but in addition to those, I'm thinking that the form inputs ("File to match by") can be moved to appear *above* the image container, which would allow the image container to be much wider than it currently is. This would further enhance the zooming feature to make it easier for the user to select the right crop area.

anissingh:
Update: Now fixes MarkUsProject/Markus#6638

anissingh:
I've added the change you suggested regarding moving the form inputs above the image, but I'm not entirely satisfied with the design I've created for the UI/UX. I was hoping you could provide some feedback and suggest improvements to make it cleaner. One option I considered is to force the image to always fill the container in terms of width. This is doable, but I wasn't sure if we want the user to be able to zoom out a lot (not sure if this would ever be beneficial, but maybe?). Another option is to make the buttons always render just to the right of the image, but I believe that this is bad UX because whenever the user zooms in, the image width expands, and so the buttons move to the right slightly (and same with zooming out).

app/views/exam_templates/_boot.js
```diff
@@ -48,39 +67,66 @@ function toggle_cover_page(id) {
...
+  $("#decrease-crop-scale").on("click", function () {
+    if (crop_scale - SCALE_CHANGE < MIN_SIZE) {
+      crop_scale = MIN_SIZE;
+    } else {
+      crop_scale -= SCALE_CHANGE;
     }
-  );
 
-  // Set crop selection if values exist.
+    jcrop_api.destroy();
```
david-yz-liu, line 97: In terms of function design, global variables are always a bit of a red flag. In this case it's okay that you're using it, but to minimize the reach/scope you can try the following:

1. Isolate the combination of `destroy` -> `config_jcrop_api` -> `set_crop_selection` into `attach_crop_box`, and call that function here and in the `increate-crop-scale` onclick. Inside `attach_crop_box`, you can guard the `destroy` call with a `if jcrop_api !== undefined` type of check.
2. Make `config_jcrop_api` *return` the Jcrop object, and then assign to `jcrop_api` inside `attach_crop_box`.
3. Modify `set_crop_selection` to take `jcrop_api` as an argument.

If I'm looking at the code correctly, this should ensure that only `attach_crop_box` really interacts with the global `jcrop_api` variable directly.

david-yz-liu (commented):
Great job @anissingh!

In terms of the overall UI, I think it's okay but not amazing. Some suggestions for improvements:

1. Make the image the full width of the enclosing div, perhaps with some additional padding. On maximized windows this will make the initial image quite large, but the zooming will still be useful if the browser window isn't fullscreen, for example.

    - To support this, you may want to change the `crop_scale` variable so that it's initialized to be the container's width when `attach_crop_box` is first called.

2. For the button icons, rather than using `+`/`-`, use the appropriate [Font Awesome](https://fontawesome.com) icons; see `fontawesome_config.js` and look for `<i></i>` tags in the codebase.
3. Also for the buttons, don't make them circular, and instead keep the same radius as other MarkUs buttons (so they'll look like rounded squares). This will make the style more consistent.
4. Instead of putting the buttons beside the image, overlay the buttons on top of the image, which is a fairly common UI pattern. Use absolute positioning to keep the buttons tracked to (e.g.) the top-right corner of the image container. If you want to get fancy, you can use a separate style for `:hover` to make the buttons pop when the user hovers over them. 😄 
5. When the image gets zoomed out, try to make it centred (equal left/right margins) rather than flushed to the left. Pick a background colour like `$disabled_area` to distinguish the background from the image.

anissingh:
@david-yz-liu I made the suggested UI/UX changes, and I'm still working on the scope of `jcrop_api` (it's turning out to be slightly more complicated than I thought, for some reason returning `jcrop_api` and assigning it to a local variable is making `Jcrop` behave differently). I should have the scope changes pushed soon, but for now, I wanted to give you a chance to review the UI/UX changes and to propose any additional feedback.

One thing I want to point out is how I overlayed the zoom buttons onto the image. This task was much more complicated than I thought because `position: absolute` does not keep the buttons in-frame when the image is scrolled. I tried multiple different techniques, including making the buttons `sticky` themselves, wrapping the buttons in a `sticky` container, making the buttons `sticky` and wrapping them in an `absolute` div, using javascript with a `scroll` event to dynamically re-position the buttons, and a little "hack" to make the buttons `fixed` and have them fix relative to a parent (rather than the window), but each of these solutions had its own issues. So I finally settled on using transformations to overlay the zoom buttons, but I am almost certain someone with more CSS experience could figure out a more elegant solution to this problem.

app/assets/stylesheets/common/_markus.scss
```diff
@@ -1366,3 +1363,54 @@ canvas {
...
 }
+
+.exam-crop-container {
+  display: flex;
+  height: 600px;
+  width: 1000px;
+}
+
+#exam-crop-img-container {
+  background-color: #f1f1f1;
```
david-yz-liu, line 1374: You shouldn't introduce new colour constants here, but instead reuse the existing ones. If you're worried about the colour being too light you can put a border or shadow around the container

app/assets/stylesheets/common/_markus.scss
```diff
@@ -1366,3 +1363,54 @@ canvas {
...
+  height: 40px;
+  margin-top: 2px;
+  min-width: 40px;
+  padding: 0;
+  transform: translate(-70px, 0);
+  width: 40px;
+  z-index: 800; // jcrop uses z-index of 690
+}
+
+.crop-scale-button:hover {
```
david-yz-liu, line 1404: I know I said to use `:hover`, but also we should make the style consistent with other buttons in MarkUs, including the ones on the page (e.g. the "save" button). So assuming that styling is already applied consistently across all buttons, you should probably remove this rule.

david-yz-liu (commented):
@anissingh UI looks good. The `transform` solution is okay, and I'm not experienced enough with CSS to have immediately come up with something better, so let's stick with that. 😄 I left a few inline comments.

app/assets/stylesheets/common/_markus.scss
anissingh: I'm actually worried about the colour of `$disabled_area` being too dark (since `Jcrop` greys out the image when there's a crop selection saved), and I didn't want to use a constant that's unrelated here (like `$line`). Do you have any suggestions?

david-yz-liu: Can you share a screenshot to illustrate the look you're worried about?

anissingh: Looking at the images side by side, I'm realizing that it's not as big of a distinction between my chosen colour and the one specified by `$disabled_area`. 

### Background with custom colour
Background with custom colour

### Background with $disabled_area colour
Background with $disabled_area colour

anissingh: I've pushed changes to use `$disabled_area` instead. I'm going to wait for your confirmation that the colour is not too dark before I resolve this conversation, however.

david-yz-liu: @anissingh okay thanks for uploading those screenshots. Yeah, I think `$disabled_area` isn't too dark, so I'm good with sticking with that 😄

app/assets/javascripts/fontawesome_config.js
```diff
@@ -41,6 +41,7 @@ import {
   faUserMinus,
   faUserPlus,
   faWarning,
+  faMinus,
```
david-yz-liu, line 44: This should go in alphabetical order with the other icons (makes it easier to keep track). Same below. Good catch on the `fawWarning`!

app/assets/stylesheets/common/_markus.scss
```diff
@@ -455,8 +455,6 @@ fieldset {
 
 .table-with-add {
   margin-left: 2.5em;
-  min-width: 500px;
```
david-yz-liu, line 458: You seem to have missed reverting the other changes to this rule.

david-yz-liu (approved):

Issue #4324, opened 2020-01-15
Feature request: grade submissions in random order


falsifian:
I would like to be able to see the submissions in a random order when I'm grading them. I'm worried that I might have biases I'm not aware of, e.g. I might be more tired and frustrated when I get to the end of the grading list, and more likely to give low grades.

(I can partly mitigate this by starting at a random point in the student list, but that's not perfect; e.g. I'll still always be grading Nancy after Mark, so if Mark's submissions make me angry, that's bad for Nancy.)

mishaschwartz:
I get what you're saying here but I'm not sure that randomizing the submission order is going to solve your problem. 

If you randomize the submissions then you're still going to be marking _someone's_ submission after Mark's and it will be unfair to that person (whether that person happens to be Nancy or anyone else). 

Perhaps a better way to mitigate bias in marking is to mark assignments anonymously (#4331)?

falsifian:
Oh, that sounds useful! I think it would still be nice to have them in random order in addition to that for the reasons mentioned, but I'm not sure whether it's worth prioritizing this request.

One other note: I grade things under an admin role, and also chose to add my TAs under an admin role too. I'm wondering whether the PR you linked would help us, or whether it only applies when people under the grader role are grading.

PR #6641, opened 2023-06-21
Added Random Incomplete Submission Button

DanielDervishi:
## Motivation and Context

Addresses #4324. 
This prevents grader fatigue from being a consistent variable in a given group's grade. Randomizing submissions also helps mitigate bias and promote fairness in the grading process. 

## Your Changes

**Description**:
The random incomplete submission button allows graders to view and grade submissions in a randomized order, instead of the default alphabetical order.

## Testing

**Rspec Tests:**
- Tests already existed for authorization, so I relied on those tests.
- Tested that `get_random_incomplete` returns nil if there are no more unmarked submissions accessible to the current user
- Tested that `get_random_incomplete` doesn't return the grouping on which this method was called
- Tested that `get_random_incomplete` doesn't return groupings without submissions
- Tested that when `get_random_incomplete` is called by a ta, it only chooses randomly from groups whose work the ta can grade as well as ones with an incomplete `current_result`
- Tested that when `get_random_incomplete` is called by an instructor, it returns a random group from all groups with an incomplete marking state

**Web interface testing:**
- Tested that when all submissions have been graded (potentially excluding the one currently being viewed), the random incomplete submission button is disabled
- Tested that pressing the random incomplete submission button indeed takes you to a random, incomplete submission that isn't the one you are currently viewing
- Tested that pressing 'ctrl+shift+r' while on the results page does the same thing as clicking the random incomplete submission button
- Tested that tas can only randomly navigate to submissions that they are allowed to grade
- Tested that the next random unmarked submissions button is not available to students
- Verified that the info modal displayed when '?' is clicked on a results page mentions the keybind corresponding to random incomplete submission when the accessing user is not a student

## Checklist


app/assets/javascripts/Results/keybinding.js
```diff
@@ -22,6 +22,14 @@ Mousetrap.bind("shift+right", function () {
   }
 });
 
+// Go to random submission with r
```
david-yz-liu, line 25: It's great that you thought about adding a keyboard shortcut. In terms of the actual shortcut to add, it would be better to make it similar to the one for "next grouping". Something like `ctrl+shift+right`. Also make sure to update this comment.

Also, make sure this keyboard shortcut appears on the help modal (accessible when you press `?`)

app/controllers/results_controller.rb
```diff
@@ -320,6 +320,16 @@ def next_grouping
...
   end
 
+  def next_random_unmarked_grouping
+    result = record
+    grouping = result.submission.grouping
+
+    next_grouping = grouping.get_next_random_unmarked(current_role)
+    next_result = next_grouping&.current_result
+
+    render json: { next_result: next_result, next_grouping: next_grouping }
```
david-yz-liu, line 330: Don't send the full object JSON, just the specific pieces of data that are necessary (result/submission/grouping id).

app/models/grouping.rb
```diff
@@ -728,8 +728,28 @@ def get_next_grouping(current_role, reversed)
...
+    groupings = get_all_unmarked(current_role)
+    groupings.order('RANDOM()').first
+  end
+
   private
 
+  def get_all_unmarked(current_role)
+    if current_role.ta?
+      # Get relevant groupings for a TA
+      groupings = current_role.groupings.joins(:current_result)
```
david-yz-liu, line 741: This is quite subtle, but `joins(:current_result)` doesn't actually work. Instead you should use `Assignment#current_results` and go from there.

spec/controllers/results_controller_spec.rb
```diff
@@ -931,6 +933,36 @@ def self.test_unauthorized(route_name)
         expect(filename).to eq "#{assignment.short_identifier}_#{complete_result.grouping.group.group_name}.pdf"
       end
     end
+    describe '#next_random_unmarked_grouping' do
```
david-yz-liu, line 936: Use a blank line to separate `describe` blocks and to separate `it` blocks

spec/models/grouping_spec.rb
```diff
@@ -1611,4 +1611,58 @@ def expect_updated_criteria_coverage_count_eq(expected_count)
       end
     end
   end
+  describe '#get_next_random_unmarked' do
+    # Code for intercepting the order call (this should allow us to see the value of groupings)
```
david-yz-liu, line 1615: It's not clear what this comment means

app/models/grouping.rb
```diff
@@ -728,8 +728,28 @@ def get_next_grouping(current_role, reversed)
...
   end
 
+  def get_next_random_unmarked(current_role)
+    groupings = get_all_unmarked(current_role)
+    groupings.order('RANDOM()').first
+  end
+
   private
 
+  def get_all_unmarked(current_role)
```
david-yz-liu, line 738: Don't separate this out into a private method, you can put the code into `get_next_random_unmarked`. For testing purposes you can just check that the return value is one of a few different possible values.

(This way you don't have to use `__send__` in a test, which is a bit of a code smell.)

app/models/grouping.rb
```diff
@@ -728,8 +728,28 @@ def get_next_grouping(current_role, reversed)
...
+    if current_role.ta?
+      # Get relevant groupings for a TA
+      groupings = current_role.groupings.joins(:current_result)
+                              .where(assignment: assignment, current_result: { marking_state: 'incomplete' })
```
david-yz-liu, line 742: Instead of `'incomplete'` and `'complete'` use the constant defined in `Result` (here and in the tests as well)

app/assets/javascripts/Components/Result/result.jsx
```diff
@@ -677,6 +677,37 @@ class Result extends React.Component {
     };
   };
 
+  nextRandomUnmarkedSubmission = () => {
```
david-yz-liu, line 680: *This is a global comment.* I thought about the name for this action (naming is hard!) and came to the following conclusions:

1. Remove the word "next" since order should not be implied.
2. Change the word "Unmarked" to "Incomplete" to be consistent with the MarkUs state naming.

app/assets/javascripts/Components/Result/submission_selector.jsx
```diff
@@ -95,6 +95,20 @@ export class SubmissionSelector extends React.Component {
     );
   }
 
+  renderNextRandomUnmarkedSubmissionButton() {
+    if (this.props.role !== "Student") {
+      return (
+        <button
```
david-yz-liu, line 101: The submission selector knows how many submissions are still ungraded. So it should be possible to disable this button when there are no other ungraded submissions (other than possible the current submission).

david-yz-liu (commented):
@DanielDervishi great work!

david-yz-liu (approved):
@DanielDervishi great work!

PR #6663, opened 2023-07-07
raise error for role API requests with nonexistent sections

pretendWhale:
## Motivation and Context

Currently, the roles API allows requests with non-existent or malformed `section_name` and sets the section to `nil`. This PR introduces an error message for this case. 

## Your Changes

**Description**:
Changed role API controller to raise an error if a section cannot be found based on the `section_name` parameter.

## Testing

Tested with the web api, and added tests for creating/updating with an invalid section name.

## Checklist

### DIFF
--- a/app/controllers/api/roles_controller.rb
+++ b/app/controllers/api/roles_controller.rb
@@ -115,8 +115,13 @@ module Api
       ApplicationRecord.transaction do
         user = User.find_by(user_name: params[:user_name])
         role = Role.new(**role_params, user: user, course: @current_course)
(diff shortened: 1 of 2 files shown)


david-yz-liu (commented):
@pretendWhale thanks! I realized as I was reviewing this that we do need a way for the user to *remove* a section from a student, e.g. with `section_name: nil`. This should be handled differently from either of the cases (1) `section_name` doesn't correspond to an existing section, and (2) `section_name` is not present in the params.

pretendWhale:
@david-yz-liu Thanks for the catch, you're totally right!

david-yz-liu (approved):

PR #6672, opened 2023-07-20
Upgrade to Ubuntu 22.04 and Ruby3

pretendWhale:
## Motivation and Context

Upgrade base docker image to use ubuntu 22.04 and upgrade ruby from 2.7 to 3.0

## Your Changes

**Description**:
Upgraded to ubuntu 22.04 and ruby 3. This requires a few additional changes. In particular:
- QR code scanning has been moved to a python function, because the gem we were using was not maintained and I wasn't able to find a suitable replacement. This also requires installing `python3-dev`
- postgres updated to version 14, the default on ubuntu 22.04
- `SortedSet` is removed in ruby3, so instances of `SortedSet` are replaced with `Set` (and sorted using the builtin `sort` method)
- modified tests for login validation to use a script with custom return values rather than mocking the return 
- Various other minor changes and test updates
 
## Testing

Ran the test suite to ensure all tests pass, and tested functionality in the web interface

## Checklist

- [x] I have updated tests for my changes. 
- [ ] I have made changes to the documentation and linked to that pull request below.


config/dummy_invalidate.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
```
david-yz-liu, line 1: If this file is only used for testing purposes, put it in the test fixtures folder

app/views/annotation_categories/_annotation_text.html.erb
```diff
@@ -66,7 +66,7 @@
     <% end %>
     <% if annotation_text[:annotation_category].nil? %>
       <div class='annotation-text-details'>
-        <%= t('activerecord.models.submission', {'count': 1}) + ': ' %>
+        <%= t('activerecord.models.submission', **{'count': 1}) + ': ' %>
```
david-yz-liu, line 69: it's better to do `t('activerecord.models.submission', count: 1)` to remove the hash entirely

Gemfile.lock
```diff
@@ -3,88 +3,88 @@ GEM
   specs:
     action_policy (0.6.5)
       ruby-next-core (>= 0.14.0)
-    actioncable (7.0.4.3)
-      actionpack (= 7.0.4.3)
-      activesupport (= 7.0.4.3)
+    actioncable (7.0.5)
```
david-yz-liu, line 6: I'm skeptical about the changes in this lock file. Please first do merge/rebase from master and then regenerate this file cleanly.

db/migrate/20121028211448_testing.rb
```diff
@@ -13,30 +13,30 @@ def self.up
     end
 
     create_table :test_support_files do |t|
-      t.string :file_name, { :null => false }
-      t.references :assignment, { :null => false }
-      t.text :description, { :null => false }
+      t.string :file_name, **{ :null => false }
```
david-yz-liu, line 16: as above, I think you can do `null: false` without the `**{...}`

app/jobs/split_pdf_job.rb
```diff
@@ -42,18 +42,24 @@ def perform(exam_template, _path, split_pdf_log, _original_filename = nil, _curr
...
+        python_exe = Rails.application.config.python
+        read_qr_py_file = Rails.root.join('lib/scanner/read_qr_code.py').to_s
+        stdout, status = Open3.capture2(python_exe, read_qr_py_file, qr_file_location)
+        if status.success?
+          m = code_regex.match(stdout)
+        else
+          m = code_regex.match(RTesseract.new(qr_file_location).to_s)
+        end
         status = ''
-
```
david-yz-liu, line 56: revert this change

app/jobs/split_pdf_job.rb
```diff
@@ -223,13 +229,12 @@ def save_pages(exam_template, partial_exams, filename = nil, split_pdf_log = nil
                   'application/pdf')
         end
         repo.commit(txn)
-
```
david-yz-liu, line 226: revert this change

db/migrate/20121028211448_testing.rb
pretendWhale: As mentioned in the meeting, this change breaks the migrations, so I've opted against it.

david-yz-liu: @pretendWhale Please push the exact change you made and I'll take a loot

pretendWhale:
@david-yz-liu thanks for the review! 

1. The migrations ran without errors, but there were test errors due to extraneous  db columns. That said, it looks like at some point errors were introduced into my `structure.sql`, which was causing the issue.
2. Fair enough, that makes sense.
3. I encountered issues with the splitting pdfs, and so yes that's what my change in `split_pdf_job.rb` is me. I missed the seeding issue (and it didn't come up in the tests) - I made an equivalent change to `exam_template.rb` which I'll push

david-yz-liu (approved):

PR #6700, opened 2023-08-08
Update Exam Matcher Package Version

anissingh:
## Motivation and Context

This pull request updates the version of the exam matcher package to the latest version, which made the package runnable. Consequently, the Python script used to interact with the package can now be removed. 

## Your Changes

**Description**:
The Python script that was previously located at `lib/scanner/run_scanner.py` was removed, and the exam matcher package is now executed on the command line (`Open3.capture2` is still used to execute it). The version of the exam matcher package specified in `requirements_scanner.txt` was updated to include the latest version. Since this version is hosted on PyPI, the git commit hash method of version checking is no longer used.

## Testing

This pull request was tested by using a subset of real student exam scans that were uploaded using the MarkUs interface. All the student numbers contained in these exams were added to the database, and automatic matching was performed. Performance was the same as it was with the previous exam matcher version, which is to be expected, as only refactoring changes were made within the package itself.

The `markus_exam_matcher` dependency needs to be pinned in this case because of the way the code that checks if the requirements are installed works (this code is found in `config/initializers/python.rb`). The relevant part of the code reads:
```
if (scanner_requirements - installed).empty?
   ...
```
which would fail if the dependency weren't pinned because installed will contain `markus_exam_matcher==0.1.0`. If there is a better way to check these dependencies, I can make that change before this PR is merged. 

## Checklist


app/jobs/split_pdf_job.rb
```diff
@@ -258,10 +258,9 @@ def save_pages(exam_template, partial_exams, filename = nil, split_pdf_log = nil
         img.destroy!
         student_info.destroy!
 
-        python_exe = Rails.application.config.python
-        read_chars_py_file = Rails.root.join('lib/scanner/run_scanner.py').to_s
+        exam_matcher_exe = "#{Rails.application.config.python} -m markus_exam_matcher"
         char_type = exam_template.cover_fields == 'id_number' ? 'digit' : 'letter'
-        stdout, status = Open3.capture2(python_exe, read_chars_py_file, student_info_file, char_type)
+        stdout, status = Open3.capture2("#{exam_matcher_exe} #{student_info_file} #{char_type}")
```
david-yz-liu, line 263: In general command line arguments should be passed individually (like it was before), not as a concatenated string. This allows the function to correctly quote the arguments (e.g. when there's spaces in filenames) so that you don't need to worry about doing so.

anissingh: This is definitely a better solution, thanks! In case it is important, the code that runs `pip freeze` in `config/initializers/python.rb` is written like so:
```
pip = "#{Rails.application.config.python} -m pip"

begin
  installed, status = Open3.capture2("#{pip} freeze")
  ...
```
Maybe this should also be changed for consistency?

david-yz-liu: Ah yes, you can change that as well. (It matters less because `Rails.application.config.python` isn't a user-provided input, but still good to follow best practice.)

anissingh: Done!

david-yz-liu (approved):

PR #6742, opened 2023-09-13
Move create and edit tag modals to React

TheTallJerry:
## Motivation and Context

This PR is part of the ongoing effort of having the frontend to be in React as much as possible. 

## Your Changes

**Description**:

The changes are centered around the modals to create and edit tags, in both the tags table and the submissions info panel (for assignments). Specifically:

- The modals for both creating and editing tags are now fully in React, alongside Jest UI tests for each. 
    - Tag name now has to be non empty before the form (on the modal) is submittable, indicated by a red background when the tag name input is empty (consistent with the rest of Markus). Similarly, the submit button is disabled until then.
- The modals previously lived in the ERB layer and partially in tags controller. Those code are removed as well. 
- The React modals are written to replicate the old ERB modals as much as possible. 
- Added validation for Tag model (rails) name being not null and max 30 chars. Removed frontend character count for Tag description as there's no not null constraint on the db level. 

New Modals

## Testing

`CreateTagModal` and `EditTagModal` are common in a lot of ways, so the common parts are refactored out into a `TagModal`. These 3 components have a test suite each. 

## Checklist

### DIFF
--- a/app/assets/javascripts/Components/__tests__/tag_modal.test.jsx
+++ b/app/assets/javascripts/Components/__tests__/tag_modal.test.jsx
@@ -5,7 +5,7 @@ import Modal from "react-modal";
 
 describe("TagModal", () => {
   let props;
-  let component;
+  let rerenderComp;
 
   beforeEach(() => {
     props = {
@@ -21,12 +21,13 @@ describe("TagModal", () => {
 
     // Set the app element for React Modal
     Modal.setAppElement("body");
-    component = render(<TagModal {...props} />);
+    const {rerender} = render(<TagModal {...props} />);
+    rerenderComp = comp => {
+      rerender(comp);
+    };
   });
 
-  afterEach(() => {
-    cleanup();
-  });
+  afterEach(cleanup);
 
   it("should close on Cancel", () => {
     fireEvent.change(screen.getByTestId("tag_name_input"), {target: {value: "Name"}});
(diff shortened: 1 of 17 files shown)


app/assets/javascripts/Components/__tests__/tag_modal.test.jsx
```diff
@@ -0,0 +1,86 @@
...
+      .queryAllByRole("textbox", {hidden: true})
+      .map(textarea => textarea.getAttribute("data-testid"));
+    expect(inputDataTestIds.sort()).toEqual(["tag_name_input", "tag_description_input"].sort());
+    // verify maxLength for the inputs
+    expect(screen.getByTestId("tag_name_input")).toHaveAttribute("maxLength", "30");
+    expect(screen.getByTestId("tag_description_input")).toHaveAttribute("maxLength", "120");
+  });
+
+  describe("When name filled", () => {
+    beforeEach(() => {
```
david-yz-liu, line 64: I'm not really sure what the best practice is for test organization is in this case, when you want to customize some props to overwrite what the default `beforeEach` is doing. The approach you're currently taking seems a bit clunky. There's a `rerender` method for react testing library, how about using that instead?

TheTallJerry: Changed to using `rerender`, still some setup needed but does get rid of the extra `cleanup`

david-yz-liu (approved):

PR #6763, opened 2023-10-05
Python init fix

pretendWhale:
## Motivation and Context

In the production environment, the `markus_exam_matcher` package installs as `markus-exam-matcher` which leads the python initializer to set `Rails.application.config.scanner_enabled = false`, even though the package is installed and can be imported/invoked with `markus_exam_matcher`.

## Your Changes

**Description**:
This PR modifies the initializer to remove hyphens and underscores from package names in the various requirements files as well as the list of installed packages, which should allow both package names to be matched.

## Testing

Manually tested with by restarting docker container, as well as using `binding.break` to test package names. NOTE: we should add unit tests for this (and other) initializers, but it requires moving the initialization script into a class. We should in the future make a separate PR which adds tests.

## Checklist

### DIFF
--- a/config/initializers/python.rb
+++ b/config/initializers/python.rb
@@ -17,6 +17,7 @@ Rails.application.config.after_initialize do
   def process_requirements(requirements_file)
     File.open(requirements_file).each_line.filter_map do |line|
       line.strip!
+      line.tr!('-_', '')
       line.start_with?('#') || line.length.zero? ? nil : line
     end
   end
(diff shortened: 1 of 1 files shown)


config/initializers/python.rb
```diff
@@ -42,6 +45,12 @@ def process_requirements(requirements_file)
              'Jupyter notebook rendering will be disabled. ' \
              "To enable notebook rendering run: #{pip} install -r #{Rails.root.join('requirements-jupyter.txt')}"
       end
+      unless (qr_requirements - installed).empty?
+        warn 'MARKUS WARNING: not all packages required to process scanned exams ' \
+             'are installed. Exam scanning will not be enabled. ' \
+             'To enable exam scanning  run: ' \
```
david-yz-liu, line 51: @pretendWhale Please delete the double space

pretendWhale: Did this for the `qr_requirements` warning. The `scanner_requirements` warning also has a double space - let me know if you think that's worth taking action on.

david-yz-liu: @pretendWhale good catch, yeah please go ahead and fix it in both

pretendWhale: @david-yz-liu Done!

david-yz-liu (approved):

PR #6764, opened 2023-10-06
Replacing the function `tas` of the `Assignment` model with a `tas` association for that model

mimischly7:
## Motivation and Context
Originally, the `Assignment` model had an instance method `tas`, responsible for retrieving all `Ta`'s associated with the given assignment. However, since -on the database- assignments and ta's are "connected" (albeit indirectly, through several intermediary tables), it makes more sense to define a `tas` association in the Assignment model.

## Your Changes

**Description**:
I removed the instance method `Assignment.tas` and instead defined a new association `tas` in the model. 
In particular, the association is defined as: `has_many :tas, through: :ta_memberships, source: :role`. Each assignment can have many ta's, and this relationship can be derived through the intermediary `TaMembership` "junction" model. Additionally, we need `source: :role` because in the `Membership` model (which `TaMembership` extends), the association that interests us is `role` (and not `tas`, which we use as the name of the association in the declaring model).
Since an association element is invoked in the same way that an instance method is invoked (dot syntax), no refactoring was needed in the parts of the code that called the former instance method. 

## Testing

I ensured all `rspec` tests run successfully. I also checked that the UI is still functioning as expected regarding the components that could have been affected.

I had a more design choice question. It appears that when we say that `@assignment.tas` (which is meant to give all tas associated with the specified assignment), this returns `Role` objects. This makes sense, as the `Ta` model extends the `Role` model. But why is the application's design so that a `Ta` is a `Role` instead of a `User`?

## Checklist


mimischly7:
Hello Professor, I am a member of MarkUs (I accepted the invitation) but for some reason, it still does not allow me to request a review, I am not sure if I am missing something.

The `reviewers` icon is not pressable (it is like it is turned off).

Screenshot 2023-10-06 at 7 42 54 PM

david-yz-liu (approved):

PR #6768, opened 2023-10-09
Fix SplitPdfJob when replacing/adding to an existing group's pages

david-yz-liu:
## Motivation and Context

An instructor who was using the new "overwrite duplicate pages" option reported that the overwritten pages were not actually being added to the group's repository (and so the updated files couldn't be collected).

## Your Changes

**Description**:

It turns out there were three bugs in the existing `SplitPdfJob` implementation.

1. The existing code that updated the group repositories (`SplitPdfJob#save_pages`) always attempted to add a file rather than ever replacing an existing file. The check `File.exist?` was incorrect; we need to use `revision.path_exists?` to ensure the path is relative to the group repository.
2. After fixing the check, I also needed to update how the `txn.replace` method was being called.
3. The check for whether to store scanned pages in the "complete" vs. "incomplete" directories was assuming that a full test paper would always be scanned. However, if an instructor wants to re-upload just a single page, we should look in the "complete" directory as well. I fixed this check.

## Testing

Tested manually in the UI and added additional test cases.

## Checklist


pretendWhale (approved):

pretendWhale:
@david-yz-liu The changes look good to me, but coveralls is complaining about master being ahead of this by a commit or so (which I think may be causing the other coveralls checks to remain pending). I don't see any reason not to merge it, but I wanted to ask before merging in case you do (if not, please feel free to just merge it).

david-yz-liu:
@pretendWhale yeah, I think there was a change to coveralls but I've seen this happen a bit. I've updated the branch to make sure it's up to date with master, which hopefully will make coveralls happy. Let's wait to see what happens after re-running it.

PR #6797, opened 2023-10-26
add websocket config to settings

pretendWhale:
## Motivation and Context

Adds websockets `allowed_request_origins` setting to the settings files, so that they can be overwritten in e.g. a `settings.local.yml`

## Your Changes

**Description**:
Added setting to settings.yml, as well as production and development

## Checklist

https://github.com/MarkUsProject/Wiki/pull/196

### DIFF
--- a/config/settings/development.yml
+++ b/config/settings/development.yml
@@ -16,6 +16,8 @@ rails:
     default_url_options:
       host: 'host.docker.internal'
       port: '3000'
+  action_cable:
+    web_socket_allowed_request_origins: <%= %w[] %>
 
 puma:
   worker_timeout: 100000000
@@ -43,5 +45,3 @@ autotest:
 lti:
   domains: <%= %w[host.docker.internal] %>
   token_endpoint: "http://host.docker.internal:3100/login/oauth2/token"
-
-web_socket_allowed_request_origins: <%= %w[host.docker.internal] %>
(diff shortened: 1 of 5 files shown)


donny-wong (approved):

config/settings/development.yml
```diff
@@ -43,3 +43,5 @@ autotest:
 lti:
   domains: <%= %w[host.docker.internal] %>
   token_endpoint: "http://host.docker.internal:3100/login/oauth2/token"
+
+web_socket_allowed_request_origins: <%= %w[host.docker.internal] %>
```
david-yz-liu, line 47: Let's nest this configuration option under `rails` -> `action_cable`, so that it better matches the structure of existing options.

david-yz-liu (commented):
@pretendWhale I left a comment about the naming. Also, I think this option needs to be added to the settings schema (settings.rb or something). And please make sure to add it to the Wiki documentation as well.

pretendWhale:
@david-yz-liu thanks for the review! Updated the naming. I did add the setting to `application.rb`, which I _think_ is sufficient? At the very least, the setting has the correct value when read from the rails console.

david-yz-liu (commented):
@pretendWhale okay this is good but also I realized I had the wrong file name, the settings *schema* is stored in `config/initializers/config.rb` and should be updated whenever a configuration setting is added/modified.

pretendWhale:
You're totally right and I forgot that file existed - thanks for the catch @david-yz-liu ! I made both `action_cable` and `web_socket_allowed_request_origins` optional because I don't think they are strictly necessary to actually having markus up, but let me know if you have other thoughts.

david-yz-liu (approved):

PR #6801, opened 2023-10-28
Prevent inactive graders from being assigned to groups

mimischly7:
## Motivation and Context

Currently, instructors have no problem adding inactive graders to groups (in the the graders tab of an assignment). This should not be the case, as inactive graders should never be assigned to a group. This is similar to the `groups` tab of assignments, where instructors are not allowed to add inactive students to a group. Thus, the motivation of this pull request is to implement a mechanism that throws a red flash message whenever an instructor tries to assign an inactive grader to a group (and ensures that the grader is not actually added). Backend modification was necessary to implement this functionality.

A short overview of the flow of data when an instructor presses the `Assign Grader(s)` button:
1. A POST HTTP request is made to the ruby backend, specifically the `global_actions` action of the `graders_controller` controller. 
2. For the ‘assign’ case of the case block (which is the one that concerns us), `assign_all_graders` is called with the `grouping_id`'s and `grader_id`’s as inputs to actually add the graders to the groups.

## Your Changes

**Description**:

In the `global_actions` method, perform a single database request to fetch all required graders, and determine whether any of the graders to be added are inactive. If so, a (newly defined) custom error is thrown (with the message of `groups.invite_member.errors.inactive_graders` from `en.yml`). 
Ultimately, any inactive graders that were selected will _not_ be added to any of the chosen groups (but if some of the selected graders were active, these will be added to the corresponding groups).
Note that for the case of _unassigning_ graders, whether the grader is active or inactive is irrelevant; instructors should be able to unassign inactive graders.

			
## Testing

Testing on the web interface with multiple different combinations of graders to add (and groups to which they are to be added), ensuring the corresponding is performed correctly.

Wrote five `RSpec` tests, testing the behavior of assigning, randomly-assigning, and unassigning (in the context of some graders being inactive).

## Checklist


app/controllers/graders_controller.rb
```diff
@@ -98,21 +98,27 @@ def global_actions
 
       case params[:global_actions]
       when 'assign'
-        if params[:skip_empty_submissions] == 'true'
-          # If the instructor wants to skip empty submissions, filter
-          # groups with empty submissions into a new array
-          filtered_grouping_ids = filter_empty_submissions(grouping_ids)
-          if filtered_grouping_ids.count != grouping_ids.count
-            found_empty_submission = true
+        begin
```
david-yz-liu, line 101: Overall the logic (for detecting inactive TAs) is correct, but it would be better to pull it out of the helper and instead put it at the start of the controller method, in the same block as Lines 80-88 that are doing the initialization of `grader_ids`. This will cause the filtering to apply right at the start. Enable the inactive-grader filter whenever an "assign" or "randomly assign" action is selected, but _not_ an "unassign". That is, we'll allow inactive TAs to be unassigned from existing groups/criteria, but not assigned to new ones.

Also if you move this check up there, you can use `flash_message` directly rather than raising an error.

config/locales/views/groups/en.yml
```diff
@@ -35,6 +35,7 @@ en:
         already_pending: 'Could not invite %{user_name}: this student already has a pending invitation for this group.'
         extension_exists: You cannot modify a group with an extension to the due date. Please contact your instructor if you wish to make changes.
         group_max_reached: 'Could not invite %{user_name}: group maximum has been reached.'
+        inactive_grader: Cannot add '%{user_names}' to group/s because these graders are inactive.
```
david-yz-liu, line 38: This message should be displayed when trying to assign graders to groups _or_ criteria. So a more general message would be "Cannot assign inactive grader(s) '%{user_names}'.

david-yz-liu (commented):
@mimischly7 good work. I left a few comments, and please also make sure to add some test cases for this change (for both group and criteria assignment).

Changelog.md
```diff
@@ -19,6 +19,8 @@
 ## [v2.3.3]
 - Fix bug where uploading scanned exam pages with overwriting option selected did not update submission files (#6768)
 - Fix bug: "Download Submissions" download link was not being rendered from partial view (#6779)
+- Bug Fix: Prevent assigning inactive graders to a group, throwing a red flash message (through ```graders_controller```)
```
david-yz-liu, line 22: This entry should be added to the "unreleased" section. Also you don't need to say the second part, just say the "Prevent assigning inactive graders to a group".

app/controllers/graders_controller.rb
```diff
@@ -87,6 +87,44 @@ def global_actions
       end
     end
 
+    # '%w' syntax is a Ruby idiom that, preferred over ["...", "...", ...] syntax for its conciseness and readability
```
david-yz-liu, line 90: You should remove all comments here. These are good for learning purposes but not in the final code.

app/controllers/graders_controller.rb
```diff
@@ -87,6 +87,44 @@ def global_actions
...
+    # It represents an array of strings
+    if %w[assign random_assign].include? params[:global_actions]
+      inactive_graders = [] # used to filter the 'grader_ids' array and to throw a custom flash message error
+
+      # Solving N+1 querying problem by executing the SQL query:
+      # SELECT * FROM roles WHERE id IN (..., ..., ..., ...)
+      # (thus making one database request in the beginning instead of one for each grader id)
+      graders = Ta.where(id: grader_ids)
```
david-yz-liu, line 98: This query can be improved in a few ways:

- use an association to only get the TAs associated with the `current_course`
- add an additional argument to the `where` to only get inactive graders
- use `pluck` to get the resulting `id`s and `user_name`s in an array. This is the value that you want for `inactive_grader_ids`. This saves instantiating `Ta` objects unnecessarily.

app/controllers/graders_controller.rb
```diff
@@ -87,6 +87,44 @@ def global_actions
...
+      # The '!' indicates that it mutates the original array
+      grader_ids.reject! { |grader_id| inactive_grader_ids.include? grader_id.to_i }
+
+      # Throw red flash message error if any inactive graders exist
+      # Note that after this check we proceed with any 'active' graders (the survivors of graders_id)
+      if inactive_graders.size > 0
+        flash_now(:error,
+                  I18n.t('groups.invite_member.errors.inactive_grader',
+                         user_names: inactive_graders.map(&:user_name).join(', ')))
+        # Decided non to return a status code of 400, because active graders may still be added, up for discussion
```
david-yz-liu, line 122: Yes, you can remove this code.

app/controllers/graders_controller.rb
```diff
@@ -113,6 +151,7 @@ def global_actions
         else
           assign_all_graders(grouping_ids, grader_ids)
         end
+
```
david-yz-liu, line 154: revert this change

config/locales/views/groups/en.yml
```diff
@@ -35,6 +35,7 @@ en:
         already_pending: 'Could not invite %{user_name}: this student already has a pending invitation for this group.'
         extension_exists: You cannot modify a group with an extension to the due date. Please contact your instructor if you wish to make changes.
         group_max_reached: 'Could not invite %{user_name}: group maximum has been reached.'
+        inactive_grader: Cannot assign inactive grader(s) '%{user_names}' due to grader inactivity.
```
david-yz-liu, line 38: Delete the phrase "due to grader inactivity". It's fine to just end with the user names.

Also, omit the quotes around the grader names.

spec/controllers/graders_controller_spec.rb
```diff
@@ -346,6 +346,14 @@
           @ta1 = create(:ta)
           @ta2 = create(:ta)
           @ta3 = create(:ta)
+
+          # Making two new TA's that are 'inactive', to test additional functionality
+          @ta4 = create(:ta)
```
david-yz-liu, line 351: you can pass in `hidden: true` to avoid setting the attribute manually and then saving.

(and throughout the new tests you added)

spec/controllers/graders_controller_spec.rb
```diff
@@ -503,6 +511,41 @@
...
+                  }
+
+          expect(response).to have_http_status(200)
+          @assignment.groupings.each do |grouping|
+            expect(grouping.tas).to eq []
+          end
+        end
+        it 'and both active and inactive graders (and multiple groupings) are selected' do
+          # In this case, the active graders should be assigned to the specified groups but the inactive
+          # ones should not, and still
```
david-yz-liu, line 530: it's not clear what "and still" means here

app/controllers/graders_controller.rb
```diff
@@ -87,6 +87,22 @@ def global_actions
...
+    if %w[assign random_assign].include? params[:global_actions]
+      inactive_graders = []
+
+      inactive_graders_hash =
+        current_course.tas.joins(:user).where(roles: { hidden: true }).pluck(:id, :user_name)
+                      .map { |x| [x[0], x[1]] }.to_h
+
+      grader_ids.reject! { |grader_id| inactive_graders_hash.key? grader_id.to_i }
+
+      if inactive_graders.size > 0
```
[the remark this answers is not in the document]
mimischly7: Apologies for my oversight Professor, I was accidentally using the old object (which now is empty). I have now updated the code to use a new hash (which was necessary to show the appropriate inactive users on the flash message).

app/controllers/graders_controller.rb
```diff
@@ -87,6 +87,22 @@ def global_actions
       end
     end
 
+    if %w[assign random_assign].include? params[:global_actions]
+      inactive_graders_hash =
+        current_course.tas.joins(:user).where(roles: { hidden: true }).pluck(:id, :user_name)
+                      .map { |x| [x[0], x[1]] }.to_h
+
+      inactive_graders_for_flash = inactive_graders_hash.select { |k| grader_ids.include? k.to_s }
```
david-yz-liu, line 95: You're mixing different types here. `grader_ids` should always be a list of integers, not strings. Make sure this is true for all test cases that you've added, as well as manually in the UI.

app/controllers/graders_controller.rb
```diff
@@ -87,6 +87,22 @@ def global_actions
...
+
+      grader_ids.reject! { |grader_id| inactive_graders_hash.key? grader_id.to_i }
```
david-yz-liu, line 97: Similarly here; the `to_i` should not be necessary

spec/controllers/graders_controller_spec.rb
```diff
@@ -503,6 +507,40 @@
           expect(@grouping1.tas.size).to eq 0
           expect(@grouping2.tas.size).to eq 0
         end
+        it 'and only inactive graders (and multiple groupings) are selected' do
+          post_as @instructor,
+                  :global_actions,
+                  params: {
+                    course_id: course.id, assignment_id: @assignment.id, global_actions: 'random_assign',
+                    groupings: [@grouping1, @grouping2, @grouping3], graders: [@ta4, @ta5],
```
david-yz-liu, line 515: See my above comment about the `graders` parameter (and check this throughout all test cases)

PR #6835, opened 2023-11-09
Fix issues with LTI launch while logged in and course creation

pretendWhale:
## Motivation and Context

Two bugs were discovered with lti functionality:

1. when launching LTI from canvas while logged in to MarkUs, users were not properly redirected to the `choose_course` path
2. When attempting to create a new course, an uncaught error would occur if a course with the same name is already in the markus database.

## Your Changes

**Description**:
1. modified the `login` method to properly redirect logged-in users.
2. Catch the case when the course already exists and display an error message if the user attempts to create a new course.

## Testing

Wrote tests for the changed methods, and tested in the web interface.

## Checklist


donny-wong (approved):

app/controllers/lti_deployments_controller.rb
```diff
@@ -180,11 +180,18 @@ def check_host
   end
 
   def create_course
-    new_course = Course.create!(name: params['name'], display_name: params['display_name'], is_hidden: true)
+    new_course = Course.find_or_initialize_by(name: params['name'], display_name: params['display_name'])
```
david-yz-liu, line 183: It is only the course `name` that is unique. You should first do a search just on `name`, and then (if a new record) update both `display_name` and `is_hidden`

app/controllers/lti_deployments_controller.rb
```diff
@@ -180,11 +180,18 @@ def check_host
...
+    unless new_course.new_record?
+      flash_message(:error, I18n.t('lti.course_exists'))
+      redirect_to choose_course_lti_deployment_path
+      return
+    end
+    new_course.update(is_hidden: true)
+    new_course.save!
```
david-yz-liu, line 190: `update` already saves. If you want to raise an error, use `.update!`

app/controllers/main_controller.rb
```diff
@@ -23,7 +23,11 @@ class MainController < ApplicationController
   def login
     # redirect to main page if user is already logged in.
     if logged_in? && !request.post?
-      if @real_user.admin_user?
+      if cookies.encrypted[:lti_data].present?
+        lti_data = JSON.parse(cookies.encrypted[:lti_data]).symbolize_keys
+        redirect_url = lti_data.key?(:lti_redirect) ? lti_data[:lti_redirect] : root_url
```
david-yz-liu, line 28: use [`Hash#fetch`](https://ruby-doc.org/core-2.5.1/Hash.html#method-i-fetch)

david-yz-liu (approved):

PR #6871, opened 2023-12-06
add plotly library and dependencies to notebook viewer

pretendWhale:
## Motivation and Context

This re-enables rendering graphics with `plotly` by adding plotly and its dependencies to the `notebook_viewer`

## Your Changes

**Description**:
Added current `plotly.js` and `require.js` libraries under `vendor/app/assets`, as well as including them in `notebook.html.erb`.

Note: using the default renderer for plotly in a jupyter notebook, plotly will include an entire copy of itself in the notebook. So if notebooks use default behaviors, we only need `require.js` to render a  notebook that uses plotly. I've included plotly in markus for completion in the case that a non-default renderer is used. This does cause a JS console error, but the plots still render.

## Testing

Manual testing to ensure that plotly rendered, both in the simple case and a more complex case from an actual MarkUs assignment. As this is all frontend/rendering, I did not add any tests.

I added the libraries under `vendor/assets/javascript` in their unminified form, but I wasn't sure if they should go there or in `app/assets`.

Plotly is about 4MB, so a notebook using the default renderer for plotly will be fairly large. I don't think that this is a problem, but it is notably larger than a notebook without plotly.

## Checklist


david-yz-liu (commented):
@pretendWhale this is good, but please use the minified version of these libraries

david-yz-liu (approved):

PR #6912, opened 2024-01-25
Allow websocket connections with external auth

pretendWhale:
## Motivation and Context

Currently, websocket connections fail when users are logged in via external auth (i.e. UTORAuth). This is because `current_user` is not added to user sessions when logged in via external auth.

## Your Changes

**Description**:
Added a check for remote authorization when attempting to create a websocket connection.

## Testing

Tested that when logged in via external auth websocket connections work in the web interface. Added a test for the case that a user is logged in via external auth.

I think the order of the checks is right - as it stands, we check for the remote auth first, then override with current user if it exists (which should catch role-switched users), then falling back to `real_user_name` if `current_user` is not set by either previous check. But I'd appreciate a second opinion.

## Checklist

### DIFF
--- a/app/channels/application_cable/connection.rb
+++ b/app/channels/application_cable/connection.rb
@@ -2,6 +2,9 @@ module ApplicationCable
   class Connection < ActionCable::Connection::Base
     identified_by :current_user
     def connect
+      if request.session[:auth_type] == 'remote'
+        self.current_user = User.find_by user_name: request.env['HTTP_X_FORWARDED_USER']
+      end
       unless request.session[:user_name].nil?
         self.current_user = User.find_by user_name: request.session[:user_name]
       end
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [unreleased]
+- Fix websocket connection when logged in with remote auth (#6912)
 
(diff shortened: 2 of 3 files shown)


app/channels/application_cable/connection.rb
```diff
@@ -2,6 +2,9 @@ module ApplicationCable
   class Connection < ActionCable::Connection::Base
     identified_by :current_user
     def connect
+      if request.session[:auth_type] == 'remote'
```
david-yz-liu, line 5: Very nice. In comparing this implementation with `MainController#login`, I note that the latter uses a `remote_auth?` helper, which checks another session key. I suspect it would be prudent to do the same here?

pretendWhale: Hi @david-yz-liu - thanks for the review! Sessions are not available in cable (hence the use of `request.session`), so we can't directly re-use `remote_auth?`. We do check the same session key here as in that method. Would you be alright with keeping it this way for simplicity, or is there another approach we should be thinking of?

david-yz-liu: @pretendWhale oh sorry you're totally right, I missed that even though I was looking at exactly this line. All good :+1:

david-yz-liu (approved):

PR #6915, opened 2024-01-26
Enable the deletion of Grade Entry Forms that have no grades

Bruce-8:
## Motivation and Context

Currently, there is no `destroy` option for marks spreadsheets (aka grade entry forms). This pull request adds both a web UI feature and new API routes for the deletion of marks spreadsheets, under the condition that there are no non-nil grades for that marks spreadsheet.

## Your Changes

**Description**:
- Added a new `destroy` action method to `grade_entry_forms_controller` for both the regular and API version.
- Modified `routes.rb` to include routes for the added action methods.
- Added a new HTML/ERB button that allows users to delete a grade entry form in the web UI.

## Testing

- Manually tested the button feature in the web UI.
- Added Rspec unit tests for valid http responses, successful deletion, and other forms of correctness for this feature. 

## Checklist

- [X] I have updated the Changelog.md file.


Changelog.md
```diff
@@ -3,6 +3,7 @@
 ## [unreleased]
 - Add workaround for CSP in Safari < 16 (#6947)
 - Improve add button labels on autotest settings page (#6948)
+- Enable the deletion of Grade Entry Forms (#6915)
```
david-yz-liu, line 6: be more specific here (grade entry forms that have no grades)

config/locales/views/grade_entry_forms/en.yml
```diff
@@ -2,6 +2,8 @@
 en:
   grade_entry_forms:
     add_column: Add a column
+    successfully_deleted: Marks Spreadsheet was successfully deleted.
+    delete: Delete
```
david-yz-liu, line 6: There should already be text defined for this under `common`, make sure to use that

app/views/grade_entry_forms/_form.html.erb
```diff
@@ -87,5 +87,10 @@
     <%= f.submit (f.object.new_record? ? t('helpers.submit.create', model: GradeEntryForm.model_name.human) :
                                          t('helpers.submit.update', model: GradeEntryForm.model_name.human)),
                  data: { disable_with: t('working') } %>
+    <% if @grade_entry_form.id != nil and @grade_entry_form.count_non_nil == 0 %>
```
david-yz-liu, line 90: Touch base with @mimischly7 about both how the button should be styled and using the `disabled` attribute

app/controllers/grade_entry_forms_controller.rb
```diff
@@ -299,4 +299,11 @@ def switch
       redirect_to student_interface_course_grade_entry_form_path(current_course, params[:id])
     end
   end
+
+  def destroy
+    @grade_entry_form = record
+    @grade_entry_form.destroy
```
david-yz-liu, line 305: You should check whether the deletion was successful or not, and flash an error message if the grade entry form could not be deleted.

You can test this case in the UI by making the Delete button (temporarily) appear even when it shouldn't, and then clicking on it.

spec/controllers/api/grade_entry_forms_controller_spec.rb
```diff
@@ -313,5 +313,37 @@
         expect(grade_entry_form.grade_entry_students.find_by(role: student).grades.first.grade).to eq(5)
       end
     end
+    context 'DELETE Destroy' do
+      it 'does not delete a non-existing assignment' do
```
david-yz-liu, line 317: should be "grade entry form" here (same in the non-api controller tests)

app/models/grade_entry_item.rb
```diff
@@ -5,7 +5,7 @@ class GradeEntryItem < ApplicationRecord
 
   has_one :course, through: :grade_entry_form
 
-  has_many :grades, dependent: :delete_all
+  has_many :grades, dependent: :destroy
```
david-yz-liu, line 8: Okay I realized after reviewing again that this change will also change what happens when the user tries to destroy a single column (on the grade entry form "update" page). Previously, we actually did allow the user to delete the column even if that would result in non-nil grades being deleted.

I don't want to change that existing behaviour in this PR. So, please revert the added `Grade` destroy callback and this line, and instead add a callback to `GradeEntryForm` itself to check for non-nil grades. See also my comment on the form below about how I want you to check for this.

Sorry for the change of plans!

app/views/grade_entry_forms/_form.html.erb
```diff
@@ -89,3 +89,12 @@
                  data: { disable_with: t('working') } %>
   </p>
 <% end %>
+
+<% if @grade_entry_form.id != nil %>
```
david-yz-liu, line 93: use `.nil?` and an `unless`

app/views/grade_entry_forms/_form.html.erb
```diff
@@ -89,3 +89,12 @@
...
+  <% deletable = @grade_entry_form.count_non_nil == 0 %>
```
david-yz-liu, line 94: While I like that you tried to use an existing method, that particular method does a lot more work than necessary. I encourage you to use an association and a query directly; note that `@grade_entry_form` already has a `grades` association, and you can filter out the non-`nil` grades, and use `exists?` to check whether there are any results. You can do this both for this disabling check and for the actual `before_destroy` callback itself (see my above comment).

app/views/grade_entry_forms/_form.html.erb
```diff
@@ -89,3 +89,12 @@
...
+  <%= button_to t(:delete),
+                course_grade_entry_form_path(@current_course.id, @grade_entry_form.id),
+                method: :delete,
+                disabled: !deletable
```
david-yz-liu, line 98: This is good, but please also add a tooltip with a message explains why the button is disabled (see Mimis' PR for assignment button deletion).

app/views/grade_entry_forms/_form.html.erb
```diff
@@ -89,3 +89,13 @@
                  data: { disable_with: t('working') } %>
   </p>
 <% end %>
+
+<% unless @grade_entry_form.id.nil? %>
```
david-yz-liu, line 93: Oh after you modified this I realized exactly what you were checking; use the `new_record?` method instead to check whether a model has been saved or not.

david-yz-liu (commented):
@Bruce-8 good work. I just left one more comment.

Regarding the failing test: I don't believe it's related to your pull request, but you should be able to fix it here by passing `ignore_query: true` to the  `have_current_path` method. Please do that in this PR.

david-yz-liu (approved):

PR #6950, opened 2024-02-07
fix grade_distribution returning multiple entries per TA

pretendWhale:
## Motivation and Context

Currently, the `grade_distribution` method calls `assignment.tas`. This is through `ta_memberships` and therefore returns individual TAs multiple times if they are assigned to grade more than one group. This causes the assignment summary's `grader distribution` graph to have multiple entries per grader, sometimes causing the display of only one grader.

## Your Changes

**Description**:
Updated `grader_distribution` to call `uniq` on the list of TAs to ensure only a single entry for each TA.

## Testing

Tested in the web interface, and added tests to ensure the content and length of data returned by `grade_distribution` are correct. 

We could put a uniqueness scope on the assignment.tas method, but I don't know if assignment.tas is called in other places where the multiple entries are required.

## Checklist


david-yz-liu (commented):
@pretendWhale thanks! This is my bad. The bug was introduced in #6764.

Please modify the association to only return unique records. This is the intended behaviour of the association in all cases.

donny-wong (approved):

david-yz-liu (approved):

Issue #5961, opened 2022-04-20
Marks Spreadsheet Upload sensitive to filename


treblacy:
For a marks spreadsheet, in the Grades tab, when using the Upload function:

If I have a file with filename "h.csv" and its content conforms to the required file format, uploading succeeds and I can see the data showing up.

If the filename is "h", same content, just a renaming, it's a "The selected file was improperly formed".

Ugh is this intended?

I'm at U of Toronto Scarborough, MarkUs version 1.14.0.DEV

mishaschwartz:
Thanks for the bug report!

Markus relies on the file extension to determine how to handle the upload since we support multiple file types. In this case though I believe the only acceptable file type is `.csv` so we can probably safely just assume that all uploaded files will be csvs. 

For now, please continue naming your files with file extensions and we'll figure out the best way to handle this in a future version.

david-yz-liu:
Closed in #6961

PR #6961, opened 2024-02-23
Allow spreadsheet uploads to not be sensitive to file extension

AinaMerch:
## Motivation and Context

To solve issue 5961 and allow spreadsheet uploads to not be sensitive to file extension

https://github.com/MarkUsProject/Markus/issues/5961

## Your Changes

**Description**:
Modified the process_file_upload method to allow the caller to specify which of the two file types are actually allowed, and in cases where there's only one type allowed, we don't check the file extension
Modified all of the places where process_file_upload is called to take advantage of this change. 
Modify the front-end file inputs by using the accept attribute to encourage the user to upload files with the expected extensions, matching the controller action.

- [X] Bug fix (non-breaking change which fixes an issue)

## Testing

Manual testing through trying to upload different kinds of files with and without extensions

## Checklist

- [X] I have verified that the CI tests have passed. 
- [X] I have reviewed the test coverage changes reported on Coveralls.


app/helpers/upload_helper.rb
```diff
@@ -1,14 +1,19 @@
...
+  def process_file_upload(allowed_filetypes = %w[.csv .yaml .yml])
     encoding = params[:encoding] || 'UTF-8'
     upload_file = params.require(:upload_file)
 
     if upload_file.size == 0
       raise StandardError, I18n.t('upload_errors.blank')
     end
 
-    filetype = File.extname(upload_file.original_filename)
+    if allowed_filetypes.size == 1
```
david-yz-liu, line 11: This check is a little awkward because there are two possible extensions for YAML. This code will work okay if you always expect users to pass `'.yml'` (but not `'.yaml'`). You can communicate this by adding to this method docstring and by removing `'.yaml'` from the default argument value

app/controllers/admin/users_controller.rb
```diff
@@ -37,16 +37,14 @@ def update
...
         flash_message(:error, e.message)
       else
-        if data[:type] == '.csv'
-          @current_job = UploadUsersJob.perform_later(EndUser,
-                                                      params[:upload_file].read,
-                                                      params[:encoding])
-          session[:job_id] = @current_job.job_id
-        end
+        @current_job = UploadUsersJob.perform_later(EndUser,
+                                                    params[:upload_file].read,
```
david-yz-liu, line 45: You've correctly identified a potential problem with `.read` calls when given a binary file (or just a file with a different text encoding, but that's a separate matter).

Looking at the implementation of `process_file_upload`, a key difference between `csv` and `yml` handling is that in the latter, there's a `read` call already made, with appropriate error handling if the `read` fails (presumably---you can test this). The former doesn't do this; I think this decision was originally made because it's more memory-efficient to parse a CSV file line-by-line rather than first reading the entire file into memory first (which `.read` does). However, this then makes error handling a bit more challenging, as it becomes spread out throughout the codebase (or just doesn't happen, e.g. here).

So instead, you can make the following change to `process_file_upload`: modify the CSV handling to call `read` on the uploaded file (mimic the YML case), and store the contents in the returned dict as `:contents`, similar again to the YML case. Then modify all places where CSV data is used to instead use `data[:content]` instead. This will also involve more cascading changes across various `from_csv` (and possibly other) methods, but at the end will result in better and more consistent error handling!

app/controllers/grade_entry_forms_controller.rb
```diff
@@ -216,19 +216,15 @@ def download
...
-        grades_file = data[:file]
-        result = @grade_entry_form.from_csv(grades_file.read, overwrite)
-        flash_message(:error, result[:invalid_lines]) unless result[:invalid_lines].empty?
-        flash_message(:success, result[:valid_lines]) unless result[:valid_lines].empty?
-      end
+      overwrite = params[:overwrite]
+      grades_file = data[:file]
+      result = @grade_entry_form.from_csv(grades_file.read, overwrite)
+      flash_message(:error, result[:invalid_lines]) unless result[:invalid_lines].empty?
+      flash_message(:success, result[:valid_lines]) unless result[:valid_lines].empty?
```
david-yz-liu, line 227: These two lines are duplicated across multiple CSV upload controller actions. This is a good opportunity to define a helper method like `flash_csv_result` and use that consistently throughout the various controllers. (This wasn't part of the original plan for this pull request, but because it's a relatively simple change and falls under the broader goal of "make the handling of controller upload actions more consistent", I'm happy to add it to this PR.)

david-yz-liu (commented):
@AinaMerch @pranavrao145 great work. You are definitely on the right track. I've left a few pointers for things you can work on, but overall I think you're pretty close in terms of code changes, and then of course the thorough testing.

On that note, when testing your changes it's good to try a valid plaintext file with no extension (like the original issue described), but the other thing you should be checking is taking a binary file (e.g. a PDF), renaming it to change the file extension to (e.g.) `.csv`, and then uploading it to each of the different controllers actions. There should not be any errors, but there *should* be flash messages indicating a problem in reading the file.

app/helpers/upload_helper.rb
```diff
@@ -1,18 +1,23 @@
...
+      filetype = allowed_filetypes[0]
+    else
+      filetype = File.extname(upload_file.original_filename)
+    end
+
     if filetype == '.csv'
       {
         type: '.csv',
-        file: upload_file,
+        contents: upload_file.read,
```
david-yz-liu, line 20: This is a good start, but take a look at how the `read` call happens in the YML case. This encoding conversion is very useful for raising an error when given binary data here, rather than when the `contents` are used later.

app/controllers/admin/users_controller.rb
```diff
@@ -37,16 +37,13 @@ def update
 
     def upload
       begin
-        data = process_file_upload
+        data = process_file_upload(['.csv'])
+        @current_job = UploadUsersJob.perform_later(EndUser,
```
david-yz-liu, line 41: See my comment in the implementation of `process_file_upload`. After you do that, I think you'll be able to move this code back into the `else` branch that you originally had.

app/helpers/application_helper.rb
```diff
@@ -38,6 +38,11 @@ def flash_allowance(type, result, flash_type = flash, most_specific: false)
     result
   end
 
+  def flash_csv_result(result)
```
david-yz-liu, line 41: Since this is associated just with CSV processing, a better location for this method is under the `MarkusCsv` class

app/controllers/students_controller.rb
```diff
@@ -106,19 +106,14 @@ def download
 
   def upload
     begin
-      data = process_file_upload
-    rescue Psych::SyntaxError => e
-      flash_message(:error, t('upload_errors.syntax_error', error: e.to_s))
+      data = process_file_upload(['.csv'])
+      @current_job = UploadRolesJob.perform_later(Student,
```
david-yz-liu, line 110: See my above comment about keeping the `else` branch

app/controllers/tas_controller.rb
```diff
@@ -62,16 +62,11 @@ def download
 
   def upload
     begin
-      data = process_file_upload
-    rescue Psych::SyntaxError => e
-      flash_message(:error, t('upload_errors.syntax_error', error: e.to_s))
+      data = process_file_upload(['.csv'])
+      @current_job = UploadRolesJob.perform_later(Ta, current_course, data[:contents], data[:encoding])
```
david-yz-liu, line 66: see above comment

spec/controllers/graders_controller_spec.rb
```diff
@@ -57,38 +57,48 @@
         #           Group 3,c7benjam
         @group_grader_map_file = fixture_file_upload('group_csvs/group_grader_map.csv')
       end
+      ['.csv', '.pdf', ''].each do |extension|
+        ext_string = extension.empty? ? 'none' : extension
+        it "and all graders and groups are valid in a file with extension #{ext_string}" do
+          file = if extension.empty?
```
david-yz-liu, line 63: the `if` isn't necessary, if `extension` is empty the filename on line 68 will still be `group_grader_map`

spec/controllers/students_controller_spec.rb
```diff
@@ -43,11 +43,15 @@
...
       end
 
-      it 'calls perform_later on a background job' do
-        expect(UploadRolesJob).to receive(:perform_later).and_return OpenStruct.new(job_id: 1)
-        post_as instructor,
-                :upload,
-                params: { course_id: course.id, upload_file: fixture_file_upload('students/form_good.csv', 'text/csv') }
+      ['.csv', '', '.pdf'].each do |extension|
+        ext_string = extension.empty? ? 'none' : extension
+        it "calls perform_later on a background job on a valid file with extension #{ext_string}" do
```
david-yz-liu, line 48: oh I realized here (and in all other descriptions) that you should use `"#{ext_string}"` rather than just `#{ext_string}`, so that the "empty string" extension doesn't make the test description look weird

david-yz-liu (commented):
@AinaMerch and @pranavrao145 great job! I just left a few comments

david-yz-liu (approved):

PR #6963, opened 2024-02-23
disallow .git folder uploads

donny-wong:
## Motivation and Context
disallow .git folder repo uploads from students. 

## Your Changes
I check if the folder being uploaded has the name ".git", if it does it will be invalidated.

## Testing
Both manual tests and unit testing. For manual test, from the Student repo UI, I uploaded a zip file that contained a folder with a .git folder in it. I also tested by directly via UI create a folder .git. 

## Checklist


app/helpers/repository_helper.rb
```diff
@@ -136,6 +136,13 @@ def add_folder(folder_path, user, repo, path: '/', txn: nil, required_files: nil
     # check if only required files are allowed for a submission
     # allowed folders = paths in required files
     if required_files.present? && required_files.none? { |file| file.starts_with?(folder_path) }
+      folder_path = format_folder_path folder_path
```
david-yz-liu, line 139: What's going on with `format_folder_path`? What is it removing?

app/helpers/repository_helper.rb
```diff
@@ -136,6 +136,13 @@ def add_folder(folder_path, user, repo, path: '/', txn: nil, required_files: nil
...
+      messages << [:invalid_folder_name, folder_path]
+      return false, messages
+    end
+
+    if folder_path.include?('/.git/')
```
david-yz-liu, line 144: Use [`each_filename`](https://ruby-doc.org/3.3.0/exts/pathname/Pathname.html#method-i-each_filename) for more precise checking for a `.git` segment

david-yz-liu (commented):
@donny-wong good work. I left a few comments, but it seems to me like another problem is if a student uploads a *file* called `.git`, not just a folder. Can you test for that and then handle as necessary?

app/helpers/repository_helper.rb
donny-wong: Thanks David, as an error message it would display the pathname variable which includes the internal storage directories. e.g. for assignment 'A0', it will include it in the pathname "/A0/subdir/...". We prefer to hide that information from the end user. That is what format_folder_path does.

PR #6989, opened 2024-03-01
association-improvement

mimischly7:
Added `dependent: :destroy` to all the associations that needed it (in the `Assignment` and `Assessment` models), changed the order of associations in Assignment to make groups-and-groupings first (this is the first thing  that should be checked when deleting an assignment), created new locale string for use in the `AnnotationCategory` and `Assignment` models, and added another rescue block in the `#destroy` action of assignments controller to handle other potential complications that could arise in assignment deletion. This last one is temporary, since we would ideally like to perform a more targetted handling of different exceptions (but for now this prevents crashing and gives an idea of the problem).

Also, wrote thorough RSPEC tests for `assignments_controller` to ensure that, upon assignment destruction, all associated entities are also destroyed appropriately, and also tested for the case that (the attempt of) destroying an Assignment raises a `StandardError`.

## Your Changes

**Description**:

## Testing

`RSPEC` (in `annotations_controller_spec`)

## Checklist

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -668,7 +668,7 @@ class AssignmentsController < ApplicationController
       flash_message(:error, I18n.t('assignments.assignment_has_groupings'))
       redirect_back fallback_location: { action: :edit, id: @assignment.id }
     rescue StandardError => e
-      flash_message(:error, "Problem with deletion: #{e.message}")
+      flash_message(:error, I18n.t('activerecord.errors.models.assignment_deletion', problem_message: e.message))
       redirect_back fallback_location: { action: :edit, id: @assignment.id }
     end
   end
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -667,6 +667,9 @@ class AssignmentsController < ApplicationController
     rescue ActiveRecord::DeleteRestrictionError
       flash_message(:error, I18n.t('assignments.assignment_has_groupings'))
       redirect_back fallback_location: { action: :edit, id: @assignment.id }
+    rescue StandardError => e
+      flash_message(:error, "Problem with deletion: #{e.message}")
+      redirect_back fallback_location: { action: :edit, id: @assignment.id }
     end
   end
 
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -72,18 +72,17 @@ class Assignment < Assessment
           dependent: :destroy,
           foreign_key: :parent_assessment_id,
           inverse_of: :parent_assignment
-  has_many :peer_reviews, dependent: :destroy, through: :groupings
-  has_many :pr_peer_reviews, through: :parent_assignment, source: :peer_reviews, dependent: :destroy
+  has_many :peer_reviews, through: :groupings
+  has_many :pr_peer_reviews, through: :parent_assignment, source: :peer_reviews
 
   has_many :current_submissions_used,
            through: :groupings,
-           source: :current_submission_used,
-           dependent: :destroy
+           source: :current_submission_used
 
-  has_many :ta_memberships, through: :groupings, dependent: :destroy
-  has_many :student_memberships, through: :groupings, dependent: :destroy
+  has_many :ta_memberships, through: :groupings
+  has_many :student_memberships, through: :groupings
 
-  has_many :submissions, through: :groupings, dependent: :destroy
+  has_many :submissions, through: :groupings
 
   has_many :notes, as: :noteable, dependent: :destroy
 
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -8,6 +8,10 @@ class Assignment < Assessment
 
   validates :due_date, presence: true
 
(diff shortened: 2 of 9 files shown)


app/models/assignment.rb
```diff
@@ -57,32 +63,35 @@ class Assignment < Assessment
...
+          inverse_of: :parent_assignment
+  has_many :peer_reviews, dependent: :destroy, through: :groupings
+  has_many :pr_peer_reviews, through: :parent_assignment, source: :peer_reviews, dependent: :destroy
 
-  has_many :current_submissions_used, through: :groupings,
-                                      source: :current_submission_used
+  has_many :current_submissions_used,
+           through: :groupings,
+           source: :current_submission_used,
+           dependent: :destroy
```
david-yz-liu, line 81: Don't add this to `through` associations (it's either redundant or misleading: in this case, for example, `groupings` can't be destroyed, and so `current_submissions_used` will also never be destroyed)

spec/controllers/assignments_controller_spec.rb
```diff
@@ -2065,5 +2075,28 @@
       expect(flash.to_hash.length).to eq(1)
       expect(response).to have_http_status(302)
     end
+    it 'should remove associated entities on destroy' do
+      # These were lazily initialized, so in order to actually be created we need to "use" them
+      checkbox_criterion
```
david-yz-liu, line 2080: Create a separate test case for each of these

app/controllers/assignments_controller.rb
```diff
@@ -667,6 +667,9 @@ def destroy
     rescue ActiveRecord::DeleteRestrictionError
       flash_message(:error, I18n.t('assignments.assignment_has_groupings'))
       redirect_back fallback_location: { action: :edit, id: @assignment.id }
+    rescue StandardError => e
+      flash_message(:error, "Problem with deletion: #{e.message}")
```
david-yz-liu, line 671: make sure to add a test case that covers these two lines of code

Also, make sure to internationalize this string

app/models/assignment.rb
```diff
@@ -57,32 +63,34 @@ class Assignment < Assessment
...
-  has_many :groups, through: :groupings, dependent: :restrict_with_exception
 
   has_many :notes, as: :noteable, dependent: :destroy
 
   has_many :exam_templates, dependent: :destroy, inverse_of: :assignment, foreign_key: :assessment_id
 
   has_many :starter_file_groups, dependent: :destroy, inverse_of: :assignment, foreign_key: :assessment_id
 
-  has_many :tas, -> { distinct }, through: :ta_memberships, source: :role
+  has_many :tas, -> { distinct }, through: :ta_memberships, source: :role, dependent: :destroy
```
david-yz-liu, line 93: There shouldn't be a `dependent: :destroy` here

app/models/assignment.rb
```diff
@@ -57,32 +63,34 @@ class Assignment < Assessment
...
+          class_name: 'Assignment',
+          dependent: :destroy,
+          foreign_key: :parent_assessment_id,
+          inverse_of: :parent_assignment
   has_many :peer_reviews, through: :groupings
   has_many :pr_peer_reviews, through: :parent_assignment, source: :peer_reviews
 
-  has_many :current_submissions_used, through: :groupings,
-                                      source: :current_submission_used
+  has_many :current_submissions_used,
```
david-yz-liu, line 78: you can revert the changes here (they are now purely stylistic since you removed the `dependent: :destroy`

spec/controllers/assignments_controller_spec.rb
```diff
@@ -2065,5 +2068,39 @@
...
     end
+
+    shared_examples 'handling associated entities upon destroy' do |entity|
+      it "should remove associated #{entity}" do
+        # NOTE: the next line assume that an `assignment` is sufficient for the factory of `entity`
+        assoc_entity = create entity, assignment: assignment
+        # Deleting the assignment - should be successful since there are not groupings
+        delete_as instructor, :destroy, params: { course_id: course.id, id: assignment.id }
+        expect(Assignment.exists?(assignment.id)).to be(false)
+        # Ensure that the associated checkbox_criterion was also removed
```
david-yz-liu, line 2079: "checkbox_criterion" -> "entity"

spec/controllers/assignments_controller_spec.rb
```diff
@@ -2065,5 +2068,39 @@
       expect(flash.to_hash.length).to eq(1)
       expect(response).to have_http_status(302)
     end
+
+    shared_examples 'handling associated entities upon destroy' do |entity|
+      it "should remove associated #{entity}" do
+        # NOTE: the next line assume that an `assignment` is sufficient for the factory of `entity`
```
david-yz-liu, line 2074: "assume" -> "assumes"

spec/controllers/assignments_controller_spec.rb
```diff
@@ -2065,5 +2068,39 @@
...
+        assoc_entity = create entity, assignment: assignment
+        # Deleting the assignment - should be successful since there are not groupings
```
david-yz-liu, line 2076: "not" -> "no"

david-yz-liu (approved):

PR #6997, opened 2024-03-22
Handle pagination in student roster LTI requests

pretendWhale:
## Motivation and Context

Canvas uses a links header similar to [github](https://docs.github.com/en/rest/using-the-rest-api/using-pagination-in-the-rest-api?apiVersion=2022-11-28) to provide pagination information. we were not correctly handling paginated data so this PR fixes that behavior.

Additionally, this PR allows instructors to be synced via the sync function. Added a checkbox on the modal for instructor syncing, which is checked by default.

## Your Changes

**Description**:
Moved acquiring roster information to its own method, and then repeatedly calls it as long as there is a `next` link in the headers.

- [x] New Feature 

## Testing

Tested in browser by forcing canvas pagination (using `?per_page=1` as a GET param). Also added tests to ensure that when paginated data is present, all data is retrieved correctly from each page and appropriate records are created. Also tested syncing instructors in browser.

## Checklist


donny-wong (changes requested):
The links are parsed incorrectly. Please have this corrected. Thank you.

donny-wong (approved):

david-yz-liu (approved):

PR #6778, opened 2023-10-20
Allow inactive groups in the graders table to be toggled for display

Bruce-8:
## Motivation and Context

Currently on the assignment groups tab, it is possible to hide students and groups that are inactive. This pull request adds the same kind of filtering to the graders tab, to toggle whether to display inactive groups on the right-hand pane.

## Your Changes

**Description**: 
Front-end:
- Modified `graders_manager.jsx` to allow for `RawGroupsTable` to display inactive groups when the icon `Display inactive groups` is clicked.  Created a `Display inactive groups` icon on the right-hand pane of the web interface. 
- Updated the hidden tooltip for the `Display inactive groups` icon to include the amount of hidden groups.

Back-end:
- Modified the `index` action of `graders_controller.rb` to include the `members` attribute of `groups` in the grader data so the front-end is able to determine whether a group is inactive or not based on its members (students).

**Type of change** (select all that apply):

- [X] New feature (non-breaking change which adds functionality)

## Testing

- Manually tested changes through the web interface.

## Checklist

- [X] I have performed a self-review of my own code.
- [x] I have verified that the pre-commit.ci checks have passed. 
- [x] I have verified that the CI tests have passed. 
- [x] I have reviewed the test coverage changes reported on Coveralls. 
- [X] I have updated the Changelog.md file.

PR #7000, opened 2024-03-26
Hide Inactive Groups in Submission Table

pranavrao145:
## Motivation and Context
Similar to #6778, this PR hides inactive groups in the Submissions table by default and adds a toggle to show inactive groups if needed.

## Your Changes
* Made Submissions table hide inactive groups by default
* Added a toggle to allow Instructors and TAs to see inactive groups if they wish
* Minor bug fixes in relevant components, as revealed by attempting to write unit tests for them

## Testing

* Amended existing RSpec to account for minor change in the information sent by the server on a specific request.
* Wrote a Jest test suite for the new front-end check box functionality

## Checklist

### DIFF
--- a/app/assets/javascripts/Components/submission_table.jsx
+++ b/app/assets/javascripts/Components/submission_table.jsx
@@ -26,6 +26,8 @@ class RawSubmissionTable extends React.Component {
       showReleaseUrlsModal: false,
       marking_states: markingStates,
       markingStateFilter: "all",
+      inactiveGroupsCount: 0,
+      filtered: [],
     };
   }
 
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/Components/submission_table.jsx
```diff
@@ -50,25 +52,42 @@ class RawSubmissionTable extends React.Component {
...
       });
   };
 
   onFilteredChange = (filtered, column) => {
     const summaryTable = this.checkboxTable.getWrappedInstance();
-    if (column.id != "marking_state") {
-      const markingStates = getMarkingStates(summaryTable.state.sortedData);
-      this.setState({marking_states: markingStates});
-    } else {
+    if (column.id == "marking_state") {
```
david-yz-liu, line 82: Don't change this, it's not related to the changes you're making

app/assets/javascripts/Components/submission_table.jsx
```diff
@@ -412,6 +437,21 @@ class RawSubmissionTable extends React.Component {
     );
   };
 
+  toggleShowInactiveGroups = event => {
+    // depends if the function is called organically or not
```
david-yz-liu, line 441: This isn't good. If you are having trouble triggering this with testing, you should first switch to using react-testing-library over enzyme

app/assets/javascripts/Components/submission_table.jsx
pranavrao145: Whoops, this was from a previous attempt to implement the functionality, will change.

app/assets/javascripts/Components/submission_table.jsx
pranavrao145: This was actually because this functionality needs to be triggered in the `fetchData` function above and I didn't want to duplicate the code, but `event` is not defined when you call it manually so I passed `null`; hence the need for a check like this. Do you have any suggestions on how I can do this cleanly without having to resort to this type of check?

david-yz-liu: @pranavrao145 ah, great. In this case make it more explicit by modifying the function to take in a boolean, and then make sure to pass in the appropriate argument where it's being used.

app/models/assignment.rb
```diff
@@ -1235,7 +1235,7 @@ def current_submission_data(current_role)
         end
       end
 
-      base[:members] = member_info.pluck('users.user_name') unless member_info.nil?
+      base[:members] = member_info.pluck('users.user_name', 'roles.hidden') unless member_info.nil?
```
david-yz-liu, line 1238: Currently `members` is `nil` if there are no members. This leads to a runtime error when there's a group with no members. You can change this to set `members` to be an empty list when there's no members.

app/assets/javascripts/Components/submission_table.jsx
```diff
@@ -412,6 +437,28 @@ class RawSubmissionTable extends React.Component {
     );
   };
 
+  toggleShowInactiveGroups = (event, init) => {
```
david-yz-liu, line 440: Okay this is better but not exactly what I meant. Don't pass in `event` or `init`, just pass in a boolean (either `true` or `false`).

During initialization you can pass in `false`; for the onclick/onchange, pass in the value of `event.target.checked`

pranavrao145: Ohh that makes a lot of sense! Will change.

PR #7003, opened 2024-03-29
Handle case where no test groups are runnable by students

mimischly7:
## Motivation and Context

In the autotest-settings page (in the instructor view), the instructor has the ability to make all test groups NOT-runnable by students. Currently, when a student presses the "Run Tests" button, a request is still sent to the auto-testing server and 0/0 is shown on the website. We want to change this so that, if no test-groups are student-runnable, we avoid the unnecessary request to the autotest server and present an appropriate message on the screen.

Screenshot 2024-03-29 at 2 18 56 AM

## Your Changes

**Description**:
An `Assignment` object has a `test_groups` association (corresponding to a `test_groups` table in the postgres database) holding information about the autotest-settings set by the instructor for that assignment, including who can run the tests, which is an array equal to one of the four following: `["instructor", "student"]`, `["instructor"]`, `["student"]`, `[]`.
In the `execute_test_run` action of the `automated_tests_controller`, we examine the `test_groups` attribute of the current assignment and if there is no test-group runnable by a student, we flash an appropriate message and return early.

## Testing

I had to modify the current tests because this pull request does change the behavior of automated testing. In particular, if none of the test-groups are runnable by students, now the test-execution job does not get enqueued (which was the whole point, since if it gets enqueued to `AutotestRunJob` then an unnecessary request will be made to the autotest server). But this made some of the existing tests for `execute_test_run` (in `automated_test_controller.rb`) fail. This is because the factory-produced assignment object used in these tests had the `test_groups` attribute/association uninitialized (due to how an `:assignment `is spawned by a factory). To test the new behavior, instead of generating a simple `:assignment`, the tests now use an `:assignment_with_criteria_and_test_results`, and in the tests with context "when the student is allowed to run tests", I had to manually update the `test_groups` attribute of the assignment to specify that the `test_group` is runnable by students. Overall, we are now testing the case where none of the test groups are runnable by students, so no job should be enqueued, and the case where at least one test-group is student-runnable, in which case the job should be enqueued as usual.

## Checklist


app/controllers/automated_tests_controller.rb
```diff
@@ -48,6 +48,16 @@ def student_interface
 
   def execute_test_run
     assignment = Assignment.find(params[:assignment_id])
+
+    # If no test groups can be run by students, flash appropriate message and return early
+    test_group_info = assignment.test_groups
```
david-yz-liu, line 53: this is a good place to use `pluck`

config/locales/views/automated_tests/en.yml
```diff
@@ -41,6 +41,7 @@ en:
     no_autotest_settings: No automated tester settings exist for this course. Please update the automated tester settings and try again.
     no_criteria: Unable to find a criterion with name "%{name}".
     no_results: No test results to display
+    no_student_runnable_tests: No test groups can be run by students
```
david-yz-liu, line 44: Change to "No tests are available for students to run."

spec/controllers/automated_tests_controller_spec.rb
```diff
@@ -2,7 +2,7 @@
   include AutomatedTestsHelper
 
   # TODO: add 'role is from a different course' shared tests to each route test below
-  let(:assignment) { create :assignment }
+  let(:assignment) { create :assignment_with_criteria_and_test_results }
```
david-yz-liu, line 5: Do not change this here; only in the block that you've changed the tests for (note that you can override a `let` with another `let` that defines the same variable).

Also, instead of using an existing factory that generates extraneous data, create a new factory that just contains test groups. In fact you can create two, one for student test groups and one with instructor test groups.

Similarly, you should modify the existing test group factory to initialize the category to `[]` (the default), but then create new factories to initialize it to `['student']`, etc. Then you can use these factories in the definitions, rather than modifying the objects in the test cases.

david-yz-liu (commented):
@mimischly7 good work, in addition to the comments I left, make the equivalent change for instructor-run tests.

spec/controllers/automated_tests_controller_spec.rb
```diff
@@ -378,9 +378,23 @@
                              token_start_date: 1.day.ago,
                              remote_autotest_settings_id: 1)
         end
+        context 'and at least one test-group can be run by students' do
+          let(:assignment) { create :assignment_with_test_groups_student_runnable }
+          # we need to "refresh" the :params variable, which was originally defined at the 'describe' level
+          let(:params) { { course_id: assignment.course.id, assignment_id: assignment.id } }
```
david-yz-liu, line 384: This shouldn't be necessary, `params` is lazily-evaluated so should use the definition of `assignment` that's been overridden. (Same with the context block below)

app/controllers/automated_tests_controller.rb
```diff
@@ -48,11 +48,25 @@ def student_interface
...
+      flash_now(:info, I18n.t('automated_tests.no_student_runnable_tests'))
+      return
+    end
+
     grouping = current_role.accepted_grouping_for(assignment.id)
     grouping.refresh_test_tokens
     allowed = flash_allowance(:error, allowance_to(:run_tests?,
                                                    current_role,
                                                    context: { assignment: assignment, grouping: grouping })).value
+
```
david-yz-liu, line 69: delete this line

app/controllers/automated_tests_controller.rb
```diff
@@ -48,11 +48,25 @@ def student_interface
 
   def execute_test_run
     assignment = Assignment.find(params[:assignment_id])
+
+    # If no test groups can be run by students, flash appropriate message and return early
+    test_group_settings = TestGroup.where(assessment_id: assignment.id).pluck(:autotest_settings)
```
david-yz-liu, line 53: use the association here, `assignment.test_groups`

app/controllers/automated_tests_controller.rb
```diff
@@ -48,11 +48,25 @@ def student_interface
...
+    # rubocop:disable Rails/Pluck
+    test_group_categories = test_group_settings.map { |settings| settings['category'] }
```
david-yz-liu, line 55: You may not be aware that there are two different `pluck` methods. The one rubocop is telling you to use is https://api.rubyonrails.org/classes/Enumerable.html#method-i-pluck

PR #7024, opened 2024-04-01
File extension rename confirmation

AinaMerch:
Currently when a student uploads a file, they can choose to rename the file. However, this also allows them to change the extension of the file, which is often not really a valid operation (e.g., changing a .docx file into a .pdf). This is a front-end change to warn students who are selecting to rename a file but change the file extension, similar to the kind of confirmation dialog you would get in Windows if you try to do this. This still allows the student to actually do it (and there are some valid use cases, e.g. renaming a .txt into a .csv), but hopefully will warn students against doing so carelessly.
## Your Changes

Checks for matching extensions when one tries to submit

**Description**:

## Testing

Manually tested
Wrote front-end tests

## Checklist

- [X] I have verified that the CI tests have passed. 
- [X] I have reviewed the test coverage changes reported on Coveralls. 
- [X] I have added tests for my changes.


app/assets/javascripts/Components/Modals/submission_file_upload_modal.jsx
```diff
@@ -17,6 +17,21 @@ class SubmissionFileUploadModal extends React.Component {
 
   onSubmit = event => {
     event.preventDefault();
+    const newFilename = this.state.renameTo;
+    const originalFilename = this.state.newFiles[0].name; // Assuming only one file is uploaded
```
david-yz-liu, line 21: This is a good thing to observe; you should *skip this check entirely* if there is more than one file being submitted. Note that in this case the "Renaming" functionality is also disabled.

app/assets/javascripts/Components/Modals/submission_file_upload_modal.jsx
```diff
@@ -17,6 +17,21 @@ class SubmissionFileUploadModal extends React.Component {
...
+    const originalExtension = originalFilename.split(".").pop();
+    const newExtension = newFilename.split(".").pop();
+
+    if (originalExtension !== newExtension) {
+      const confirmChange = window.confirm(
+        "Changing the file extension may render the file unusable. Are you sure you want to proceed?"
```
david-yz-liu, line 27: This string should be internationalized

david-yz-liu (commented):
@AinaMerch good work! I left a few inline comments, and also make sure to update the changelog with an entry to reflect this change.

david-yz-liu (approved):

PR #7256, opened 2024-10-03
Feature: Option to retain old grading data 

### LINKED ISSUE #5212: Feature request: do not overwrite existing fields on recollection
Hello,

This term, a number of students asking for remarks had made small errors that our TAs had to manually fix. In order to run tests on the updated files, we had to recollect the submissions. This deleted the content of all existing fields, not just the ones that were auto-tested. This includes the remark request itself and any feedback the TA had previously given.

While there may be a good use case for clearing all fields on recollection, I think there should definitely be an option for keeping the values of existing fields.

Thanks for your time,
Sean

pranavrao145:
## Proposed Changes
Closes #5212.

## Type of Change

| ✨ *New feature* (non-breaking change that adds functionality)                          |     X     |

## Checklist

- [X] I have added tests for my changes, if applicable.
  - Associated PR: https://github.com/MarkUsProject/Wiki/pull/221

## Questions and Comments
Within inline comments.


app/jobs/submissions_job.rb
```diff
@@ -5,6 +5,101 @@ def add_warning_messages(messages)
...
+      end
+    end
+
+    # amend submission file ids for annotations (new submission files are
+    # created for each submission)
+    new_result.reload.annotations.each do |annotation|
+      filename = annotation.submission_file.filename
+      path = annotation.submission_file.path
+
+      new_submission_file = new_submission.submission_files.where(filename: filename, path: path).first
```
pranavrao145, line 50: I wonder whether there is a better way to do this (rather than by filename and path).

spec/jobs/submissions_job_spec.rb
```diff
@@ -185,6 +185,182 @@
...
+        it 'creates the correct number of new marks for each result' do
+          @new_results.zip(@original_results).each do |new_result, old_result|
+            expect(new_result.marks.size).to eq(old_result.marks.size)
+            expect(new_result.marks.ids).not_to eq(old_result.marks.ids)
+          end
+        end
+
+        it 'retains the correct mark values for each result' do
+          @new_results.zip(@original_results).each do |new_result, old_result|
+            expect(new_result.marks.map(&:mark)).to eq(old_result.marks.map(&:mark))
```
pranavrao145, line 305: This is a bit flaky - 1% of the time the marks come out in different orders and causes the test to fail. Is sorting the arrays and comparing solely the content still a good test?

app/assets/javascripts/Components/repo_browser.jsx
```diff
@@ -168,15 +191,26 @@ class ManualCollectionForm extends React.Component {
               {I18n.t("submissions.collect.apply_late_penalty")}
             </label>
           </p>
-          <button
-            type="submit"
-            name="commit"
-            onClick={event => {
```
pranavrao145, line 174: I moved this out of the `onClick` logic and into the `onSubmit` logic for the form so I could test this properly, otherwise there were errors about a lack of implementation for `onSubmit`

app/jobs/submissions_job.rb
```diff
@@ -5,6 +5,101 @@ def add_warning_messages(messages)
...
+                               .last
+
+    result_data = [
+      { old: old_result.annotations, new: new_result.annotations },
+      { old: old_result.marks, new: new_result.marks },
+      { old: old_result.extra_marks, new: new_result.extra_marks }
+    ]
+
+    # copy over data from old result to new result
+    result_data.each do |result_set|
```
pranavrao145, line 31: Same copy logic for annotations, marks, extra marks.

jest.config.js
```diff
@@ -60,7 +60,9 @@ module.exports = {
   // globalTeardown: undefined,
 
   // A set of global variables that need to be available in all test environments
-  // globals: {},
+  globals: {
+    AUTH_TOKEN: "test_auth_token",
```
pranavrao145, line 64: Need this for Jest to work, `ManualCollectionForm` has a constant in it that's normally defined on layout render but not in component-specific testing.

app/jobs/submissions_job.rb
```diff
@@ -5,6 +5,101 @@ def add_warning_messages(messages)
...
+      # if we can't find a matching file for a file-bound annotation, we assume
+      # the file has been modified or removed in the collected revision and
+      # discard this annotation
+      if new_submission_file.nil?
+        annotation.destroy!
+        next
+      end
+
+      annotation.update(submission_file_id: new_submission_file.id)
+      add_warning_messages(annotation.errors.full_messages) if annotation.errors.present?
```
pranavrao145, line 61: The approach I've taken to error handling is just adding all error messages to the flash message using the function already provided in this file - not sure if this is the best way to handle this.

app/jobs/submissions_job.rb
david-yz-liu: Overall, there are enough differences in behaviour that I don't think this loop is worth it.

1. Only existing _marks_ need to be destroyed (this is the only model with some pre-existing objects)
2. Only _annotations_ need to have a separate check for an existing submission file (which you've correctly noted below may trigger a case where an annotation _isn't_ copied over).
3. For _extra marks_, you should copy over all point-based marks (these are set manually by the grader), but _ignore_ all percentage-based marks (these are set by the late penalty, which is something that the instructor decides to apply, or not, for the newly-collected submission). Please make sure to document the latter case, as it's a bit subtle.

david-yz-liu (commented):
Hi @pranavrao145, great work! I just did a quick review of the back-end implementation and left some comments. Please work on those, and then I'll do another review in a few days.

pranavrao145:
I'm just going to convert this back to a draft while I work on addressing comments just so it doesn't run tests on intermediate work, and I'll make it ready for review again once I'm finished!

app/jobs/submissions_job.rb
```diff
@@ -5,6 +5,101 @@ def add_warning_messages(messages)
     Rails.logger.error msg
   end
 
+  def copy_old_grading_data(new_submission, grouping)
```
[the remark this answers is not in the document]
pranavrao145: Hi @david-yz-liu, thanks for the feedback! Code is looking a lot cleaner now. One question, seems the flash message on error is already taken care of implicitly when the Submissions model raises an exception, and it displays a clean message in red. However, this stops progress on all submissions if any one of them has a problem.

Would you like me to explicitly rescue Submission class errors here and add it only as a warning-level alert (using `add_warning_messages`), or just leave it at the error-level alert with a stop in processing other submissions? Content-wise, messages on both are pretty much the same (neither give the group info), so I thought I'd verify what the intended behaviour is.

app/jobs/submissions_job.rb
pranavrao145: This has also been fixed, and I've added docs both within the code and on the Wiki!

app/assets/javascripts/Components/Modals/collect_submissions_modal.jsx
```diff
@@ -96,15 +102,45 @@ class CollectSubmissionsModal extends React.Component {
...
+                    <input
+                      type="checkbox"
+                      defaultChecked={this.state.retain_existing_grading}
+                      name="retain_existing_grading"
+                      data-testid="chk_retain_existing_grading"
+                      onChange={this.handleRetainExistingGradingChange}
+                    />
+                  </label>
+                  &nbsp;
+                  {this.state.retain_existing_grading ? (
```
david-yz-liu, line 130: Keep the checkbox label the same for both checked and unchecked states. Instead, if the option is unchecked, show a warning div (same styling as warning flash messages) with the extra warning message. Since the div will have good styling already, you don't need to bold the warning message.

app/assets/javascripts/Components/__tests__/repo_browser_manual_collection_form.test.jsx
```diff
@@ -0,0 +1,72 @@
+import {render, screen, fireEvent} from "@testing-library/react";
+import {ManualCollectionForm} from "../repo_browser";
+
+// workaround needed for using i18n in jest tests, see
+// https://github.com/fnando/i18n/issues/26#issuecomment-1235751777
+jest.mock("i18n-js", () => {
```
david-yz-liu, line 6: Ah yes, can you actually move these mocks into one of the Jest setup files (and remove them from existing places in the codebase)?

app/controllers/submissions_controller.rb
```diff
@@ -170,9 +170,16 @@ def manually_collect_and_begin_grading
                          else
                            params[:apply_late_penalty]
                          end
+    retain_existing_grading = if params[:retain_existing_grading].nil?
```
david-yz-liu, line 173: Similar to below, here I suspect you can do a `params[:retain_existing_grading] == 'true'`. You can modify the `apply_late_penalty` setting above as well.

david-yz-liu (commented):
@pranavrao145 great work. I just left a few minor comments.

A wrinkle: after some discussion, I now understand that a common use case for this feature is that when an instructor receives a remark request, that's what triggers a new submission to be collected (i.e., the submission files need to be changed). Given this, I do think it's best to copy over both the original result and new remark result after all. Can you please modify the code to make this work (you might want to refactor again to create a `Result` instance method). Sorry about this!

app/assets/javascripts/Components/__tests__/repo_browser_manual_collection_form.test.jsx
pranavrao145: As it turns out, apparently we don't need these at all! Seems that there was already a Jest setup in `jest_env_setup.js` which defines `I18n` for all Jest tests globally. When writing this and previous tests, I never considered it may be implicitly defined so I didn't try without explicitly defining it - should have looked in some other files first, sorry about that! I've gone ahead and removed this setup from all files that use it.

config/locales/views/submissions/en.yml
```diff
@@ -16,11 +16,14 @@ en:
...
       could_not_collect_some_due: Could not collect submissions for some groupings and assignment %{assignment_identifier} - the collection date has not been reached for these groupings.
       could_not_collect_some_released: Could not collect submissions for some groupings and assignment %{assignment_identifier} - the submission has already been released.
+      full_overwrite_warning: Collecting and grading this revision will overwrite any previous collections/grading done for this group on this assignment.  Are you sure you want to do this?
+      grading_data_will_be_lost: 'Warning: without selecting this option, all old grading data for the collected submissions will be lost'
       manual_collection: Manual Collection
-      override_existing_html: 'Recollect previously collected submissions? <strong>Warning: any previous grading work will be lost.</strong>'
-      overwrite_warning: Collecting and grading this revision will overwrite any previous collections/grading done for this group on this assignment.  Are you sure you want to do this?
+      override_existing: Recollect previously collected submissions?
```
david-yz-liu, line 22: This message should be updated now (since remark data is now also copied)

app/assets/javascripts/Components/repo_browser.jsx
```diff
@@ -168,15 +191,26 @@ class ManualCollectionForm extends React.Component {
...
+              name="retain_existing_grading"
+              data-testid="chk_retain_existing_grading"
+              checked={this.state.retainExistingGrading}
+              hidden={!this.props.collected_revision_id}
+              disabled={!this.props.collected_revision_id} // prevent from sending info on submit
+              onChange={e => {
+                this.setState({retainExistingGrading: e.target.checked});
+              }}
+            />
+            <label
```
david-yz-liu, line 206: Use `htmlFor` attribute to link this label to the corresponding `input`; make sure to do this in the modal as well.

FYI, you can test this in the UI by *clicking on the label itself*: when properly linked, clicking the label should act the same as checking the checkbox itself.

app/controllers/submissions_controller.rb
```diff
@@ -167,14 +167,13 @@ def manually_collect_and_begin_grading
     end
 
     @revision_identifier = params[:current_revision_identifier]
-    apply_late_penalty = if params[:apply_late_penalty].nil?
-                           false
-                         else
-                           params[:apply_late_penalty]
-                         end
+    apply_late_penalty = params[:apply_late_penalty] == 'true'
+    retain_existing_grading = params[:retain_existing_grading] == 'true'
```
david-yz-liu, line 171: I'm not sure if you manually tested this after making this change, but it seems like the parameter value is `"on"`, not `"true"`, which is a bit strange.

app/models/submission.rb
```diff
@@ -255,6 +255,52 @@ def make_remark_result
...
+    # whether it's been submitted or not, remark request is going to be copied
+    if old_submission.remark_request.present?
+      self.remark_request = old_submission.remark_request
+      self.remark_request_timestamp = old_submission.remark_request_timestamp
+      self.save!
+
+      self.create_result # additional result for remark data
+
+      # if there's already a remark result created as well
+      if old_submission.has_remark?
```
david-yz-liu, line 298: This check is redundant with the `old_submission.remark_request.present?` check above (unless I'm missing something?)

app/models/result.rb
```diff
@@ -135,6 +135,43 @@ def self.get_total_extra_marks(result_ids, max_mark: nil, user_visibility: :ta_v
...
+      mark_dup = mark.dup
+      mark_dup.update!(result_id: self.id)
+    end
+
+    old_result.annotations.each do |annotation|
+      # annotations are associated with files; if a file for an annotation doesn't exist
+      # we just skip adding this annotation to the  new result
+      annotation_filename = annotation.submission_file.filename
+      annotation_path = annotation.submission_file.path
+      new_submission_file = self.submission.submission_files.where(filename: annotation_filename,
```
david-yz-liu, line 156: the search is okay, but use `find_by` instead of `.where(...).first`

app/models/result.rb
```diff
@@ -135,6 +135,43 @@ def self.get_total_extra_marks(result_ids, max_mark: nil, user_visibility: :ta_v
...
+                remark_request_submitted_at: old_result.remark_request_submitted_at)
+
+    old_result.marks.each do |mark|
+      mark_dup = mark.dup
+      mark_dup.update!(result_id: self.id)
+    end
+
+    old_result.annotations.each do |annotation|
+      # annotations are associated with files; if a file for an annotation doesn't exist
+      # we just skip adding this annotation to the  new result
```
david-yz-liu, line 153: delete the extra space

david-yz-liu (commented):
@pranavrao145 great work! I just left a few last comments, but I think the end is in sight. 😄

app/controllers/submissions_controller.rb
pranavrao145: Ah whoops, I did not manually test this in this iteration of changes - sorry about that! However, I recall now that I actually tried to make this change a couple of weeks ago when messing around with this, and I did run into this problem of it being `on` instead of `true`, which is why I left it the way it was. I will go ahead and revert this change for now - thanks for the catch.

config/locales/views/submissions/en.yml
pranavrao145: Ah yes, I will actually make this a confirmation message rather than a warning now because we're pretty much copying everything.

app/models/submission.rb
pranavrao145: So this is a bit odd, even to me - I'm overall trying to handle retaining data from two cases here:

1. In the case where the student is writing the remark request, has saved it, but has not submitted it to the instructor yet, the `remark_request` attribute and associated timestamp is populated on the submission, but no remark _result_ has been created yet; so I just check the attribute manually.
2. Where the student has _submitted_ the remark request to the instructor, everything in case 1 happens, but also a new remark result _is_ created, so we have to run the copy on that. Since `has_remark?` basically checks for if that remark result exists or not, I used it here to verify this case.

One thing worth noting is that `has_remark?` has the following comments:

> Returns a remark result that may or may not have been submitted. If you want the submitted remark result then the submitted_remark association should be used instead.

However, unless I'm misinterpreting what it means for a remark to not be submitted yet, I don't think this is fully accurate. From what I observed, I'm not actually sure it's possible to have a remark result without submitting it. To me, the function `has_remark?` seems to do practically the same thing as `submitted_remark`, but in a different way. Specifically, they both return the remark result (which I'm assuming is always the latest result if there is more than one result on the submission); `submitted_remark` gets it using the `remark_request_submitted_at` timestamp on `Result` and `has_remark` returns the last result directly _contingent_ on there being 2 results minimum). Wasn't sure which one to use so I just went with `has_remark?`, but `submitted_remark` might be better for readability.

Do you have any guidance on the best way to handle both of the above cases (or if I am even interpreting this correctly)?

david-yz-liu: Ah I see, yes that makes sense. In this case I believe you can do the following:

1. *Always* copy the `remark_request` and `remark_request_timestamp` attributes from the old submission to the new submission; this doesn't need to be guarded by a conditional.
2. Move the `self.create_result` to inside the `old_submission.has_remark?` check. You can also simplify a bit by first doing something like `old_remark_result = old_submission.remark_result` and then checking whether `old_remark_result.present?`

david-yz-liu (approved):
@pranavrao145 great job getting this completed! 🎉

PR #7311, opened 2024-11-28
Scale rubric criterion levels on max_mark update

### LINKED ISSUE #7296: Rubic criterion isn't populated correctly with a denominator > 4
When I add a rubric criterion and set the "out of" value to be 8, the default levels aren't given the correct values.  The values are always 0, 1, 2, 3, 4.  With a denominator of 8 they should be 0, 2, 4, 6, 8.

pranavrao145:
## Proposed Changes
Fixes #7296 by changing the order in which changes are processed when `max_mark` is updated on a rubric criterion.

## Type of Change

| 🐛 *Bug fix* (non-breaking change that fixes an issue)                                  |     X     |

## Checklist

- [X] I have added tests for my changes, if applicable.

- [ ] I have reviewed the test coverage changes reported by Coveralls.


david-yz-liu (commented):
@pranavrao145 thanks for doing this. I realized now why the code was written the way it was originally, as we didn't want to accidentally override the instructor's inputs on the form (for example, if the instructor changes the criterion max mark _and_ the level weights at the same time).

This is a bit more work, but can you please modify the back-end algorithm so that

1. for each existing level, if the `levels_attributes` has a mark that is *different* from the current level mark, use the new `levels_attributes` mark directly (no scaling); but if the `levels_attributes` mark is the same as the current level mark, scale that.
2. for each *new* level, use the instructor-provided mark in `levels_attributes` (no scaling)

I think the easiest way to do this is to keep the original order of criterion/level updating, but prior to the level updating mutate the `levels_attributes` hash to compute the new level marks according to my above description.

app/models/rubric_criterion.rb
```diff
@@ -1,54 +1,49 @@
...
     end
+
     self.update(levels_attributes: levels_attributes)
   end
 
   def level_with_mark_closest_to(mark)
     self.levels.min_by { |m| (m.mark - mark).abs }
   end
 
-  def scale_marks_if_max_mark_changed
```
pranavrao145, line 36: Decided to do this manually, from my obeservations this wasn't getting triggered except on update anyways.

app/models/rubric_criterion.rb
```diff
@@ -1,54 +1,49 @@
...
-  before_validation :scale_marks_if_max_mark_changed
   before_save :round_max_mark
 
   validates :levels, presence: true
 
   DEFAULT_MAX_MARK = 4
   # Checks whether the passed in param's level_attributes have unique name and marks.
   # Skips the uniqueness validation if true.
-  def update_levels(levels_attributes)
+  def update_levels(levels_attributes, scale = 1.0)
```
pranavrao145, line 9: Previously we were looking at changes to `max_mark` using `self.changes` (which I figured was a callback thing, not sure if that's available/how that works here) - to avoid some weird logic with the same I just made it a parameter the caller can very easily calculate.

app/models/rubric_criterion.rb
```diff
@@ -1,54 +1,49 @@
...
     s1, s2 = Set[], Set[]
     should_skip = true
 
-    # Check if there's a dup in marks or names
     levels_attributes.each_value do |val|
-      if s1.include?(val[:mark]) || s2.include?(val[:name])
-        should_skip = false
-        break
+      # Check if there's a dup in marks or names
+      unless should_skip
```
david-yz-liu, line 15: This branch will never evaluate since `should_skip` is initialized to `true`

app/models/rubric_criterion.rb
david-yz-liu: You shouldn't delete this; there may be other ways of updating `RubricCriterion` other than the controller method you're working on.

Instead, you can *skip* this validation in the controller method when calling `update` on the criterion.

david-yz-liu (approved):

PR #7420, opened 2025-02-06
Improved dashboard querying for TA data in overview graph

hemmatio:
## Proposed Changes
*(Describe your changes here. Also describe the motivation for your changes: what problem do they solve, or how do they improve the application or codebase? If this pull request fixes an open issue, [use a keyword to link this pull request to the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).)*

Currently (especially with a large amount of graders), rendering the graphs in the dashboard view for an assignment is inefficient. This is due to a number of inefficient queries made obtaining completion counts.

- Cached versions of existing functions:
  - `def get_num_marked_from_cache(assignment)`  
  - `def get_num_assigned_from_cache(assignment)`  
- These functions take advantage of cached data by the new`marked_result_ids_for(assignment)` function.  
- In the controller, the TA grade distribution code now calls these cached versions instead of the previous `get_num_marked` and `get_num_assigned` methods. This reduces the number of queries made per TA.

<details>
image
Before: 4 queries per TA

<br>
<br>
<br>
<br>

image
After: 1 query per TA
</details>

<details>

</details>

## Type of Change

| ♻️ *Refactoring* (internal change to codebase, without changing functionality)          |    X      |

## Checklist

- [x] If this is my first contribution, I have added myself to the list of contributors.

## Questions and Comments
*(Include any questions or comments you have regarding your changes.)*
02/06/25: This is still WIP. Current "fix" is very shoddy.
02/09/25: Correctly uses assignment id to query data. Only one query is created to obtain all results associated with the TA.
02/17/25: Added testcases. One potential thing to note: The cached versions of the functions cannot be used interchangeably with their non-cached versions. This is the main reason why I added them to the ta class, to separate them to be for their very specific ta-defined usages.
03/10/25: Logic for bulk caching moved into `assignment.rb`. Only one query is made per assignment rather than per TA. Still need to update tests (removing unneeded tests, testing bulk caching behaviour, etc). Also need to fix issue with the grade distribution array appending everything to the left.
03/16/25: Issue fixed! Still need to work on testing.
03/24/25: Switched to test-driven-development to ensure correctness of function when graders are assigned to specific criteria. Still need to create more tests to confirm behaviour. Also need to remove the excessive comments, they are mostly there for my own use.
03/31/25: Some tests failing. Also need to cover helper methods for correct behaviour.


app/models/ta.rb
```diff
@@ -24,14 +24,25 @@ def get_groupings_by_assignment(assignment)
              .includes(:students, :tas, :group, :assignment)
   end
 
+  def marked_result_ids_for(assignment)
```
david-yz-liu, line 27: So overall you have a good idea, but this should really be an `assignment` method. This is because the true N+1 query is the fact that there's one query made _per TA_, and this method does not help address that problem.

You can instead write an assignment method that uses the same instance variable caching idea, but stores a hash mapping TA id to the two lists (what you currently call `total_results` and `marked_result_ids`). As a tip, you can first use `pluck` to extract just the columns you want (ta id, result id, and marking state) and then [`Enumerable#group_by`](https://ruby-doc.org/core-3.0.2/Enumerable.html#method-i-group_by).

Finally, you made a good comment about not wanting to change the behaviour of the existing assignment methods. What you can do is add to it an optional argument (say, `bulk: false`) which, when `true`, will run the query to get all TAs and set the instance attribute. But when it's `false` you can default to just using the original code.

Finally, you should look closely at how `Assignment#get_num_marked` currently works; it has a separate case for `is_criteria_mark?`. This part has different logic for considering which results a TA has completed marking, based on the criteria the TA has been assigned rather than the result marking state.

hemmatio:
Waiting for CI/CD tests before requesting a review.

hemmatio:
Waiting for CI/CD tests before requesting a review.

app/models/assignment.rb
```diff
@@ -826,7 +826,10 @@ def is_criteria_mark?(ta_id)
     assign_graders_to_criteria && self.criterion_ta_associations.where(ta_id: ta_id).any?
   end
 
-  def get_num_assigned(ta_id = nil)
+  def get_num_assigned(ta_id = nil, bulk: false)
+    if bulk
+      return @cache_ta_results[ta_id][:total_results].size
```
david-yz-liu, line 831: In general, no methods except `cache_ta_results` should access the instance attribute `@cache_ta_results`, since it might not be defined.

But, what you should do is ensure `cache_ta_results` always *returns* the correct hash (i.e., `@cache_ta_results`); that way it can be used anywhere you want to use the instance attribute. Basically, the instance attribute is private but the method `cache_ta_results` is public.

So in this case, you can write `return cache_ta_results[ta_id][:total_results].size` instead.

david-yz-liu: But also, direct access is bad: use `Hash#dig` instead, as you do elsewhere.

david-yz-liu: Finally, for the `bulk: true` option, only do it when `ta_id` is not `nil`. That is, move this into the `else` branch below.

app/models/assignment.rb
```diff
@@ -850,7 +853,67 @@ def get_num_valid
              .count(&:is_valid?)
   end
 
-  def get_num_marked(ta_id = nil)
+  def marked_result_ids_for(ta_id)
+    cache_ta_results
+    @cache_ta_results.dig(ta_id, :marked_result_ids) || []
```
david-yz-liu, line 858: Same comment as above regarding the use of the instance attribute

app/models/assignment.rb
```diff
@@ -850,7 +853,67 @@ def get_num_valid
...
+
+  def cache_ta_results
+    return @cache_ta_results if defined? @cache_ta_results
+
+    data = current_results.joins(grouping: :tas).pluck('tas.id', 'results.id', 'results.marking_state')
+    # Group results by TA ID
+    grouped_data = data.group_by { |ta_id, _result_id, _marking_state| ta_id }
+
+    results_by_id = current_results.includes(:marks).index_by(&:id)
+    @cache_ta_results = grouped_data.each_with_object({}) do |(ta_id, results), hash|
```
david-yz-liu, line 869: You don't need to use `each_with_object` here. Instead, you can use `Hash.[]` and inside pass a list of lists, where each inner list is of the form `[key, value]`. There are a few other places in the MarkUs codebase that do this, but as a hint the overall structure is something like

```ruby
@cache_ta_results = Hash[grouped_data.map |ta_id, results| do
  ...
  end
]
```

app/models/assignment.rb
```diff
@@ -850,7 +853,67 @@ def get_num_valid
...
+      total_results = results.map { |_, result_id, _| result_id }
+
+      if self.assign_graders_to_criteria
+        # Get the list of criteria IDs assigned to this TA
+        assigned_criteria_ids = self.criteria.flat_map do |criterion|
```
david-yz-liu, line 874: This part still has an N+1 query because it's repeated once per TA.

It is safe to pull this out of the loop, and create a `Hash` mapping `ta_id` to the list of criterion ids.

Also, note that you can directly access `self.criterion_ta_associations` (because that's an association defined for the `Assignment` model).

app/models/assignment.rb
```diff
@@ -850,7 +853,67 @@ def get_num_valid
...
+        if criteria_count == 0
+          # If the TA has no assigned criteria, fallback to using marking_state to determine marked results
+          complete_results = results.select do |_, _, marking_state|
+            marking_state ==
+                      Result::MARKING_STATES[:complete]
+          end
+          marked_result_ids = complete_results.map { |_, result_id, _| result_id }
+        else
+          # Count results where all assigned criteria have been marked
+          marked_result_ids = total_results.select do |result_id|
```
david-yz-liu, line 888: Similar to my above comment, this part is okay but will still result in an N+1 query because of the looping over results.

Instead, use one query to get *a count of non-nil marks for the assigned criteria, and for each result*. A good starting point is something like `Result.joins(:marks).where(id: ..., mark: {criterion_id: ..., mark: ...})` and then use a [grouping operation](https://guides.rubyonrails.org/active_record_querying.html#grouping).

Note: even if you do the above, there actually will still be an N+1 query because you'll need to do this once per TA. Not ideal, but I think resolving that N+1 query will be harder so I think it's best to forgo it for this PR to get this merged in, and then I'll give it to another student to continue this work in the future.

david-yz-liu (commented):
@hemmatio nice work! Handling the assign-to-criteria case is good, but I've left some comments overall to improve the code even further.

app/models/assignment.rb
```diff
@@ -827,9 +827,11 @@ def is_criteria_mark?(ta_id)
     assign_graders_to_criteria && self.criterion_ta_associations.where(ta_id: ta_id).any?
   end
 
-  def get_num_assigned(ta_id = nil)
+  def get_num_assigned(ta_id = nil, bulk: false)
     if ta_id.nil?
       groupings.size
+    elsif bulk
+      cache_ta_results.dig(ta_id, :total_results)&.size
```
david-yz-liu, line 834: You'll need to use an `|| 0` to ensure an integer is always returned (rather than `nil`)

app/models/assignment.rb
```diff
@@ -851,7 +853,72 @@ def get_num_valid
...
+  def cache_ta_results
+    return @cache_ta_results if defined? @cache_ta_results
+
+    data = current_results.joins(grouping: :tas).pluck('tas.id', 'results.id', 'results.marking_state')
+    # Group results by TA ID
+    grouped_data = data.group_by { |ta_id, _result_id, _marking_state| ta_id }
+
+    results_by_id = current_results.includes(:marks).index_by(&:id)
+
+    ta_to_criteria = self.criterion_ta_associations.group_by(&:ta_id).transform_values do |assocs|
```
david-yz-liu, line 869: It's a bit cleaner to do a `.pluck([:ta_id, :criterion_id])` and then use a `group_by`, like you're doing for `grouped_data` above. This prevents a bunch of `CriterionTaAssociation` model objects from being created.

app/models/assignment.rb
```diff
@@ -851,7 +853,72 @@ def get_num_valid
...
+                                    .group('results.id')
+                                    .having('count(marks.id) = ?', criteria_count)
+                                    .pluck('results.id')
+        end
+      else
+        # Grading not by criterion: count only completed results
+        marked_result_ids = results.select { |_, _, marking_state| marking_state == Result::MARKING_STATES[:complete] }
+                                   .map { |_, result_id, _| result_id }
+      end
+      # Save the results into the cache
```
david-yz-liu, line 912: I don't think this comment really applies any more, and you can safely remove it

app/models/assignment.rb
```diff
@@ -851,7 +853,72 @@ def get_num_valid
...
+        if criteria_count == 0
+          # If the TA has no assigned criteria, fallback to using marking_state to determine marked results
+          complete_results = results.select do |_, _, marking_state|
+            marking_state ==
+                      Result::MARKING_STATES[:complete]
+          end
+          marked_result_ids = complete_results.map { |_, result_id, _| result_id }
+        else
+          # Count results where all assigned criteria have been marked
+          total_results.select do |result_id|
```
david-yz-liu, line 890: This block isn't necessary any more (you've replaced it with the query on Line 900 to initialize `marked_result_ids`).

app/models/assignment.rb
```diff
@@ -851,7 +853,72 @@ def get_num_valid
...
+      end
+      # Save the results into the cache
+      [ta_id, { total_results: results, marked_result_ids: marked_result_ids }]
+    end.to_h
+  end
+
+  def get_num_marked(ta_id = nil, bulk: false)
+    if bulk
+      cache_ta_results
+      return @cache_ta_results[ta_id][:marked_result_ids].size
```
david-yz-liu, line 920: recall my previous comment; also, use `dig` (and return a default value of 0, not `nil`)

david-yz-liu (commented):
@hemmatio nice work; I left a few minor comments to tidy up, but I think this is almost ready to merge. 👍

app/models/assignment.rb
```diff
@@ -851,7 +853,59 @@ def get_num_valid
...
+    cache_ta_results.dig(ta_id, :marked_result_ids) || []
+  end
+
+  def cache_ta_results
+    return @cache_ta_results if defined? @cache_ta_results
+
+    data = current_results.joins(grouping: :tas).pluck('tas.id', 'results.id', 'results.marking_state')
+    # Group results by TA ID
+    grouped_data = data.group_by { |ta_id, _result_id, _marking_state| ta_id }
+    current_results.includes(:marks).index_by(&:id)
```
david-yz-liu, line 866: @hemmatio I'm not sure what the purpose of this line is any more. It looks like you just deleted the variable (maybe because of a rubocop issue?) but in general if a variable is unused, you should almost always deleting the entire assignment statement (unless there are side effects; I don't think that applies here).

david-yz-liu (approved):
@hemmatio great job!

Issue #7677, opened 2025-09-25
Indicate if due date extension has apply late penalty selected or not on the Groups table view


donny-wong:
**Description:**

When providing due date extension to a group, there is a checkbox option to apply the late submission policy when collecting the group's submission. 

Here is the Due Date Extension modal in question:

<img width="632" height="424" alt="Image" src="https://github.com/user-attachments/assets/2948a3a8-101d-4544-a84d-63cabb5bb13a" />

It would be beneficial to also display this information on the Groups table view, otherwise you would need to refer to the modal.

Currently, here is an example of how it looks when one group, 0_apply_penalty_true, has the checkbox selected and the other, 0_apply_penalty_false, doesn't.  

<img width="1272" height="292" alt="Image" src="https://github.com/user-attachments/assets/6201ba32-b451-4d4d-b0f5-14c5c9c81a0c" />

PR #7689, opened 2025-10-03
ISSUE-7677: Display late penalty selection

Naragod:
## Description

We wish to notify users if a late penalty has been applied to an extension at first glance. Currently, users have to click the due date extension link, before being able to view the application of late penalties to their assignments. More information can be found in the issue [description](https://github.com/MarkUsProject/Markus/issues/7677#event-20046560532).

## Implementation

We already keep track of a due date extension's late penalty. Let's display it to the user when viewing the group table assignments.

## Todo

[] Implement new test cases that check the due date extension penalty column is properly displayed.


app/javascript/Components/groups_manager.jsx
```diff
@@ -563,6 +563,21 @@ class RawGroupsTable extends React.Component {
...
     },
+    {
+      Header: "Penalty",
+      id: "late_penalty_applied",
+      accessor: "extension",
+      Cell: row => {
+        return row.original.extension.apply_penalty ? (
+          <FontAwesomeIcon icon="fa-solid fa-square-check" />
+        ) : (
+          <FontAwesomeIcon icon="fa-solid fa-xmark" />
```
Naragod: I picked font awesome icons I thought appropriate. Please let me know if it'd be better to use a different icon to display to our users.

app/javascript/Components/groups_manager.jsx
```diff
@@ -563,6 +563,31 @@ class RawGroupsTable extends React.Component {
...
+        );
+      },
+      minWidth: 30,
+      filterable: false,
+      sortMethod: (a, b) => {
```
Naragod: Given the one time use, small size, and simplicity of this function, I thought it'd be a good idea to implement it directly in the group_manager file. Let me know if it'd be better to place this sorting function inside the [table_helpers](https://github.com/MarkUsProject/Markus/blob/master/app/javascript/Components/Helpers/table_helpers.jsx) file instead.

Naragod:
Hi @david-yz-liu,

I have taken into account your comment and a few things remains unclear:

* The issue's original intent was to demonstrate to the user whether or not a penalty was applied to a user's late submission. I am unsure why we would want to add the text `late submissions accepted` when the time extension is already displayed, i.e. `1 week`. Do we want to make it more clear to the user that a time extension has been granted?

* I was under the impression that submitting an assignment after it had been extended was a late submission. I am unsure what you mean by `groups with extension but no late submissions`, can you please clarify? I perhaps lack the context to understand exactly what you mean.

donny-wong:
Extension due date allows submissions to be submitted without being considered late. By selecting "Apply late submission policy when collecting this group's submissions" checkbox, the late submission policy will take into effect once the extension due date has been reached.

For the class in general, the late submission policy will be applied to the original due date.

Changelog.md
```diff
@@ -8,6 +8,7 @@
 - Added new loading spinner icon for tables (#7602)
 - Update message and page displaying cannot create new course via external LTI tool (#7669)
 - Provide file viewer the option to render Microsoft files (#7676)
+- Display late submission selection and add accompanying filter (#7677)
```
david-yz-liu: For the number at the end of the entry, we put the PR number, not the issue number

app/javascript/Components/groups_manager.jsx
```diff
@@ -559,20 +559,27 @@ class RawGroupsTable extends React.Component {
...
-              }
+            const lateSubmissionText = row.original.extension.apply_penalty
+              ? `(${I18n.t("groups.late_submissions_accepted")})`
+              : "";
+
+            if (!val) {
+              return null;
+            }
+            // don't build these strings dynamically or they will be missed by the i18n-tasks checkers.
+            if (key === "weeks") {
```
david-yz-liu: In general prefer using the existing style, which is using `else if` rather than separate if statements. This part didn't need to be changed other than adding the `lateSubmissionText` part.

app/javascript/Components/groups_manager.jsx
```diff
@@ -599,8 +606,39 @@ class RawGroupsTable extends React.Component {
...
+        }
+        const applyPenalty = row._original.extension.apply_penalty;
+        const {withExtension, withLateSubmission} = JSON.parse(filter.value);
+        // If there is an extension applied, the extension object will contain a property called hours
+        const hasExtension = Object.hasOwn(row._original.extension, "hours");
+
+        if (!withExtension) {
+          return !hasExtension;
+        }
+        if (withLateSubmission) {
```
david-yz-liu, line 602: Same comment as above. I can tell this is your preferred style, but this is not the preferred style for this code base.

config/locales/views/groups/en.yml
```diff
@@ -28,6 +28,10 @@ en:
     grace_day_over_limit: You have assigned one or more students who has less grace credits than the amount already used by group %{group} for this assignment.
     group_name_already_in_use: This name is already in use for this assignment.
     group_name_already_in_use_diff_assignment: The group can't be renamed, because it would result in the loss of submission data. Try a name that is not in use for another assignment.
+    groups_with_extension:
+      with_late_submission: Group With Extension and With Late Submission
```
david-yz-liu: Please make the following wording changes:

- "Group" -> "Groups"
- "With" -> "with"
- "and With Late Submission" -> "(Late Submissions Accepted)" (to align with the individual cell contents; "With Late Submission" is actually a bit misleading, because the configuration option means that late submissions are _accepted_, but the group may or may not actually submit late).

Similar comments for the `without_late_submission` key, perhaps using "Late Submissions Not Accepted".

config/locales/views/groups/en.yml
```diff
@@ -45,6 +49,7 @@ en:
       success: Invitation(s) successful.
     is_not_valid: Group is not valid
     is_valid: Group is valid
+    late_submissions_accepted: Late Submissions Accepted
```
david-yz-liu: Make this text lowercase (since it appears in the table)

Changelog.md
Naragod: Got it 👍

app/javascript/Components/groups_manager.jsx
Naragod: Got it 👍

config/locales/views/groups/en.yml
Naragod: Done

config/locales/views/groups/en.yml
Naragod: Done. Please let me know if there any other changes you'd like me to make.

app/javascript/Components/groups_manager.jsx
Naragod: Got it 👍

PR #7754, opened 2025-12-03
ISSUE-7711: Add GET - api/.../groups/test_results api route

Naragod:
## Proposed Changes
We wish to give users the ability to obtain test results through the API. This PR adds the following endpoint: `GET - /api/courses/:course_id/assignments/:assignment_id/groups/:id/test_results`.

<details>

</details>

<details>

</details>

## Type of Change

| ✨ *New feature* (non-breaking change that adds functionality)                          |       X   |

## Checklist

## Questions and Comments

This PR references:
* [Wiki](https://github.com/MarkUsProject/Wiki/pull/249) - Add `GET - api/../groups/test_results` endpoint documentation

### DIFF
--- a/app/controllers/api/groups_controller.rb
+++ b/app/controllers/api/groups_controller.rb
@@ -204,46 +204,28 @@ module Api
     end
 
     def test_results
-      test_runs = grouping&.test_runs&.includes(test_group_results: [:test_group,
-                                                                     :test_results])&.order(created_at: :desc)
+      return render_no_grouping_error unless grouping
 
-      if test_runs.blank?
-        return render 'shared/http_status',
-                      locals: { code: '404', message: 'No test results found for this group' },
-                      status: :not_found
-      end
+      # Use the existing Assignment#summary_test_results method filtered for this specific group
+      # This ensures format consistency with the UI download (summary_test_result_json)
+      group_name = grouping.group.group_name
+      results = assignment.summary_test_results([group_name])
 
-      results_data = test_runs.map do |test_run|
-        {
-          id: test_run.id,
-          status: test_run.status,
-          created_at: test_run.created_at,
-          problems: test_run.problems,
-          test_groups: test_run.test_group_results.map do |test_group_result|
-            {
-              name: test_group_result.test_group.name,
-              marks_earned: test_group_result.marks_earned,
-              marks_total: test_group_result.marks_total,
-              time: test_group_result.time,
-              tests: test_group_result.test_results.order(:position).map do |test|
-                {
-                  name: test.name,
-                  status: test.status,
-                  marks_earned: test.marks_earned,
-                  marks_total: test.marks_total,
-                  output: test.output,
-                  time: test.time
-                }
-              end
-            }
-          end
-        }
-      end
+      return render_no_grouping_error if results.blank?
+
+      # Group by test_group name to match the summary_test_result_json format
+      results_by_test_group = results.group_by(&:name)
 
       respond_to do |format|
-        format.xml { render xml: results_data.to_xml(root: 'test_runs', skip_types: 'true') }
-        format.json { render json: results_data }
+        format.xml { render xml: results_by_test_group.to_xml(root: 'test_results', skip_types: 'true') }
+        format.json { render json: results_by_test_group }
       end
+    end
+
+    def render_no_grouping_error
+      render 'shared/http_status',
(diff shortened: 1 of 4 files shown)


app/controllers/api/groups_controller.rb
```diff
@@ -203,6 +203,49 @@ def annotations
       end
     end
 
+    def test_results
```
david-yz-liu, line 206: This implementation does work, but I would like to see this align with the implementation of the JSON download through the UI, as I don't think it's good to have two different JSON formats exposed.

The existing JSON export makes use of `Assignment#summary_test_results`. You could add an optional parameter for a group id (or perhaps a list of group ids, which allows more possibilities in the future) to filter the existing query.

Note that the above method only returns the results for the latest test run. I think this is fine for now, since this is aligned with the existing JSON functionality. In the future we could extend this with an option to download all test runs.

Naragod: Hi David,

Would it be better to update the API call to follow the exiting JSON download structure, or should we update the existing implementation to download all test runs? 

Here are the pros and cons of each:

```
UI format:

Pros:
 - Simpler
 - Grouped by group_name → test_group_name

New API Format:

Pros:
- Returns all test runs for the group
- Is structured as an array of test_runs with nested test_groups
- Includes metadata like test_run id, status, created_at
``` 

I have updated the new API implementation to use the existing the default UI implementation and added the option `include_metadata` to use the new API JSON format, which returns all test runs.

We can also update the exiting UI download implementation to provide all test runs and meta data.

app/controllers/api/groups_controller.rb
```diff
@@ -441,8 +530,6 @@ def add_test_run
       end
     end
 
-    private
```
david-yz-liu, line 444: Revert this change

app/controllers/api/groups_controller.rb
david-yz-liu: Hi Mateo, for this PR I want to use exclusively what is already implemented and so please remove the other code from this PR.

If we decide to make changes in the future (to add all test runs) then I would like to do so by modifying the existing method and have it apply to both API and the UI exports. We would also have a discussion about exactly what the API parameters should be (I find the proposed `include_metadata` to be overly broad).

So please remove the code that you've added and just use the implementation in `render_legacy_test_results` (by the way, I also do not believe this is an appropriate name, as I do not consider the existing implementation to be "legacy").

app/controllers/api/groups_controller.rb
Naragod: Done

app/controllers/api/groups_controller.rb
Naragod: Sounds good. Please take another look.

app/controllers/api/groups_controller.rb
david-yz-liu: I'm not sure you understood what I meant by this. It seems the line `private` and a blank line were deleted. This shouldn't be necessary, as this PR doesn't introduce a requirement to access the private methods outside of this class.

To address this, please add back the two lines (`private` and the blank line beneath it).

app/controllers/api/groups_controller.rb
```diff
@@ -203,6 +203,32 @@ def annotations
...
     end
 
+    def test_results
+      return render_no_grouping_error unless grouping
+
+      # Use the existing Assignment#summary_test_results method and filter for this specific group
+      # This ensures format consistency with the UI download (summary_test_result_json)
+      group_name = grouping.group.group_name
+      all_results = assignment.summary_test_results
+      results = all_results.select { |r| r.group_name == group_name }
```
david-yz-liu: It is better to filter within the `Assignment#summary_test_results` method itself to avoid loading a lot more test results. I think I saw an earlier comment about managing the nested queries in `Assignment#summary_test_results`; one suggestion is that you can add a `where` call immediately before Line 688 in `assignment.rb`.

app/controllers/api/groups_controller.rb
Naragod: You were right. I accidentally deleted the private keyword. It has been re-added.

app/controllers/api/groups_controller.rb
Naragod: Done. Please take another look.

PR #7968, opened 2026-05-29
Fix navigation warnings on autotest manager and starter file manager pages, closes 7641

philipkukulak:
## Proposed Changes
On the _Settings > Automated Tests_ page, a warning should pop up if a user attempts to navigate away from the page after making changes and before saving them.

Currently, it appears even after the user has saved their changes. This happens because the _Settings > Automated Tests_ page is rendered with `react`, which is rendered after the `jquery` click handler in `_navigation_warning.js.erb` is registered: `submit_clicked` can never be flipped to `true` because the submit button doesn't exist when that handler is registered.

Furthermore, the `window.onbeforeunload` handler as defined in the `_navigation_warning.js.erb` partial has no awareness of the _Automated Tests_ page `react` structure. The page and the partial within it are tracking 'dirtiness' in separate state variables.

With help from `claude`, the cleanest fix is to override the `window.onbeforeunload` handler in `autotest_manager.jsx` and to gate the initialization of the page so that `fetchData` doesn't instantly 'dirty' the page when `react` fires `onChange` upon first receipt of existing form data.

The fix for the _Settings -> Starter Files_ page is far simpler, since it is not rendered using `react`: we bind the handler to `.on('click', ...)` instead.

<details>

https://github.com/user-attachments/assets/6b94ec2a-916c-451f-b92a-e9a28377850b

https://github.com/user-attachments/assets/f8eab892-e9dc-4a40-8f8a-b9d7af19f5a2

</details>

<details>
n/a
</details>

## Type of Change

| 🐛 *Bug fix* (non-breaking change that fixes an issue)                                  |     x     |

## Checklist

- [ ] I have updated the project documentation, if applicable.

## Questions and Comments
I had the thought that we could also take this a step further and track changes to the page that are 'undone', e.g. checking a box and then unchecking it, and make sure the navigation warning doesn't pop up in that case. I came to the judgment that this would bloat the code more than we could justify an improvement so slight.

~~I also did not touch the _Settings -> Starter Files_ page (on which this bug is also present) but if it is decided upon review that we'd like to bundle that fix into this PR, I would be happy to do so.~~ Applied this fix to the _Settings -> Starter Files_ page.

### DIFF
--- a/app/javascript/Components/autotest_manager.jsx
+++ b/app/javascript/Components/autotest_manager.jsx
@@ -77,7 +77,7 @@ class AutotestManager extends React.Component {
           },
         },
       },
-      formData: {},
+      formData: null,
       enable_test: true,
       enable_student_tests: true,
       token_start_date: "",
@@ -98,6 +98,16 @@ class AutotestManager extends React.Component {
 
   componentDidMount() {
     this.fetchData();
+    // takes over from _navigation_warning.js.erb to use React state as the source of truth
+    window.onbeforeunload = () => {
+      if (this.state.form_changed) {
+        return I18n.t("uncommitted_changes_warning");
+      }
+    };
+  }
+
+  componentWillUnmount() {
+    window.onbeforeunload = null;
   }
 
   fetchData = () => {
@@ -205,7 +215,10 @@ class AutotestManager extends React.Component {
   };
 
   handleFormChange = data => {
-    this.setState({formData: data.formData}, () => this.toggleFormChanged(true));
+    this.setState({formData: data.formData});
+    if (this.state.formData !== null) {
+      this.toggleFormChanged(true);
+    }
   };
 
   toggleEnableTest = () => {
@@ -488,7 +501,7 @@ class AutotestManager extends React.Component {
             disabled={!this.state.enable_test}
             schema={this.state.schema}
             uiSchema={this.state.uiSchema}
-            formData={this.state.formData}
+            formData={this.state.formData ?? {}}
             onChange={this.handleFormChange}
             validator={this.props.validator}
             templates={{
@@ -625,6 +638,8 @@ function MoveUpButton(props) {
     </button>
   );
 }
+
+export {AutotestManager};
 
 export function makeAutotestManager(elem, props) {
   const validator = customizeValidator({ajvOptionsOverrides: {discriminator: true}});
--- a/app/views/automated_tests/_form.html.erb
+++ b/app/views/automated_tests/_form.html.erb
@@ -5,9 +5,6 @@
         { course_id: <%= @current_course.id %>, assignment_id: <%= @assignment.id %>});
     });
   <% end %>
-  <%= render partial: 'shared/navigation_warning',
-             formats: [:js],
-             handlers: [:erb] %>
 <% end %>
 
 <div id='autotest-manager'></div>
--- a/Changelog.md
+++ b/Changelog.md
@@ -45,6 +45,7 @@
 - Fixed filter Canvas Test Student from roster sync (#7926)
 - Fix: include original total mark in JSON response for remark requests (#7945)
 - Fixed `(hidden)` assignment labeling for assignments with `visible_on` and/or `visible_until` set (#7944)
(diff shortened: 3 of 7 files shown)


app/javascript/Components/autotest_manager.jsx
```diff
@@ -205,7 +225,10 @@ class AutotestManager extends React.Component {
   };
 
   handleFormChange = data => {
-    this.setState({formData: data.formData}, () => this.toggleFormChanged(true));
+    this.setState({formData: data.formData});
+    if (!this._initializing) {
```
david-yz-liu: Overall I don't think this `_initializing` attribute is necessary. You can check the current value of `this.state.formData` before doing any updates, treating `Object.keys(this.state.formData).length === 0` as the "initializing" condition.

david-yz-liu (commented):
Nice work, @philipkukulak. I left one inline comment, and also it seems like you should now be able to remove the loading of the `_navigation_warning.js.erb` partial from this page.

Please go ahead and also make the same change to the starter files page.

app/javascript/Components/autotest_manager.jsx
philipkukulak: Wouldn't `Object.keys(this.state.formData).length > 0` if any form data already exists, e.g. some tester is already configured? That would mark the page 'dirty' by default whenever there's persistent data. I'm thinking of 'initialization' as the initial fetch for page contents rather than the contents of the page itself.

david-yz-liu: I'm also considering "initialization" as the initial fetch. The initial state of this component sets `formData: {}`. After the data is fetched this will change. (For a new assignment I suppose it's possible the fetched value may be `{}` as well, I'm not sure. If so, you can change the initial `formData` state to `null` instead to disambiguate it.)

philipkukulak: Ahhh okay I think I understand. I'll re-request review once I pushed these changes. Thanks David.

philipkukulak:
Removed the `_navigation_warning` partial from _Settings -> Automated Tests_ since everything is handled by a `react` component now. Applied the fix to the _Settings -> Starter Files_ page (much easier to fix here since there is no `react` on that page.) Also changed the initialization gate on the _Settings -> Automated Tests_ per David's recommendations. Tested it locally. It works! Pretty cool.

Note that overall test coverage has gone down because we've created a new test suite (`autotest_manager.test.jsx`) that only tests our bug fix.

app/views/automated_tests/_form.html.erb
```diff
@@ -5,9 +5,7 @@
         { course_id: <%= @current_course.id %>, assignment_id: <%= @assignment.id %>});
     });
   <% end %>
-  <%= render partial: 'shared/navigation_warning',
-             formats: [:js],
-             handlers: [:erb] %>
+
```
david-yz-liu: Delete this blank line

app/views/shared/_navigation_warning.js.erb
```diff
@@ -17,17 +17,7 @@
...
-      });
-    }
-
-    if (document.readyState === "loading") {
-      document.addEventListener("DOMContentLoaded", domContentLoadedCB);
-    } else {
-      domContentLoadedCB()
-    }
-  })();
+  $(document).on('click', 'input[type="submit"]', function () {
```
david-yz-liu: This change is quite different from the one for the autotest manager and not what I expected (this file is used by other views as well, so should not be changed).

The starter file page does use React, with the root component being `StarterFileManager`. Please make the same type of change you made for the `AutotestManager` in that component.

philipkukulak: Sorry, I was unclear; the starter file page doesn't use `rjsf` to render form contents like the autotester settings page does, which is the library that fires an `onChange` event right out the gate when `fetchData` is called.

About to push a commit in which the change to the starter files page mirrors that in the autotester settings page.

david-yz-liu (commented):
@philipkukulak okay looks good, but you should also now be able to remove the `shared/navigation_warning` include on the starter files page.

david-yz-liu (approved):

PR #8007, opened 2026-06-18
Support all annotation types in add_annotations API (#7999)

### LINKED ISSUE #7999: Support all annotation types in GroupsController#add_annotations API
Currently the `GroupsController#add_annotations` API route only supports annotations for plaintext files (it hard-codes creation of `TextAnnotation` objects). I'd like to modify this to support all annotation types, probably with a `type` field in the annotation data to disambiguate the types, analogous to how they are stored in the database.

philipkukulak:
## Proposed Changes

The API endpoint `POST add_annotations` now supports all annotation types. Co-authored with Claude. Closes #7999.

In achieving this, it was found that there was no validation of annotation type to file type for which we are seeking to annotate. This means a request to apply a `PdfAnnotation` to a `.py` file would succeed, which would break rendering in the UI. To resolve this, some pre-existing type derivation logic was lifted out of the UI `AnnotationsController` and is now being used in both the UI and API controllers; the API also uses it as a part of the validation chain. 

As a side effect of this change, `add_existing_annotation`'s specs were lacking coverage due to a mix in record creation between annotation types: `HtmlAnnotation` used `create!`, and all other annotations used `create`. This is now consolidated. There was no coverage of the failure path for this record creation, so this was added here for good measure.

Another small bug that was fixed is that prior to this change, `POST`ing a `TextAnnotation` to a non-text file could create an annotation that would never be displayed on that file. Users will now see a `422` response if they try to do this.

The API endpoint itself now either accepts an explicit annotation type, or derives it implicitly. In the former case, it expects a match between the annotation type and the file being annotated. In the latter, it simply reconciles it from the file being annotated.

<details>

https://github.com/MarkUsProject/Wiki/pull/267

</details>

## Type of Change

| ✨ *New feature* (non-breaking change that adds functionality)                          |     x     |

## Checklist

## Questions and Comments

Here's a summary of the smoke tests for the changes to the endpoint:

### Smoke test — `add_annotations` API

Local server (`/csc108`), authenticated instructor API key. Seeded one file per type on a
version-used submission: `hello.py`, `nemo.jpg`, `possimus.pdf`, `ipsa.ipynb`. Each result
confirmed by reading back through `GET …/groups/:id/annotations`. The `type` field uses the
full STI class names (`TextAnnotation`/`ImageAnnotation`/`PdfAnnotation`/`HtmlAnnotation`).

| # | Scenario | `type` sent | Target file | Expected | Result |
|---|----------|-------------|-------------|----------|--------|
| 1 | Text annotation | `TextAnnotation` | `hello.py` | 200, `TextAnnotation` | ✅ 200 — `TextAnnotation`, line/col persisted |
| 2 | Image annotation | `ImageAnnotation` | `nemo.jpg` | 200, `ImageAnnotation` | ✅ 200 — `ImageAnnotation`, x/y persisted |
| 3 | PDF annotation | `PdfAnnotation` | `possimus.pdf` | 200, `PdfAnnotation` | ✅ 200 — `PdfAnnotation`, page 2 |
| 4 | HTML annotation | `HtmlAnnotation` | `ipsa.ipynb` | 200, `HtmlAnnotation` | ✅ 200 — `HtmlAnnotation`, nodes persisted |
| 5 | Type omitted → derived from file | *(none)* | `possimus.pdf` | 200, `PdfAnnotation` | ✅ 200 — `PdfAnnotation`, page 3 |
| 6 | Type ↔ file mismatch | `TextAnnotation` | `possimus.pdf` | 422 | ✅ 422 — `Annotation type 'TextAnnotation' does not match the type of file 'possimus.pdf'` |
| 7 | Unknown / invalid type | `bogus` | `hello.py` | 422 | ✅ 422 — `Invalid annotation type: bogus` |
| 8 | Missing required field | *(derived pdf)* | `possimus.pdf` (no `page`) | 422 | ✅ 422 — `Missing required fields for PdfAnnotation: page` |
| 9 | Nonexistent file | `TextAnnotation` | `nope.txt` | 422 | ✅ 422 — `Submission file not found: nope.txt` |
| 10 | Batch atomicity (1 valid + 1 invalid) | `TextAnnotation`, `TextAnnotation` | `hello.py`, `possimus.pdf` | 422, nothing written | ✅ 422 — DB count unchanged; valid item not written |
| 11 | Empty annotations array | — | — | (edge) | ✅ 200 — no-op, nothing written |

**Net:** all 5 valid writes persisted with the correct STI type on the correct file; all 4
bad-input paths failed closed with specific 422 messages and left the DB unchanged; batch
validation is atomic (rejects the whole request before any insert).

> Note: `type` accepts the STI class names; a short code such as `"text"` is rejected as
> `Invalid annotation type: text` (consistent with the `add_test_results` endpoint).

> Note: API error `description` values are HTML-escaped (e.g. `&#39;` for `'`) by the shared
> `shared/http_status.json.erb` renderer — pre-existing, app-wide behavior, not introduced by
> this change.

### DIFF
--- a/app/controllers/annotations_controller.rb
+++ b/app/controllers/annotations_controller.rb
@@ -5,7 +5,6 @@ class AnnotationsController < ApplicationController
     result = Result.find(params[:result_id])
     submission = result.submission
     submission_file = submission.submission_files.find(params[:submission_file_id])
-    rmd_convert = submission_file.is_rmd? && Rails.application.config.rmd_convert_enabled
 
     base_attributes = {
       submission_file_id: submission_file.id,
@@ -16,44 +15,12 @@ class AnnotationsController < ApplicationController
       result_id: params[:result_id]
     }
 
-    if submission_file.is_supported_image?
-      @annotation = result.annotations.create(
-        type: 'ImageAnnotation',
-        x1: params[:x1],
-        y1: params[:y1],
-        x2: params[:x2],
-        y2: params[:y2],
-        **base_attributes
-      )
-    elsif submission_file.is_pdf?
-      @annotation = result.annotations.create(
-        type: 'PdfAnnotation',
-        x1: params[:x1],
-        y1: params[:y1],
-        x2: params[:x2],
-        y2: params[:y2],
-        page: params[:page],
-        **base_attributes
-      )
-    elsif submission_file.is_pynb? || rmd_convert
-      @annotation = result.annotations.create!(
-        type: 'HtmlAnnotation',
-        start_node: params[:start_node],
-        start_offset: params[:start_offset],
-        end_node: params[:end_node],
-        end_offset: params[:end_offset],
-        **base_attributes
-      )
-    else
-      @annotation = result.annotations.create(
-        type: 'TextAnnotation',
-        line_start: params[:line_start],
-        line_end: params[:line_end],
-        column_start: params[:column_start],
-        column_end: params[:column_end],
-        **base_attributes
-      )
-    end
+    annotation_type = submission_file.annotation_type
+    @annotation = result.annotations.create!(
+      type: annotation_type,
+      **type_specific_annotation_attributes(annotation_type),
+      **base_attributes
+    )
     render :create
   end
 
@@ -61,7 +28,6 @@ class AnnotationsController < ApplicationController
     result = Result.find(params[:result_id])
     submission = result.submission
     submission_file = submission.submission_files.find(params[:submission_file_id])
-    rmd_convert = submission_file.is_rmd? && Rails.application.config.rmd_convert_enabled
 
     d = result.grouping.assignment.annotation_categories.find_by(id: params[:category_id])&.flexible_criterion_id
 
@@ -90,44 +56,12 @@ class AnnotationsController < ApplicationController
       is_remark: !result.remark_request_submitted_at.nil?,
       submission_file_id: submission_file.id
     }
-    if submission_file.is_supported_image?
-      @annotation = result.annotations.create!(
-        type: 'ImageAnnotation',
-        x1: params[:x1],
-        x2: params[:x2],
-        y1: params[:y1],
-        y2: params[:y2],
-        **base_attributes
-      )
-    elsif submission_file.is_pdf?
-      @annotation = result.annotations.create!(
-        type: 'PdfAnnotation',
-        x1: params[:x1],
-        x2: params[:x2],
-        y1: params[:y1],
-        y2: params[:y2],
-        page: params[:page],
-        **base_attributes
-      )
-    elsif submission_file.is_pynb? || rmd_convert
-      @annotation = result.annotations.create!(
-        type: 'HtmlAnnotation',
-        start_node: params[:start_node],
-        start_offset: params[:start_offset],
-        end_node: params[:end_node],
-        end_offset: params[:end_offset],
-        **base_attributes
-      )
-    else
-      @annotation = result.annotations.create!(
-        type: 'TextAnnotation',
-        line_start: params[:line_start],
-        line_end: params[:line_end],
-        column_start: params[:column_start],
-        column_end: params[:column_end],
-        **base_attributes
-      )
-    end
+    annotation_type = submission_file.annotation_type
+    @annotation = result.annotations.create!(
+      type: annotation_type,
+      **type_specific_annotation_attributes(annotation_type),
+      **base_attributes
+    )
   end
 
   def destroy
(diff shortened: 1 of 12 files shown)


app/controllers/api/groups_controller.rb
```diff
@@ -5,6 +5,17 @@ class GroupsController < MainApiController
     # Define default fields to display for index and show methods
     DEFAULT_FIELDS = [:id, :group_name].freeze
 
+    # The annotation types accepted by add_annotations (single-table-inheritance class names,
+    # matching the values reported by the annotations endpoint), mapped to the fields that must
+    # be present for each. Mirrors each subclass's model validations, enforced in add_annotations
+    # because insert_all bypasses them.
+    REQUIRED_ANNOTATION_FIELDS = {
```
david-yz-liu: Okay, so I think defining this is good, but really this information belongs in each model. It's a bit awkward to translate from string name to class, but since we're always deriving the expected class from the submission file anyway, I suggest the following changes:

1. Modify `SubmissionFile.annotation_type` to be `SubmissionFile.annotation_class` and return the class itself (e.g., `TextAnnotation`) rather than the string (`"TextAnnotation"`).
2. You can go from the class to its name with the `.name` method
3. Define the symbol lists in each of the `Annotation` subclasses and then access them instead of `REQUIRED_ANNOTATION_FIELDS[expected_type]` below
4. There's a use of `REQUIRED_ANNOTATION_FIELDS.key?` below, but I don't think we need it. It's sufficient to check against the expect type and report the "does not match the type" error, without having a separate "Invalid annotation type" error.

app/controllers/api/groups_controller.rb
```diff
@@ -238,12 +249,45 @@ def add_annotations
         return page_not_found('Marking for that submission is already completed')
       end
 
+      annots = annotations_params
+      submission_files = result.submission.submission_files.index_by(&:filename)
+
+      # Validation pass: insert_all bypasses model validations, so guard the input here
```
david-yz-liu: A separate pass isn't necessary here. This check can happen inside the existing loop below; the early return will ensure nothing is inserted into the database, as that only happens after the loop

app/controllers/api/groups_controller.rb
```diff
@@ -238,12 +249,45 @@ def add_annotations
...
+            return add_annotations_error("Invalid annotation type: #{requested}")
+          end
+          if requested != expected_type
+            return add_annotations_error(
+              "Annotation type '#{requested}' does not match the type of file '#{annot_params[:filename]}'"
+            )
+          end
+        end
+
+        missing = REQUIRED_ANNOTATION_FIELDS[expected_type].select { |field| annot_params[field].blank? }
```
david-yz-liu: Let's be stricter here.

1. If there are any required fields that are missing, report an error (this is what the previous `params.require` call would have done)
2. If there are additional annotation fields that are "for the wrong type" (e.g. passing `line_start` for a `PdfAnnotation`) also report an error and return

Then, extract the annotation attributes (you can use the Ruby [`Hash#slice` method](https://ruby-doc.org/core-3.1.0/Hash.html#method-i-slice) and use that to create the hash that's appended to `annotations`, rather than passing all attributes.

app/controllers/api/groups_controller.rb
```diff
@@ -507,15 +559,31 @@ def grouping
     end
 
     def annotations_params
-      params.require(annotations: [
+      params.permit(annotations: [
```
david-yz-liu: I'd like to keep the `require`, which was strict with the param requirement. I assume you changed this to allow the inner hashes to not have all of the attributes.

I think you can achieve both the strict `:annotations` key and the permissive inner structure by using the new [`expect` method](https://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-expect)

app/controllers/annotations_controller.rb
```diff
@@ -16,52 +15,19 @@ def add_existing_annotation
...
-        line_end: params[:line_end],
-        column_start: params[:column_start],
-        column_end: params[:column_end],
-        **base_attributes
-      )
-    end
+    annotation_type = submission_file.annotation_type
+    @annotation = result.annotations.create!(
+      type: annotation_type,
+      **type_specific_annotation_attributes(annotation_type),
```
david-yz-liu: Right, so this is the part where I'd like the logic to be the same for both controller methods.

1. Get the list of symbols from a model method
2. Call `Hash#slice` to get the values from `params` (no need to define a separate method)

app/controllers/api/groups_controller.rb
philipkukulak: Great notes, thanks David. Fixing this

app/controllers/api/groups_controller.rb
philipkukulak: I've moved validation into the main loop, but one of the specs in this PR (valid item creating a new category, followed by invalid item) started failing because `find_or_create_by` to insert annotation categories will commit as soon as it is called, thereby breaking atomicity and leaving an orphaned annotation category for that assignment.

To mitigate this, I've wrapped the entire loop in an `ActiveRecord::Base.transaction` that rolls back whenever any kind of error is caught.

app/controllers/api/groups_controller.rb
philipkukulak: Gotcha

app/controllers/api/groups_controller.rb
philipkukulak: Oh nice! Yes, that's why it was changed. Fixing.

app/controllers/annotations_controller.rb
philipkukulak: Done!

philipkukulak:
Thanks for the detailed review, David. I hope the new commits nail down your concerns.

david-yz-liu (approved):
@philipkukulak this is great, thanks!

PR #8008, opened 2026-06-18
feat: grades upload flow

### LINKED ISSUE #7794: Add the ability to upload grades directly to assignments
There should be an option to upload grades directly to assignments, similar to how they're uploaded to 'Marks Spreadsheet' which do not have access to groups made for specific assignments.

It is quite easy for us to make grading scripts, but we currently need to manually enter grades for each student/group.

YheChen:
## Proposed Changes

Adds CSV upload support for criterion marks from the assignment Grades tab.

This PR adds an Upload button for instructors on the assignment Grades tab. The upload flow accepts a CSV in the same general format as the existing Grades tab CSV export / grade entry form uploads, updates marks for matching assignment criteria, supports an overwrite option, and ignores total mark and bonus/deduction columns.

Also adds backend import handling, authorization, translations, and model/controller tests.

Fixes #7794

<details>

## Upload Button
image

## Upload Modal
image

</details>

<details>

https://github.com/MarkUsProject/Wiki/pull/268

</details>

## Type of Change

| ✨ *New feature* (non-breaking change that adds functionality)                          | X        |
| 🎨 *User interface change* (change to user interface; provide screenshots)              | X        |

## Checklist

- [x] If this is my first contribution, I have added myself to the [list of contributors](https://github.com/MarkUsProject/Markus/blob/master/doc/markus-contributors.txt).

## Questions and Comments


app/policies/assignment_policy.rb
```diff
@@ -2,6 +2,7 @@
 class AssignmentPolicy < ApplicationPolicy
   default_rule :manage?
   alias_rule :summary?, to: :view?
+  alias_rule :upload_grades?, to: :manage?
```
david-yz-liu: This isn't necessary because of the `default_rule :manage?` call

app/models/assignment.rb
```diff
@@ -866,6 +866,64 @@ def summary_csv(role)
     end
   end
 
+  def import_marks_from_csv(grades_data, overwrite, role)
+    results_by_group_name = current_results
+                            .joins(grouping: :group)
+                            .includes(:marks, grouping: :group)
+                            .index_by { |result| result.grouping.group.group_name }
+    results_by_user_name = current_results
```
david-yz-liu: Don't include this. The group name is required; the list of users is optional, and should be ignored if it's in the CSV upload. There is no need to have a fallback query (see my further comment below on `result_for_uploaded_marks_row`)

app/models/assignment.rb
```diff
@@ -866,6 +866,64 @@ def summary_csv(role)
...
+        rescue CsvInvalidLineError
+          criterion_columns = :invalid
+          raise
+        end
+        next
+      end
+
+      raise CsvInvalidLineError if criterion_columns == :invalid
+
+      result = result_for_uploaded_marks_row(row, group_name_index, user_name_index,
```
david-yz-liu: I wouldn't define a helper function here, it's not worth it. Do a lookup on `row[group_name_index]`. This will either be a `Result` or `nil`, and is the only thin you need to use.

app/models/assignment.rb
```diff
@@ -866,6 +866,64 @@ def summary_csv(role)
...
+                                             results_by_group_name, results_by_user_name, results_by_id)
+      raise CsvInvalidLineError if result.nil? || result.released_to_students?
+
+      Mark.transaction do
+        criterion_columns.each do |column_index, criterion|
+          mark = result.marks.find_or_initialize_by(criterion: criterion)
+          next if !overwrite && !mark.mark.nil?
+
+          mark_value = parse_uploaded_mark(row[column_index])
```
david-yz-liu: No need to make this a helper function, inline the logic here

app/models/assignment.rb
```diff
@@ -866,6 +866,64 @@ def summary_csv(role)
...
+          unless mark.update(mark: mark_value,
```
david-yz-liu: Doing this one record at a time will be quite slow, even with the transaction. Modify the logic to accumulate a set of updates and use [`upsert_all`](https://api.rubyonrails.org/v7.0/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-upsert_all)

app/models/assignment.rb
```diff
@@ -866,6 +866,64 @@ def summary_csv(role)
...
+
+      if headers.nil?
+        headers = row
+        group_name_index = headers.index(Group.human_attribute_name(:group_name)) || 0
+        user_name_index = headers.index(User.human_attribute_name(:user_name)) || 1
+        next
+      elsif criterion_columns.nil?
+        begin
+          criterion_columns = assignment_upload_criterion_columns(headers, row, criteria_by_name)
+        rescue CsvInvalidLineError
```
david-yz-liu: I'm not sure what the purpose of this exception handling and re-raising is

app/models/assignment.rb
```diff
@@ -866,6 +866,64 @@ def summary_csv(role)
...
+      next if row.empty?
+
+      if headers.nil?
+        headers = row
+        group_name_index = headers.index(Group.human_attribute_name(:group_name)) || 0
+        user_name_index = headers.index(User.human_attribute_name(:user_name)) || 1
+        next
+      elsif criterion_columns.nil?
+        begin
+          criterion_columns = assignment_upload_criterion_columns(headers, row, criteria_by_name)
```
david-yz-liu: Similar to my above comment, I think this is overly complex and doesn't require a helper. Ignore the "totals" row; determine `Criterion` object based on the name, which is found only in the headers row.

You can ignore the "totals" row by skipping any row with a blank "group name" column, which I believe you already have some logic to do.

app/models/assignment.rb
```diff
@@ -881,6 +939,68 @@ def next_criterion_position
     criteria.exists? ? criteria.last.position + 1 : 1
   end
 
+  def criteria_by_upload_name
```
david-yz-liu: This helper function isn't necessary; again, inline the code above.

First, to create a hash use the pattern of

```
ta_criteria.map do |criteria|
  ...

  # end with an array of [key, value]
end.to_h
```

Second, the most complex part of this logic is the conditional adding of "bonus". This appears elsewhere in the code for generating CSV files, so actually what I would suggest is to define a `Criterion` helper method called `export_name` that handles this logic, and then just call `criterion.export_name` here. Also modify the existing places in the CSV exporting to use this method.

The `.to_s` and `strip` calls are unnecessary. They are not found in the corresponding CSV export method.

config/locales/views/assignments/en.yml
```diff
@@ -109,3 +109,6 @@ en:
     upload_config_help_html: "<p>This zip file will configure and create this assignment.</p><p>This does not configure student or grader settings.<br>For more information please visit <a href=https://github.com/MarkUsProject/Wiki/blob/%{markus_version}/Instructor-Guide--Importing-and-Exporting-Data.md#%{section_id}>this page</a>.</p>"
     upload_file_requirement: Filename %{file_name} is not allowed for this assignment.
     upload_file_requirement_in_folder: Filename %{file_name} in %{file_path} is not allowed for this assignment.
+    upload_grades:
+      invalid_criteria: 'No matching criteria were found in the uploaded CSV. Unknown criteria: %{criteria}'
```
david-yz-liu: Reword this message to `"The following criteria were not found for this assignment: %{criteria}"`

david-yz-liu (commented):
@YheChen nice work, I left a few comments. Please also make sure to update the screenshots in the PR description and the documentation Wiki.

spec/controllers/assignments_controller_spec.rb
```diff
@@ -692,6 +735,7 @@
       example_form_params[:assignment][:submission_rule_attributes][:periods_attributes] = submission_rule.id
       example_form_params
     end
+
```
david-yz-liu: Please revert the blank line changes in this file (here and below)

app/models/assignment.rb
```diff
@@ -867,6 +867,108 @@ def summary_csv(role)
...
 
+  def import_marks_from_csv(grades_data, overwrite, role)
+    results_by_group_name = current_results
+                            .joins(grouping: :group)
+                            .includes(grouping: :group)
+                            .index_by { |result| result.grouping.group.group_name }
+    criteria_by_name = ta_criteria.index_by(&:export_name)
+    ignored_headers = [Group.human_attribute_name(:group_name),
+                       I18n.t('results.total_mark'),
+                       'Bonus/Deductions'] +
```
david-yz-liu: Sorry I just noticed this; please internationalize this string (and replace here and in the CSV export)

app/models/assignment.rb
```diff
@@ -867,6 +867,108 @@ def summary_csv(role)
...
+          criterion = criteria_by_name[header]
+          next [i, criterion] unless criterion.nil?
+
+          unknown_columns << header if header.present? && ignored_headers.exclude?(header)
+          nil
+        end
+
+        if uploaded_criterion_columns.empty? || unknown_columns.present?
+          criterion_columns = []
+          raise CsvInvalidLineError,
```
david-yz-liu: Ah okay I understand a bit better now. When the criteria columns are invalid, we shouldn't raise a `CsvInvalidLineError` (which is handled by `MarkusCsv.parse`), but instead you can just raise the string directly. This will terminate this entire function; then, handle the error in the controller and report the specific error message to the user.

app/models/assignment.rb
```diff
@@ -867,6 +867,108 @@ def summary_csv(role)
...
+                I18n.t('assignments.upload_grades.invalid_criteria', criteria: unknown_columns.join(', '))
+        end
+        criterion_columns = uploaded_criterion_columns
+        next
+      end
+
+      raise CsvInvalidLineError if criterion_columns.empty?
```
david-yz-liu: Doing the above comment will allow you to remove this check

app/models/assignment.rb
```diff
@@ -867,6 +867,108 @@ def summary_csv(role)
...
+        ]
+      end
+
+      row_updates.each do |key, mark, attrs|
+        if mark&.[](:id)
+          mark_updates[mark[:id]] = attrs.merge(id: mark[:id])
+        else
+          new_mark_updates[key] = attrs.merge(created_at: now)
+        end
+        marks_by_result_and_criterion[key] = { id: mark&.[](:id), mark: attrs[:mark] }
```
david-yz-liu: I don't think this line has any effect (`marks_by_result_and_criterion` is not used further in this function)

app/models/assignment.rb
```diff
@@ -867,6 +867,108 @@ def summary_csv(role)
...
+            updated_at: now
+          }
+        ]
+      end
+
+      row_updates.each do |key, mark, attrs|
+        if mark&.[](:id)
+          mark_updates[mark[:id]] = attrs.merge(id: mark[:id])
+        else
+          new_mark_updates[key] = attrs.merge(created_at: now)
```
david-yz-liu: Hmm I don't think the `created_at` timestamp is necessary? I thought this was address in Rails v7 (https://github.com/rails/rails/pull/43003).

spec/controllers/assignments_controller_spec.rb
YheChen: FYI I discovered that RuboCop had a rule `RSpec/EmptyLineAfterFinalLet` that was creating these blank lines, which I bypassed in my commit by doing 
```SKIP=rubocop git commit -m "Remove incidental controller spec blank lines"```

david-yz-liu (approved):
Nice work, @YheChen!

PR #8054, opened 2026-07-13
Refactored Channel classes' authorization 

mrafie1:
## Proposed Changes
*(Describe your changes here. Also describe the motivation for your changes: what problem do they solve, or how do they improve the application or codebase? If this pull request fixes an open issue, [use a keyword to link this pull request to the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).)*

- Updated policies to include the record and context validation checks that were previously handled within subscribed methods in channel child classes.
- Refactored the channel superclass:
  - Added automatic auth handling for children and auth state checks.
    - Runs authorization automatically with `before_subscribe :authorize_channel`
    - Checks if authorization occurs with `before_subscribe :verify_authorized` and custom `verify_authorized` method
  - Introduced methods for setting up context, defining target records, performing auth, and verifying auth completion.
- Refactored channel child classes:
  - Removed auth logic from subscribed methods.
  - Added dedicated methods for defining context and auth target records.

**UPDATE JUL 16**

- Moved `Assignment` method to classes that only required it
  - Additionally, moved `authorize:...` context to classes that only required it
- Removed `record` from `assignment_policy.rb` and replaced it with `assignment` that is passed in as context
- Updated all relevant policy files to accept context passed in with `authorize!`
  - Example line 9 of `app/policies/assignment_policy.rb` explicitly accepts `:assignment`
 - Changed `channel.rb`'s `authorize_channel` method to reject if current_role is nil. See response to third comment for more information as I found this change to be a little complicated
 - Changed `channel.rb` definition of `real_user` to use `connection.real_user` instead of `connection.current_user`
 - Changed `connection.rb` to fit the definitions in `session_handler.rb`
 - Changed all channel tests to now use `real_user` whose values are the same as `current_user`
   - Reasoning: No rspec test seemed to be testing for when real_user is different than current_user (in cases of impersonation)

<details>

</details>

## Type of Change

| ♻️ *Refactoring* (internal change to codebase, without changing functionality)          |     x     |
| 📖 *Documentation update* (change that updates documentation)                           |          |

## Checklist

- [ ] I have added tests for my changes, if applicable.
- [ ] I have updated the project documentation, if applicable.

- [ ] I have reviewed the test coverage changes reported by Coveralls.

## Questions and Comments
*(Include any questions or comments you have regarding your changes.)*

About your comment regarding `verify_authorized`: 
- The function exists in ActionPolicy::Controller [link](https://www.rubydoc.info/gems/action_policy/ActionPolicy/Controller?utm_source=chatgpt.com)
- It is specific to Controllers only; "Controller concern..." 
- ActionPolicy::Behaviour [link](https://www.rubydoc.info/gems/action_policy/ActionPolicy/Behaviour?utm_source=chatgpt.com) can be used anywhere; "..Could be included anywhere to perform authorization..." however does not include `verify_authorized`

It would be wiser to implement our own `verify_authorized` method, which only checks if authorization occurs, and not perform authorization itself. (In the context of controllers, `before_action { authorize! }` automatically authorizes, and `verify_authorized` checks to see if that function is called`

Questions:

- In my `channel.rb` file, what are your opinions on the methods defined such as `current_role`, `real_user`, etc? I feel like there must be a better way to define variables without using many 'one-line methods'


app/channels/application_cable/channel.rb
```diff
@@ -1,4 +1,51 @@
...
+
+    authorize :role, through: :current_role
+    authorize :real_user, through: :real_user
+    before_subscribe :authorize_channel
+    before_subscribe :verify_authorized
+
+    private
+
+    def authorize_channel
+      @authorization_checked = true
```
david-yz-liu: Seems like you should put this after the `authorize!` call

app/channels/application_cable/channel.rb
```diff
@@ -1,4 +1,51 @@
...
+
+    def authorization_rule
+      raise NotImplementedError
+    end
+
+    def current_role
+      Role.find_by(user: current_user, course: course)
+    end
+
+    def real_user
```
david-yz-liu, line 37: Overall, defining small methods in here is okay (especially the ones that are meant to be overridden). You can see similar methods defined in `session_helper.rb` for our controllers.

However, for `real_user` in particular, this definition isn't correct (I think the original implementation also had this bug). There's a separate session header. Take a look at `session_handler.rb`.

app/policies/submission_policy.rb
```diff
@@ -10,7 +10,7 @@ class SubmissionPolicy < ApplicationPolicy
   authorize :from_codeviewer, :view_token, optional: true
 
   def manage?
-    check?(:manage_submissions?, role)
+    !role.nil? && check?(:manage_submissions?, role)
```
david-yz-liu: I'm a bit surprised by this check. I thought `ApplicationPolicy` already has a `pre_check :role_exists?` that will handle this?

app/policies/assignment_policy.rb
```diff
@@ -17,7 +17,7 @@ def switch?
   end
 
   def manage?
-    check?(:manage_assessments?, role)
+    (!role.nil? || !record.nil?) && check?(:manage_assessments?, role)
```
david-yz-liu: See my comment below about `role.nil?`.

For `record.nil?`, can you dig deeper into how record` is defined here? There's one definition in `session_handler.rb` (again, for controllers); it seems like maybe we'd need to define this in the relevant channels as well?

app/channels/application_cable/channel.rb
```diff
@@ -1,4 +1,51 @@
...
+      connection.current_user
+    end
+
+    def course
+      Course.find_by(id: params[:course_id])
+    end
+
+    def assignment
```
david-yz-liu: I think this is only being used in one of the channels, let's define this method there for now. We can lift it up later if needed.

app/policies/submission_policy.rb
mrafie1: I am deleting this from all policies. This is because it seems to be necessary to have this 'logic' inside the authorize_channel function, as some test cases in `test_runs_channel_spec.rb` test whether giving a course_id value of nil (which causes role to be nil) will stop the channel from being subscribed to.
 
This means I would have to go through *all* (relevant) policies to add this 'nil checking' so I ended up keeping the 'role nil checking' inside the `authorize_channel` function.

app/policies/assignment_policy.rb
mrafie1: For this, I ended up removing `record` and replacing it with `assignment` instead and updated the policy to declare that they need `:assignment`

My initial thought about using record was because originally, we checked if `assignment` is nil. Since we end up passing it into our `authorize!` function, `record` is representing assignment ([link](https://www.rubydoc.info/gems/action_policy/ActionPolicy/Behaviour))

app/channels/application_cable/channel.rb
mrafie1: From the documentation, it seems that verify_authorized only checks that authorization occurs ([Link](https://www.rubydoc.info/gems/action_policy/ActionPolicy/Controller)). In the context of the controllers, it seems that it is there to make sure each child controller calls `before_action {authorize!}`

But since we already 'automatically' call authorization from the parent, the children don't really need to do anything. Is it safe to remove verify_authorized entirely?

app/policies/submission_policy.rb
mrafie1: UPDATE: However, I think I will keep the logic to check if 'assignment' is nil in the file `assignment_policy.rb`.

This is because if I were to put this logic inside the parent `channel.rb` file, not all channel actions require the assignment to not be nil, which could cause some connections to be rejected when not intended to be rejected.

mrafie1:
Hi David, took a look at some of the feedback. I wrote down some comments on some of the feedback comments. I will let you know when I update my PR with the new changes.

mrafie1:
Hi David. I pushed all of my changes and addressed all of your comments. I tried my best to thoroughly explain all of my changes, but let me know if there's something in particular that needs to be addressed. 

Note: I will change the changelog description to be more descriptive.

One question: I ended up changing the definition of `real_user` from `connection.current_user` to `connection.real_user`. Then I ended up changing the channel tests by adding `real_user` in the connection stub and setting its value to be identical to `current_user`

I wanted to know if this is okay. My understanding of the differentiation between current/real users is to handle impersonation logins. The channel tests don't handle this kind of case.

app/channels/application_cable/channel.rb
david-yz-liu: Hmm okay, you can go ahead and remove the `@authorization_checked` variable and `verify_authorized` entirely.

app/channels/collect_submissions_channel.rb
```diff
@@ -1,20 +1,19 @@
...
-    end
-    unless allowed_to?(:collect_submissions?, with: SubmissionPolicy, context: { real_user: current_user,
-                                                                                 role: role })
-      reject
-      return
-    end
     stream_for current_user
   end
 
+  private
```
david-yz-liu: Don't make the `unsubscribed` methods private (throughout)

app/policies/assignment_policy.rb
```diff
@@ -17,7 +17,7 @@ def switch?
   end
 
   def manage?
-    check?(:manage_assessments?, role)
+    !assignment.nil? && check?(:manage_assessments?, role)
```
david-yz-liu: You should not be changing the existing policy. Put the `assignment.nil?` check directly in the channel that requires this check.

app/policies/assignment_policy.rb
```diff
@@ -6,7 +6,7 @@ class AssignmentPolicy < ApplicationPolicy
   alias_rule :show?, :peer_review?, to: :student?
   alias_rule :starter_file?, :download_starter_file_mappings?, :download_sample_starter_files?,
              :populate_starter_file_manager?, to: :read_starter_files?
-  authorize :assessment, :test_run_id, optional: true
+  authorize :assessment, :test_run_id, :role, :real_user, :assignment, optional: true
```
david-yz-liu: I don't see why you have added `:role` and `:real_user` to these policies; they are inherited from `ApplicationPolicy`.

david-yz-liu (commented):
Hi @mrafie1, good progress, and I left a few more comments. The changes to the tests are fine.

app/policies/assignment_policy.rb
```diff
@@ -6,7 +6,7 @@ class AssignmentPolicy < ApplicationPolicy
   alias_rule :show?, :peer_review?, to: :student?
   alias_rule :starter_file?, :download_starter_file_mappings?, :download_sample_starter_files?,
              :populate_starter_file_manager?, to: :read_starter_files?
-  authorize :assessment, :test_run_id, optional: true
+  authorize :assessment, :test_run_id, :assignment, optional: true
```
david-yz-liu: I don't think adding `assignment` here is necessary any more (there are no more references to this variable).

Changelog.md
```diff
@@ -46,6 +46,7 @@
 - Updated images: moved externally-hosted images into repository, ensured images are vertically centred and displayed as new paragraphs, and cropped images to remove blank space (#8053)
 
 ### 🔧 Internal changes
+- Refactored Action Cable channel authorization into `channel.rb` superclass (#8054)
```
david-yz-liu: Oh I missed this before, but please change `channel.rb` to just `Channel` (the former is the filename, whereas the latter is that actual class name).

david-yz-liu (approved):
Nice work, @mrafie1!

PR #8128, opened 2026-08-10
Issue 6225: Allow multiple periods in Assignments API routes

### LINKED ISSUE #6225: Allow Assignments API routes to accept nested parameters for SubmissionRules
Currently, the create and update routes on the Assignments API do not accept nested parameters. Notably, this means that when creating an assignment via API only one `Period` can be created for the associated `SubmissionRule`. We could update these methods to accept nested parameters to allow multiple periods.

sophia-huynh:
## Proposed Changes
Updated the `POST /api/courses/:course_id/assignments` and `PUT /api/courses/:course_id/assignments/:id` API routes to accept the parameters `submission_rule_type` and `submission_rule_periods`, the latter allowing for one or more submission rule periods.

While there was formerly an implementation that would allow for a single submission rule to be included with a single period, this _did not work_ and was undocumented. When the related parameters were provided, no errors were raised, but no updates were made. The related parameters have been removed and replaced with `submission_rule_periods`, as they weren't documented or functional in the first place.

Closes #6225 

<details>

</details>

## Type of Change

| ✨ *New feature* (non-breaking change that adds functionality)                          |    X     |
| 🐛 *Bug fix* (non-breaking change that fixes an issue)                                  |    X     |
| 📖 *Documentation update* (change that updates documentation)                           |          |

## Checklist

- [x] If this is my first contribution, I have added myself to the [list of contributors](https://github.com/MarkUsProject/Markus/blob/master/doc/markus-contributors.txt).

## Questions and Comments
Formerly, there were checks for `submission_rule.valid?` which _never_ succeeded due to the relationships/dependencies between submission rules, periods, and assignments. Thus, the rules were never updated (and the code within the conditional also wouldn't work correctly), and then a success was returned afterwards.


app/controllers/api/assignments_controller.rb
```diff
@@ -230,22 +232,66 @@ def update_test_specs
     # Defaults to NoLateSubmissionRule
     def get_submission_rule(params)
       if params[:submission_rule_type] == 'GracePeriod'
+        if params[:submission_rule_periods].nil? ||
```
david-yz-liu: So overall we can make use of Rails support for [nested attributes](https://api.rubyonrails.org/classes/ActiveRecord/NestedAttributes/ClassMethods.html) to have a more nested structure for these params:

```ruby
{
  submission_rule_attributes: {
    type: ...,
    period_attributes: [
      { hours: ..., ... },
      ...
    ]
  }
}
```

This should allow you to create a new `SubmissionRule` object directly from the `submission_rule_attributes`, and if the creation fails due to a validation error, an error should be reported automatically, eliminating the need to check for particular keys manually,

app/controllers/api/assignments_controller.rb
```diff
@@ -138,15 +136,19 @@ def update
       # Update the submission rule if provided
       unless params[:submission_rule_type].nil?
         submission_rule = get_submission_rule(params)
+
```
david-yz-liu: Remove this blank line

app/controllers/api/assignments_controller.rb
```diff
@@ -138,15 +136,19 @@ def update
...
           render 'shared/http_status', locals: { code: '500', message:
             HttpStatusHelper::ERROR_CODE['message']['500'] }, status: :internal_server_error
           return
-        elsif submission_rule.valid?
-          # If it's a valid submission rule, replace the existing one
-          assignment.submission_rule.destroy
-          assignment.submission_rule = submission_rule
         end
+
+        unless assignment.update(submission_rule: submission_rule)
```
david-yz-liu, line 146: Catching `submission_rule.valid?` is great. However, you should keep the logic of deleting the old submission rule if it's replaced by a new one.

app/controllers/api/assignments_controller.rb
sophia-huynh: Revised to use nested attributes. However, error checking for missing attributes within periods was still required - or I can't find a way around it.

Without it, the creation *doesn't* fail properly, and checking `.errors.any?` seems to always give an error even for correctly created rules (probably for a similar reason as `valid?`'s failures.) It *does* raise an error when `update` is called*, but the error isn't handled elegantly and a traceback gets sent in the response.
\* `can't modify frozen attributes` is the error that gets raised... for some reason I haven't been able to completely figure out. If it was just from the validation failing, then `update` should have returned `false` instead.

app/controllers/api/assignments_controller.rb
```diff
@@ -230,28 +233,35 @@ def update_test_specs
...
+      end
+
+      if params[:submission_rule_periods].nil?
+        return
+      end
+
+      permitted_params = params.permit(
+        submission_rule_periods: [:hours, :deduction, :interval, :_destroy]
+      )
+      if permitted_params[:submission_rule_periods].any? do |period|
```
david-yz-liu: Okay, I'm good with the check but the syntax here is kind of hard to read. I would use a more explicit loop instead:

```ruby
permitted_params[:submission_rule_periods].each do |period|
  if ...
    return
  end
end
```

app/controllers/api/assignments_controller.rb
```diff
@@ -230,28 +233,35 @@ def update_test_specs
...
+        required.any? { |key| !period.key?(key) }
+      end
+        return
       end
 
-      submission_rule
+      SubmissionRule.create(
```
david-yz-liu: Hm, previously we used `[...SubmissionRule].new`, which defers the database saving to later. I'd prefer keeping the same behaviour unless there's a particular reason we need to save in this method.

(Saving will run the validations, but I don't think the code using `get_submission_rule` is checking for validation errors on the submission rule instance returned.)

david-yz-liu (approved):
Thanks, @sophia-huynh!

PR #8150, opened 2026-09-03
Fix production JS builds compiling JSX in development mode

Naragod:
## Proposed Changes

Production webpack builds compile JSX in development mode. #8042 moved JSX to the automatic runtime. In development mode, that emits `jsxDEV` calls. Production React drops that function. The bundle then crashes on its first component. Every page loses its JavaScript after that ("Routes is not defined", empty admin Courses and Users tables). Deployed v2.10.2 instances hit this. Dev servers and Jest kept working.

Cause: Babel picks its mode from the build shell's `BABEL_ENV` or `NODE_ENV`. Webpack's `mode` plays no part. Deploy hosts leave both unset, so Babel ran in development mode.

Two changes:
- `webpack.production.js` sets `BABEL_ENV` plus `NODE_ENV` to `production` at the top. A production build now compiles right on any host.
- `babel.config.js` gives `@babel/preset-react` an explicit `development` flag. Dev and test keep `jsxDEV`.

Verification:
- Before: a production bundle contains `jsxDEV`. A headless Chromium login crashes with `(0, fF.jsxDEV) is not a function` and shows zero tables.
- After: the bundle has zero `jsxDEV`. The same headless login renders the course list with a clean console. `window.Routes` works.
- Jest passes (the test env keeps the development runtime).

## Type of Change

| Type                                                                                    | Applies? |
|-----------------------------------------------------------------------------------------|----------|
| 🚨 *Breaking change* (fix or feature that would cause existing functionality to change) |          |
| ✨ *New feature* (non-breaking change that adds functionality)                          |          |
| 🐛 *Bug fix* (non-breaking change that fixes an issue)                                  | X        |
| 🎨 *User interface change* (change to user interface; provide screenshots)              |          |
| ♻️ *Refactoring* (internal change to codebase, without changing functionality)          |          |
| 🚦 *Test update* (change that *only* adds or modifies tests)                            |          |
| 📦 *Dependency update* (change that updates a dependency)                               |          |
| 📖 *Documentation update* (change that updates documentation)                           |          |
| 🔧 *Internal* (change that *only* affects developers or continuous integration)         | X        |

## Checklist

Before opening your pull request:

- [x] I have performed a self-review of my changes.
  - Check that all changed files included in this pull request are intentional changes.
  - Check that all changes are relevant to the purpose of this pull request, as described above.
- [ ] I have added tests for my changes, if applicable.
  - This is **required** for all bug fixes and new features.
- [ ] I have updated the project documentation, if applicable.
[...]

PR #8162, opened 2026-09-09
Fix production JS builds compiling JSX in development mode

Naragod:
Cherry-picks #8150 onto `release` so the branch builds a correct production bundle on its own.

### Why

v2.10.2 shipped without this fix. `@babel/preset-react` compiled JSX in development mode, so the production bundle called `jsxDEV`, which React's production build does not provide. The bundle died on its first component and pages lost their React tables.

The workaround on the servers was an asset rebuild with `NODE_ENV=production` exported by hand. This commit makes the build set it itself.

### Scope

| File | Change |
|---|---|
| `babel.config.js` | `preset-react` receives `development: isDevelopmentEnv \|\| isTestEnv` |
| `webpack.production.js` | sets `BABEL_ENV` and `NODE_ENV` before webpack loads |
| `Changelog.md` | one bug-fix line under `[unreleased]` |

### Verification

- Clean cherry-pick of `ec4bf2b` from master. Zero conflicts.
- `babel.config.js` and `webpack.production.js` md5-match master byte for byte.
- Both files also md5-match the copies running on `ben` and `tiny`, where the rebuilt bundle carries zero `jsxDEV` calls.
- Original PR #8150 passed jest, rspec, brakeman, CodeQL, and pre-commit on master.

### Tag

The `v2.10.2` tag stays where it is. This commit lands on `release` only, so deployed boxes stop being a hand patch. #8150 keeps its v2.10.3 milestone.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,6 +9,7 @@
 ### ✨ New features and improvements
 
 ### 🐛 Bug fixes
+- Fixed production webpack builds compiling JSX in development mode (#8150)
 
 ### 🔧 Internal changes
 
--- a/babel.config.js
+++ b/babel.config.js
@@ -33,7 +33,12 @@ module.exports = function (api) {
           exclude: ["transform-typeof-symbol"],
         },
       ],
-      require("@babel/preset-react").default,
+      [
+        require("@babel/preset-react").default,
+        {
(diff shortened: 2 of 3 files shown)


david-yz-liu (approved):
Looks good to me, thanks @Naragod

PR #8173, opened 2026-09-17
TICKET-13745: Fix marks release upsert

Naragod:
## Proposed Changes

Grade release on a marks spreadsheet fails. Unrelease fails too. Both raise this error:

```
PG::NotNullViolation: ERROR:  null value in column "assessment_id" of relation "grade_entry_students" violates not-null constraint
DETAIL:  Failing row contains (35718, t, 2026-09-16 21:32:29.446139, 2026-09-16 21:32:29.446139, null, 39608).
```

Here is what happens. MarkUs hands Postgres one row per student. Each row holds a row id, a role id, and the release flag. The form id stays out. Postgres checks the row for empty columns first. It finds an empty form id, so the write dies. The column is `assessment_id`. The call is `GradeEntryStudent.upsert_all` in `app/controllers/grade_entry_forms_controller.rb`.

Rows have skipped that column since 2019. #7965 made it `NOT NULL`. v2.10.1 shipped that change. Grade release has failed since.

Reported in RT #13745.

What we changed:

- Every row now carries `assessment_id`. Release writes again.
- Unrelease writes again.
- Release email waits for the write to land.
- A guard clause replaces the `else` branch.
- The raised error rolls the write back on its own.
- Release sends the student email.
- Unrelease stays quiet.
- New specs check the write on both paths.
- New specs check the email on a failed write.
- New specs check the email on an unrelease.

Students got mail about released marks after a failed write. They got the same mail on an unrelease. The old specs counted mail alone. Mail went out after each rollback, so they stayed green.

## Type of Change

| Type                                                        | Applies? |
|-------------------------------------------------------------|----------|
| 🚨 *Breaking change* (changes how MarkUs works today)        |          |
| ✨ *New feature* (adds something new)                        |          |
| 🐛 *Bug fix* (fixes a broken thing)                          | X        |
| 🎨 *User interface change* (add screenshots)                 |          |
| ♻️ *Refactoring* (tidies code, same behaviour)               | X        |
| 🚦 *Test update* (tests alone)                               |          |
| 📦 *Dependency update* (bumps a package)                     |          |
| 📖 *Documentation update* (docs alone)                       |          |
| 🔧 *Internal* (dev setup or CI alone)                        |          |

## Checklist

Before opening this pull request:

- [x] I read my own changes.
- [x] I added tests.
- [ ] I updated the docs, where docs apply.
[...]

PR #8174, opened 2026-09-17
TICKET-13745: Scope marks release emails to the form

Naragod:
## Proposed Changes

Releasing marks on one spreadsheet sends email to students on a different spreadsheet.

Here is what happens. The browser posts a list of row ids. The write step filters that list by course and by form, then saves the matching rows. The email step reads the raw list from the browser instead. It skips both filters. The list is `params[:students]`. The call is `GradeEntryStudent.where(id: params[:students])` in `app/controllers/grade_entry_forms_controller.rb`.

We measured it. Post form B row ids against form A. The write changes 0 rows. The email step still queues 3 messages, each naming form A. The flash says 0 grades changed on the same request.

#8173 added the form filter to the write step. The email step kept reading the raw list.

What we changed:

- The write block returns the row ids it saved.
- The email step reads those ids.
- Email now matches the count the flash reports.
- A new spec checks that students on another form get 0 email.
- A new spec checks that another form keeps its grades unreleased.

Both specs sit in the shared example, so the instructor run and the grader run each cover them.

Removing the fix turns both specs red. The email count changes when it holds today.

## Type of Change

| Type                                                        | Applies? |
|-------------------------------------------------------------|----------|
| 🚨 *Breaking change* (changes how MarkUs works today)        |          |
| ✨ *New feature* (adds something new)                        |          |
| 🐛 *Bug fix* (fixes a broken thing)                          | X        |
| 🎨 *User interface change* (add screenshots)                 |          |
| ♻️ *Refactoring* (tidies code, same behaviour)               |          |
| 🚦 *Test update* (tests alone)                               |          |
| 📦 *Dependency update* (bumps a package)                     |          |
| 📖 *Documentation update* (docs alone)                       |          |
| 🔧 *Internal* (dev setup or CI alone)                        |          |

## Checklist

Before opening this pull request:

- [x] I read my own changes.
- [x] I added tests.
- [ ] I updated the docs, where docs apply.
- [ ] First time here? Add your name to the [list of contributors](https://github.com/MarkUsProject/Markus/blob/master/doc/markus-contributors.txt).

After opening this pull request:

- [x] I updated `Changelog.md`. Every change requires a line.
- [x] The pre-commit.ci checks pass.
- [x] The CI tests pass.
- [x] I read the coverage change from Coveralls.
- [x] I asked a maintainer for a [review](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review).


donny-wong (approved):

david-yz-liu (approved):
Thanks @Naragod!

